PR #2, opened 2015-09-09
Fix #1 'not exists' issue and subscribe whole collection

### LINKED ISSUE #1: 'not exist' filters don't work if used alone
When subscribing to Kuzzle by filtering documents with only a 'not exist' condition, then:

* documents containing the filtered field are correctly rejected
* documents that don't contain it are not forwarded to the subscriber

Filter example:
```json
{
  "not":
    {
      "exists":
        {
          "field": "some field name"
        }
    }
}
```

The problem lies in the DSL, where we test each filter on fields. When the document don't contain any filtered field, it is automatically rejected.  
If the filter contains any other condition, then everything work fine.

The fix involves a refactor of the way filters are stored. This is already planned to support collection-wide filters.

AnthonySendra:
With this PR, we're now able to perfom a filter with just a filter 'not exists' like

```json
{
  "not":
    {
      "exists":
        {
          "field": "some field name"
        }
    }
}
```

We can also perform a subscribe on a whole collection by providing an empty object.
More information in issue #1

### DIFF
--- a/docs/API.AMQP.md
+++ b/docs/API.AMQP.md
@@ -132,6 +132,8 @@ How subscription works:
 }
 ```
 
+**Note:** If an empty body is provided, the subscription is performed on the whole collection
+
 **Response:**
 
 ```javascript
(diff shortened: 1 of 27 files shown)


j33f:
+1

scottinet:
+1

PR #11, opened 2015-09-17
add eslint and fix problems

AnthonySendra:
* Add eslint configuration
* Fix problem detected by eslint
* Create a new grunt task for eslint

**Why using Eslint & Jshint ?**
Eslint is very useful for test unused variable, undeclared variable, etc. But for unit testing is to restrictive (return error for undeclared variable `describe`, or `it` from mocha).

We use Eslint for all code in `lib` folder and Jshint for everything else.


scottinet:
+1 (nvm the codecov alert, this is on a not-yet tested part of the code)

PR #12, opened 2015-09-17
DSL Code Coverage

scottinet:
Reached 100% code coverage on the DSL code files (lib/api/dsl)

Bugfixes:

* (index.js) removeFilterPath was removing all filtered fields instead of only the one matching the filter path
* (index.js) getFiltersPathsRecursively didn't return every paths of a filter if nested 'and' or 'or' operand were present
* (methods.js) getFormattedFilters: the async callback wasn't called when en empty filter was given along with other non-empty filters


AnthonySendra:
LGTM (great work btw ;))

PR #19, opened 2015-09-18
Added code coverage for the worker loader.

scottinet:
Added code coverage for the worker loader.

Refactored it so that it truly loads workers groups sequentially, to allow some workers to depend on others.


j33f:
+1

Issue #22, opened 2015-09-21
Public socket frames


AnthonySendra:
When no connection is provided, the current implementation of the notifier core component broadcasts notifications to all connected clients. This allows clients to secretly listens to messages not adressed to them in the first place.  
A solution might be to get connections attached to the room to notify and to send the notification to each one of these connections.

PR #23, opened 2015-09-21
fix problem with public frame socket #22

AnthonySendra:
* Fix problem issue #22 
* Fix functional testing with multi sockets simulation
* Fix unit testing

Before this PR the room notification was made with broadcast. Now, when a user subscribe to a room with websocket, we use function`socket.join` for later use `io.to('room').emit('event', data)` function.
A call to function `leave` was also added in `removeSubscription` function.

### DIFF
--- a/features/support/apiWebsocket.js
+++ b/features/support/apiWebsocket.js
@@ -10,13 +10,12 @@ module.exports = {
   world: null,
   responses: null,
   subscribedRooms: {},
+  listSockets: {},
 
   init: function (world) {
     this.world = world;
 
-    if ( !this.socket ) {
-      this.socket = io(config.url, { 'force new connection': true });
-    }
+    initSocket.call(this, 'client1');
   },
 
   disconnect: function () {
@@ -147,7 +146,7 @@ module.exports = {
     return emit.call(this, 'admin', msg );
   },
 
-  subscribe: function (filters) {
+  subscribe: function (filters, socketName) {
     var
       msg = {
         action: 'on',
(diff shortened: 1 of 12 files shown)


features/support/apiWebsocket.js
```diff
@@ -170,18 +169,23 @@ module.exports = {
         requestId: room
       };
 
-    this.socket.removeListener(this.subscribedRooms[room], this.subscribedRooms[room].listener);
-    delete this.subscribedRooms[room];
-    return emit.call(this, 'subscribe', msg, false);
+    socketName = initSocket.call(this, socketName);
+
+    console.log('room', room);
```
ballinette: is this log trace needed or is it a debug trace that you forgot to clean-up ?

test/api/core/notifier/send.test.js
```diff
@@ -125,7 +125,7 @@ describe('Test: notifier.send', function () {
     (Notifier.__get__('send')).call(kuzzle, room, response);
     should(responded).be.true();
     should(mockupio.emitted).be.true();
-    should(mockupio.id).be.undefined();
+    //should(mockupio.id).be.undefined();
```
ballinette: why is this line commented ?

features/support/apiWebsocket.js
AnthonySendra: Oops. Fixed.

test/api/core/notifier/send.test.js
AnthonySendra: Because before, the implementation was wrong. Now it works and the mockupio.id must be defined.
A line is added for check 'not.bet.undefined()'

ballinette:
OK, +1 for this PR

PR #25, opened 2015-09-29
Added code coverage for the elasticsearch service

scottinet:
Added code coverage for the elasticsearch service:

* made the init() function return this in all cases
* made the import() function exit immediatly if the requestObject.body object is missing

Unrelated: fixed the disconnect() function in the websocket functional tests. 
Functional tests crashed during websocket tests, making cucumber think everything went fine.


AnthonySendra:
LGTM

ballinette:
+1

PR #28, opened 2015-10-02
Nodejs update v0.12 => v4.1.1

scottinet:
Removed the profiling service : it uses an old version of nodegrind because the later versions aren't compatible with nodejs 0.12+
And since the nodegrind project hasn't been maintained for almost a year, and since there aren't any other alternative (except for pm2), it has been decided to remove this service completly.


docker-compose.yml
```diff
@@ -1,5 +1,5 @@
 kuzzle:
-  image: kuzzleio/kuzzle
+  image: kuzzleio/dev
```
AnthonySendra: I think it's still `kuzzleio/kuzzle` https://hub.docker.com/r/kuzzleio/kuzzle/~/dockerfile/

docker-compose/test.yml
```diff
@@ -1,5 +1,5 @@
 kuzzle:
-  image: kuzzleio/kuzzle
+  image: kuzzleio/dev
```
AnthonySendra: I think it's still `kuzzleio/kuzzle` https://hub.docker.com/r/kuzzleio/kuzzle/~/dockerfile/

docker-compose.yml
scottinet: Good catch & fixed

AnthonySendra:
LGTM

j33f:
maybe specify the node engine version in package.json ?

ballinette:
+1

PR #30, opened 2015-10-06
Extract logger from Kuzzle into Plugin

AnthonySendra:
PluginsManager creation with the first plugin: [logger](https://github.com/kuzzleio/kuzzle-plugin-logger)
All references to `kuzzle.log` are switched to `kuzzle.pluginsManager.trigger` except for called made before plugin initialization or when an error occured and plugins are not initialized.

Switch from captains-log to [winston](https://github.com/winstonjs/winston) (because captains-log depend on winston)


lib/api/core/pluginsManager.js
```diff
@@ -0,0 +1,326 @@
...
+    // Add a fake function in first position for return the 'data' to other chained functions
+    preparedPipes.unshift(function (callback) { callback(null, data); });
+  }
+  else {
+    deferred.resolve(data);
+    return deferred.promise;
+  }
+
+  async.waterfall(preparedPipes, function (error, result) {
+    deferred.resolve(result);
```
scottinet: should reject if error is not null
The way I see it, a plugin shall call callback(error) when something goes wrong or simply if it does not validate the query. TriggerPipes' callers should also reject the query if the promise is rejected

scottinet: Proposal: to avoid badly coded plugins to freeze Kuzzle, maybe we should add a timeout and automatically reject the promise with an explicit error when it is reached?

lib/api/core/pluginsManager.js
```diff
@@ -0,0 +1,326 @@
...
+  _.forEach(plugins, function (plugin, name) {
+    if (plugin.url) {
+      pluginInstallId = plugin.url;
+    }
+    else if (plugin.version) {
+      pluginInstallId = name + '@' + plugin.version;
+    }
+    else {
+      this.kuzzle.log.error('Plugin', name, 'has no version. The version is mandatory if there is no URL.');
+      return true;
```
scottinet: Does the return value give the installation status of the plugin?
If so, is this normal to return ``true`` here?

AnthonySendra: We return true because we're in a loop and we want to pass to the next iteration. In lodash, if we return false in `forEach`, the loop is stoped.
The installation status is `newInstalled` and it's return after the loop.

lib/api/prepareDb.js
```diff
@@ -84,21 +86,23 @@ var importFixtures = function () {
...
+      'Here is the original error: ' + e.message
+    );
 
     deferred.resolve();
     return deferred.promise;
   }
 
   async.each(Object.keys(fixtures), function (collection, callback) {
-    this.log.info('== Importing fixtures for collection', collection, '...');
+    this.pluginsManager.trigger('== Importing fixtures for collection ' + collection + '...');
```
benoitvidis: missing first param?

AnthonySendra: Yep!

lib/api/core/pluginsManager.js
scottinet: Oops, didn't pay attention to the _.forEach :-)

lib/api/core/servers.js
```diff
@@ -18,16 +18,15 @@ module.exports = {
...
-
   kuzzle.router.initRouterHttp();
 
   var server = http.createServer(function (request, response) {
     kuzzle.router.routeHttp(request, response);
   });
-
   server.listen(port);
 
+  kuzzle.pluginsManager.trigger('server:httpStarted', 'Starting: HTTP server on port');
```
scottinet: nitpicking: missing ``port`` information

lib/services/internalbroker.js
```diff
@@ -236,12 +236,18 @@ function clientConnect() {
 
   this.client.socket.on('error', function (e) {
     this.close();
-    this.log.warn('Error while trying to connect to ws://' + this.host + ':' + this.port, ': ', e.message);
-    this.log.warn('==> RECONNECTING IN ', this.client.retryInterval, 'ms');
+
+    this.kuzzle.pluginsManager.trigger('internalBroker:error',
```
scottinet: Nitpicking: this isn't actually an error, but a warning.
I downgraded this message level some time ago because workers and servers start order is random, and I didn't want people to fret over an "error" raised whenever a worker starts before its server.
As a side note for a later time: we should probably try to detect this case and make this message silent when this happens.

scottinet:
+1

docs/plugins.md
```diff
@@ -0,0 +1,150 @@
...
+
+By default, the logger plugin is enabled and configured for using the service `winston` (refer to the kuzzle-plugin-logger documentation for more information).  
+
+# How to create a plugin
+
+A plugin is a Javascript module that can be installed with NPM or via a public GIT repository.
+
+## Configuration
+
+The module must have a `package.json` file with a `pluginInfo` entry. The optional `defaultConfig` will be copied in files `config/defaultPlugins.json` and `config/customPlugins.json` in Kuzzle.
```
ballinette: just to be sure.... the config/defaultPlugins.json and config/customPlugins.json are not gitinored (only config/plugins/ folder is).
So these generated files would be commited... is it what we want ?

AnthonySendra: The customPlugin.json file is gitignored (config/.gitignore).
If the user specify a plugin in customPlugin.json, the defaultConfig will be copied only in customPlugin.json.
If a plugin is defined in defaultPlugins.json (like logger), the defaultConfig will be copied only in defaultPlugin.json
If a plugin is defined in both files (i.e the user wants to override the service logger), the defaultConfig will be copied in both files.

ballinette: OK, understood. Did not see the config/.gitignore file, sorry.
Maybe we should reformulate this to make the idea more clear, but I don't know how to tell it. So let's keep as this for now ;)

ballinette:
Maybe we could precise in a MD file the list of triggers called by kuzzle  ? (plugins.md or a dedicated file ?)

ballinette:
Great job ;)
OK for me, my remark above can be treated later

AnthonySendra:
I've added an internal task for list all plugins events in the documentation.

PR #33, opened 2015-10-07
updated lint rules

scottinet:
* Updated jshint and eslint to accept ES6 features
* Sorted eslint rules in alphabetical order to make them easier to find
* Added new eslint rules aiming to catch potential errors and enforce our established coding rules
* Fixed new rules errors


ballinette:
looks good for me, but maybe need a stronger view from code developer ?

AnthonySendra:
I have a doubt about mandatory 'var' on top. Maybe we can just display a warning or something ? Sometimes it makes sens if the var is not on top.

```js
var foo = function (object) {
  if (!object) {
    return false;
  }

  var bar = [object];
  something(bar);

  return bar;
};
```

If we define `var bar` on top, we define it even if we don't use it letter, because of `return false`. What is your opinion about it ?

scottinet:
Well, that's what we defined in our CONTRIBUTING.md documentation file, so unless we add this rule to our exception list of the npm coding rules, I think that we should enforce it.
That said, coming from other languages, I don't quite grasp the importance of this rule. To me, it's a really unnecessary rule, as long as we have tests preventing hoistering abuses. And it's even more true since we should have block scoping with ES6 (well, as soon as Node implements it).

That said, the example you give can be easily solved by declaring the variable, performing the necessary checks, and then initializing the variable:

```js
function foo(object) {
  var bar;

  if (!object) {
    return false;
  }

  bar = [object];
  something(bar);

  return bar;
};
```

I'm not a fan about downgrading this check level to "warning" either, I prefer to reserve warnings for harmless rules like too many spaces, lonely ifs, undefined initialization, etc.

So my take on this is either we keep this rule, or we drop it altogether.

What do you think?

benoitvidis:
The idea of the rule is to reflect what node is doing and prevent
developers to think the bar variable would not be defined.
I agree this is not beautiful but it enforces consistency and clarity.

+1 to keep the rule.

PR #40, opened 2015-10-12
Make documentation up to date (small corrections on MD files)

ballinette:


benoitvidis:
+1 but to me, the config/perf/visualize.md doc would deserve a wording review.

ballinette:
yes, we should add this file to translation trello after PR validation ;)

PR #43, opened 2015-10-12
Update dsl specs

ballinette:
* update FilterTree schema according to latest change
*  use double quotes for json within MD files (just to have a good syntax coloration in github)


benoitvidis:
+1

PR #44, opened 2015-10-15
fixed random worker test timeout

scottinet:
Cause: kuzzle instances in unit tests are using a client-sided internal broker. The last workerNotifications test was using a broker.add on the core notifier room, already listened by the main Kuzzle server. The load balancer randomly attributed the test message to the Kuzzle main server instead of the test intance, causing the test to timeout randomly.


ballinette:
+1

AnthonySendra:
+1

PR #45, opened 2015-10-20
Statistics

scottinet:
* Added the "statistics" core component: now Kuzzle monitors ongoing/completed/failed requests by protocols, and active connections. Statistics are saved every 10s by default and are kept for 1hr
* Added ``getStats`` admin action, allowing to retrieve the last saved statistics frame
* Added ``getAllStats`` admin action, allowing to retrieve all saved statistics
* Refactored the way we use the Redis service to allow multiple cache databases
* Added new redis service capabilities to handle volatile keys
* Updated unit and functional tests
* Updated API documentations
* Fixed uses of captains-log in the bin/kuzzle* scripts
* Fixed jshint grunt configuration to use the .jshintrc file instead of the defautl configuration


test/api/controllers/routerController/routeMQListener.test.js
```diff
@@ -6,6 +6,7 @@
 var
   should = require('should'),
   winston = require('winston'),
+  q = require('q'),
   params = require('rc')('kuzzle'),
   Kuzzle = require('root-require')('lib/api/Kuzzle'),
```
benoitvidis: The root-require seems to be missing in the package.json.

However, this module is marked as test only and can be replaced quite easilly using the built-in `require.main.require('./lib/api/Kuzzle')`

scottinet: root-require is in the devDependencies part of the package.json file, as it is only used for unit testing.

However it seems that we can indeed remove this by using require.main. I'll prepare another pull requests.

features/AMQP.feature
```diff
@@ -136,3 +136,14 @@ Feature: Test AMQP API
...
     Then I unsubscribe
     And I should receive a "off" notification
+
+  @usingAMQP
+  Scenario: Getting the last statistics frame
+    When I get the last statistics frame
+    Then I get at least 1 statistic frame
+
+  @usingAMQP
+  Scenario: Getting the last statistics frame
```
ballinette: It should be "Getting all statistics frame", no ?

features/MQTT.feature
```diff
@@ -135,3 +135,14 @@ Feature: Test MQTT API
...
     Then I unsubscribe
     And I should receive a "off" notification
+
+  @usingMQTT
+  Scenario: Getting the last statistics frame
+    When I get the last statistics frame
+    Then I get at least 1 statistic frame
+
+  @usingMQTT
+  Scenario: Getting the last statistics frame
```
ballinette: same remark as above

features/REST.feature
```diff
@@ -66,3 +66,14 @@ Feature: Test REST API
...
     When I write the document "documentGrace"
     Then I find a document with "Grace" in field "firstName"
+
+  @usingREST
+  Scenario: Getting the last statistics frame
+    When I get the last statistics frame
+    Then I get at least 1 statistic frame
+
+  @usingREST
+  Scenario: Getting the last statistics frame
```
ballinette: same remark as above

features/STOMP.feature
```diff
@@ -136,3 +136,14 @@ Feature: Test STOMP API
...
     Then I unsubscribe
     And I should receive a "off" notification
+
+  @usingSTOMP
+  Scenario: Getting the last statistics frame
+    When I get the last statistics frame
+    Then I get at least 1 statistic frame
+
+  @usingSTOMP
+  Scenario: Getting the last statistics frame
```
ballinette: same remark as above

features/Websocket.feature
```diff
@@ -135,3 +135,14 @@ Feature: Test websocket API
...
     Then I unsubscribe socket "client1"
     And I should receive a "off" notification
+
+  @usingWebsocket
+  Scenario: Getting the last statistics frame
+    When I get the last statistics frame
+    Then I get at least 1 statistic frame
+
+  @usingWebsocket
+  Scenario: Getting the last statistics frame
```
ballinette: same remark as above

docs/API.REST.md
```diff
@@ -697,3 +699,122 @@ Bulk import only works on documents in our persistent data storage layer.
...
+  result: {
+    count: <number of found documents>
+    _source: {                      // Your original count query
+      ...
+    },
+    collection: '<data collection>',
+    action: 'getStats',
+    controller: 'admin',
+    statistics: {
+      "YYYY-MM-DDTHH:mm:ss.mmmZ": {
```
AnthonySendra: I'm not sure about the date is the key for the nested object. That force in js to make something like `Object.keys()` for retrieve the date and for access to the embed object. Or maybe I miss something ?

scottinet: This was the easiest way I found to be able to easily sort frames when getting multiple statistics frames. And since I decided that getStats and getAllStats should return the same answer structure, here it is.

Another way to format these responses could be to return an array of statistics:

```
{
  error: null,
  result: {
    ...
  statistics: [
    {"timestamp": "YYYY-MM-DDTHH:mm:ss.mmm",   ... statistics informations ...},
    {"timestamp": "YYYY-MM-DDTHH:mm:ss.mmm",   ... statistics informations ...},
    {"timestamp": "YYYY-MM-DDTHH:mm:ss.mmm",   ... statistics informations ...},
    ...
  ],
```

ballinette:
OK for me.

docs/API.REST.md
scottinet: One last possibility would be to leave the response for ``getAllStats`` as it is, and to reformat the one for ``getStats`` like this:

```
{
  error: null,
  result: {
    ...
    statistics: {
      timestamp: "YYYY-MM-DDTHH:mm:ss.mmm",
      ... statistics informations ...
    }
}
```

This would allow reading the result of this request without resorting to ``Object.keys``, as it seems a bit superfluous for only one key.

PR #49, opened 2015-10-26
Check rest request

scottinet:
When transmitting a REST query with an invalid content encoding, the user receive a "The body can't be empty" response, instead of a more explicit error. This PR fixes that.


ballinette:
+1
PS  : guys, I'd like tihs PR to be quickly merged, because I need to include this new error case in the error management refactor... ;)

j33f:
+1

PR #50, opened 2015-10-26
KUZ-194 error codes

ballinette:
Re-factoring of response JSON  format, with a "HTTP-like" status code, and a complex error object with message, error count and a stack for "internal server errors".

* Update ResponseObject and its constructor to set the "status" attribute.
* Create custom Kuzzle Error protypes
* Update documentation and unit tests to match the new response format
* Add unit test for good status code initialization while creating a ResponeObject.

NB: still to do (will be done in another PR): handle "partial success" response (for bulk requests with some errors)


scottinet:
+1, seems great!

benoitvidis:
+1

PR #52, opened 2015-10-27
Add status code for partial error on bulk actions.

ballinette:


docs/status-codes.md
```diff
@@ -32,6 +32,8 @@ It is used to inform the client about the real status of his request (if an erro
 #### 2xx Success
 
 * ``200``: standard status for a successful request.
+* ``206``: the request (tipically a bulk import) is partially successful, but some actions encountered an error.
+(in this case, details of errors are returned within _error.stack_)
```
scottinet: When I see the "stack" word in an error context, I can only think of stack traces.
Maybe we should find a more suitable word for the error list, like "errors" or something like that.

benoitvidis:
+1

docs/status-codes.md
ballinette: good remark.

changed into "error.errors" ;)

scottinet:
+1

PR #54, opened 2015-10-29
reinit response buffer between each test scenario

scottinet:
In functional tests, checking if a notification is received is done by reading the response buffer of the tested API.
Problem is, this buffer is not reinitialized between test scenarios, meaning that if a notification takes more time than usual to reach the test client (we're speaking of a few milliseconds), then the previous response is read instead, leading to a false "wrong notification received" error.

This PR aims to correct this behavior by reinitializing the response buffer between each test scenario.

### DIFF
--- a/docs/API.REST.md
+++ b/docs/API.REST.md
@@ -724,7 +724,6 @@ These statistics include:
 {
   error: null,                      // Assuming everything went well
   result: {
-    count: <number of found documents>
     _source: {                      // Your original count query
       ...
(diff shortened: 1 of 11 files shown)


scottinet:
It seems that I got that wrong. Do not merge this pull request yet.

scottinet:
Doing some tests, trying to figure out why STOMP functional tests randomly fail... and only in Travis. o_O

PR #57, opened 2015-10-30
Add ids filter subscription

AnthonySendra:
Add realtime ids filter

```js
kuzzle
  .dataCollectionFactory('user')
  .subscribe({
    ids: {
      values: ['myDocumentId']
    }
  })

### CODE REVIEW

File: lib/api/core/models/requestObject.js [hunk 1]
```diff
@@ -35,10 +35,13 @@ function construct (object, additionalData, protocol) {
   if (object.body && object.body._id !== undefined) {
```
scottinet: This means that we take the first found document ID, in this order: document body => document headers => additional data

I wonder if this shouldn't be the other way around: first the _id present in additional data, if any, then the one in the headers, and lastly, if no other _id was found, the one in the document body.

What do you think?

benoitvidis:
+1

jenow:
+1

scottinet:
+1

PR #59, opened 2015-11-02
fix notification not sent when using createOrUpdate

scottinet:
* added createOrUpdate action handling in the worker notification manager
* added functional tests aiming to catch this bug should he appears again


AnthonySendra:
+1

ballinette:
+1

PR #62, opened 2015-11-04
[BREAKING CHANGE] subscription refactoring

scottinet:
* removed the ``room name`` notion in the subscribe controller. It was hard to comprehend for newcomers, as ``countSubscription`` and ``subscribe`` use the Room ID, and ``unsubscribe`` use the Room Name, for no apparent reason (from a user point of view). Now, there is only the room ID. Moreover, the ``room name`` was taken from the subscription ``requestId``, making it even harder to understand.
* requests now allow metadata. Metadata are volatile data, ignored by Kuzzle, but forwarded to responses and notifications. This allow users to pass some useful information to other users when performing queries

:warning: After this PR is merged to the master branch, expect ``Kuzzle Tournament`` demo to stop function. This is due to an ugly trick using the room name to know which user a tournament game client should stop render after a disconnection. This will be fixed once subscription metadata are implemented.


docs/API.AMQP.md
```diff
@@ -290,7 +333,9 @@ It works with the room unique ID Kuzzle returns to you when you make a subscript
 ```javascript
 {
   /*
-  Optional
+    Required. If your query doesn't include a requestId field, Kuzzle
```
ballinette: clientId instead of requestId, I guess ? ;)

anyway, is it really required, as we already have a "replyTo" header in the message that should be used to send back the query response ?

lib/api/core/hotelClerk.js
```diff
@@ -204,29 +198,43 @@ module.exports = function HotelClerk (kuzzle) {
...
+    roomId;
+
+  /*
+   Ensure the resulting room id is identical even if the order of filters terms is changed.
+   This should do the trick for now, but a deep sort() would probably be better.
+    */
+  if (filters) {
+    Object.keys(filters).forEach(key => sortedFilters.push([key, filters[key]]));
+    stringifiedFilters = stringify(sortedFilters.sort((a, b) => {
+      return a[1] - b[1];
```
scottinet: ewww

Don't mind this, I'm fixing it

docs/API.AMQP.md
scottinet: it's indead optional

benoitvidis:
+1

docs/API.STOMP.md
```diff
@@ -286,7 +330,9 @@ It works with the room unique ID Kuzzle returns to you when you make a subscript
 ```javascript
 {
   /*
-  Optionnal
+    Required. If your query doesn't include a clientId field, Kuzzle
```
ballinette: Same remark as for AMQP: clientId is optionnal, here.

ballinette:
+1

docs/API.AMQP.md
```diff
@@ -109,7 +156,7 @@ The matching criteria you pass on to Kuzzle are [filters](./filters.md).
 
 How subscription works:  
 :arrow_right: You send a subscription query to Kuzzle  
-:arrow_left: Kuzzle responds to you with a room name and a room unique ID  
+:arrow_left: Kuzzle responds to you with a room unique ID  
 :arrow_right: You subscribe to the queue you provided in the ``replyTo`` metadata
 :arrow_left: When a document matches your room criteria, Kuzzle sends you a ``response``
```
AnthonySendra: Nitpick in this file: missing newline

PR #65, opened 2015-11-06
KUZ-300 No answer on non-persitent REST write call

benoitvidis:
Fixes the case where the following query would not return any anwser.

    curl -H "Content-Type: application/json" -d '{"persist":false, "body":{"t":1}}' -D- "http://localhost:7512/api/test"

### DIFF
--- a/test/api/controllers/routerController/executeFromRest.test.js
+++ b/test/api/controllers/routerController/executeFromRest.test.js
@@ -11,13 +11,14 @@ var
   Kuzzle = require.main.require('lib/api/Kuzzle'),
   rewire = require('rewire'),
   RouterController = rewire('../../../../lib/api/controllers/routerController'),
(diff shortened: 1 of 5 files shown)


test/api/controllers/routerController/executeFromRest.test.js
```diff
@@ -137,14 +138,63 @@ describe('Test: routerController.executeFromRest', function () {
       catch (e) {
         done(e);
       }
-    }, 10);
+    }, 5);
 
-    setTimeout(function () {
-      if (timer === false) {
+    setTimeout(() => {
```
scottinet: This can be done more simply, by using the should method "this.timeout(<timeout in ms>)"

If the done() callback is not called before the timeout is reached, the test fails

scottinet:
+1

ballinette:
+1

PR #66, opened 2015-11-09
Subscription metadata

scottinet:
This pull request implements the metadata in subscriptions feature.

It works by providing metadata in the subscription request. These metadata are forwarded to users who have subscribed to the same room.

It is then stored in the customer's information, and forwarded to other users when this customer leaves the room.


benoitvidis:
+1

ballinette:
+1

PR #69, opened 2015-11-10
fix filter ids with value and Dockerfile

AnthonySendra:
* Fix dockerfile for add the code in container
* Fix ids filter with value in currified function name
* Fix parameters in pluginsManager.trigger function in internalbroker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
 FROM kuzzleio/dev
 MAINTAINER Kuzzle <support@kuzzle.io>
 
-# resolves onbuild triggers into the final image
+ADD ./ /var/app/
+
+RUN npm install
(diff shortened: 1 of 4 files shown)


scottinet:
+1

Dockerfile
```diff
@@ -1,4 +1,6 @@
 FROM kuzzleio/dev
 MAINTAINER Kuzzle <support@kuzzle.io>
 
-# resolves onbuild triggers into the final image
```
benoitvidis: If we explicitly add the current directory on build, the ONBUILD command becomes useless on the dev container. I suppose it should be removed from it to avoid cloning the repository.

benoitvidis:
+1

Dockerfile
scottinet: Already planned, but we need this PR to be merged first. :-)

ballinette:
+1

PR #72, opened 2015-11-16
Normalize realtime responses

scottinet:
Realtime response objects are expected to be different than regular response objects, mainly because of the 'roomId' key field.

But these responses weren't normalized, mainly when counting subscriptions:
```json
 {
  "error": null,
  "status": 200,
  "response": <number>
}
```

No roomId, and an integer ``result`` instead of an object like all other responses. This poses a problem when coding SDK in languages more strictly typed than JS, forcing developers to handle subscription count response differently than all other responses.

This PR fixes this by normalizing realtime responses:

* all RT responses return a ``roomId`` field 
* all RT responses are formatted the same way, with ``result`` being an object
* RT responses now return a ``requestId`` and a request ``timestamp``, like with regular responses
* RT responses now return the following technical fields: ``controller``, ``action``, ``protocol`` and ``metadata``

### DIFF
--- a/docs/API.AMQP.md
+++ b/docs/API.AMQP.md
@@ -233,7 +233,11 @@ How subscription works:
   status: 200,                      // Assuming everything went well
   error: null,                      // Assuming everything went well
   result: {
-    roomId: 'unique Kuzzle room ID'
+    roomId: 'unique Kuzzle room ID',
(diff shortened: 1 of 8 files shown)


ballinette:
+1

jenow:
+1

PR #75, opened 2015-11-18
added createCollection and truncate operations

scottinet:
* added the createCollection API method: create a new empty collection 
* added the truncate API method: truncate an existing collection
* added unit tests and functional tests
* updated API documentation
* simplified the initialization of REST routes, eliminating a lot of repetition of code
* bugfix: errors from Elasticsearch coming from workers weren't correctly delivered to the users because of a formatting problem

PR #76, opened 2015-11-18
Statistics

jenow:
Added getLastStat, getStats now takes a date as a parameter and return all stats after it

### DIFF
--- a/docs/API.AMQP.md
+++ b/docs/API.AMQP.md
@@ -1230,7 +1230,7 @@ These statistics include:
 * the number of completed requests since the last frame
 * the number of failed requests since the last frame
 
-**Topic:** ``admin.getLastStat``
+**Topic:** ``admin.getLastStats``
 **replyTo queue header:** Required.
 
 **Query:**
@@ -1238,7 +1238,7 @@ These statistics include:
 ```javascript
 {
   controller: 'admin',
-  action: 'getStats'
+  action: 'getLastStats'
 }
 ```
 
@@ -1250,7 +1250,7 @@ These statistics include:
   error: null,                      // Assuming everything went well
   result: {
     _source: {}                     // Your original query
-    action: 'getLastStat',
+    action: 'getLastStats',
     controller: 'admin',
     statistics: {
       "YYYY-MM-DDTHH:mm:ss.mmmZ": {
@@ -1283,7 +1283,7 @@ These statistics include:
 
 ### Getting the statistics from a date
 
-Kuzzle monitors its internal activities and make snapshots of them. This command allows getting the last stored statistics frame from a date.
+This command allows getting statistics frames saved/stored after a provided timestamp
 
 These statistics include:
 
@@ -1313,8 +1313,12 @@ These statistics include:
   */
   requestId: <Unique query ID>
 
+  /*
+  Optional: Kuzzle will return all statistics if nor the startTime and stopTime are defined
+  */
   body: {
-    since: 424242424242
+    startTime: <timestamp>,
+    stopTime: <timestamp>
   }
 }
 ```
@@ -1326,7 +1330,7 @@ These statistics include:
   status: 200,                      // Assuming everything went well
   error: null,                      // Assuming everything went well
   result: {
-    _source: { since: 424242424242 }                     // Your original query
+    _source: { startTime: <timestamp>, stopTime: <timestamp> }                     // Your original query
     action: 'getStats',
     controller: 'admin',
     statistics: {
--- a/docs/API.AMQP.md
+++ b/docs/API.AMQP.md
@@ -1312,7 +1312,7 @@ These statistics include:
   requestId: <Unique query ID>
 
   body: {
-    timestamp: 424242424242
+    since: 424242424242
   }
 }
 ```
@@ -1324,7 +1324,7 @@ These statistics include:
   status: 200,                      // Assuming everything went well
   error: null,                      // Assuming everything went well
   result: {
-    _source: { timestamp: 424242424242 }                     // Your original query
+    _source: { since: 424242424242 }                     // Your original query
     action: 'getStats',
     controller: 'admin',
     statistics: {
--- a/docs/API.AMQP.md
+++ b/docs/API.AMQP.md
@@ -35,6 +35,7 @@ The current implementation of our MQ Broker service uses [RabbitMQ](https://www.
   * [Performing a bulk import](#performing-a-bulk-import-on-a-data-collection)
   * [Performing a global bulk import](#performing-a-global-bulk-import)
   * [Getting the last statistics frame](#getting-the-last-statistics-frame)
+  * [Getting the statistics from a date](#getting-the-statistics-from-a-date)
   * [Getting all stored statistics](#getting-all-stored-statistics)
   * [Listing all known data collections](#listing-all-known-data-collections)
   * [Getting the current Kuzzle timestamp](#getting-the-current-kuzzle-timestamp)
@@ -1337,7 +1338,7 @@ These statistics include:
 * the number of completed requests since the last frame
 * the number of failed requests since the last frame
 
-**Topic:** ``admin..getStats``
+**Topic:** ``admin.getLastStat``
 
 **Exchange name:** ``amq.topic``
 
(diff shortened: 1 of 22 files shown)


lib/api/controllers/routerController.js
```diff
@@ -54,7 +54,7 @@ module.exports = function RouterController (kuzzle) {
       executeFromRest.call(kuzzle, params, request, response);
     });
 
-    api.get('/_getStats', function (request, response) {
+    api.post('/_getStats', function (request, response) {
```
ballinette: I'd better keep a GET method and add the timestamp as a URI parameters :
api.get('/_getStats/:timestamp', ....

So we can get the stats from a timestamp with the request :
GET http://kuzzle:xxxx/api/_getStats/12341564897

jenow: This would mean we want a stat with that exact timestamp but the purpose of this is to get the stats SINCE that timestamp so i thought it's better to use POST

features/AMQP.feature
```diff
@@ -154,6 +154,17 @@ Feature: Test AMQP API
     Then I get at least 1 statistic frame
 
   @usingAMQP
+  Scenario: Getting the statistics frame from a date
+    When I get the statistics frame from a date
+    Then I get at least 1 statistic frame
+
+  @usingAMQP
+  Scenario: Getting the statistics frame from a date when there is no stats in cache
```
ballinette: Not sure to understand the meaning of this test ?

lib/api/controllers/routerController.js
ballinette: OK, so, we should use another parameter name as "timestamp" within the body ;)

ballinette:
Note: the code within routerController will need to be refactored after merging PR #75 ;)

jenow:
I have removed the test (that was a mistake) and changed the 'timestamp' parameter as 'since'

lib/api/controllers/routerController.js
scottinet: With Kevin we talked about this yesterday, and we reached the conclusion that POST was the best verb to use... but we didn't think about changing the argument name nor the route name.

The problem with a ``getStatsSince`` route name is that it returns the last known frame if no argument is provided, and this is not coherent with what the route name suggests.  
So if we go for this solution, we might have to change the route behavior too, I think.

However, changing the argument name from ``timestamp`` to something like ``startTime`` would be easier to understand, so I think this is a good suggestion. This do not solve the GET vs POST debate, though.

docs/API.AMQP.md
```diff
@@ -1227,6 +1228,7 @@ These statistics include:
 * the number of completed requests since the last frame
 * the number of failed requests since the last frame
 
+**Topic:** ``admin.getLastStat``
```
scottinet: should be ``getLastStats``, as it returns the last statistics frame (one frame containing multiple statistics)

docs/API.AMQP.md
```diff
@@ -1245,9 +1247,84 @@ These statistics include:
...
+    requestId: '<unique request identifier>'
+  }
+}
+```
+
+---
+
+### Getting the statistics from a date
+
+Kuzzle monitors its internal activities and make snapshots of them. This command allows getting the last stored statistics frame from a date.
```
scottinet: should be "This command allows getting statistics frames saved after a provided date (timestamp?)"

lib/api/controllers/adminController.js
```diff
@@ -57,6 +57,17 @@ module.exports = function AdminController (kuzzle) {
    * @param {RequestObject} requestObject
    * @returns {Promise} promise resolved as a ResponseObject
    */
+  this.getLastStat = function (requestObject) {
+    kuzzle.pluginsManager.trigger('data:getLastStat', requestObject);
+    return kuzzle.statistics.getLastStat(requestObject);
+  };
+
+  /**
+   * Returns the last statistics frame
```
scottinet: should be "Returns all stored statistics frames"

lib/api/core/statistics.js
```diff
@@ -145,6 +145,55 @@ module.exports = function (kuzzle) {
   };
 
   /**
+   * Gets the lasts saved statistics frame from a date
```
scottinet: should be "Gets stored statistics frames from a date"

lib/api/core/statistics.js
```diff
@@ -145,6 +145,55 @@ module.exports = function (kuzzle) {
...
+      since;
+
+    currentDate = new Date().getTime();
+    since = requestObject.data.body.since;
+    if (since > currentDate) {
+      deferred.resolve(new ResponseObject(requestObject, response));
+      return deferred.promise;
+    }
+    if (!this.lastFrame) {
+      if (!since || currentDate >= new Date(since).getTime()) {
```
scottinet: the second condition is always ``false`` as it has already been evaluated line 165

scottinet: the second condition is always ``true`` because we catched the other cases line 165

lib/api/core/statistics.js
```diff
@@ -145,6 +145,55 @@ module.exports = function (kuzzle) {
...
+
+    this.kuzzle.services.list.statsCache.getAllKeys()
+      .then(keys => {
+        frames = keys;
+        // Statistics keys are timestamp. Ordering them guarantees stats frames to be returned in the right order
+        return this.kuzzle.services.list.statsCache.mget(keys.sort());
+      })
+      .then(values => {
+        values.forEach((v, idx) => {
+          if (!since || (new Date(since).getTime()) <= (new Date(Number(frames[idx])).getTime())) {
```
scottinet: ``new Date(since).getTime()`` should only be calculated once instead of at each iteration.

Moreover, since this argument is provided by users, it should also be tested against the NaN value (the result of a problem while trying to converting the argument to a Epoch value) to return an empty stats result instead of all results like it does in this current implementation.

I think that putting this test line 165 along with the one testing against a timestamp in the future would be the right place.

lib/api/core/statistics.js
```diff
@@ -145,6 +145,55 @@ module.exports = function (kuzzle) {
...
+            response.statistics[(new Date(Number(frames[idx]))).toISOString()] = JSON.parse(v);
```
scottinet: ``new Date(Number(frames[idx])`` should only be calculated once

lib/api/core/statistics.js
```diff
@@ -145,6 +145,55 @@ module.exports = function (kuzzle) {
...
+        statistics: {}
+      },
+      frames,
+      currentDate,
+      since;
+
+    currentDate = new Date().getTime();
+    since = requestObject.data.body.since;
+    if (since > currentDate) {
+      deferred.resolve(new ResponseObject(requestObject, response));
```
scottinet: Since this is the response in the case of an incorrect value, maybe it should reject the promise with a ``BadRequestError``?

lib/api/controllers/routerController.js
ballinette: ihmo, GET is the better way to retrieve content when there is just one parameter, it's more HTTP-logic compliant (GET for fetching content, POST for sending content or executing an action).
But... for more complex queries (like for an advanced search, for example), it's not possible to do it properly with GET so we have to do a POST anyway..

So, for the getStats query, it would be OK for me to do a POST if we can have more than one filter parameters.... for example, it could be interesting to have, at least:
* startTime (required, or just act like a getLastStats if not given).
* stopTime (optional, use "now" by default).

What do you think about this suggestion ?

lib/api/core/statistics.js
jenow: I don't think it is an incorrect argument, we only assume there is no statistics in the future therefore it returns an empty array and do not throw an error as if it was a malformatted date

lib/api/controllers/routerController.js
scottinet: I like it

jenow: Alright, not bad.
I prefer to return all statistics if startTime is not given though (act as if the startTime equals Jan 1st 1970)

lib/api/core/statistics.js
ballinette: quite agree with @jenow 
BadRequestError are for malformed queries (for example, missing a required paramereter, or wrong format).
Here, the request is well formed, but we just have no content to retrieve.

lib/api/controllers/routerController.js
ballinette: also OK for a "1970-01-01" default value for starTime... just make it clear in the doc ;)

so.... finally, getAllStats would be an alias for getStats with no parameters ?

jenow: In that case yes

lib/api/core/statistics.js
```diff
@@ -145,23 +146,41 @@ module.exports = function (kuzzle) {
...
+    }
+    if (startTime >= currentDate) {
+      deferred.resolve(new ResponseObject(requestObject, response));
+      return deferred.promise;
+    }
     if (!this.lastFrame) {
-      // We return at least the current statistics frame
-      response.statistics[(new Date()).toISOString()] = this.currentStats;
-
+      if (!stopTime || stopTime >= currentDate) {
```
ballinette: I guess we should here append the currentStats, and not get back only this...
So remove the deffered.resolve() just behind ?

jenow: Which on exactly ? I'm not sure to understand what you mean

ballinette: forget it, I dit not notice that if this.lastFrame is null, it means that we have only the currentStats to return.

So, OK for me ;)

ballinette:
+1

lib/api/core/statistics.js
```diff
@@ -145,23 +146,41 @@ module.exports = function (kuzzle) {
...
+
+    currentDate = new Date().getTime();
+    if (requestObject.data.body.startTime !== undefined) {
+      startTime = new Date(requestObject.data.body.startTime).getTime();
+    }
+    if (requestObject.data.body.stopTime !== undefined) {
+      stopTime = new Date(requestObject.data.body.stopTime).getTime();
+    }
+    if ((startTime !== undefined && isNaN(startTime)) || (stopTime !== undefined && isNaN(stopTime))) {
+      deferred.reject(new ResponseObject(requestObject, new InternalError('Invalid time value')));
```
scottinet: shouldn't this be a BadRequestError instead of an InternalError?

scottinet:
apart from the error type returned when an incorrect start/stop time is provided: +1 :-)

lib/api/core/statistics.js
ballinette: yes, indeed.

jenow:
I've changed it to BadRequestError

PR #81, opened 2015-11-23
Geodistance

j33f:


lib/api/dsl/methods.js
```diff
@@ -7,8 +7,202 @@ var
...
+  /**
+   * Generate a valid usable distance
+   *
+   * @param {String} the given distance
+   * @return {Object} the distance in meters
+   */
+  getDistance: function(distance) {
+    var tmp;
+    if (_.isString(distance)) {
+      tmp = distance.replace(/^([0-9., ]*)([a-z])/, '$1 $2').replace(/ /,'').replace(/,/,'.').replace(/-/,'').toLowerCase();
```
benoitvidis: 1. As-is, the `replace(/-/, '')` looks undeeded to me (not allowed in the first regex).
2. I suppose the first regex should be `/^([0-9., ]+)([a-z]+)/i` to allow uppercase units and let the user enter some multi-char units (i.e. km).

lib/api/dsl/methods.js
```diff
@@ -7,8 +7,202 @@ var
   q = require('q'),
   util = require('util'),
   geohash = require('ngeohash'),
+  units = require('node-units'),
```
scottinet: This library has not been updated for 3 years, and it uses an even more outdated version of ``numerizer`` as a dependency.

[convert-units](https://www.npmjs.com/package/convert-units) seems more up-to-date.

lib/api/dsl/methods.js
```diff
@@ -551,11 +665,89 @@ module.exports = methods = {
...
+      deferred.reject(new BadRequestError('Missing filter'));
+      return deferred.promise;
+    }
+
+    fieldName = Object.keys(filter)[0];
+    if (fieldName === 'distance') {
+      fieldName = Object.keys(filter)[1];
+    }
+
+    geoFilter = filter[fieldName];
```
scottinet: If the filter contains only a ``distance`` member and nothing else, then ``fieldName`` is undefined. Since this argument comes from users, this test must be tested.

lib/api/dsl/methods.js
```diff
@@ -551,11 +665,89 @@ module.exports = methods = {
...
+    // we need an exception for the linter
+    /* jshint camelcase: false */
+    if (geoFilter.lat_lon) {
+      geoFilter.latLon = geoFilter.lat_lon;
+      delete geoFilter.lat_lon;
+    }
+    /* jshint camelcase: true */
+
+    try {
+      if (!filter.distance) {
```
scottinet: shouldn't this be tested before initializing ``fieldName``?

lib/api/dsl/methods.js
```diff
@@ -597,7 +789,9 @@ module.exports = methods = {
 
     deferred.reject(new KuzzleError('regexp is not implemented yet.'));
     return deferred.promise;
-  }
+  },
+
+  _geoUtil: geoUtil // in order to allow tests
```
scottinet: Use ``rewire`` to test "private" methods or variables inside a module, instead of exposing them needlessly.

lib/api/dsl/methods.js
j33f: yep but since the units conversions dont use to move that much... is it really important ?

scottinet: Nope, as long as it doesn't use deprecated methods.

I was just curious about this particular library and found convert-units, which seemed more recent and with what seemed to be a cleaner API. But this is a minor "issue".

scottinet:
+1

ballinette:
+1

PR #82, opened 2015-11-24
+geoDistanceRange DSL function 

j33f:
Adds geoDistanceRange DSL function 
+ related tests and doc +change Kuzzle version in package.json


lib/api/dsl/methods.js
```diff
@@ -751,11 +747,93 @@ module.exports = methods = {
...
+    fieldName = Object.keys(filter)[0];
+    if (fieldName === 'from' || fieldName === 'to') {
+      fieldName = Object.keys(filter)[1];
+    }
+    if (fieldName === 'from' || fieldName === 'to') {
+      fieldName = Object.keys(filter)[2];
+    }
 
-    deferred.reject(new KuzzleError('geoDistanceRange is not implemented yet.'));
+    geoFilter = filter[fieldName];
```
scottinet: As it is, ``fieldName`` may be undefined, leading to a ``TypeError`` when initializing ``geoFilter``.

For instance with this filter:

```json
{ "from": "foobar" }
```

j33f: yep! corrected

ballinette:
+1

benoitvidis:
+1

PR #84, opened 2015-11-25
Hotel clerk bugfix

scottinet:
There is a race condition in the ``hotelClerk`` core component, occuring when a room's destruction is underway and a new subscription on that room arrives.

What happens is that, before destroying a room, there is a deep clean up in the DSL to remove any associated filter. During that delay, when a new subscription arrives, it uses the associated room object, as it already exists.  
When the DSL cleaning finishes, the room is destroyed, resulting in unpredictable results.

This PR solves this race condition using the ``destroyed`` flag, already used to solve a previous race condition (when multiple almost simultaneous ``unsubscribe`` occur).  
If a room is marked for destruction, the room creation is delayed until the destruction is effective.


ballinette:
+1

benoitvidis:
+1

stafyniaksacha:
+1

PR #85, opened 2015-11-26
Plugin controllers

ballinette:
Implement plugin controllers

a sample plugin is proposed here : https://github.com/kuzzleio/kuzzle-plugin-helloworld

PS : please do not comment on English language level, the doc file will be reviewed and corrected by Julie then ;)


docs/plugins.md
```diff
@@ -138,10 +139,117 @@ module.exports = function () {
...
+module.exports = [
+  {verb: 'get', url: '/foo/:name', controller: 'mycontroller', action: 'myAction'},
+  {verb: 'post', url: '/foo', controller: 'mycontroller', action: 'myAction'},
+];
+```
+
+_NB: you can describe any routes as you want, according to the actions you need to implement.<br>
+For each action, you can declare either a GET action, or a POST action, or both of them._
+
+* the controller code, implementing your actions.
```
scottinet: may we add something along the line "Action methods must return a promise"?

lib/api/core/pluginsContext.js
```diff
@@ -0,0 +1,13 @@
...
+  RealTimeResponseObject: require('./models/realTimeResponseObject'),
+
+  KuzzleError: require('./errors/kuzzleError'),
+  BadRequestError: require('./errors/badRequestError'),
+  ForbiddenError: require('./errors/forbiddenError'),
+  InternalError: require('./errors/internalError'),
+  NotFoundError: require('./errors/notFoundError'),
+  PartialError: require('./errors/partialError'),
+  ServiceUnavailableError: require('./errors/serviceUnavailableError'),
+  UnauthorizedError: require('./errors/unauthorizedError')
```
scottinet: As new errors get added regularly, it may be best to load them dynamically, by requiring all javascript files in the errors/ folder for instance. Or by using an index.js file.

What do you think?

scottinet: As new errors get added regularly, it may be best to load them dynamically, by requiring all javascript files in the errors/ folder for instance. Or by using an index.js file.

What do you think?

ballinette: yes, good idea.

so we could have someting like :
module.exports = {
  ResponseObject: require('./models/responseObject'),
  RealTimeResponseObject: require('./models/realTimeResponseObject'),
  errors: require('./errors')
};

and then, in the plugin controller, create an error  with :
var error = new context.errors.InternalError();

is it what you mean ?

scottinet: Exactly. But this solution requires an index.js file, which need to be manually updated when a new error is updated. This may not be a problem though.

I was also wondering if we could simply list all .js files in the errors folder and require them.

scottinet:
+1, looks good!

docs/plugins.md
ballinette: added

benoitvidis:
+1

lib/api/core/pluginsContext.js
ballinette: OK, just trying something... ;)

ballinette:
@scottinet  : I've refactored the plugin context constructor according to your suggestion (see lattest commit: https://github.com/kuzzleio/kuzzle/commit/1d2d99f3ae64ce69c3e6bd81963afc5713b4910e )
what do you think about it ?

scottinet:
I like it! :-)

PR #89, opened 2015-11-27
Geo polygon

j33f:
- adds the geoPolygon filter
- modify the dependency to geolib to a custom version of geolib with better 'isPointInside' functions set
- some refactor to follow the new way of calculating and returned value of the new "isPointInside' functions

### DIFF
--- a/lib/api/dsl/methods.js
+++ b/lib/api/dsl/methods.js
@@ -109,6 +109,70 @@ var geoUtil = {
   },
 
   /**
+   * Construct a valid usable BBox
+   *
+   * @param {Object} the given object
+   * @return {Object} the valid usable BBox object
+   */
+  constructPolygon: function(geoFilter) {
+    var i,
+      tmp,
+      point,
+      entry,
+      polygon = [];
+
+    if (geoFilter.points === undefined) {
+      throw new BadRequestError('No point list found');
+    }
+
+    if (!_.isArray(geoFilter.points)) {
+      throw new BadRequestError('A polygon must be in array format');
+    }
+
+    if (geoFilter.points.length < 3) {
+      throw new BadRequestError('A polygon must have at least 3 points');
+    }    
+
+    for (i in geoFilter.points) {
+      entry = geoFilter.points[i];
+
+      if (_.isArray(entry)) {
+        if (entry.length >= 2) {
+          point = {lat: entry[0], lon: entry[1]};
+        } 
+        else {
+          throw new BadRequestError('Unable to parse some point coordinates into polygon');
+        }
+      }
+      else if (_.isObject(entry)) {
+        if (entry.lat !== undefined && entry.lon !== undefined) {
+          point = {lat: entry.lat, lon: entry.lon};
+        } 
+        else {
+          throw new BadRequestError('Unable to parse some point coordinates into polygon');
+        }
+      }
+      else if(_.isString(entry)) {
+        if (/^[-.0-9]+,\s*[-.0-9]+$/.test(entry)) {
+          tmp = entry.match(/^([-.0-9]+),\s*([-.0-9]+)$/);
+          point = {lat: tmp[1], lon: tmp[2]};
+        }
+        else if (/^[0-9a-z]{4,}$/.test(entry)) {
+          tmp = geohash.decode(geoFilter.topLeft);
+          point = {lat: tmp.latitude, lon: tmp.longitude};
+        }
+        else {
+          throw new BadRequestError('Unable to parse some point coordinates into polygon');
+        }
+      }
+
+      polygon.push(geoUtil.constructPoint(point));
+    }
+
+    return polygon;
+  },
+
+  /**
    * Construct a valid usable point
    *
    * @param {Object} the given object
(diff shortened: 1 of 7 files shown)


lib/api/dsl/methods.js
```diff
@@ -109,6 +109,72 @@ var geoUtil = {
...
+
+    if (!_.isArray(geoFilter.points)) {
+      throw new BadRequestError('A polygon must be in array format');
+    }
+
+    if (geoFilter.points.length < 3) {
+      throw new BadRequestError('A polygon must have at least 3 points');
+    }
+
+    /*eslint guard-for-in: 0 */
```
benoitvidis: I don't think we need an exception for eslint here.

We can add a if.. hasOwnProperty as suggested or use an alternative to for..in like Array.prototype.forEach

lib/api/dsl/methods.js
```diff
@@ -857,12 +926,75 @@ module.exports = methods = {
...
+
+      polygon = geoUtil.constructPolygon(geoFilter);
+
+    }
+    catch (err) {
+      deferred.reject(err);
+      return deferred.promise;
+    }
+
+    /*eslint guard-for-in: 0 */
```
benoitvidis: Same remark regarding eslint exception.

benoitvidis:
For the rest, +1

lib/api/dsl/methods.js
scottinet: eslint throws an error with for...in only if there isn't a test for hasOwnProperty, so I will go further than that: this comment should be removed and the eslint error should be corrected :-)

lib/api/dsl/methods.js
```diff
@@ -109,6 +109,72 @@ var geoUtil = {
...
+    for (i in geoFilter.points) {
+      entry = geoFilter.points[i];
+
+      if (_.isArray(entry)) {
+        if (entry.length >= 2) {
+          point = {lat: entry[0], lon: entry[1]};
+        } 
+        else {
+          throw new BadRequestError('Unable to parse some point coordinates into polygon');
```
scottinet: suggestion: if the parsed document is really complex or if there is lots of points, it may be useful for the user to return the incorrect point in the exception, to ease diagnostics.
Something along these lines: 'Unable to parse the following coordinate: ' + entry

What do you think?

lib/api/dsl/methods.js
```diff
@@ -109,6 +109,72 @@ var geoUtil = {
...
+        }
+      }
+      else if (_.isObject(entry)) {
+        if (entry.lat !== undefined && entry.lon !== undefined) {
+          point = {lat: entry.lat, lon: entry.lon};
+        } 
+        else {
+          throw new BadRequestError('Unable to parse some point coordinates into polygon');
```
scottinet: same as above (simple suggestion)

lib/api/dsl/methods.js
```diff
@@ -109,6 +109,72 @@ var geoUtil = {
...
+        if (/^[-.0-9]+,\s*[-.0-9]+$/.test(entry)) {
+          tmp = entry.match(/^([-.0-9]+),\s*([-.0-9]+)$/);
+          point = {lat: tmp[1], lon: tmp[2]};
+        }
+        else if (/^[0-9a-z]{4,}$/.test(entry)) {
+          tmp = geohash.decode(geoFilter.topLeft);
+          point = {lat: tmp.latitude, lon: tmp.longitude};
+        }
+        else {
+          throw new BadRequestError('Unable to parse some point coordinates into polygon');
```
scottinet: ditto

lib/api/dsl/methods.js
```diff
@@ -1039,6 +1171,7 @@ function deepExtend(filters1, filters2) {
 
   resultFilters = _.clone(filters1);
 
+  /*eslint guard-for-in: 0 */
```
scottinet: same here: this comment must be removed

lib/api/dsl/operators.js
```diff
@@ -272,6 +270,48 @@ module.exports = operators = {
...
+      result;
+
+    if (!operators.exists(field + '.lat', value, document)) {
+      return false;
+    }
+    if (!operators.exists(field + '.lon', value, document)) {
+      return false;
+    }
+    
+    /*eslint guard-for-in: 0 */
```
scottinet: comment must be removed

scottinet:
once the eslint errors are fixed: +1

lib/api/dsl/methods.js
j33f: not a bad idea

lib/api/dsl/methods.js
```diff
@@ -690,15 +760,17 @@ module.exports = methods = {
     }
 
     try {
-      /*eslint guard-for-in: 0 */
       for (i in Object.keys(filter)) {
```
AnthonySendra: Can we do something like:

```
Object.keys(foo).forEach(function (bar) {
  ...
});
```

In order to standardize with the rest of code ?

And for avoid lines
```
if (!Object.keys(filter).hasOwnProperty(i)) {
   continue;
}
```

j33f: array.foreach is asynchronious, and we need to have something synchronious here.

Two cases: 
* transform the ```if (!Object.keys(filter).hasOwnProperty(i)) {
   continue;
}``` into ``` if (Object.keys(filter).hasOwnProperty(i)) { ... ```
* keep the guard-for-in exception since it is an array, and for in statements will only get array members...

AnthonySendra: `.forEach` is absolutely not asynchronous, it's blocking.

stafyniaksacha:
+1

PR #92, opened 2015-12-02
Add new action list subscriptions

AnthonySendra:
Add a new action list subscription in Subscribe container. This function allows to get a list of rooms (with how many subscribers) for all controllers.

**bonus**: Refactor `features/support/api*`. No need to add a new action in each protocols. Just add it in `ApiRT`. And if it makes sense, also in `ApiREST`.


scottinet:
+1

benoitvidis:
+1

PR #97, opened 2015-12-04
Kuz 305 requests send realtime messages

scottinet:
* Fix: the responseListener core component waited for queries not forwarded to workers, effectively creating a memleak
* Publishing messages is now done using the ``publish`` route from the ``write`` controller, instead of using a ``persist`` flag in the ``create`` route
* ``create``, ``createOrUpdate`` and ``delete`` requests now notify subscribed users as-if the requests itself was a realtime message
* additional arguments can be provided in the ``subscribe.on`` request, allowing to filter notifications depending on their state, scope, or if this is a user event
* kuzzle rooms are now multi-channel, each channel corresponding to a subscription configuration. A user can subscribe multiple times on the same room using different channels
* subscribers now have to listen to the ``channel`` tag provided in the subscribe.on response, instead of the ``roomId``. The ``channel`` tag is only that: the channel on which the user needs to listen in order to receive notifications


benoitvidis:
+1

xbill82:
[here](https://github.com/kuzzleio/kuzzle/blob/KUZ-305-RequestsSendRealtimeMessages/docs/API.WebSocket.md#subscribing-to-documents) The doc says

    There are 4 types of notifications you can receive:

But there are actually a lot more. I would rather change it to

    You can receive the following types of notifications.

scottinet:
Agreed, I missed this particular line when I updated the documentation. Fix incoming...

scottinet:
Done. I also removed the hook publishing the request as a realtime message on ``update`` requests, since these kind of requests only contain part of the document, forcing Kuzzle to perform a filter match that can't return the right rooms.

docs/API.AMQP.md
```diff
@@ -261,7 +269,44 @@ How subscription works:
...
   metadata: {
     // query metadata
-  }
+  },
+
+  /*
+  The "scope" argument filters document modifications when a change has been
+  detected in the database.
+  You may receive notifications only when documents enter your
+  scope (scope: 'in'), when documents leave it (scope: 'out'), or
```
ballinette: for (scope: 'in'), isn't it rather "when documents enter or stay in your scope" ?

scottinet: It is. Documentation updated.

ballinette:
+1

xbill82:
:+1:

PR #102, opened 2015-12-14
Lowercase the 'state', 'scope' and 'users' before comparison

jenow:
I added a toLowerCase before comparison of requestObject.state, requestObject.scope and requestObject.users to allow devs to use, for example, enums (Scope.OPEN ...)

### DIFF
--- a/lib/api/core/hotelClerk.js
+++ b/lib/api/core/hotelClerk.js
@@ -369,7 +369,7 @@ function createChannelConfiguration (requestObject) {
     deferred = q.defer(),
     channel = {};
 
-  if (requestObject.state && ['all', 'done', 'pending'].indexOf(requestObject.state) === -1) {
+  if (requestObject.state && ['all', 'done', 'pending'].indexOf(requestObject.state.toLowerCase()) === -1) {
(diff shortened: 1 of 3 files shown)


AnthonySendra:
+1

scottinet:
+1

scottinet:
Question: for the sake of consistency, shouldn't this be extended to the following request arguments: ``action`` and ``controller``?

The API documentation should also be updated with this new feature.

benoitvidis:
As far as I can see, the update is just done on the error checks.

I guess we are missing something like
```javascript
[channel.state, channel.scope, channel.users].map(value => { return value.toLowerCase(); });
```
somewhere before the deferred.resolve call.

ballinette:
agree with Benoit.
or, better, noramlize these attributes in RequestObject constructor, instead ?

jenow:
We have to decide wether or not we want the actions and controller case sensitive or not.
For options like state, scope, users which have fixed values it is natural not to be case sensitive.

jenow:
I did move the lowercasing of 'scope', 'states' and 'users' in the requestObject constructor. Some actions are using camelCase so for the consistancy I didn't do it on controller, action and collection either.

ballinette:
+1

scottinet:
+1

PR #103, opened 2015-12-14
bugfix: filter removal was not migrated to the new "or" filter format

scottinet:
* the recursive filter paths getter was still considering the content of a "or" filter to be an object, instead of an array
* the hotelClerk room destruction method destroyed a room only if the DSL successfully cleaned up the filters, instead of destroying it whenever the DSL finishes its work


AnthonySendra:
+1

j33f:
+1

PR #105, opened 2015-12-16
Kuz 335 list realtime collections

scottinet:
* implemented the possibility to get realtime collections using the existing "listCollections" API method
* updated unit tests, functional tests and API documentation
* tidied up the ``readController`` unit tests
* removed remnants of the old ``roomName`` argument in the ``hotelClerk.removeCustomerFromAllRooms`` unit tests
* added missing unit tests on the ``removeRooms`` method of the ``adminController``


benoitvidis:
+1
NB: taken into account on the new api doc already (in progress).

scottinet:
@benoitvidis : wait before updating the new API documentation. Right now, this new request is the only one not taking its arguments from the document body, and this is wrong. I'm changing that.

scottinet:
Fixed.

ballinette:
+1

lib/api/controllers/readController.js
```diff
@@ -23,9 +23,35 @@ module.exports = function ReadController (kuzzle) {
   };
 
   this.listCollections = function (requestObject) {
```
AnthonySendra: Here, we need to return only allowed collections

AnthonySendra:
+1

PR #106, opened 2015-12-16
Kuz315 api documentation

benoitvidis:
The new doc is online on http://kuzzleio.github.io/kuzzle-api-documentation

May be worth validating at the same time as this PR.


ballinette:
+1 and great, I will be able to document the authentication mechanism with that ;)

scottinet:
Thanks A LOT for doing this work, it really doesn't seem fun.
The API documentation is now much more organized, and it's a breeze to update it, we really needed that!

+1 obviously

jenow:
+1
Now I can even give url to the slate into the sdk code documentation for the user to have a nice documentation display :)

stafyniaksacha:
+1

PR #108, opened 2015-12-17
[BREAKING] Kuz 318 mutli index

stafyniaksacha:
* Support multi-index (persistent & realtime layers)
* Add : CRUDL operations on indexes
* Add :  _createIndex role policy
* Add :  _createCollection role policy
* Add :  internal cache for index hierarchy (used by when checking policies)
* Refactor : ReadEngine & WriteEngine allways fulify promises with a ResponseObject : no need to recreate it in worker when Response is delayed
* Refactor : Change api route to /api/v1
* Refactor : Mock elasticsearch in units tests


features/support/apiREST.js
```diff
@@ -13,16 +13,16 @@ ApiREST.prototype.init = function (world) {
 ApiREST.prototype.disconnect = function () {};
 
 ApiREST.prototype.pathApi = function (path) {
-  return config.url + '/api/' + path;
+  return config.url + '/api/v1/' + path;
```
scottinet: :+1: API version number will be required by protocol plugins, to make sure that a protocol plugin will be compatible to a given version of Kuzzle.

It may be interesting to also have a minor version number though, like 'v1.0', to distinguish between minor and major changes.

lib/api/core/models/security/role.js
```diff
@@ -39,6 +52,15 @@ Role.prototype.isActionAllowed = function (requestObject, context) {
     return false;
   }
 
+  if (requestObject.persist &&
+    this.indexes._canCreate !== undefined &&
+    !this.indexes._canCreate &&
+    _.contains(['import', 'create', 'createCollection', 'putMapping', 'createOrUpdate'], requestObject.action) &&
```
AnthonySendra: Can we find a better solution than list actions ? I'm not sure we'll remember to add a newly added collection here.  Maybe we can split the `adminController` into two files and have all these 5 methods in a specific controller and list available  method for this controller. Just a proposition, maybe we can discuss

lib/api/core/responseListener.js
```diff
@@ -3,7 +3,7 @@ var ResponseObject = require('./models/responseObject');
 module.exports = function ResponseListener (kuzzle, responseQueue) {
   var
     whitelist = {
-      admin: ['deleteCollection', 'putMapping', 'truncateCollection'],
+      admin: ['deleteCollection', 'putMapping', 'truncateCollection', 'createIndex', 'deleteIndex', 'deleteIndexes'],
```
AnthonySendra: Same as above. I'm not comfortable with maintain a list

lib/api/core/hotelClerk.js
```diff
@@ -578,13 +604,13 @@ function cleanUpRooms (roomId) {
...
         this.kuzzle.pluginsManager.trigger('room:remove', roomId);
+        delete this.rooms[roomId];
         deferred.resolve(roomId);
       })
       .catch(error => {
         this.kuzzle.pluginsManager.trigger('log:error', error);
         deferred.reject(error);
-      })
-      .finally(() => delete this.rooms[roomId]);
+      });
```
scottinet: While merging it seems you rolled back a bugfix: the room deletion should occur in the ``finally`` statement to make sure it is deleted even if ``dsl.removeRoom`` fails.

lib/api/dsl/methods.js
```diff
@@ -8,8 +8,231 @@ var
   util = require('util'),
   geohash = require('ngeohash'),
   units = require('node-units'),
-  methods,
-  geoUtil = {};
+  methods;
+
+var geoUtil = {
```
AnthonySendra: Can we move the `geoUtil` object below all methods ?  Or maybe like you suggest, in other file

lib/api/core/responseListener.js
scottinet: Currently under refactoring, it'll soon be removed.

lib/api/dsl/methods.js
scottinet: +1 with another file then, to make the methods one lighter

lib/api/prepareDb.js
```diff
@@ -39,30 +39,35 @@ function importMapping() {
...
-    }.bind(this), function (error) {
-      callbackCollection(error);
+  async.each(Object.keys(mappings), (index, callbackIndex) => {
+    async.each(Object.keys(mappings[index]), (collection, callbackCollection) => {
+      this.pluginsManager.trigger('log:info', '== Importing mapping for collection ' + index + ':' + collection + '...');
+
+      async.each(mappings[index][collection], (mapping, callbackMapping) => {
+        var mappingOptions = {
+          action: 'putMapping',
+          persist: true,
```
scottinet: I think this flag is unused and should be removed.

lib/api/prepareDb.js
```diff
@@ -101,23 +106,28 @@ function importFixtures() {
...
-        callback();
-      })
-      .catch(function (error) {
-        callback('Fixture import error' + error);
-      });
+  async.each(Object.keys(fixtures), (index, callbackIndex) => {
+    async.each(Object.keys(fixtures[index]), (collection, callback) => {
+      var fixture = {
+        action: 'import',
+        persist: true,
```
scottinet: same as above

lib/services/elasticsearch.js
```diff
@@ -147,9 +147,27 @@ module.exports = function (kuzzle, options) {
...
+
+    this.client.create(data)
+      .then(function (result) {
+        if (kuzzle.indexes[data.index] === undefined) {
+          kuzzle.indexes[data.index] = [];
+        }
+
+        if (kuzzle.indexes[data.index].indexOf(data.type) === -1) {
+          kuzzle.indexes[data.index].push(data.type);
+        }
```
scottinet: These few lines seems to be repeated in all ES function: maybe they should be isolated in a dedicated private method?

lib/api/core/responseListener.js
AnthonySendra: :+1:

AnthonySendra:
Great ! :+1:

j33f:
+10

lib/api/controllers/funnelController.js
```diff
@@ -37,24 +37,25 @@ module.exports = function FunnelController (kuzzle) {
     var
       deferred = q.defer();
 
+    context.kuzzle = kuzzle;
```
benoitvidis: We cannot put kuzzle in the context. The context describes the current connection (cf HttpContext in [Java](https://docs.oracle.com/javase/7/docs/api/javax/xml/ws/spi/http/HttpContext.html) or [.net](https://msdn.microsoft.com/en-us/library/system.web.httpcontext(v=vs.110).aspx))

The context object is later one given to the role validation, and passed to the role validation methods that the bo users may have defined. We do not want an end-user to be able to modify kuzzle's behaviour on runtime.

As far as I can see, the only use is to check if the requested index & collection exist. If we only need the `kuzzle.indexes` collection, I would rather pass it to `role.prototype.isActionAllowed` method.

lib/api/core/models/security/role.js
```diff
@@ -12,6 +12,19 @@ function Role () {
   this.closures = {};
 }
 
+function isIndexExist(requestObject, context) {
```
benoitvidis: nitpicking: isIndexExisting (or doesIndexExist ; I am not sure we should really need to strictly comply to the isxxx convention for  all methods that return a boolean).

benoitvidis:
++++ :+1: :checkered_flag:

scottinet:
Tremendous work. +1!

PR #109, opened 2015-12-17
Auhentication controller

ballinette:
* create an authentication controlller, using passport.js
* enable by default kuzzle-plugin-basicauth for authentication with username/password
* add a light password hash management methods (for now, only a "hard-coded" sha1 HMAC)
* fix some bugs about loadFromCache and persistToCache for repositories objects
* add a method to UserRepository for JWT token generation
* add documentation (see also https://github.com/kuzzleio/kuzzle-api-documentation/pull/1 )


ballinette:
NB : still to do : implement unit tests and functionnnal testings.
This PR is created to let you comment and give your feedbacks and any suggestion to improvments.
Wait for test implementation before merging (meanwhile, you can still work on Kuz-277-authentication if you need authentication)

ballinette:
Still to do : functionnal testing
But the PR is ready to merge as soon as enough people validate it ;)

lib/api/core/auth/passportWrapper.js
```diff
@@ -0,0 +1,35 @@
...
+
+module.exports = function PassportWrapper (kuzzle) {
+
+  this.authenticate = function(request, strategy) {
+    var
+      deferred = q.defer();
+
+    try {
+      if (!passport._strategy(strategy)) {
+        return Promise.reject(new Error('Unknown authentication strategy "' + strategy + '"'));
```
scottinet: should return q.reject() instead, to ensure promise compatibility (for instance, Promise doesn't have a ``finally`` method)

scottinet: should be a ``xxxError`` (internal or badrequest, depending from where comes the ``strategy`` argument?), instead of a generic ``Error``

lib/api/core/auth/passportWrapper.js
```diff
@@ -0,0 +1,35 @@
...
+        }
+        else if (user === false) {
+          deferred.reject(new Error(info.message));
+        }
+        else {
+          deferred.resolve(user);
+        }
+      })(request);
+    } catch (err) {
+      deferred.reject(new Error('Error authenticating user'));
```
scottinet: shouldn't this error be a ``ForbiddenError`` instead?

lib/api/core/auth/passwordManager.js
```diff
@@ -0,0 +1,21 @@
+var crypto = require('crypto');
+
+module.exports = function PassportManager () {
+
+  this.secret = 'jbsB8r69PFP39KdLtjVr25Z22';
```
scottinet: Magic number. Has this specific value any significance? If so, can you add a comment to explain it?
If not, is it possible to generate it from a configuration parameter?

lib/api/core/auth/passwordManager.js
```diff
@@ -0,0 +1,21 @@
...
+  this.algo = 'sha1';
+  this.digest = 'hex';
+
+  this.encryptPassword = function(password) {
+    var hashedPassword = crypto.createHmac(this.algo, this.secret).update(password).digest(this.digest);
+    return Promise.resolve(hashedPassword);
```
scottinet: should return ``q(hashedPassword)`` instead of ``Promise.resolve`` to ensure promises compatibility.

lib/api/core/auth/passwordManager.js
```diff
@@ -0,0 +1,21 @@
...
+  };
+
+  this.checkPassword = function(password, hash) {
+    return this.encryptPassword(password)
+    .then(function(value) {
+      return Promise.resolve(hash === value);
```
scottinet: should return ``q(hash === value)`` instead of ``Promise.resolve``

lib/api/core/auth/passwordManager.js
ballinette: it's just a first step to have something usable.
We need to improve the passwordManager and of course let the secret, the algo be configurable, + probably add other features.
But I prefer working step by step and delay this to a later PR ;)

scottinet: Ahh ok!
I couldn't understand what that meant, but it's now clearer. Can you add a TODO comment then?

lib/api/core/auth/passportWrapper.js
ballinette: Hmm right. I guess the BadRequestError is more appropriate here, as the strategy must be set and mapped to an existing one.

lib/api/core/auth/passportWrapper.js
j33f: +1 AuthenticationError since the error is related to the authentication itself, not the ressource access

ballinette: hmm for the line 21, I agree to set a ForbiddenError.
Bug within the catch, maybe an InternalError is more appropriate...

lib/api/core/auth/passwordManager.js
ballinette: ok, done

lib/api/core/auth/passportWrapper.js
j33f: forbidden says "this user does not have the right to access to the requested ressource", is it the right context here ?

ballinette: ok, right, I add this ;)

ballinette: @scottinet @j33f 
hmm well, if I try to follow http-like codes, the best error type should be a 401 - UnauthorizedError... your advice ?

j33f:
+1

lib/api/controllers/routerController.js
```diff
@@ -23,6 +23,7 @@ module.exports = function RouterController (kuzzle) {
       api = new Router(),
       coverage,
       routes = [
+        {verb: 'post', url: '/_login', controller: 'auth', action: 'login'},
```
stafyniaksacha: Can the post route be with other post routes ?

test/api/controllers/authController.test.js
```diff
@@ -0,0 +1,90 @@
...
+        var decodedToken = jwt.verify(response.data.jwt, params.jsonWebToken.secret);
+        should(decodedToken._id).be.equal('jdoe');
+        done();
+      })
+      .catch(function (error) {
+        done(error);
+      });
+  });
+
+  it('should reject if authentication failure', function () {
```
stafyniaksacha: need to rename test to : should reject if authentication use unexisting strategy

docs/plugins.md
```diff
@@ -40,6 +40,10 @@ A plugin configuration can have attributes:
 
 By default, the logger plugin is enabled and configured to use the service `winston` (refer to the kuzzle-plugin-logger documentation for more information).  
 
+## Basic Auth
```
benoitvidis: To me, Basic auth is the one used by HTTP, i.e. [.htpasswd-like](https://en.wikipedia.org/wiki/Basic_access_authentication).
Could we rename it to something more generic like `LoginPassword` or `Password`?

lib/api/controllers/routerController.js
ballinette: sure

docs/plugins.md
ballinette: @benoitvidis  
Or we can name the plugin "auth-passport-local", to use the same name as passport strategy ?

lib/api/core/auth/passwordManager.js
```diff
@@ -0,0 +1,30 @@
...
+- improve globally the manager to enable more features
+*****/
+
+var
+  q = require('q'),
+  crypto = require('crypto');
+
+module.exports = function PassportManager () {
+
+  this.secret = 'jbsB8r69PFP39KdLtjVr25Z22';
```
AnthonySendra: Note: Can easily be done by adding this secret in .kuzzlerc. Then, can access with `kuzzle.config.myVariable`

lib/api/core/models/repositories/userRepository.js
```diff
@@ -81,6 +107,31 @@ module.exports = function (kuzzle) {
     return deferred.promise;
   };
 
+  UserRepository.prototype.loadByUsernameAndPassword = function (username, password) {
```
AnthonySendra: I'm not sure to get it. So when we want to install new auth plugin with another strategy, we have to add a function here for this specific authentication ?

ballinette: Not exactly.
We could also use UserRepository.prototype.load(username), and then compare user's password with the one given by the user. But for this feature especifically, I found cleaner and easier to manage password encryption/decryption within kuzzle kernel, and so just call loadByUsernameAndPassword() from the plugin.

For other authentication plugin, using external services, the strategy will always reponse with a strategy-dependant identifier.
So we will have 2 choices : 
- either map directly this identifier to the kuzzle user's id, and just call UserRepository.prototype.load(id)
- or map it to another kuzzle user's field, and we have to implement only one method UserRepository.prototype.search('identifier-name', id);

test/api/controllers/authController.test.js
ballinette: No, we just test here that any authentifcation failure reject the promise. I only used the unexisting strategy as a failure example.
But you're right, it's badly written.
Fixing it ;)

ballinette: done

benoitvidis:
Great! +1

scottinet:
+1!

lib/api/controllers/funnelController.js
```diff
@@ -5,6 +5,7 @@ var
   SubscribeController = require('./subscribeController'),
   BulkController = require('./bulkController'),
   AdminController = require('./adminController'),
+  AuthController = require('./authController'),
```
stafyniaksacha: Did the authentication controller can be in kuzzle-plugin-auth-passport-local ?

PR #110, opened 2015-12-21
Worker responses refactor

scottinet:
To remove unnecessary Kuzzle dependencies from the router controller and to prepare the incoming protocol plugins, this pull request changes the way the router controller handles worker responses, following @stafyniaksacha suggestion:

* renamed ``core/responseListener`` to ``core/workerListener`` to make it easier to understand what it does
* removed all references to the ``workerListener`` from the router controller. Also removed empty promises handling. Promises resulting from calls to the funnel controller now always resolve to the according responseObject
* the main Kuzzle object now contains a global ``workerListener`` instance
* controllers behind the funnel controller now always return a promise resolving to the awaited response, either directly, or by returning the result of ``workerListener.add``
* the ``workerListener`` now awaits worker responses using the internal broker, and resolve/reject the stored promises, letting the router controller respond to the requesting client. It doesn't notify directly anymore.

### DIFF
--- a/lib/api/controllers/funnelController.js
+++ b/lib/api/controllers/funnelController.js
@@ -44,21 +44,21 @@ module.exports = function FunnelController (kuzzle) {
         if (!this[requestObject.controller] ||
           !this[requestObject.controller][requestObject.action] ||
           typeof this[requestObject.controller][requestObject.action] !== 'function') {
-          return Promise.reject(new BadRequestError('No corresponding action ' + requestObject.action + ' in controller ' + requestObject.controller));
+          return q.reject(new BadRequestError('No corresponding action ' + requestObject.action + ' in controller ' + requestObject.controller));
         }
 
         // check if the current user is allowed to process
         if (context.user.profile.isActionAllowed(requestObject, context, kuzzle.indexes) === false) {
-          return Promise.reject(new UnauthorizedError('Unauthorized action [' + requestObject.index + '/' + requestObject.collection + '/' + requestObject.controller + '/' + requestObject.action + '] for user ' + context.user._id, 401));
+          return q.reject(new UnauthorizedError('Unauthorized action [' + requestObject.index + '/' + requestObject.collection + '/' + requestObject.controller + '/' + requestObject.action + '] for user ' + context.user._id, 401));
         }
 
         return this[requestObject.controller][requestObject.action](requestObject, context);
       })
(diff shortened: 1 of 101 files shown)


lib/api/controllers/funnelController.js
```diff
@@ -44,21 +44,21 @@ module.exports = function FunnelController (kuzzle) {
...
         // check if the current user is allowed to process
         if (context.user.profile.isActionAllowed(requestObject, context, kuzzle.indexes) === false) {
-          return Promise.reject(new UnauthorizedError('Unauthorized action [' + requestObject.index + '/' + requestObject.collection + '/' + requestObject.controller + '/' + requestObject.action + '] for user ' + context.user._id, 401));
+          return q.reject(new UnauthorizedError('Unauthorized action [' + requestObject.index + '/' + requestObject.collection + '/' + requestObject.controller + '/' + requestObject.action + '] for user ' + context.user._id, 401));
         }
 
         return this[requestObject.controller][requestObject.action](requestObject, context);
       })
-      .then(function (responseObject) {
+      .then(responseObject => {
```
xbill82: Yeah! I like this :+1:

AnthonySendra:
Quick question: Why is the function `kuzzle.workerListener.add(requestObject);` directly called in `adminController` ? Why don't we pass by a hook that will call `kuzzle.workerListener.add(requestObject);` like for `write` worker ?

AnthonySendra:
+1

scottinet:
@AnthonySendra > because the main goal of this PR is to fix the data flow.

It was like this:

**Direct calls:**
client ===> Router controller    ===> direct operation ===> response ===> router controller ===> client

**Worker calls:**
client ===> router controller ===> worker ==> response ==> responseListener ===> client

Now, worker calls are forwarded to the client via the router controller. To do this we use promise, resolved by the ``workerListener`` when the worker gets back with the response.

And in order to use promises, we have to somehow store them in the ``workerListener``. I do that by letting the ``workerListener.add`` function creating the promise and returning it.

If we use hooks like you propose, then controllers will have to create the promise, and find a way to forward it to the ``workerListener``. It produces more lines of code, and it complicates things a lot.

xbill82:
Yo! Seems like a great job! As far as I can understand, I say :+1: :+1: :+1:

j33f:
+1

PR #111, opened 2015-12-22
Protocol plugins prerequisites

scottinet:
*  The router controller now directly respond to requesters, without using the notifier core component anymore. This is to set clients<=>router communication apart from the rest of Kuzzle, making easier to implement protocol plugins
* The notifier core component now only broadcasts messages and does not accept a connection argument, since the router now respond directly to clients
* Added the new api/serverInfo route. The main goal is to provide requesters with the API version of Kuzzle, ensuring that plugins and SDKs are compatible. This command also returns a lot of useful informations about Kuzzle, its plugins, the system and active services.
* Added unit and functional tests
* Split the functional tests keywords implementation into multiple files in order to clean things up


j33f:
+1

stafyniaksacha:
+1

AnthonySendra:
+1
(thanks for splitting `step_definitions` files ...)

ballinette:
+1

PR #114, opened 2016-01-06
Plugin protocol implementation

scottinet:
* Implemented protocol plugins support
* Added a new plugin configuration argument: "loadedBy", taking "all", "server" or "worker". Its purpose is to limit plugin installation & loading only by the corresponding kuzzle instance type
* Added socket.io plugin as a default plugin
* removed socket.io dependency
* updated unit and functional tests
* updated plugin documentation
* REST protocol is now linked to the port 7511 instead of 7512
* Reworked plugin installation so that servers and workers can act independantly. This removes the "normal" errors we got when starting Kuzzle
* Added a banner to inform users when a Kuzzle instance is up, running, and accepting requests

Others:
* removed winston dependency


docs/plugins.md
```diff
@@ -64,13 +101,46 @@ The module must have a `package.json` file with a `pluginInfo` entry. The option
   }
 ```
 
+The `loadedBy` option tells Kuzzle to install and load the plugin only by corresponding instance types.  
+The accepted values are: `all`, `server` and `worker`.
```
ballinette: Is "loadedBy" now required, or do we consider that it has a default value to "all" if not set ?

scottinet: It is defaulted to `all`. I'll add that in the documentation.

scottinet:
Travis seems to have a hard time creating indexes in functional tests without hitting cucumber's timeout, so I'm gonna migrate index creations to test fixtures.

docs/plugins.md
```diff
@@ -64,13 +101,46 @@ The module must have a `package.json` file with a `pluginInfo` entry. The option
...
 ## Architecture
 
 Your main javascript file in your plugin must have a function `init` and expose a `hooks` and/or a `pipes` and/or a `controllers` object. All functions defined in these files must be exposed as main object.
 
-### Hooks
 
-Hook events are triggered and are not blocking functions. Typically, if we want to log something, we will use hook events.
+## The plugin init function
+
+All plugins must expose a ``init`` function. Its purpose is to initialize the plugins according to its configuratin.
```
ballinette: configuration ;)

docs/plugins.md
```diff
@@ -240,29 +309,134 @@ Sample:
...
+* With REST protocol, we use the routes configured in _routes.js_.  
+These routes are automatically prefixed with "\_plugin/" + the plugin name.
+
 Samples:
 
 GET action:
 
 ```
-GET http://kuzzle:7512/api/_plugin/myplugin/foo/John%20Doe
+GET http://kuzzle:7511/api/_plugin/myplugin/foo/John%20Doe
```
ballinette: should fix URL with api version :  http://kuzzle:7511/api/1.0/_plugin/myplugin/foo/John%20Doe

docs/plugins.md
scottinet: ok

docs/plugins.md
scottinet: Comes from your previous merge from master :-)
I'll fix it in an upcoming commit.

features/support/apiREST.js
```diff
@@ -14,13 +14,11 @@ ApiREST.prototype.init = function (world) {
 ApiREST.prototype.disconnect = function () {};
 
 ApiREST.prototype.pathApi = function (path) {
-  var basePath = '/api';
-
-  if (apiVersion) {
-    basePath += '/' + apiVersion;
-  }
+  return config.url + '/api/1.0/' + path;
```
ballinette: hmm why "1.0" is hard-coded ?
Should it be rather "return config.url + '/api/' + apiVersion + path;" ?

ballinette:
+1  ! great ;)

features/support/apiREST.js
scottinet: Because only Kuzzle knows its API version, and functional tests are clients to Kuzzle. 
To get the API version, the only solution would be to perform a "serverInfo", but that would mean starting to use Kuzzle before effectively testing it.
So... hardcoded. :-)

ballinette: Forgot it, i read it too fast and didn't notice that we were within feautres/ ;)

scottinet:
Now Kuzzle workers keep dying during Travis tests. I'm looking into it (although I can't reproduce this behavior testing locally)

scottinet:
Kuzzle workers are now stable. The problem was indeed coming from plugin installations timeout.

Now, there is still the problem of the first functional test failing in Travis. And I just figured it out: kuzzle starts during unit tests. And when Travis launches functional tests, Kuzzle is still installing plugins... Only after the first cucumber timeout is Kuzzle up & running.

I'm gonna add a curl test to check that Kuzzle is up before running functional tests in our travis docker container.

scottinet:
The code coverage reduction comes mainly from added logic in the untested plugin installation.

If that's a problem I'll check if this can be tested with stubs, but it'll requires lots of work to throughly test this.

@ballinette > since I added new functionalities in the plugins manager, you may want to take a look to confirm your "+1"

ballinette:
It's probably also because you removed some tested code in the routerController.
Not an issue imho ;)

docs/plugins.md
```diff
@@ -7,32 +34,35 @@ There are several types of plugins:
...
 
 * `url`: a git URL where the plugin can be found and cloned.
-* `version`: a version corresponding to the version given in the file `package.json` in the plugin module.
+* `version`: the NPM package version to download
 * `customConfig`: config for the plugin. Each plugin has a different configuration (required or optional), check the corresponding plugin documentation for more information.
 * `defaultConfig`: Don't edit this attribute. The defaultConfig is provided by the plugin itself. If you need to change the configuration, edit the `customConfig` attribute
 
 **Note:**
-* URL or version are required. The URL is checked first, so if you have set a version and an URL, the version will be ignored.
+* A `url` or `version` parameter is required. The URL is checked first, so if you have set both a version and an URL, the version will be ignored.
```
j33f: replace "A `url`" by "An `url` ?

ballinette:
+1 confirmed ;)

docs/plugins.md
```diff
@@ -95,22 +167,23 @@ module.exports = function () {
...
   }
 }
 ```
 
-### Pipes
+### Pipe plugins
 
-When an event pipe is triggered, we are waiting for all the functions attached on this event. A function attached on a pipe event has access to the data and can even change them.
-A function must take in its last parameter a callback. This callback must be called at the end of the function with `callback(error, object)`:
+When an pipe event is triggered, we are waiting for all plugins attached on this event. A plugin attached on a pipe event has access to the data and can even change them.
```
j33f: "When **a** pipe event" instead ?

docs/plugins.md
```diff
@@ -95,22 +167,23 @@ module.exports = function () {
...
+A pipe plugin must take in its last parameter a callback. This callback must be called at the end of the function with `callback(error, object)`:
```
j33f: "A pipe plugin constructor" ? maybe

docs/plugins.md
```diff
@@ -240,29 +309,134 @@ Sample:
...
 ```
 
+### Protocol plugins
+
+Kuzzle core only supports REST communications. All other supported protocols are implemented as protocol plugins.  
+By default, the Kuzzle official docker image is shipped with the [Socket.io](https://github.com/kuzzleio/kuzzle-plugin-socketio) protocol.
+
+#### How it works
+
+Protocol plugins allow Kuzzle to support any existing protocol. These plugins ensure a two-ways communication between clients and Kuzzle.
```
j33f: two-way (without plural)

features/support/apiREST.js
```diff
@@ -14,13 +14,11 @@ ApiREST.prototype.init = function (world) {
...
+};
 
-  return config.url + basePath + '/' + path;
+ApiREST.prototype.pathBaseApi = function (path) {
```
j33f: "baseApiPath" sounds more logical to me

lib/api/core/pluginsManager.js
```diff
@@ -71,10 +116,22 @@ module.exports = function PluginsManager (kuzzle) {
         return true;
       }
 
+      /*
+       Plugins can be loaded by servers or workers. The "loadedBy" property tells by what type of kuzzle instance
```
j33f: "The "loadedBy" property tells by **wich** type of kuzzle instance" ?

docs/plugins.md
scottinet: Not before a highlighted keyword

docs/plugins.md
scottinet: ok

docs/plugins.md
scottinet: ok

docs/plugins.md
scottinet: ok

features/support/apiREST.js
scottinet: ok

docs/plugins.md
j33f: ok

lib/api/core/pluginsManager.js
scottinet: no

j33f:
purrrrfect :smiley_cat:

scottinet:
A commit is coming with corrections asked by Jeff + plugin managers initialization unit tests

scottinet:
Added unit tests to the plugin manager initialization anyway :-p

j33f:
+1

jenow:
+1

PR #122, opened 2016-01-12
Kuzzle CLI

scottinet:
- The CLI is now capable of installing plugins. Simple call `bin/kuzzle.js install` to do it
- Removed plugins installation from the plugin manager
- Removed obsolete app-start.js and app-perf.js scripts ( :warning: you'll have to update your docker image to make this work )


scottinet:
I'm going to merge this one as soon as checks finish, because I had to adapt our docker images for this PR to work.

PR #123, opened 2016-01-12
Crudl role

AnthonySendra:
Add CRUD for manage role according to the new API


j33f:
+1

lib/api/controllers/securityController.js
```diff
@@ -0,0 +1,61 @@
...
+   * @returns {Promise}
+   */
+  this.getRole = function (requestObject) {
+    kuzzle.pluginsManager.trigger('data:getRole', requestObject);
+
+    return kuzzle.repositories.role.loadRole(
+      kuzzle.repositories.role.getRoleFromRequestObject(requestObject)
+      )
+      .then(role => {
+        return Promise.resolve(new ResponseObject(requestObject, role || {}));
```
benoitvidis: Looks like the response does not follow the new responseObject format. We get the role object directly in the result, not in _source:
```
{ status: 200,
  error: null,
  requestId: '6904a370-baa5-11e5-94df-d3e4c4b5fa6d',
  controller: 'security',
  action: 'getRole',
  metadata: { iwant: 'to break free', we: [ 'will', 'rock', 'you' ] },
  state: 'done',
  scope: null,
  result: { indexes: { fakeIndex1: [Object] }, closures: {}, _id: 'test' } }
```

lib/api/core/models/repositories/roleRepository.js
```diff
@@ -118,6 +118,53 @@ module.exports = function (kuzzle) {
...
+    if (!role._id) {
+      deferred.reject(new BadRequestError('Missing role id'));
+      return deferred.promise;
+    }
+
+    if (this.roles[role._id]) {
+      deferred.resolve(this.roles[role._id]);
+    }
+    else {
+      return this.loadOneFromDatabase(role._id);
```
ballinette: better user this.load(role._id)
It will load the role from cache if exists, or from database if not in cache

lib/api/core/models/repositories/repository.js
```diff
@@ -252,6 +313,27 @@ Repository.prototype.persistToDatabase = function (object) {
 };
 
 /**
+ * Delete repository from database according to its id
+ * @param id
+ */
+Repository.prototype.deleteFromDatabase = function (id) {
```
ballinette: need also a deleteFromCache and a delete methods, for data that are in db and in cache.

lib/api/controllers/securityController.js
AnthonySendra: Because this is not a ES document (for the user) I'm wondering if we have to add a `_source` attribute (we don't have `_id`, `_version`, etc)
Like for getStats and other function like this, we don't have a `_source` attribute.
Do think it's better to add `_source` in any cases ? If we add it here, we have to add everywhere

lib/api/core/models/repositories/repository.js
AnthonySendra: We've already discussed with @benoitvidis and the conclusion was to let each repo (profile, role, user) to delete from its memory and the meta repository for delete from the database.
the method `deleteRole` delete the role from memory and next call this `deleteFromDatabase`.
@benoitvidis maybe I misunderstood ?

lib/api/core/models/repositories/roleRepository.js
AnthonySendra: We can't use `load` function from repository. There is an extract of this function

```
if (this.cacheEngine === null) {
    return this.loadOneFromDatabase(id);
  }

  this.loadFromCache(id, opts).then()
```

And for RoleRepository the cache is undefined. But like you can see line 133 in roleRepository, I check if the role is not in memory before load it from database.

lib/api/controllers/securityController.js
```diff
@@ -0,0 +1,79 @@
...
+
+    return kuzzle.repositories.role.loadRole(
+      kuzzle.repositories.role.getRoleFromRequestObject(requestObject)
+      )
+      .then(role => {
+        if (role && role.closures) {
+          delete role.closures;
+        }
+
+        return Promise.resolve(new ResponseObject(requestObject, role || {}));
```
scottinet: should return `q(new ResponseObject...` to harmonize the promise objects we use throughout the entire project

lib/api/controllers/securityController.js
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * Get a specific role according to the given id
+   * @param {RequestObject} requestObject
+   * @returns {Promise}
+   */
+  this.getRole = function (requestObject) {
+    kuzzle.pluginsManager.trigger('data:getRole', requestObject);
+
+    return kuzzle.repositories.role.loadRole(
```
scottinet: Should add a `catch` handler, as `kuzzle.repositories.role.loadRole` may return a rejected promise.

lib/api/controllers/securityController.js
```diff
@@ -0,0 +1,79 @@
...
+            var parsedRole = {};
+
+            parsedRole._id = role._id;
+            parsedRole._source = {indexes: role.indexes};
+
+            return parsedRole;
+          });
+        }
+
+        return Promise.resolve(response);
```
scottinet: Should return `q(response);`

lib/api/controllers/securityController.js
```diff
@@ -0,0 +1,79 @@
...
+   * @returns {Promise}
+   */
+  this.putRole = function (requestObject) {
+    kuzzle.pluginsManager.trigger('data:putRole', requestObject);
+
+    return kuzzle.repositories.role.validateAndSaveRole(
+      kuzzle.repositories.role.getRoleFromRequestObject(requestObject)
+      )
+      .then(result => {
+        return Promise.resolve(new ResponseObject(requestObject, result));
```
scottinet: Should return `q(new ResponseObject(requestObject, result));`

lib/api/controllers/securityController.js
```diff
@@ -0,0 +1,79 @@
...
+
+module.exports = function SecurityController (kuzzle) {
+
+  /**
+   * Get a specific role according to the given id
+   * @param {RequestObject} requestObject
+   * @returns {Promise}
+   */
+  this.getRole = function (requestObject) {
+    kuzzle.pluginsManager.trigger('data:getRole', requestObject);
```
scottinet: Question: I wonder if plugin hooks on API security routes shouldn't be "security:..." instead of "data:..."

scottinet:
+1

lib/api/core/models/repositories/roleRepository.js
ballinette: hmm actually, the cacheEngine is NOT undefined (cf line 18 :  cacheEngine: kuzzle.services.list.userCache )

What I don't understand is: why should the cache mechanism be different btw. UserRepository vs. RoleRepository ?
No matter for me anyway, but I suggest that you talk about that with @benoitvidis  to be sure to have an homogeneous code code within each repository prototypes

lib/api/core/models/repositories/repository.js
```diff
@@ -109,6 +109,67 @@ Repository.prototype.loadMultiFromDatabase = function (ids) {
...
+        object = new this.ObjectConstructor();
+        data = _.extend({}, document._source);
+        data._id = document._id;
+
+        promises.push(this.hydrate(object, data));
+      });
+
+      return q.all(promises)
+        .then(roles => {
+          deferred.resolve(new ResponseObject(requestObject, {hits: roles, total: roles.length}));
```
xbill82: Manipulating `roles` in a general purpose repository.

xbill82:
:+1:

stafyniaksacha:
+1

lib/api/core/models/repositories/userRepository.js
```diff
@@ -219,7 +219,13 @@ module.exports = function (kuzzle) {
     return result;
   };
 
-  UserRepository.prototype.serializeToDatabase = UserRepository.prototype.serializeToCache;
+  UserRepository.prototype.serializeToDatabase = function (user) {
```
stafyniaksacha: Can be simplified to `UserRepository.prototype.serializeToDatabase = UserRepository.prototype.serializeToCache`

PR #124, opened 2016-01-13
Fixed statistics TTL

scottinet:
Statistics TTL was multiplied by 1000 to convert it to milliseconds... but Redis API TTL is in seconds.
So we were saving 1000 times too many stats frames.


AnthonySendra:
Already done in this PR https://github.com/kuzzleio/kuzzle/pull/123/files#diff-2befc9a592f07df5bf0a52d53cdc1d63L9 (because it was mandatory for my PR)

scottinet:
Cool!

I'm closing this one then.

PR #127, opened 2016-01-14
Kuz 345 sdk add login method

stafyniaksacha:
* add token repository (whitelist)
* englobe user data in token object
* add tokenized connexion controller (garbage collector, which clean expired connexions in hotelclerk)
* add logout method


lib/api/core/models/repositories/repository.js
```diff
@@ -169,14 +169,14 @@ Repository.prototype.loadFromCache = function (id, opts) {
 Repository.prototype.load = function (id, opts) {
   var deferred = q.defer();
 
-  if (this.cacheEngine === null) {
+  if (!Boolean(this.cacheEngine)) {
```
AnthonySendra: Why don't just check `if (this.cacheEngine)` ?

lib/api/controllers/authController.js
```diff
@@ -7,26 +7,53 @@ module.exports = function AuthController (kuzzle) {
...
   this.passport = new PassportWrapper(kuzzle);
 
-  this.login = function (requestObject) {
+  this.logout = function (requestObject, context) {
+    var
+      deferred = q.defer(),
+      response;
+
+    kuzzle.pluginsManager.trigger('auth:logout', context)
+      .then(kuzzle.repositories.token.expire(context.token))
```
scottinet: This doesn't work.
According to [promises specifications](https://promisesaplus.com/) followed by Q, the `then` method takes two arguments: `onFulfilled` and `onRejected`.
If these arguments aren't functions, they are ignored.

What happens here is that `kuzzle.repositories.token.expire` is executed but its result is not considered for the next then()/catch() statements. So the execution of this call stack only depends on the `pluginsManager.trigger` statement.

This should be changed to: `.then(() => { return kuzzle.repositories.token.expire(context.token); })`

lib/api/controllers/authController.js
```diff
@@ -7,26 +7,53 @@ module.exports = function AuthController (kuzzle) {
...
+      .then(() => {
+        if (Boolean(context.connection.id)) {
```
scottinet: Should distinguish between REST actions (using `context.connection.type === 'rest'`) and a `logout()` request executed without a `context.connection.id`, which should be a `BadRequestError`.

Moreover, using generic `throws` leads to generic error, and Kuzzle errors are preferred. :-)

lib/api/core/hotelClerk.js
```diff
@@ -790,3 +793,44 @@ function subscribeToRoom (roomId, requestObject, context) {
...
+  this.add = function(token, context) {
+    tokenizedConnections[token._id] = {
+      expiresAt: token.expiresAt,
+      connection: context.connection
+    };
+
+    runTimer.call(this);
+  };
+
+  runTimer = function() {
```
scottinet: Nitpicking: looks good because I know what it is, but in a few months this may not be this clear. Maybe adding a comment explaining that this is a token garbage collector might help?

lib/api/core/models/repositories/tokenRepository.js
```diff
@@ -0,0 +1,261 @@
...
+      decodedToken = jwt.verify(token, kuzzle.config.jsonWebToken.secret);
+
+      if (decodedToken._id === 'admin') {
+        return this.admin();
+      }
+
+      this.load(token)
+        .then(userToken => {
+          if (userToken === null) {
+            error = new UnauthorizedError('Token invalid', 401);
```
scottinet: Nitpicking: "Invalid token"

features/support/hooks.js
```diff
@@ -71,12 +71,11 @@ var myHooks = function () {
...
-        callback();
-      }.bind(this))
-      .catch(function () {
-        callback();
-      });
+      .then(() => {
+        return this.api.createIndex('%kuzzle');
+      })
+      .then(() => callback())
+      .catch(() => callback());
```
scottinet: Nitpicking: you can use `q.finally`, which is executed once the promise is either fulfilled or rejected.

lib/api/controllers/authController.js
```diff
@@ -7,26 +7,46 @@ module.exports = function AuthController (kuzzle) {
...
+    kuzzle.pluginsManager.trigger('auth:logout', context)
+      .then(() => {
+        return kuzzle.repositories.token.expire(context.token);
+      })
+      .then(() => {
+        response = new ResponseObject(requestObject);
+        deferred.resolve(response);
+      })
+      .catch(function (err) {
+        deferred.reject(err);
```
scottinet: All Kuzzle responses are `ResponseObject`, successes and errors alike.
So a new `ResponseObject` must be created and provided to the rejected promise.

lib/api/controllers/authController.js
```diff
@@ -7,26 +7,46 @@ module.exports = function AuthController (kuzzle) {
...
+      });
+
+    return deferred.promise;
```
scottinet: Nitpicking: a deferred promise is not necessary here, returning `kuzzle.pluginsManager.trigger` should do.

lib/api/controllers/authController.js
```diff
@@ -7,26 +7,46 @@ module.exports = function AuthController (kuzzle) {
 
   this.passport = new PassportWrapper(kuzzle);
 
-  this.login = function (requestObject) {
+  this.logout = function (requestObject, context) {
+    var
+      deferred = q.defer(),
+      response;
```
scottinet: Nitpicking: should be moved inside the appropriate `then()` callback, to restrict variable scoping.

lib/api/controllers/authController.js
```diff
@@ -7,26 +7,46 @@ module.exports = function AuthController (kuzzle) {
...
+        deferred.resolve(response);
+      })
+      .catch(function (err) {
+        deferred.reject(err);
+      });
+
+    return deferred.promise;
+  };
+
+  this.login = function (requestObject, context) {
```
scottinet: Same remarks than for the `logout` method:

- (nitpicking) Using a deferred promise is not necessary
- In case of rejection, must return a `ResponseObject`

scottinet:
+1!

lib/api/core/hotelClerk.js
```diff
@@ -790,3 +793,61 @@ function subscribeToRoom (roomId, requestObject, context) {
 
   return deferred.promise;
 }
+
+
+/**
+ * Authentication token white-list.
+ *
+ * Maintain a list of valid token.
```
xbill82: Should be

    Maintains a list of valid tokens

lib/api/core/hotelClerk.js
```diff
@@ -790,3 +793,61 @@ function subscribeToRoom (roomId, requestObject, context) {
...
+ * When a token is invalidated, it remove all room subscriptions from connexion
```
xbill82: Should be

    When a token is invalidated, it removes all room subscriptions from connection.

jenow:
+1

xbill82:
:+1: !

lib/api/core/models/repositories/repository.js
```diff
@@ -230,14 +230,14 @@ Repository.prototype.loadFromCache = function (id, opts) {
...
-  if (this.cacheEngine === null) {
+  if (!this.cacheEngine) {
     return this.loadOneFromDatabase(id);
   }
 
   this.loadFromCache(id, opts)
     .then(function (object) {
       if (object === null) {
-        if (this.readEngine === null) {
+        if (!Boolean(this.readEngine)) {
```
benoitvidis: I don't get the condition. To me, the readEngine should either be set to null or to an readEngine service instance.
If the readEngine is an instance of ES service, my understanding here is that the object won't ever be fetched if not found in the cache.

PR #129, opened 2016-01-15
Action GET on /myIndex/myCollection/_search

AnthonySendra:
In case the user make a GET on url `/myIndex/myCollection/_search` or `/myIndex/myCollection/_bulk`, et, now we return a error.

Before that, on GET on url `/myIndex/myCollection/_search`  the response was something like

```
result: {
   hits: {
      hits: [ ... ],
      ...
   }
}
```

It's just an optional security

### DIFF
--- a/lib/api/controllers/routerController.js
+++ b/lib/api/controllers/routerController.js
@@ -358,6 +358,18 @@ function executeFromRest(params, request, response) {
   response.setHeader('Access-Control-Allow-Origin', '*');
   response.setHeader('Access-Control-Allow-Headers', 'X-Requested-With');
 
+  // Just in case the user make a GET on url /mainindex/test/_search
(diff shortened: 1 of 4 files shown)


scottinet:
For the record: the bug is not in our REST routes but in the ElasticSearch SDK: it creates a REST URL from the provided arguments, and if the document ID of a get route is `_search` it ends up with a search URL with an empty body.

Could be interesting to check if the bug is still here in ES 2.x, but in the meantime, we'll have to prevent looking for documents named `_search` in the `readEngine` service instead of the REST routes. Especially considering that it can be reproduced with other protocols ;-)

scottinet:
+1

xbill82:
:+1:

jenow:
+1

stafyniaksacha:
+1 & merge

PR #132, opened 2016-01-21
Kuz362 internal index security

scottinet:
- added a new `internalIndex` global configuration
- replaced all explicit references to the `%kuzzle` index with the new `internalIndex` configuration
- index wildcards now resolve all indexes except the internal one. Rights on the internal index must now be explicitly granted
- security bugfix: the `security` controller cannot be invoked anymore if no explicit rights have been granted on the internal index 
- added explicit rights on the internal index to the default `guest` and `admin` roles
- role and profile creation now use the user's context instead of a generic one


lib/api/core/models/repositories/roleRepository.js
```diff
@@ -173,7 +173,7 @@ module.exports = function (kuzzle) {
    */
   RoleRepository.prototype.validateAndSaveRole = function (role) {
```
stafyniaksacha: Why we don't give the real context there ?

lib/api/core/models/repositories/roleRepository.js
```diff
@@ -173,7 +173,7 @@ module.exports = function (kuzzle) {
...
     var context = {
-      connection: {type: 'websocket'},
+      connection: {type: 'internal'},
```
stafyniaksacha: Why we don't give the real context there ?

scottinet: I was afraid of potential side-effects, and too lazy to investigate further :stuck_out_tongue: 

It seems this function is only called by the security controller, so I'll add the real context to this one, and to other similar functions used by profiles and users management.

jenow:
+1

stafyniaksacha:
+1

ballinette:
+1

scottinet:
Got 3 approvals => merging

PR #135, opened 2016-01-25
Kuz373 write engine

scottinet:
* Kuzzle server instances now don't call the `writeEngine` directly anymore, but instead use write workers
* the global `repository` module now use a `writeLayer` option instead of a `writeEngine` service. The new layer only exposes a `execute` function, taking a `RequestObject` as argument, and returns a promise. A default `writeLayer` is provided directly in the repository implementation, pointing to write workers
* rewrote most of unit tests made on the `prepareDb` module, as they didn't test much
* removed unnecessary global variables in the `prepareDb` module
* fixed several bugs in `prepareDb.importMapping` 
* `prepareDb.importFixtures` now filters partial errors fired when documents already exist
* `repository` documentation updated


j33f:
+1

lib/api/prepareDb.js
```diff
@@ -1,24 +1,23 @@
...
-  actualIndexes = [],
   env = {
     fixtures: 'FIXTURES',
     mappings: 'DEFAULT_MAPPING'
-  },
-  data = {};
+  };
 
 module.exports = function () {
+  this.actualIndexes = [];
```
xbill82: Didn't you mean `currentIndexes`?

scottinet: Not quite. I didn't think about it when I refactored this piece of code, but now that you mention it, this variable is poorly named.
It contains all known indexes at the start of Kuzzle. So it should really be named `knownIndexes`.
But now that we maintain a cache containing all known indexes to handle multi-indexes, this variable should be removed altogether, and we should use this cache instead.

Problem is: I'm currently reworking this index cache because of a bug discovered yesterday by @AnthonySendra. So I propose to keep this variable for now, and in my next PR I'll remove it completely.

What do you think about this?

xbill82: :+1:

xbill82:
The PR is :+1: for me!

bin/kuzzle-start.js
```diff
@@ -41,6 +39,14 @@ module.exports = function () {
...
 ████████████████████████████████████`);
     })
+    .then(() => {
+      /*
+       Waits for at least one write worker to be connected to the server before trying to use them
+       */
+      return kuzzle.services.list.broker.waitForListeners(kuzzle.config.queues.workerWriteTaskQueue);
+    })
+    .then(() => { return kuzzle.cleanDb(); })
+    .then(() => { return kuzzle.prepareDb(); })
```
AnthonySendra: I'm not sure about running cleanDb after the message "Kuzzle is started". Because just after, we have to wait for workers

AnthonySendra:
+1

scottinet:
3 "+1" obtained => I'm merging this PR

PR #136, opened 2016-01-26
Kuz 369 new "replace" route

dbengsch:
Addition of action replace in controller write.
Creation of a new notification type.
Unitary and Functional tests.

### DIFF
--- a/lib/api/core/notifier.js
+++ b/lib/api/core/notifier.js
@@ -55,7 +55,7 @@ module.exports = function NotifierController (kuzzle) {
             if (response.action === 'publish') {
               response.state = 'done';
             }
-            else if (response.action === 'create' || response.action === 'createOrUpdate') {
+            else if (response.action === 'create' || response.action === 'createOrUpdate' || response.action === 'replace') {
               /*
                since we have the complete document, we use the cache to avoid performing another testFilter when
                notifying about document creations
@@ -100,6 +100,7 @@ function workerNotification (serializedResponseObject) {
     create: notifyDocumentCreate,
     createOrUpdate: (responseObject.data && responseObject.data.created) ? notifyDocumentCreate : notifyDocumentUpdate,
     update: notifyDocumentUpdate,
+    replace: notifyDocumentReplace,
     delete: notifyDocumentDelete,
     deleteByQuery: notifyDocumentDelete
   };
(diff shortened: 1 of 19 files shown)


ballinette:
+1

lib/api/core/notifier.js
```diff
@@ -144,13 +145,56 @@ function notifyDocumentCreate (responseObject) {
...
  * @return {Promise} number of notified rooms
  */
+function notifyDocumentReplace (responseObject) {
+  var
+    matchedRooms,
+    deferred = q.defer();
+
+  responseObject.scope = 'in';
+
+  this.dsl.testFilters(responseObject)
```
scottinet: We already pushed the result of `dsl.testFilters` in the `publish` method, to avoid calling it twice, as it is an expensive method performance-wise.
So this line must be replaced by a cache retrieval, like it is done in the `notifyDocumentCreate` method.

AnthonySendra:
+1

lib/services/elasticsearch.js
```diff
@@ -264,6 +264,47 @@ module.exports = function (kuzzle, options) {
...
+    this.client.exists({
+      index: data.index,
+      type: data.type,
+      id: data.id
+    })
+      .then((exists) => {
+        if (exists) {
+          this.client.index(data)
+            .then(function (result) {
+              indexCache.add.call(kuzzle, data.index, data.type);
```
scottinet: Since the document exists, then so do its index and collection. So there is no need to add them in the cache.

Moreover, the index cache has been overhauled in PR #137

test/api/controllers/writeController.test.js
```diff
@@ -30,6 +30,7 @@ describe('Test: write controller', function () {
     should(kuzzle.funnel.write.create(requestObject)).be.rejected();
     should(kuzzle.funnel.write.createOrUpdate(requestObject)).be.rejected();
     should(kuzzle.funnel.write.update(requestObject)).be.rejected();
+    should(kuzzle.funnel.write.replace(requestObject)).be.rejected();
```
scottinet: Hm, well, this is awkward: this test cannot work as it is.
ShouldJS only test the result of promises when returning them, like this:
```js
return should(<some method returning a promise>).be.rejected();
```

So this test must be split into 4 distinct tests to work.

lib/services/elasticsearch.js
dbengsch: indexCache call removed

scottinet:
+1

PR #137, opened 2016-01-26
fixed index cache feature

scottinet:
Bugfix: the index cache was handled by the elasticsearch service, called both by the server instance and its workers. Problem is: workers didn't have access to the index cache, thus leading to non-catched errors.

* created a new `indexCache` core component
* removed all index cache management from the elasticsearch service
* replaced `kuzzle.indexes` with `kuzzle.indexCache`, using the new core component
* the index cache is now refreshed by the `write` and `admin` controllers
* `cleanDb` and `prepareDb` now use and maintain the index cache
* `cleanDb` and `prepareDb` are now independant to Kuzzle to avoid adding unnecessary objects to the main context
* added new console messages to make the Kuzzle startup sequence clearer

### DIFF
--- a/lib/api/core/models/security/role.js
+++ b/lib/api/core/models/security/role.js
@@ -13,17 +13,17 @@ function Role () {
 }
 
 function doesIndexExist(requestObject, indexes) {
-  return indexes[requestObject.index] !== undefined;
+  return indexes && indexes[requestObject.index] !== undefined;
 }
 
 function doesCollectionExist(requestObject, indexes) {
-  return _.contains(indexes[requestObject.index], requestObject.collection);
+  return indexes && _.contains(indexes[requestObject.index], requestObject.collection);
 }
 
 /**
  * @param requestObject
- * @param context
- * @param indexes
+ * @param context - user's context
+ * @param indexes - list of existing indexes/collections
  * @returns {*}
  */
 Role.prototype.isActionAllowed = function (requestObject, context, indexes) {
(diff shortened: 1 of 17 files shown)


stafyniaksacha:
+1!

lib/api/core/models/security/role.js
```diff
@@ -53,10 +53,10 @@ Role.prototype.isActionAllowed = function (requestObject, context, indexes) {
     return false;
   }
 
-  if (this.indexes._canCreate !== undefined &&
+  if (!doesIndexExist(requestObject, indexes) &&
+    this.indexes._canCreate !== undefined &&
     !this.indexes._canCreate &&
-    _.contains(['import', 'create', 'createCollection', 'putMapping', 'createOrUpdate'], requestObject.action) &&
-    !doesIndexExist(requestObject, indexes)) {
+    _.contains(['import', 'create', 'createCollection', 'putMapping', 'createOrUpdate'], requestObject.action)) {
```
AnthonySendra: (Quick note: dependencies with @dbengsch PR #138 )

AnthonySendra:
+1

lib/api/core/models/security/role.js
scottinet: Indeed. Good news is: merging this will be easy as in this PR, I only moved the last test to the first position, to make it clearer that it handles the case when an index or a collection doesn't exist.

benoitvidis:
+1

PR #138, opened 2016-01-27
Kuz 369 put verb

dbengsch:
change route putMapping by createOrReplaceMapping
change route putRole by createOrReplaceMapping
change route putProfile by createOrReplaceProfile

/!\ Has incidence on API doc and SDK implementations


lib/api/controllers/routerController.js
```diff
@@ -153,13 +153,13 @@ module.exports = function RouterController (kuzzle) {
...
         {verb: 'delete', url: '/:index/:collection', controller: 'admin', action: 'deleteCollection'},
 
-        {verb: 'put', url: '/roles/:id', controller: 'security', action: 'putRole'},
-        {verb: 'put', url: '/profiles/:id', controller: 'security', action: 'putProfile'},
+        {verb: 'put', url: '/roles/:id', controller: 'security', action: 'createOrReplaceRole'},
+        {verb: 'put', url: '/profiles/:id', controller: 'security', action: 'createOrReplaceProfile'},
         {verb: 'put', url: '/:index', controller: 'admin', action: 'createIndex'},
         {verb: 'put', url: '/:index/:collection', controller: 'write', action: 'createCollection'},
-        {verb: 'put', url: '/:index/:collection/_mapping', controller: 'admin', action: 'putMapping'},
+        {verb: 'put', url: '/:index/:collection/_mapping', controller: 'admin', action: 'createOrReplaceMapping'},
```
AnthonySendra: I have a doubt about the decision `createOrReplaceMapping`. Because we can't replace a mapping, just add some rules. Are you sure this is what we decided ?

dbengsch: I quiet remember the last argument exposed as "createOrReplaceSomeRulesIfPossibleInMapping" was a bit too long and finalized as "createOrReplaceMapping" would do the work.

scottinet: This route cannot be named `createOrReplace`, since we cannot replace a whole mapping.
This route behave exactly like an `update` does, so I suggest we name it like that.

dbengsch: updateMapping ?

scottinet: I would prefer an `update` action in a `mapping` controller, but that would mean reworking the whole API to make it homogeneous.
So for now `updateMapping` will do. :-)

scottinet:
+1

stafyniaksacha:
+1

AnthonySendra:
+1

PR #139, opened 2016-01-27
subscription on null filters bugfix

scottinet:
Subscriptions made on empty filters and on null/undefined filters generated different room IDs, whereas the same room ID should be generated.


xbill82:
:+1:

jenow:
+1

stafyniaksacha:
+1

PR #145, opened 2016-02-01
Bugfixes

scottinet:
- Moved `getCurrentUser` from the `security` controller to the `auth` one
- Changed numerous hook names in security controller from `data:...` to `security:...`
- Bugfix: profiles weren't correctly linked to users, and roles to profiles, due to the parent ID not been correctly copied by the `repository` core component
- Bugfix: profile searches in functional tests didn't prefix the role id
- Bugfix: the `prepareDb` mapping import still used the `putMapping` API route instead of the `updateMapping` one
- Bugfix: `%kuzzle/profiles` and `%kuzzle/users` collections used default elasticsearch mapping. This lead to problems with profiles and roles searches, as `%kuzzle/users/profile` and `%kuzzle/profiles/roles` fields are storing documents ID while being indexed. This means that looking for roles or profiles would potentially return false negatives. Kuzzle's `prepareDb` module now creates the internal index, and the `users` and `profiles` collections with appropriate mappings

### DIFF
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -64,7 +64,7 @@ module.exports = function AuthController (kuzzle) {
           return q(new ResponseObject(requestObject, {found: false}));
         }
 
-        if (requestObject.data.body.hydrate !== undefined && requestObject.data.body.hydrate === false) {
+        if (!requestObject.data.body.hydrate) {
           response = {
             _id: user._id,
             _source: kuzzle.repositories.user.serializeToDatabase(user)
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -1,7 +1,6 @@
 var
   q = require('q'),
   PassportWrapper = require('../core/auth/passportWrapper'),
-  BadRequestError = require('../core/errors/badRequestError'),
   ResponseObject = require('../core/models/responseObject');
 
 module.exports = function AuthController (kuzzle) {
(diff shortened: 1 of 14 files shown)


stafyniaksacha:
+1

lib/api/controllers/authController.js
```diff
@@ -44,4 +44,41 @@ module.exports = function AuthController (kuzzle) {
...
+
+    return kuzzle.repositories.user.load(requestObject.data._id)
+      .then(user => {
+        var response;
+
+        if (!user) {
+          return q(new ResponseObject(requestObject, {found: false}));
+        }
+
+        if (requestObject.data.body.hydrate !== undefined && requestObject.data.body.hydrate === false) {
```
AnthonySendra: On other routes, by default, hydrate is false. `if (!requestObject.data.body.hydrate)` is enough ?

AnthonySendra:
+1

ballinette:
+1

scottinet:
:warning: DON'T MERGE :warning: 

I found a new bug in functional tests (fixed), leading me to a new one in Kuzzle: the `profiles` field in user documents and the `roles` field in profile documents are indexed, meaning they are parsed by elasticsearch.
This means that looking for roles or profiles using the `searchProfiles` or `searchRoles` routes do not work whenever a profile or role ID contains what ES considers to be a separator.

I'm adding an internal index preparation in the `prepareDb` module with mappings forcing these fields to not be indexed. This will take me some time.

PR #149, opened 2016-02-01
Allows createUser to create a user without passing an id.

benoitvidis:
Forces updateUser to actually use the update action

### DIFF
--- a/lib/api/controllers/securityController.js
+++ b/lib/api/controllers/securityController.js
@@ -1,6 +1,7 @@
 var
   _ = require('lodash'),
   User = require('../core/models/security/user'),
+  uuid = require('node-uuid'),
   q = require('q'),
   BadRequestError = require('../core/errors/badRequestError'),
   ResponseObject = require('../core/models/responseObject');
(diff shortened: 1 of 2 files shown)


ballinette:
+1

lib/api/controllers/securityController.js
```diff
@@ -329,6 +330,9 @@ module.exports = function SecurityController (kuzzle) {
         if (newRequestObject.data._id !== undefined) {
           pojoUser._id = newRequestObject.data._id;
         }
+        else {
+          pojoUser._id = uuid.v4();
```
scottinet: Why don't we put `'anonymous'` instead, so that the user repository attributes an anonymous profile?
That would be more efficient than reading the cache, and then the database, to search for a purposely unknown user.

benoitvidis: This was actually the situation this PR tries to solve.

The createUser action should create a real user when no id is provided and not update the anonymous user.

scottinet: Oh, ok. I clearly misunderstood the purpose of this PR, sorry :-)

scottinet:
+1

xbill82:
:+1:

PR #152, opened 2016-02-03
OAUTH authentication

jenow:
- Added route for login with a strategy
- Added a trigger to `passport:loadScope` for authentication
- Added additional data to get the redirect url from any OAUTH service to the authentication method


ballinette:
:+1:

benoitvidis:
+1

PR #154, opened 2016-02-03
Token expiration notification

scottinet:
- expired tokens now triggers a new `jwtTokenExpired` notification to all channels the user subscribed to
- bugfix: tokens were expired immediately instead of at their expiration time
- bugfix: expired tokens were kept in cache instead of being cleaned up
- moved token management from the `hotelClerk` to the new `tokenManager` core component
- added a new `protocol:notify` hook, allowing protocols to notify 1 user on all its channels by delivering a payload
- updated plugins documentation
- updated the socket.io plugin to version 1.0.6
- added unit & functional tests
- hooks configuration: moved some `data:...` hooks to `security:...`, and added all missing `security:...` hooks

### DIFF
--- a/config/defaultPlugins.json
+++ b/config/defaultPlugins.json
@@ -11,7 +11,7 @@
     }
   },
   "kuzzle-plugin-socketio": {
-    "version": "1.0.5",
+    "version": "1.0.6",
     "activated": true,
     "name": "kuzzle-plugin-socketio",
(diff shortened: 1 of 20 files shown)


j33f:
:+1:

AnthonySendra:
+1

ballinette:
:+1:

PR #156, opened 2016-02-04
Read user should not send the password

xbill82:
**Warning** This PR should be merged only after the authentication plugins have been updated.

The `getCurrentUser` call in the `authController` currently sends the user's (encrypted) password through the wire. This is bad, isn't it?

### DIFF
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -80,6 +80,7 @@ module.exports = function AuthController (kuzzle) {
             _id: user._id,
             _source: user
           };
+          delete response._source.password;
         }
         delete response._source._id;
 
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -65,7 +65,7 @@ module.exports = function AuthController (kuzzle) {
         var response;
 
         if (!user) {
-          return q(new ResponseObject(requestObject, {found: false}));
+          return q({found: false});
         }
 
         if (requestObject.data.body.hydrate !== undefined && requestObject.data.body.hydrate === false) {
@@ -82,10 +82,10 @@ module.exports = function AuthController (kuzzle) {
         }
         delete response._source._id;
 
-        return q(new ResponseObject(requestObject, response));
+        return  kuzzle.pluginsManager.trigger('auth:getCurrentUser', response);
       })
       .then(response => {
-        return  kuzzle.pluginsManager.trigger('auth:getCurrentUser', response);
+        return q(new ResponseObject(requestObject, response));
       });
   };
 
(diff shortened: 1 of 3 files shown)


xbill82:
After discussion with @scottinet we decided to delegate the password deletion to the authentication plugins. To enable this, the hooks have been moved to the bottom of the read calls for users, namely

* `getCurrentUser`
* `searchUsers`
* `getUser`

stafyniaksacha:
+1

scottinet:
+1

ballinette:
OK, so, before merging this, we should update https://github.com/kuzzleio/kuzzle-plugin-auth-passport-local and add a pipe on "auth:getCurrentUser" event that remove the password from the responseObject ?

scottinet:
@ballinette > Yes. And also on `security:searchUsers` and on `security:getUser`

ballinette:
OK, I'm on it ;)

lib/api/controllers/authController.js
```diff
@@ -84,6 +83,9 @@ module.exports = function AuthController (kuzzle) {
         delete response._source._id;
 
         return q(new ResponseObject(requestObject, response));
+      })
+      .then(response => {
```
ballinette: shouldn't we better : 
* 1st call the trigger on "response"
* then return the ResponseObject with the modified response
?

I mean:
    return kuzzle.repositories.user.load(requestObject.data._id)
      .then(user => {
        var response;
(....)
        delete response._source._id;

        return  kuzzle.pluginsManager.trigger('auth:getCurrentUser', response);
      })
      .then(response => {
        return q(new ResponseObject(requestObject, response));
      });
  };

ballinette:
@xbill82 I made the PR on plugin auth local that removes the password: https://github.com/kuzzleio/kuzzle-plugin-auth-passport-local/pull/6
... but it's necessary to fix the trigger calls according to my remark above

* then merge and the plugin PR and deploy a new plugin version
* then fix the plugin version within defaultPlugins.json

scottinet:
Since each Kuzzle version embarks specific plugin versions, I think you can simply deploy the new plugin version, and then update the `defaultPlugins.json` on this PR.

lib/api/controllers/authController.js
xbill82: :+1: If @scottinet is ok anyone can make the change!

ballinette:
mmmhh changed made on triggers call...
.... but this PR will have conflicts with PR #158 from @AnthonySendra 

I suggest to cancel this one and just add a trigger on  formatUserForSerialization() function on Anthony's PR.

xbill82:
So, should we close this PR?

ballinette:
Yep.
The stuff it solves must be rewritten while fixing issue #159

PR #158, opened 2016-02-04
Fix security

AnthonySendra:
**Bugfix**
* Always return `_source` attribute (also in nested profile/role)
* Return `NotFoundError` on `getUser`, `getProfile` and `getRole` when document with id doesn't exist
* Store in memory (cache) profiles and roles
* In functionnal testing, on `searchUsers`, `searchProfiles` and `searchRoles`, only check results for `world.idPrefix`. Avoid to mix with role/user/profil from real user data

**Clean**
* In `authController` use `getUser` for function `kuzzle.funnel.security.getCurrentUser` instead of reimplement the function

### DIFF
--- a/test/api/controllers/securityController/users.test.js
+++ b/test/api/controllers/securityController/users.test.js
@@ -75,7 +75,7 @@ describe('Test: security controller - users', function () {
         .catch(error => { done(error); });
     });
 
-    it('should reject wit NotFoundError when the user is not found', () => {
+    it('should reject with NotFoundError when the user is not found', () => {
       var promise = kuzzle.funnel.security.getUser(new RequestObject({
         body: { _id: 'i.dont.exist' }
       }));
--- a/test/api/controllers/securityController/users.test.js
+++ b/test/api/controllers/securityController/users.test.js
@@ -75,7 +75,7 @@ describe('Test: security controller - users', function () {
         .catch(error => { done(error); });
     });
 
-    it('should respond found:false when the user is not found', () => {
+    it('should reject wit NotFoundError when the user is not found', () => {
       var promise = kuzzle.funnel.security.getUser(new RequestObject({
         body: { _id: 'i.dont.exist' }
       }));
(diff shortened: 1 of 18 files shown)


stafyniaksacha:
+1

ballinette:
not directly related to the changes you made, but I found something strange within formatProfileForSerialization and formatUserForSerialization functions:
if "hydrate" is true, the response will be overrided and all its additional attributes deleted.

Instead of,
  var formatUserForSerialization = function (user, hydrate) {
    if (hydrate) {
      return {
        _id: user._id,
        _source: {
          profile: formatProfileForSerialization(user.profile, true)
        }
      };
    }
    return user;
  };

we should have something like:

  var formatUserForSerialization = function (user, hydrate) {
    var response = { _id: user._id, _source: user }
    if (hydrate) {
      response._source.profile = formatProfileForSerialization(user.profile, true);
      };
    }
    delete response._source._id;
    return response;
  };

ballinette:
:+1:

AnthonySendra:
Close the PR for update it with feedback from issue #159

scottinet:
About the last issue you raised, the one about the `authController`: I reimplemented the function because of the `security:getUser` hook. I don't think this is a good idea to fire this hook when performing a `auth:getCurrentUser` action.

To avoid repetition of code, some of the `getUser` code could be moved to the user repository, and both `auth:getCurrentUser` and `security:getUser` could invoke this code.

What do you think?

AnthonySendra:
@scottinet Yep, good point, I'm on it.

ballinette:
agree with @scottinet

test/api/controllers/securityController/users.test.js
```diff
@@ -74,31 +75,12 @@ describe('Test: security controller - users', function () {
...
-
-          done();
-
-        })
-        .catch(error => { done(error); });
-    });
-
-    it('should respond found:false when the user is not found', done => {
-      kuzzle.funnel.security.getUser(new RequestObject({
+    it('should reject wit NotFoundError when the user is not found', () => {
```
dbengsch: typo on `with` ?

scottinet:
+1

AnthonySendra:
As discussed with @scottinet, the hook `auth:getCurrentUser` is removed because plugin can know (or in the future, we have to let the plugin know) who is the user and can do something specific when he tries to get himself.

PR #160, opened 2016-02-05
Add new routes createRole createProfile in security

AnthonySendra:
* Add missing routes in `routerController`
* Add `createRole` and `createProfile` in `securityController`

**Note:** come with [kuzzle-api-documentation PR#22](https://github.com/kuzzleio/kuzzle-api-documentation/pull/22)


ballinette:
+1

lib/api/controllers/securityController.js
```diff
@@ -385,3 +388,31 @@ module.exports = function SecurityController (kuzzle) {
   };
 
 };
+
+/*eslint vars-on-top: 0*/
```
scottinet: Question: instead of having an eslint guard, why do you not define `function createOrReplaceRole (...)`?

lib/api/controllers/securityController.js
```diff
@@ -385,3 +388,31 @@ module.exports = function SecurityController (kuzzle) {
...
+
+var createOrReplaceRole = function (requestObject, context, opts) {
+  this.pluginsManager.trigger('security:createRole', requestObject);
```
scottinet: Either the `security:createOrReplace*` are removed, and they should be removed from `config/hooks.js` too, or this hook must depend on the `opts` content

stafyniaksacha:
+1

benoitvidis:
+1

PR #161, opened 2016-02-08
KUZ-363 : Roles: add a indexes._canDelete right

ballinette:
* fix bug about index creation rights: it's no more possible to create an index if we don't are allowed to (+ add unit test)
* add a "_canDelete" role on indexes and collections
* refactor the "deleteIndexes" action: it now delete only the indexes we are allowed to, and give the list of deleted indexes within the response.
* add unit tests.


ballinette:
See also PR https://github.com/kuzzleio/kuzzle-api-documentation/pull/24 in API Documentation

lib/services/elasticsearch.js
```diff
@@ -542,21 +542,28 @@ module.exports = function (kuzzle, options) {
...
    * @return {Promise}
    */
   this.deleteIndexes = function (requestObject) {
+    var
+      allowedIndexes = requestObject.data.body.indexes,
+      deletedIndexes = [];
+
     delete requestObject.data.body;
 
     return this.listIndexes(requestObject)
```
scottinet: Since we already use `readEngine.listIndexes` in the `adminController` to get all the indexes that can be deleted, is it necessary to perform another `listIndexes` in the `writeEngine.deleteIndexes` method?

scottinet:
I'm a bit uneasy with the `deleteIndexes` route. If a user with rights on `%kuzzle` execute this API method, he'll also remove the internal index.
Maybe this route should exclude the internal index from the list? The idea behind this proposal is to exclude the internal index from any API route executing mass operations on indexes, like we prevented the internal index to be resolved by the index wildcard.

What do you think about this?

lib/services/elasticsearch.js
ballinette: yes, because it's not exactly the same: the adminController list all indexes that the user is allowed to READ.
Here, we need the indexes that the user is allowed to delete.

ballinette:
Totally agree, and it's actually what the code does ;)
see lines 108-113 of adminController, with the condition "availableIndex !== internalIndex" (where internalIndex is "%kuzzle"):

    if ((availableIndex === roleIndex ||
          roleIndex === '*' && availableIndex !== internalIndex) &&
          _.indexOf(allowedIndexes, availableIndex) === -1
    ) {
        allowedIndexes.push(availableIndex);
    }

lib/services/elasticsearch.js
ballinette: oh no, sorry, I didn't not understand your remark

scottinet:
epic fail

lib/services/elasticsearch.js
ballinette: hmm well, I need also to refactor cleanDB, that uses internally the deleteIndexes without filters... 
work in progress...

benoitvidis:
+1

lib/services/elasticsearch.js
ballinette: done, with the fix in cleanDB

scottinet:
+1

AnthonySendra:
+1

PR #166, opened 2016-02-11
token manager bugfix

scottinet:
Bugfix: the token manager didn't check the availability of the `connection` part of a connection context before adding the token to the cache. When the token expired, the token garbage collector tried to access to this connection object, and crashed.
This bug caused some unit tests to fail randomly.


ballinette:
+1

dbengsch:
+1

j33f:
:+1:

PR #170, opened 2016-02-15
Fix #159: format user for serialization hydrate

### LINKED ISSUE #159: securityController -- formatUserForSerialization and formatProfileForSerialization do not hydrate properly
I found something strange within securityController, in formatProfileForSerialization and formatUserForSerialization functions (see https://github.com/kuzzleio/kuzzle/blob/master/lib/api/controllers/securityController.js#L34 ):

if "hydrate" is true, the response will be overrided and all its additional attributes deleted.

Instead of,

    var formatUserForSerialization = function (user, hydrate) {
      if (hydrate) {
        return {
          _id: user._id,
          _source: {
            profile: formatProfileForSerialization(user.profile, true)
          }
        };
      }
      return user;
    };

we should have something like:

    var formatUserForSerialization = function (user, hydrate) {
      var response = { _id: user._id, _source: user }
      if (hydrate) {
        response._source.profile = formatProfileForSerialization(user.profile, true);
        };
      }
      delete response._source._id;
      return response;
    };

ballinette:
fix issue #159 
+ add a trigger on formatUserForSerialization() method to permit plugins to alter user data before sending them to client (for example, hide user password: PR following up to plugin kuzzle-plugin-auth-passport-local)


AnthonySendra:
+1

scottinet:
+1

PR #171, opened 2016-02-15
follow #159: invoke formatUserForSerialization() also for createUser and createOrReplaceUser

ballinette:
Because these methods also give back the user object in the response, we need to format user data and trigger the security:formatUserForSerialization pipe, to permit plugins to, for example, remove the password from the ResposeObject.


benoitvidis:
+1

stafyniaksacha:
+1

PR #179, opened 2016-02-29
add auth:checkToken into the anonymous & default role

stafyniaksacha:
Used to check if we can restore authentication from token, same policy as login


j33f:
:+1:

ballinette:
+1

j33f:
+1

PR #183, opened 2016-03-04
Enhance closures with fetch

AnthonySendra:
Refactor `isActionAllowed` in order to add the capability to fetch documents and to inject them into the test to execute.

Add magic words:
* $currentId: in `args` the value corresponds to the id of the current. For exemple with a delete action, the `$currentId` corresponds to the document id aimed to delete.
* $currentUserId: shortcut for `context.token.user._id`
* $requestObject : The request object, can be both used in fetch parameters and test function

Example with `get` fetch :
```json
"write": {
  "actions": {
    "create": true,
    "delete": {
      "args": {
        "document": {
          "collection": "messages",
          "index": "klack",
          "action": {
            "get": "$currentId"
          }
        }
      },
      "test": "return document.content.user.id === $currentUserId"
    }
  }
}
```
Example with `mget` fetch :
```json
"write": {
  "actions": {
    "create": true,
    "delete": {
      "args": {
        "documents": {
          "index": "klack",
          "collection": "messages",
          "action": {
            "mget": [
              "AVNB4LVmQxRehRYp0SDn",
              "AVNB5ioCQxRehRYp0SED"
            ]
          }
        }
      },
      "test": "return args.documents[0].content.user.id === $currentUserId"
    }
  }
}
```
Example with `search` fetch :
```json
"write": {
  "actions": {
    "create": true,
    "delete": {
      "args": {
        "documents": {
          "index": "klack",
          "collection": "messages",
          "action": {
            "search": {
              "filter": {
                "ids": {
                  "values": [
                    "AVNBmrAkQxRehRYp0SDN"
                  ]
                }
              }
            }
          }
        }
      },
      "test": "return args.documents[0].content.user.id === $currentUserId"
    }
  }
}
```

Example with `search` and the $requestObject arg :
```json
"create": {
  "args": {
    "documents": {
      "index": "klack",
      "collection": "channels",
      "action": {
        "search": {
          "filter": {
            "term": {
              "name": "$requestObject.data.body.name"
            }
          }
        }
      }
    }
  },
  "test": "return args.documents.length === 0;"
}
```


lib/api/controllers/readController.js
```diff
@@ -25,32 +25,11 @@ module.exports = function ReadController (kuzzle) {
     return kuzzle.services.list.readEngine.count(requestObject);
   };
 
-  this.listCollections = function (requestObject, context) {
+  this.listCollections = function (requestObject) {
```
ballinette: why did you remove filterCollections() method ?
We need to list  only collections that the user is allowed to read, that was the purpose of this method (that probably should be fixed to follow your changes).

I don't see how this kind of filter is implemented in your PR.

lib/api/controllers/readController.js
```diff
@@ -60,22 +39,20 @@ module.exports = function ReadController (kuzzle) {
     if (type === 'stored') {
       return kuzzle.services.list.readEngine.listCollections(requestObject).then(response => {
         response.data.body.type = type;
-        return q(filterCollections(response, 'stored'));
+        return q(response);
```
ballinette: see comment above

lib/api/controllers/readController.js
```diff
@@ -60,22 +39,20 @@ module.exports = function ReadController (kuzzle) {
...
       });
     }
 
     realtimeCollections = kuzzle.hotelClerk.getRealtimeCollections();
 
     if (type === 'realtime') {
-      return q(filterCollections(new ResponseObject(requestObject, {type, collections: {realtime: realtimeCollections}}), 'realtime'));
+      return q(new ResponseObject(requestObject, {type, collections: {realtime: realtimeCollections}}), 'realtime');
```
ballinette: see comment above

lib/api/controllers/readController.js
dbengsch: As discussed with other team members, the listCollection action should only be restricted by the action rights. Filter collections can be more harmful as a user experience than a list of collections they can't read (as the unauthorized error will trigger to let them know).

ballinette: hmm well, I don't really agree with that... If a user can see a collection that he is not allowed to use, it can be a security issue. It's safer to show only the things he can use.

Anyway, if we decide to change that:
1. this has nothing to do with actual PR (it's not closure-related) and should be proposed in another PR
2. to be consistent, we have to to the same with listIndexes action.

lib/api/core/models/security/role.js
```diff
@@ -352,3 +345,193 @@ Role.prototype.validateDefinition = function (context) {
...
+      this.closures[path].getArgsDefinitions = global.eval('(function (requestObject) {return ' +
+        JSON
+          .stringify(actionRights.args)
+          .replace(/"\$(requestObject\.data\.[a-zA-Z0-9_\-.]*[a-zA-Z0-9])"/g, '$1')
+          .replace('"$currentId"', 'requestObject.data._id') +
+        ';})');
+      /* jshint evil: false */
+    }
+    catch (err) {
+      error = new InternalError('Error parsing rights for role ' + this._id + ' (' + path.join('/') + ') :' + actionRights);
```
scottinet: If I understand correctly, a parse error occurs if an administrator has configured an invalid closure.
If so, then instead of an `InternalError`, I would use either a new `ParseError` or a new `InvalidRightError` (or `InvalidClosureError`) error.
To me, `InternalError` should be reserved to unhandled Kuzzle errors

lib/api/core/models/security/role.js
```diff
@@ -352,3 +345,193 @@ Role.prototype.validateDefinition = function (context) {
...
+            actionRights.test +
+            '\nreturn false;\n })($requestObject, context, $currentUserId, args)');
+
+          this.closures[path].test = sandboxScript;
+        }
+        catch (err) {
+          message = `Error parsing closure rights for role ${this._id} (${path.join('/')}): ${actionRights}`;
+
+          kuzzle.pluginsManager.trigger('log:error', message);
+          error = new InternalError(message);
```
scottinet: Same here, I think that another kind of error should be used.

lib/api/core/models/security/role.js
```diff
@@ -352,3 +345,193 @@ Role.prototype.validateDefinition = function (context) {
...
+ * @param {String} argName - the argument name set by user (i.e.: document, documents...)
+ * @param {Object} argDefinition - definition set by user. Like {collection: 'messages', index: 'chat', action: { get: '$currentId' }}
+ * @returns {Function} the function that will execute a get/mget/search on readEngine
+ */
+function factoryFunctionClosure (argName, argDefinition) {
+  var methodName;
+
+  if (!argDefinition.collection || !argDefinition.index || !argDefinition.action || _.isEmpty(argDefinition.action)) {
+    this.pluginsManager.trigger('log:error', `Bad format in closure rights for ${argName}`);
+    return callback => callback(null, {});
```
scottinet: Why are we returning a silent failure?
Wouldn't it be better to return `callback => callback(new BadRequestError(...))`?

lib/api/core/models/security/role.js
```diff
@@ -352,3 +345,193 @@ Role.prototype.validateDefinition = function (context) {
...
+  }
+
+  methodName = Object.keys(argDefinition.action)[0];
+
+  if (['get', 'mget', 'search'].indexOf(methodName) === -1) {
+    this.pluginsManager.trigger('log:error', `Try to use an unauthorized function (${methodName}) in closure rights check`);
+    return callback => callback(null, {});
```
scottinet: same here

lib/api/core/models/security/role.js
```diff
@@ -352,3 +345,193 @@ Role.prototype.validateDefinition = function (context) {
...
+
+  // Build the object that will be passed to parallel
+  _.forEach(argsDefinitions, (argDefinition, argName) => {
+    argsFunctions[argName] = factoryFunctionClosure.call(this, argName, argDefinition);
+  });
+
+  async.parallel(argsFunctions, (error, results) => {
+    if (error) {
+      // In case we have an error we want to return an empty object because the error can come from readEngine
+      return deferred.resolve({});
```
scottinet: If the closure fails, shouldn't we reject the promise?

scottinet:
Apart from the error types I have noted: +1 :)

lib/api/controllers/readController.js
AnthonySendra: As discussed, the `listIndex` was refactored in PR #185

stafyniaksacha:
+1

benoitvidis:
+1

PR #185, opened 2016-03-08
Remove filter collection on listCollection once it is allowed in funnel

dbengsch:


j33f:
+1

dbengsch:
On hold, waiting AnS and SeC to discuss about it.

scottinet:
+1
As discussed this morning, it is way simpler to show collections and let users receive a security error if attempting to perform an unauthorized action on them, than hiding them.

As @ballinette mentioned this morning, it should also be done on the `listIndexes` API route

stafyniaksacha:
Ok for me +1

PR #194, opened 2016-03-17
CLI like a virgin

j33f:
Put the "like a virgin" (reset) process into the CLI + small CLI refactor

Note: the env var are still usable for this to be done

* a fixture and a mapping files can be specified 
* the reset process can be done at startup along with fixtures and mapping from the ./bin/kuzzle start comand


bin/README.md
```diff
@@ -38,13 +38,47 @@ $ kuzzle disable <service> <PID|all>
...
+
+# Reset Kuzzle
+
+```
+$ kuzzle likeAvirgin
+```
+
+will allow you to reset Kuzzle and restore it as if it is freshly installed.
+
+If you need to be connected as an administrative user, an interrective prompt will ask you for a username and a password
```
dbengsch: interactive

PR #204, opened 2016-03-18
[BETA] Load plugin by path (path added to configuration and is prioritary ov…

dbengsch:
…er url and npm package


AnthonySendra:
+1

PR #208, opened 2016-03-21
CLI + Remote actions refactor

j33f:
Put the "like a virgin" (reset) process into the CLI + CLI / remote actions refactor

# Concerning the Like a virgin functionnality
a fixture and a mapping files can be specified
the reset process can be done at startup along with fixtures and mapping from the ./bin/kuzzle start comand

Note: the env var are still usable, same thing for the fixtures and mappings, but the CLI arguments have precedence

# Concerning the remote actions 

- a queue is defined for all remotes actions
- the remote action controller logic resides in a file in the lib/api/controllers/remoteActions folder
- the remote action must be added in the lib/api/controllers/remoteActionsController.js file

To call the controller, an entry point is defined in api/remoteActions. Each entry points must have : 
- a prepareData function wich prepare the data from the caller script (ie the CLI) in order to pass them to the controller
- an onListenCB function wich is a callback called when the action is done
- a timeOutCB function wich is a callback called when the kuzzle server is not reachable
- an isPidMandatory var wich is a boolean indicating if we must provide a PID when calling the action or not

The remote actions can be called like : 
``` 
kuzzle.remoteActions.do('remoteActionName', params); like in the bin/commands/enable.js file
```


dbengsch:
+1

bin/README.md
```diff
@@ -47,9 +47,41 @@ $ kuzzle createFirstAdmin
 ```
 
 will guide you through the creation process of the first admin user and fix the rights to other user types if needed.
+<<<<<<< HEAD
```
benoitvidis: I guess it should not be here

bin/README.md
```diff
@@ -47,9 +47,41 @@ $ kuzzle createFirstAdmin
...
 
+# Reset Kuzzle
+
+```
+$ kuzzle likeAvirgin
+```
+
+will allow you to reset Kuzzle and restore it as if it is freshly installed.
+
+If you need to be connected as an administrative user, an interrective prompt will ask you for a username and a password
```
scottinet: nitpicking: "interactive" instead of "interrective"

bin/README.md
```diff
@@ -47,9 +47,41 @@ $ kuzzle createFirstAdmin
...
 
 **Note:** This command is interactive and let you choose to reset the roles rights or not.
 
+# Reset Kuzzle
+
+```
+$ kuzzle likeAvirgin
+```
+
+will allow you to reset Kuzzle and restore it as if it is freshly installed.
```
scottinet: To make it clearer, I would delete the note below, and include a full description of what this command does.

Something like this:
"Interactive command resetting Kuzzle data, and eventually Kuzzle rights"

What do you think?

bin/commands/reset.js
```diff
@@ -0,0 +1,54 @@
...
+      console.log(error('[✖] The file ' + params.mappings + ' cannot be opened... aborting.'));
+      process.exit(1);
+    }
+  }
+
+
+  console.log(warn('[ℹ] You are about to reset Kuzzle and make it like a virgin.'));
+  console.log(warn('[ℹ] This operation cannot be undone.\n'));
+
+  if (readlineSync.question('[❓] Are you sure ? If so, please type "I am sure" (if not just press [Enter]): ') === 'I am sure') {
```
scottinet: nitpicking: no space before interrogation mark. I would also remove the parenthesis, as any other input than "I am sure" will abort the process :-)

bin/kuzzle
```diff
@@ -8,38 +8,49 @@
...
 
-// $ kuzzle.js start
+// $ kuzzle start
 program
   .command('start')
   .option('--port [port]', 'Kuzzle port number', parseInt)
   .option('--server', 'start an instance without workers')
   .option('--worker', 'spawn a single set of workers instead of starting a Kuzzle instance')
+  .option('--likeAvirgin', 'reset before start, so that Kuzzle is like a virgin, touched for the very first time')
+  .option('--fixtures <fixtures>', 'allow you to specify a fixtures file')
```
scottinet: nitpicking: I would use "load" instead of "specify" to make this option clearer

lib/api/controllers/routerController.js
```diff
@@ -188,6 +188,7 @@ module.exports = function RouterController(kuzzle) {
         {verb: 'post', url: '/:index/:collection', controller: 'write', action: 'publish'},
 
         {verb: 'delete', url: '/_deleteIndexes', controller: 'admin', action: 'deleteIndexes'},
+        {verb: 'delete', url: '/_cleanDb', controller: 'admin', action: 'cleanDb'},
```
scottinet: This should be removed: we don't want to expose these actions in our API

lib/api/controllers/routerController.js
```diff
@@ -202,6 +203,7 @@ module.exports = function RouterController(kuzzle) {
         {verb: 'put', url: '/profiles/:id/_createOrReplace', controller: 'security', action: 'createOrReplaceProfile'},
         {verb: 'put', url: '/users/:id', controller: 'security', action: 'createOrReplaceUser'},
         {verb: 'put', url: '/:index', controller: 'admin', action: 'createIndex'},
+        {verb: 'put', url: '/_prepareDb', controller: 'admin', action: 'prepareDb'},
```
scottinet: ditto

bin/README.md
j33f: the goal to this note is to alert users that they will be asked if they want to reset the roles or not but it could be removed since the sentence above already says it

j33f:
Ready

jenow:
+1

lib/api/remoteActions/index.js
```diff
@@ -0,0 +1,77 @@
+/*eslint no-console: 0*/
+var 
+  availableRemoteActions = require('../../config/remoteActions'),
```
benoitvidis: nitpicking:
Would it be possible to avoid having to maintain an action list as a setting?

Something like this?:

```js
(action, params, args) => {
  if (remoteActions[action] === undefined) {
    try {
      remoteActions[action] = require('./' + action);
    }
    catch(e) {
        console.log('The action "' + action + '" doesn\'t exist');
        return process.exit(1);
    }
  }
}
```

bin/commands/disable.js
```diff
@@ -6,6 +6,7 @@ var
 module.exports = function () {
 
   var kuzzle = new Kuzzle(false);
-  kuzzle.enable(params, true);
+
+  kuzzle.remoteActions(kuzzle, 'enableServices', params, {enable: false});
```
benoitvidis: I am a bit disturbed by passing the `kuzzle` object to itself down to the final action.

Could Kuzzle constructor pass itself to the remoteActions?

benoitvidis:
I was a little afraid that this topic could bring some security issue. 
While I am still not 100% comfortable to allow anyone who has access to the cli to drop all the indexes, it is true that if one can access it, he most likely is able to reach ES directly too.

So, apart from my nitpicking points, +1

lib/api/remoteActions/index.js
j33f: good point ! I planned to but forgot...

bin/commands/disable.js
j33f: good point too

j33f:
yep, indeed, but if you can access the cli, you can do whatever you want on all part of kuzzle... so yes, it looks unconfortable, but it is pragmatic

bin/commands/start.js
```diff
@@ -58,25 +61,62 @@ module.exports = function () {
...
-         ████████████████████████████████████
-         ██        WORKER CONNECTED        ██
-         ██    ...PREPARING DATABASE...    ██
-         ████████████████████████████████████`);
+ ████████████████████████████████████
+ ██        WORKER CONNECTED        ██
+ ██    ...PREPARING DATABASE...    ██
+ ████████████████████████████████████`);
+
+        //return kuzzle.cleanDb(kuzzle, (args.likeAvirgin));
```
ballinette: some commented code to remove ?

ballinette:
why do you have both "lib/api/remoteActions/" and "lib/api/controllers/remoteActions/" folders ?

bin/commands/start.js
j33f: ooops

j33f:
"lib/api/remoteActions/" is a kind of "router" for the CLI : it handle the remote action commands request, format them then put then into a queue then wait for a reply. "lib/api/controllers/remoteActions/" is for the controller itself : it performs the actions then respond to the router

j33f:
You can merge :)

PR #216, opened 2016-03-25
fixed performance issue + updated dependencies

scottinet:
- Hydrated roles weren't stored in the profile cache, forcing a call to Elasticsearch on each request
- Updated dependencies
- Fixed breaking changes between lodash 3.10 and 4.6
- Fixed new errors fired by ESLint 2
- Deactivated ESLint `consistent-return` rule, as it appears to be bugged for the moment
- Removed `async` use from `RequestObject.checkInformation`, as it was overkill


j33f:
+1

benoitvidis:
+1

PR #217, opened 2016-03-30
Prevent event loop saturation

scottinet:
Prior to this pull request, when flooding Kuzzle (with >2000 requests/second), the NodeJS event loop was gradually filled with promises, Kuzzle was becoming slower and slower, eventually crashing if the flood persisted. If the flood stopped before Kuzzle crashes, it took several minutes to Kuzzle to recover from this state.

The purpose of this pull request is to make Kuzzle resilient under stressful conditions:

* Replaced promises with callback in protocol plugins, in the router controller and in the funnel controller. This was made to keep tight control on the event loop, as promises go in it, while callbacks don't.
* Added two properties in the `.kuzzlerc` configuration file:
  * `maxConcurrentRequests` configures the maximum number of requests that can be executed at the same time. All requests submitted above that number are queued to be executed at a later time
  * `maxRetainedRequests` configures the maximum number of requests that can be queued. Once this limit is reached, subsequent requests are discarded with a "service temporarily unavailable" error (status code 503)
* The funnel controller now has 2 states:
  * normal state: when a request is submitted, if it's under the `maxConcurrentRequests` limit, it's executed immediately, as before
  * overloaded state: the funnel controller enters this state if the `maxConcurrentRequests` limit is reached. In this state, new requests are queued and a background task is started to replay delayed requests as soon as possible. The background task stops automatically when the funnel switch back to normal state
* Moved token retrieval from the router controller to the request execution part in the funnel controller. This avoids repetition of code, and it helps controlling the event loop by performing these tasks once it was made sure that the request could be executed
* Added a new `server:overload` hook, fired at most each 500ms. The purpose of this hook is to inform potential plugins, or Kuzzle itself, about the overload state. The overload percentage is provided to the listeners (for instance: '28.63%'). This percentage reflects the number of queued requests compared to the `maxRetainedRequests` property
* Updated the SocketIO protocol plugin to use a callback instead of a promise for the `getRouter().execute` plugin context method
* Updated protocol plugins documentation


dbengsch:
+1

j33f:
+1

ballinette:
:+1:

PR #223, opened 2016-03-31
removed useless MQ functional tests

scottinet:
Functional tests involving the `jwtTokenExpired` event were executed using the websocket protocol instead of AMQP/MQTT/STOMP.

These tests cannot be ported to MQ protocols as of now, because of the way they are currently implemented, using RabbitMQ.

Instead of working on the way we implemented MQ protocols, we'll wait for them to be migrated to protocol plugins: we'll then get rid of RabbitMQ, and each one of these protocol will act as a server, allowing to handle rooms and channels as we do with websockets.

Until then, this pull request removes these useless functional tests.


ballinette:
+1

benoitvidis:
+1

Issue #215, opened 2016-03-25
Response when deleting user is not returning id of deleted object


mbmadaj:
When I deleteUser I do no get result with id of deleted object.

Easily reproducible, with JS SDK as example:

```
kuzzle
  .security
  .createUser('myuser', userContent, options, function(error, response) {
    // result is a KuzzleUser object
    console.log(error)
    console.log(response)
  });
```
```
kuzzle
  .security
  .deleteUser('myuser', function(error, result) {
    console.log(error) // undefined - which is expected
    console.log(result) // undefined - which is not expected
  });
```

When I call directly deleteUser with query

```
kuzzle.query({controller: 'security', action: 'deleteUser'}, {"_id": 'myuser'}, function (err, res) {
  console.log(err)
  console.log(res) // no key _id under result (and no value, the reason SDK returns undefined)
});
```

Probable reason, in response empty {} is returned
https://github.com/kuzzleio/kuzzle/blob/ce5415b998537ac85975d83d398a701373e8b40a/lib/api/controllers/securityController.js#L333

mbmadaj:
The deletion process works as expected, when I call for 2nd time deletion of just deleted user I receive error that user was not found.

scottinet:
Got it.

We also have multiple bugs with other security routes, like updateRole, updateProfile and updateUser, as they don't provide the updated document in the response.

Check out the `securityRoutesFixes` branch. Implementation is finished, I'm now working on the unit tests.

scottinet:
The issue has been merged on the `develop` branch and the `master` will be updated soon.

PR #225, opened 2016-04-04
Security routes fixes

scottinet:
Fixes issue #215 and more: several security weren't returning any useful data in the response, namely updateRole, updateProfile, updateUser, and deleteUser

Started to update repositories to make them return raw objects instead of ResponseObject.


ballinette:
+1

dbengsch:
+1

PR #226, opened 2016-04-04
fix updateUser in securityController: persist method need a hydrated …

stafyniaksacha:
…user to work

### DIFF
--- a/lib/api/controllers/securityController.js
+++ b/lib/api/controllers/securityController.js
@@ -380,6 +380,8 @@ module.exports = function SecurityController (kuzzle) {
    * @returns {Promise}
    */
   this.updateUser = function (requestObject) {
+    var updatedUser = {};
+
     return kuzzle.pluginsManager.trigger('security:updateUser', requestObject)
       .then(newRequestObject => {
         if (!newRequestObject.data._id) {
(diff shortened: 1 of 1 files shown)


ballinette:
missing unit test for this, but +1

scottinet:
Is this pull request still relevant?

lib/api/controllers/securityController.js
```diff
@@ -388,7 +390,19 @@ module.exports = function SecurityController (kuzzle) {
 
         return kuzzle.repositories.user.load(newRequestObject.data._id)
           .then(user => {
-            return kuzzle.repositories.user.persist(_.extend(user, newRequestObject.data.body), { database: { method: 'update' } });
```
benoitvidis: Isn't `user`already hydrated at this point?

stafyniaksacha: Nope, the `Repository.load` method didn't return a real `User` object: it contains data in a `_source` subfield (seems to be the raw output of elastic search) so the `_.extend` didn't work as we want here.

On my way, we need to plan a refactor on the Repositories methods to ensure the validity of theirs return values.

To reproduce the bug, just create a new user with admin profile, update it with anonymous profile. Then get it and see that the profile is switched to default.

benoitvidis: Then, if so, the problem lies within the repository code.
The UserRepository::load method [_should_ resolve to a hydrated object](https://github.com/kuzzleio/kuzzle/blob/develop/lib%2Fapi%2Fcore%2Fmodels%2Frepositories%2Frepository.js#L55).

I would rather fix the repository than putting the logic in the controller.

stafyniaksacha: Yes, it's the way we are going to.

@j33f is implementing this.

So this PR will not be relevant after.

PR #232, opened 2016-04-08
Kuz 485 roles refactoring

ballinette:
Roles refactoring

Implements https://jira.kaliop.net/browse/KUZ-485

# About roles and profiles

Roles are now only related to controllers and actions (no more index/collection). They follow this syntax:
```
{
  _id: "<role_id>"
  controllers: {
    <controller_name>: {
      actions: {
        <action_name>: <true/false/closure> (*)
      }
    }
  }
}
```

(*) closures keep actual syntax and behaviour

A profile contain 1 to many roles, with optional restrictions to some indexes/collections, like that:
```
{
  _id: "<profile_id>"
  roles: [
    {_id: "<role1>"},
    {
      _id: "<role2>",
      restrictedTo: [{index: "<index1"}]
    },
    {
      _id: "<role3>",
      restrictedTo: [
        {index: "<index1>"},
        {index: "<index2>", collections: ["collectionA", "collectionB"]}
      ],
      allowInternalIndex: true (*)
    },
  ]
}
```

# (*) about allowInternalIndex

Ideally, we should always disallow any action to internal index (we should not have to use it explicitly, and if it is needed for a low level adminitration operation, it could be done by using elasticsearch directly).
... but removing this feature implies to change the cleanSecurity() method in feature tests, and it would break too many thing. So for now, I just keep this behaviour by adding a "allowInternalIndex"  flag.
Removing it should be part to another PR.

# Dependencies

To work correcrly, this Pull-request need following ones to be merged before:
* #253 
* #255 

(but it is already ready for reviewing)


benoitvidis:
Looks the ProfileRepository::searchProfiles method does not reflect the new serialized format for the roles yet.

ballinette:
@benoitvidis searchProfiles method fixed ;)

features/AMQP.feature
```diff
@@ -246,7 +246,7 @@ Feature: Test AMQP API
 
   @usingAMQP @cleanSecurity
   Scenario: login user
-    Given I create a user "user1" with id "user1-id"
+    Given I create a user "useradmin" with id "user1-id"
```
ballinette: "user1" with admin role renamed to "useradmin", to avoid conflicts with users created to test roles.

features/step_definitions/authentication.js
```diff
@@ -17,7 +17,7 @@ var apiSteps = function () {
           return false;
         }
 
-        if (this.currentUser === null) {
+        if (this.currentUser === null || this.currentUser === undefined) {
```
ballinette: if user was logged-in, then logged-out, this.currentUser has been unset is. set to "undefined"

benoitvidis:
+1

dbengsch:
+1

.kuzzlerc
```diff
@@ -64,134 +64,54 @@
...
-                    "*": true
-                    }
-                  }
-                }
-              }
-            }
         }
       }
     },
     "anonymous": {
```
scottinet: The default roles `anonymous` and `default` should be able to execute a `serverInfo` action (not part of this PR, but it's the right time to fix that :-) )

features/Websocket.feature
```diff
@@ -312,38 +312,140 @@ Feature: Test websocket API
...
   @usingWebsocket @cleanSecurity @unsubscribe
   Scenario: token expiration
     Given A room subscription listening to "lastName" having value "Hopper"
-    Given I create a user "user1" with id "user1-id"
+    Given I create a user "useradmin" with id "user1-id"
     When I log in as user1-id:testpwd expiring in 1s
     Then I wait 1s
     And I should receive a "jwtTokenExpired" notification
 
+  @usingWebsocket @cleanSecurity @unsubscribe
```
scottinet: You don't seem to subscribe in this test, is the `@unsubscribe` hook necessary?

features/step_definitions/role.js
```diff
@@ -214,6 +214,87 @@ var apiSteps = function () {
...
+          return true;
+        }
+        callback(error);
+      });
+  });
+
+  this.Then(/^I'm ?(not)* allowed to search for documents in index "([^"]*)" and collection "([^"]*)"$/, function (not, index, collection, callback) {
+    this.api.search({}, index, collection)
+      .then(body => {
+        if (not && body.status === 403) {
```
scottinet: Is it possible to enter the `then` part of this promise with a 4xx status code?
This test seems to duplicate the one in the `catch` statement (same goes for the other new tests below)

lib/api/controllers/adminController.js
```diff
@@ -119,33 +120,33 @@ module.exports = function AdminController (kuzzle) {
...
-                }
-              });
-            }
-          });
-        }
-
+        response.indexes.filter(index => _.includes(requestObject.data.body.indexes, index));
+
+        async.each(response.indexes, (index, callback) => {
+          context.token.user.profile.isActionAllowed({controller: 'admin', action: 'deleteIndex', index: index}, context, kuzzle)
```
scottinet: if the `isActionAllowed` returns a rejected promise, the async callback is never called

lib/api/controllers/adminController.js
```diff
@@ -119,33 +120,33 @@ module.exports = function AdminController (kuzzle) {
...
-                      allowedIndexes.push(availableIndex);
-                    }
-                  });
-                }
-              });
-            }
-          });
-        }
-
+        response.indexes.filter(index => _.includes(requestObject.data.body.indexes, index));
```
scottinet: `Array.filter()` does not mutate the original array, meaning this line does not appear to do anything.

lib/api/core/models/repositories/profileRepository.js
```diff
@@ -186,7 +203,21 @@ module.exports = function (kuzzle) {
       }
     });
 
-    result.roles = extractRoleIds(profile.roles);
+    result.roles = [];
+    async.each(profile.roles, (role, callback) => {
```
scottinet: I may be wrong, but I'm under the impression that  this method will return an empty `result.roles` object, since getting roles from profiles will be done asynchronously with async

lib/api/core/models/security/role.js
```diff
@@ -346,6 +217,78 @@ Role.prototype.validateDefinition = function (context) {
...
+    return q(true);
+  }
+
+  async.some(this.restrictedTo, (restriction, callback) => {
+    callback(checkIndexRestriction(requestObject, restriction));
+  }, result => {
+    isAllowed = result;
+  });
+
+  return q(isAllowed);
```
scottinet: Should return a deferred promise, resolved only when the async call completes.

features/Websocket.feature
ballinette: was just a copy-past from the former line, but yes, you're right, @unsubscribe  is not needed here.

features/step_definitions/role.js
ballinette: you're right, I first thought that the promise resolved event with a 4xx status code, so I made these lines... and forgot to clean them later...

.kuzzlerc
ballinette: Yep, I can add it ;)

ballinette: done.

ballinette:
@scottinet all your remarks are fixed ;)

scottinet:
Great!
The new roles and profiles format ROCKS!

PR #238, opened 2016-04-19
ResponseObject refactor

scottinet:
The purpose of this pull request is to remove all `ResponseObject` creation and to make controllers do it instead. This simplifies considerably the code architecture, mainly because now we don't have to wonder whether the data to process is encapsulated inside a `ResponseObject` or not, and because there is a neat separation between a user request, the data emanating from it, and the response forwarded to the user.

This is the new expected behavior:
* the router controller translate user queries to `RequestObject` objects
* controllers receive a `RequestObject` and forward it to the right Kuzzle components
* these components respond with raw data
* controllers analyze these data and forward a `ResponseObject` to the router controller
* if an error occurs, it is forwarded as an `Error` object (or a derivated class) to controllers, who then reject their promise with an errored `ResponseObject` 
*  error derivated classes are now serializable, allowing workers to forward them as-is to Kuzzle server instances

## Changes made to the `ResponseObject` model

* Removed `state` and `scope` from the `ResponseObject` model. Only the `NotificationObject` model hold these information
* Due to errors being serializable, `ResponseObject` now rely on the `status` code instead of on the `isError()` method to detect an error. Objects with a status code outside the 2xx range are treated as errors.
* `ResponseObject` containing an error now have the `count`, `stack` and `_source` fields undefined instead of null if these fields are absent

## Changes made to notifications

* the `RealtimeResponseObject` model has been renamed to `NotificationObject`
* `NotificationObject` objects now contain the `index` and `collection` related to the notification
* the notifier core component is now the only component building `NotificationObject` objects from raw data
* Workers forwarded response objects both to controllers and to the notifier core component. Now, they forward raw data only to the controllers, who in turn call explicitly the notifier core component if necessary. This was made to reduce network traffic and to make this behavior explicit instead of implicit (leading to potential side-effects when working with Kuzzle code)

## Other related changes

* the `HotelClerk` and the `DSL` core components now process raw data instead of model objects (response or request objects)
* removed code made to distinguish between raw data and `ResponseObject` in the repositories components

## Misc.

* fixed lots of security related unit tests. These were effectively testing nothing
* the `security/role` component now enforce the `test` attribute in closures
* removed some `q.defer()` calls when they were unnecessary, simplifying the code. In some rare cases, this also has the benefits to remove objects stored unnecessarily in memory when a method exitted early due to an error
* updated the minimal NodeJS version from `4.1.1` to `4.3.1`. This was made to avoid the `4.3.0` version, which is very unstable in debug mode
* made unit tests on the `RemoteActionsController` controller work in isolation to avoid random failures when the internal broker doesn't have time to start


lib/api/controllers/remoteActions/prepareDb.js
```diff
@@ -33,45 +27,31 @@ module.exports = function PrepareDb (kuzzle, request) {
...
       });
-  } else {
-    response = new ResponseObject(request);
-    response.data.body = { isWorker: true };
-    deferred.resolve(response);
-    return deferred.promise;
   }
 
-  return deferred.promise;
+  return q({isWorker: true});
```
j33f: Why ```{isWorker: true}``` ?

lib/api/core/models/responseObject.js
```diff
@@ -44,7 +42,7 @@ function construct (requestObject, response) {
     Object.keys(o).forEach(attr => { this[attr] = o[attr]; });
     return this;
   }
-  else if (util.isError(o)) {
+  else if (this.status < 200 || this.status >= 300) {
```
j33f: 3xxx statuses are not errors!

lib/api/controllers/remoteActions/prepareDb.js
scottinet: Why not?
I wanted to return relevant data and found nothing more interesting than that. :-)

lib/api/core/models/responseObject.js
scottinet: Good point!
I'm fixing it right away.

j33f:
+1

dbengsch:
+1

PR #240, opened 2016-04-20
Unit test stubs

scottinet:
This is a proof-of-concept of replacing our custom mocks with Sinon.js, simplifying test creation and maintenance.

If you like what you see, go through the usual review procedure, and we'll see how we convert other tests to Sinon.js.
Before merging this pull request, please make sure that everyone in the team has at least took notice of its existence. :-)

If you disagree with this pull request, please discuss why so we can find other solutions.


jenow:
+1 love it
can even check arguments on spies with .withArgs(..), like mockito does

test/api/controllers/adminController.test.js
```diff
@@ -99,41 +56,46 @@ describe('Test: admin controller', function () {
...
     it('should add the new collection to the cache', () => {
       this.timeout(50);
+      sandbox.stub(kuzzle.workerListener, 'add').resolves({});
+      sandbox.spy(kuzzle.indexCache, 'add');
+      sandbox.spy(kuzzle.indexCache, 'remove');
+      sandbox.spy(kuzzle.indexCache, 'reset');
 
       return kuzzle.funnel.controllers.admin.updateMapping(requestObject)
         .then(response => {
           should(response).be.instanceof(ResponseObject);
```
AnthonySendra: Did we have already a problem here? missing `done()`? Because we're in asynchronous

AnthonySendra:
Seems really really BETTER!

benoitvidis:
Much cleaner. +1 obviously

j33f:
yes !

test/api/controllers/adminController.test.js
```diff
@@ -99,41 +56,46 @@ describe('Test: admin controller', function () {
 
     it('should add the new collection to the cache', () => {
       this.timeout(50);
+      sandbox.stub(kuzzle.workerListener, 'add').resolves({});
+      sandbox.spy(kuzzle.indexCache, 'add');
```
ballinette: As far as I understand how sinon.js works, I wonder if we should rather use always "stub" instead of "spy":
In a unit-test, we don't really want that the method kuzzle.indexCache.add is actually executed... we just want to check that its is called by the updateMapping() method.

Using only "stubs", we ensure that only the code we want to test is actually executed.

PR #243, opened 2016-04-20
add the list of available routes to api root url

j33f:
add the list of available routes by controller and by verb to api root url + correct a typo in api definition


scottinet:
+1

dbengsch:
+1

PR #248, opened 2016-04-25
New updateSelf action in auth controller

dbengsch:
* Add new `updateSelf` action in `auth` controller.
* `data.body._id` was overwriting `data._id` when `_.extend` was called in `security` controller.
* Some boyscout changes.


scottinet:
Ok I'm going to be really, REALLY nitpicky here. I don't know if this is just me, but whenever I see `selfUpdate` I understand that the system will trigger an automatic update, while `updateSelf` tells to me that I'm able to update myself.

Now that I've said that: +1 :stuck_out_tongue:

jenow:
+1

lib/api/controllers/authController.js
```diff
@@ -91,4 +109,37 @@ module.exports = function AuthController (kuzzle) {
...
+          return q.reject(new BadRequestError('profile can not be provided in selfUpdate'));
+        }
+
+        if (modifiedRequestObject.data.body._id) {
+          return q.reject(new BadRequestError('_id can not be part of the body'));
+        }
+
+        return kuzzle.repositories.user.load(context.token.user._id);
+      })
+      .then(user => kuzzle.repositories.user.persist(_.extend(user, modifiedRequestObject.data.body), { database: { method: 'update' } }))
```
stafyniaksacha: Be careful with using `_.extend`: if user already contain a field it will not be overwritten by the request.
You have to use `_.merge` instead.

By this way, we need a functional test to test it : 
* Update user with a new field `foo: bar` 
* Then update user with an existing field `foo: baz`

dbengsch: If it is the case, it is also present in security controller in every update method

dbengsch: In fact, it does :

ballinette: the difference btw. _.extend and _.merge is only about child objects. _.extend will replace child objects, as _.merge will merge them.
see http://stackoverflow.com/questions/19965844/lodash-difference-between-extend-assign-and-merge.

example:
console.log(_.extend({a:{x: '1'}},{a:{y: '2'}}));
> { a: { y: '2' } }
console.log(_.merge({a:{x: '1'}},{a:{y: '2'}}));
> { a: { x: '1', y: '2' } }

In our case, I wonder if merge should be a better choice: if we have a user with nested attribute a.x = 1, and we want to update him with a new nested attribute a.y = 2, we still want to keep a.x, I guess...
(so it means that we should also fix the securityController as well, maybe...)

j33f:
+1

lib/api/controllers/authController.js
dbengsch: It means that you can't remove a property of a nested object once it is here, you can only nullify it. In particular, it could be really tricky when updating a role.

dbengsch:
Renamed it to `updateSelf`, I come along with your arguments and think it will also be more consistant with other update actions.

lib/api/controllers/authController.js
ballinette: ok, good point ;)

ballinette:
+1

PR #253, opened 2016-04-27
Stores plugins configuration in database

scottinet:
Instead of configuring plugins using files, this pull request now stores them in the database, under the `kuzzle.config.internalIndex` index (defaults to `%kuzzle`).

# Changelog

## General

Renamed some plugins configuration:
- `version` is now `npmVersion`
- `url` is now `gitUrl`

This has been made to make these options clearer. And also because `commander` has several issues with options named `version` because it causes conflicts with internal variable names :-1: 

## Kuzzle configuration

- removed the `/config` directory, as this PR obsoletes it
- added a list of default plugins in the `.kuzzlerc` file. This list must include a way of installing these plugins (through the `npmVersion`, `gitUrl` or `path` tag) and if they should be activated upon their installation or not
- added the plugins configuration collection name in the `.kuzzlerc`
- added several (and sometimes unrelated) explanations/comments in the `.kuzzlerc` file

## New `internalEngine` service

Added a new `internalEngine` service. This service differs from the regular read/write engines by not using any plugin, by not being accessible under the `kuzzle.services.list` object (it's instead accessible via `kuzzle.internalEngine`), and by being initialized before everything else during startup.

It has been added because normal services might trigger plugin hooks, and because of that, they are initialized _after_ plugins. Since we need to make the plugins manager able to access the database, I was deadlocked. Thanks to @AnthonySendra and @ballinette for providing me with this solution.
Another PR will be submitted to make other Kuzzle core components use this internalEngine instead of the regular read/write engines.

## Command-line interface

- `kuzzle` (with no other argument) now falls back to `kuzzle --help`
- added a `kuzzle plugins` command, allowing to manage plugins. This command allows installing, removing, (de)activating plugins, and also getting or updating their configuration (full documentation here: https://github.com/kuzzleio/kuzzle-guide/pull/5 )
- `kuzzle plugins` uses the database to store and manage plugins configuration
- `kuzzle plugins --install` will install all plugins listed in the database. If no plugins are listed, it will load the default plugins setted in the `.kuzzlerc` file
- `kuzzle install` is now an alias to `kuzzle plugins --install` to ensure backward compatibility with our docker containers

## Plugins manager

The plugins manager now gets its plugins list from the database, using the `internalEngine`.
Also, the `pluginsManager.init` method now returns a promise, resolved once plugins are loaded. 

## Unit tests

While playing with services initialization, I was able to make dummy-mode work without any service loaded, and I found out that several unit tests needed to access the elasticsearch service to work. 
Since I thought it was not a good thing, I fixed these unit tests by mocking Kuzzle services with sinon.js

## prepareDB

The `prepareDB` module skipped roles, profiles and users creation if the internal index already exist.
Since the plugins CLI now creates the internal index beforehand, this module has been refactored a bit:

- it skips index creation if the index already exists, and continue with roles creation
- then it skips roles creation if the roles collection already exists, and continue with profiles creation
- then it skips profiles creation if the profiles collection already exists, and continue with users creation
- and then it creates the users collection only if it doesn't exist already

## Unrelated / Boyscout

The `hotelClerk.listSubscriptions` method has been simplified: some code was unreachable due to the way async and promises work.


scottinet:
It just occured to me that adding several API routes might be helpful. I'm thinking about these ones:

- `admin/addPlugin`
- `admin/deletePlugin`
- `admin/updatePluginConfig`

If you agree, I'll update this PR with these new functionalities (and the corresponding documentation, of course)

scottinet:
I'll start by extending the CLI so that it'll be possible to install plugins by doing `kuzzle install <plugin name> --version 1.2.3`

ballinette:
+1

scottinet:
For the record: the CLI now features a full range of plugins management capabilities.
And I have added a new Jira ticket to add API routes, including plugins hot loading.

This pull request is now ready to be reviewed.

benoitvidis:
+1

PR #254, opened 2016-04-28
Redis Cluster + Worker & Services catch error

AnthonySendra:
## About Redis Cluster
This PR allow to prepare for future needs about failover. The new used library `ioredis` allow to init a `client` with a cluster.

The configuration has changed. Now we have an entry `nodes` OR an entry `node`, that define if a cluster will be used.

```js
{
  nodes: [
    {host: 'foobar', port: 6379},
    {host: 'foobar2', port: 6379}
  ]
}
```

Remember, during the configuration initialization in `/lib/config/models/cache.js`, a `databases` entry is added.

```js
{
  nodes: [
    {host: 'foobar', port: 6379},
    {host: 'foobar2', port: 6379}
  ],
  databases: ['notificationCache', 'statsCache', 'userCache', 'internalCache', 'tokenCache', 'memoryStorage']
}
```

## About Worker And Services

`then`/`catch` worker and service initialization


lib/config/models/cache.js
```diff
@@ -1,7 +1,26 @@
...
+    cache.node = {
+      host: process.env.CACHE_HOST,
+      port: process.env.CACHE_PORT
+    };
+  }
+  else if (params.cache) {
+    if (params.cache.nodes) {
+      cache.nodes = params.cache.nodes;
+    }
+    else if (params.cache) {
```
ballinette: shouldn't is be rather "else if (params.cache.node)" ?

ballinette:
+1

lib/services/internalbroker.js
```diff
@@ -40,7 +40,7 @@ module.exports = function (kuzzle, options) {
    * @return Promise when initialization is complete
    */
   this.init = function () {
-    return (this.isServer) ? serverStart.call(this) : clientConnect.call(this);
+    return (this.isServer) ? q(serverStart.call(this)) : q(clientConnect.call(this));
```
scottinet: Unnecessary: both these functions already return promises, resolved either when the ipc server is started or when the client is connected to a kuzzle server

package.json
```diff
@@ -51,7 +52,6 @@
     "rc": "1.1.6",
     "readline-sync": "^1.3.0",
     "redis": "2.6.0-1",
```
scottinet: Is there a reason for keeping the "redis" module?

scottinet: Is there a reason for keeping the "redis" module?

scottinet:
Besides my remarks: +1!

xbill82:
Seems legit to me. +1 (should I merge or do you still have some fixes to add?)

PR #255, opened 2016-04-28
Error Response refactor

ballinette:
# Summary
Reject error objects instead of ResponseObjects when an error occur in a controller.

# Reason
After   #238 , the ResponseObject is only created by each controller.
Then, even in case of error, the controller's promise rejects with a ResponseObject containing the error.

But this includes a regression while an error is occured by another component than a controller.
See for instance https://github.com/kuzzleio/kuzzle/blob/develop/lib/api/controllers/funnelController.js#L103 :
1. call requestObject.checkInformation()
2. check if the action is allowed

It would be ugly to generate a ResponeObject from the RequestObject item
So the best behaviour is to reject only Error objects, and generate the ResponeObject containing this error only within the execute() merhod or the funnelController.

This way also, we always call the callback with a valid ResponseObject as "response" parameter, so it simplifies the protocols implementations.

# Breaking Changes
This PR breaks the actual implementation of SocketIO plugin.
So, if you agree with this PR, we must before:
* merge the related PR in the SocketIO plugin: https://github.com/kuzzleio/kuzzle-plugin-socketio/pull/6
* upgrade the plugin's version in dependencies
(NB: should be done after merging #253 )


lib/api/controllers/adminController.js
```diff
@@ -24,7 +24,7 @@ module.exports = function AdminController (kuzzle) {
         kuzzle.indexCache.add(modifiedRequestObject.index, modifiedRequestObject.collection);
         return kuzzle.pluginsManager.trigger('data:afterUpdateMapping', new ResponseObject(modifiedRequestObject, response));
       })
-      .catch(err => q.reject(new ResponseObject(modifiedRequestObject || requestObject, err)));
+      .catch(err => q.reject(err));
```
scottinet: This line, and all lines alike, should be removed. The rejected promise is already forwarded to the caller, we intercepted errors to reject a ResponseObject instead of the original error.

ballinette: right.
fixed

scottinet:
+1

benoitvidis:
+1

AnthonySendra:
+1
Waiting for tests and conflicts resolutions.

PR #256, opened 2016-04-29
Fix plugin controller

stafyniaksacha:
- Add remoteActionController in pluginContext
- Fix plugin route registration
- Fix controller plugin call context
- Fix likeAVirgin remote action (clean index hook wasn't refactored)

### DIFF
--- a/lib/api/controllers/routerController.js
+++ b/lib/api/controllers/routerController.js
@@ -145,6 +145,7 @@ module.exports = function RouterController (kuzzle) {
       api = new Router(),
       apiPlugin = new Router(),
       handleRoute,
+      extractRouteInfo,
       coverage;
 
     routes = [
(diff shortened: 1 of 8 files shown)


lib/api/controllers/routerController.js
```diff
@@ -370,8 +370,26 @@ module.exports = function RouterController (kuzzle) {
     });
 
     // create and mount a new router for plugins
-    this.pluginRouter = new Router();
-    api.use('/_plugin', this.pluginRouter);
+    api.use('/_plugin', apiPlugin);
+
+    kuzzle.pluginsManager.routes.forEach(route => {
+      apiPlugin[route.verb](route.url, function (request, response) {
```
ballinette: the callback function is the same as for non-plugin routes: it should be factorized to avoid code duplication.

scottinet:
+1

lib/api/controllers/routerController.js
```diff
@@ -390,6 +408,17 @@ module.exports = function RouterController (kuzzle) {
         _routes[route.controller][route.verb].push('/api/' + kuzzle.config.apiVersion + route.url);
       });
 
+      kuzzle.pluginsManager.routes.forEach(route => {
```
ballinette: To avoid code duplication, this code should be factorized with piece of code just above, probably with a method taking 2 arguments:
* routes (either _routes_ or _kuzzle.pluginsManager.routes_)
* prefix (empty string in the 1st case, _"/_plugin"_ in the second).

ballinette:
besides my remarks about duplicated code, +1

jenow:
+1

PR #257, opened 2016-05-02
Feature/kuz 463 es2 autorefresh

benoitvidis:
This PR is a workaround for the [NRT 1s delay of elasticsearch](https://www.elastic.co/guide/en/elasticsearch/guide/current/near-real-time.html).

Kuzzle users can set an `autoRefresh` flag on some indexes.  
When set to `true`, all write actions coming from the service will be immediately followed by a `refresh` call, making the documents(s) immediately (un)available for search.

Additionally, in order to persist the `autoRefresh` flag, the services configuration is now persisted in ES.

# Changelog

### Configuration

The autoRefresh is a simple index => boolean object stored in memory in the writeEngine.  
Note that the engine is the Worker(s) one.  

The configuration can be set using .kuzzlerc to mark some indexes as autorefreshed. The default configuration is to disable this functionality as it can lead to some performance issues.

Users can also modify \_provided they are granted to\_ the autoRefresh using the `admin:setAutoRefresh` action.

### getAutoRefresh & setAutoRefresh actions

The `admin` controller is added 2 new actions:

* `getAutoRefresh`: retrieves the current autoRefresh value set for the given index. Note that even being a "get" action, it still needs to be passed down to the worker(s), which are the only instances that keep the flag updated.
* `setAutoRefresh`: sets the autoRefresh flag for the given index.

## ElasticSearch service tests

All tests for Kuzzle es service now use sinon.

## Services init

* switched from a blacklist logic to a whitelist for services initialization
* In dummy mode, no service is initialized at all

## Services dependencies for tests

Now that Kuzzle dummy mode does not init any service by default, the tests needed to be adapted to remove some remaining depencies.

# Additional remark

## ElasticSearch 5

Some development is [currenty in progress to implement a `block until refresh` ability in Elasticsearch](https://github.com/elastic/elasticsearch/pull/17986).

If/once it is merged in the main stable branch, Kuzzle new version should remove the autoRefresh ability.


.kuzzlerc
```diff
@@ -19,6 +19,13 @@
   },
   // Host and port for the writeEngine service: the database called when we write a data. Can be an array.
   "writeEngine": {
+    // autoRefresh: an index => boolean container object.
+    // If the autoRefresh is set to true for a given index, all write calls will be followed by a refreh request.
```
scottinet: nitpicking: missing "s" in the last "refresh" word

.kuzzlerc
```diff
@@ -19,6 +19,13 @@
...
+    //
+    // /!\ Enabling the autoRefresh can lead to performance issued.
```
scottinet: nitpicking: "performance issues"

lib/api/controllers/adminController.js
```diff
@@ -255,4 +256,61 @@ module.exports = function AdminController (kuzzle) {
...
+        // even if being a get method, the autorefresh is set in the worker side
+        // we need to ask it the status
+        return kuzzle.workerListener.add(modifiedRequestObject);
+      })
+      .then(response => kuzzle.pluginsManager.trigger('data:afterGetAutoRefresh', new ResponseObject(modifiedRequestObject, response)))
+      .then(response => {
+        response.data.body = response.data.body.response;
+        return response;
+      })
+      .catch(err => q.reject(new ResponseObject(modifiedRequestObject || requestObject, err)));
```
scottinet: About rejecting a response object: you may want to synchronize with @ballinette: https://github.com/kuzzleio/kuzzle/pull/255

lib/api/controllers/adminController.js
```diff
@@ -255,4 +256,61 @@ module.exports = function AdminController (kuzzle) {
...
+
+  /**
+   * Sets elasticsearch autorefresh on/off for current index.
+   * /!\ Can lead to some performances issues.
+   * cf https://www.elastic.co/guide/en/elasticsearch/guide/current/near-real-time.html
+   *
+   * @param {RequestObject} requestObject
+   * @returns {*}
+   */
+  this.setAutoRefresh = function (requestObject) {
```
scottinet: Shouldn't this update `.kuzzlerc`too,  to keep the option for the index indefinitly ?
Or does this explicitly set the autoRefresh option of an index only until the next Kuzzle restart?

benoitvidis: Hum.. I don't like the idea of modifying the `.kuzzlerc` but we could think of persisting the update by storing it in es, something alike what you did for the plugins configuration.

ballinette: +1 about benoit's suggestion

benoitvidis: OK, I propose we go for this solution. I added a warning on the dependencies for this PR.

benoitvidis: Done

j33f:
+1 nice job

lib/config/hooks.js
```diff
@@ -13,6 +13,8 @@ module.exports = {
   'data:beforeDeleteIndex': ['write:add'],
   'data:beforeDeleteIndexes': ['write:add'],
   'data:beforeRefreshIndex': ['write:add'],
+  'data:beforeSetAutoRefresh': ['write:broadcast'],
+  'data:beforeGetAutoRefresh': ['write:add'],
```
ballinette: hmm OK, after carefully reading the code, I understand the reason of this "write:add".... but it looks very anti-intuitive. 
Maybe we should rename it in "notify" or something else ?

ballinette:
despite my very little remark about hook naming, + :100:  ;)

nik9000:
> Some development is currenty in progress to implement a block until refresh ability in Elasticsearch.
> 
> If/once it is merged in the main stable branch, Kuzzle new version should remove the autoRefresh ability.

Can someone from Kuzzle come and comment over at https://github.com/elastic/elasticsearch/issues/1063#issuecomment-218185247 ? I'm trying to figure out if the solution I proposed for `?block_until_refresh` is going to solve the issue for many folks. For the most part the thing you are doing (adding `?refresh` to bulk requests) should be safe until you get to pretty high ingestion rates. And when you have *very* high ingestion rates my solution sort of breaks down because I get worried about blocking 100s of requests.

lib/api/core/models/repositories/tokenRepository.js
```diff
@@ -76,14 +76,12 @@ module.exports = function (kuzzle) {
 
     if (!user || user._id === null) {
       error = new InternalError('Unknown User : cannot generate token');
-      deferred.reject(error);
-      return deferred.promise;
+      return q.reject(error);
```
scottinet: Minor/nitpicking: In that case, the line `deferred = q.defer()` should be put after these checks, to avoid creating a promise whose sole purpose is to be garbage collected.

lib/api/remoteActions/index.js
```diff
@@ -54,7 +55,10 @@ module.exports = function RemoteActions (kuzzle) {
     }
 
     this.kuzzle.config = config(params);
-    this.kuzzle.services.init({blacklist: ['mqBroker', 'logger', 'notificationCache']});
+    // the only service we need is the internal broker
+    this.kuzzle.services.list = {
+      broker: new InternalBroker(kuzzle, {})
+    };
     this.kuzzle.pluginsManager = new PluginsManager(this.kuzzle);
     this.kuzzle.pluginsManager.init(this.kuzzle.isServer, true);
```
scottinet: I think this is my fault: in one of my previous PR, I changed the `pluginsManager.init` method to make it return a Promise, because it now needs to fetch plugins configuration from the database.
So `pluginsManager.run()` should only be started after `pluginsManager.init()` promise resolves.

Sorry about that :-(

lib/config/hooks.js
benoitvidis: what about "emit"?

scottinet:
Great!

+1

PR #258, opened 2016-05-03
remove unnecessary passport local plugin default config

ballinette:
This default settings are already set within the plugin's configuration file, so it is not needed here.


scottinet:
+1

j33f:
+1

PR #260, opened 2016-05-09
Path plugins configuration now taken from plugins dir

scottinet:
The `package.json` file was taken from the `node_modules` directory even for plugins installed using the `--path` option.
This pull request fixes this by updating the plugin's configuration from its source directory instead.

It is still required to provide the absolute plugin path.

edit: found out that eslint was not checking the `bin/` directory, so I've added that. And fixed eslint errors.


benoitvidis:
+1

j33f:
+1

ballinette:
+1

PR #274, opened 2016-05-18
KUZ-541 Allow plugins to register an event on multiple functions

scottinet:
This pull request allows a plugin to be plugged on an event with different functions.

The `hooks` and `pipes` properties now accept either a string or an array of strings:
```js
{
  hooks: {
    'foo:bar': 'someMethod',
    'bar:foo': ['some', 'other', 'methods']
  }
}
```

Unit tests on worker plugins and on the `pluginsManager.run()` method have also been rewritten to use Sinon.js instead of the handmade mocks . Well... except for that humongous PM2 mock, I didn't have the courage to refactor this one.

### DIFF
--- a/lib/api/core/plugins/pluginsManager.js
+++ b/lib/api/core/plugins/pluginsManager.js
@@ -331,7 +331,7 @@ function initPipes (pipes, plugin, pipeWarnTime, pipeTimeout) {
 
       if (pipeTimeout) {
         pipeTimeoutTimer = setTimeout(() => {
-          var errorMsg = `Timeout error. Pipe plugin ${plugin.name} exceeded ${pipeTimeout} ms to execute. Aborting pipe`;
+          var errorMsg = `Timeout error. Pipe plugin ${plugin.name} exceeded ${pipeTimeout}ms to execute. Aborting pipe`;
           this.trigger('log:error', errorMsg);
 
           callback(new GatewayTimeoutError(errorMsg));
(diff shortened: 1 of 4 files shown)


lib/api/core/plugins/pluginsManager.js
```diff
@@ -313,60 +313,78 @@ function pm2Init (workers, pluginsWorker, isDummy) {
 }
 
 function initPipes (pipes, plugin, pipeWarnTime, pipeTimeout) {
-  if (plugin.config && plugin.config.pipeWarnTime !== undefined) {
-    pipeWarnTime = plugin.config.pipeWarnTime;
-  }
-  if (plugin.config && plugin.config.pipeTimeout !== undefined) {
-    pipeTimeout = plugin.config.pipeTimeout;
-  }
+  var register = (event, fn) => {
```
AnthonySendra: Generally, I'm not really comfortable with function creation inside other functions. It's just a personal point of view, but can we create instead a private function inside the same file? Something like:
`registerPipes`
`registerHooks`

I find it difficult to read and to understand. We can discuss if you're not agree.

ballinette:
+1

lib/api/core/plugins/pluginsManager.js
j33f: I agree

scottinet: I'm fine with this too. Inc another commit :-)

AnthonySendra:
+1

dbengsch:
lib/api/core/plugins/workerReady.js and its tests should also be adapted in order to call more than one function at a time (line 40 to 43)

PR #275, opened 2016-05-18
fix-KUZ-533 Update a user profile was buggy

j33f:
this is a quick fix before repositories / hydratation refactor

### DIFF
--- a/lib/api/core/auth/formatProcessing.js
+++ b/lib/api/core/auth/formatProcessing.js
@@ -64,8 +64,6 @@ module.exports = {
   formatUserForSerialization: function (kuzzle, user, hydrate) {
     return kuzzle.pluginsManager.trigger('security:formatUserForSerialization', user)
       .then(triggeredUser => {
-        var deferred = q.defer();
-        
         if (typeof triggeredUser.profile === 'string') {
           return kuzzle.repositories.profile.loadProfile(triggeredUser.profile)
             .then(profile => {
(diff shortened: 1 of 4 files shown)


ballinette:
+1

lib/api/core/auth/formatProcessing.js
```diff
@@ -64,7 +64,24 @@ module.exports = {
   formatUserForSerialization: function (kuzzle, user, hydrate) {
     return kuzzle.pluginsManager.trigger('security:formatUserForSerialization', user)
       .then(triggeredUser => {
-        var response = triggeredUser;
+        var deferred = q.defer();
```
scottinet: Avoid use of `defer()` when it can be avoided. This improves readability and limits the risks of unresolved promises.

In this case, this function can be written like this:
```js
       if (typeof triggeredUser.profile === 'string') {
         return kuzzle.repositories.profile.loadProfile(triggeredUser.profile)
           .then(profile => {
             triggeredUser.profile = profile;
             return q(triggeredUser);
           });
       }

       return q(triggeredUser);
```

j33f: as french people say : "Bien vu !"

stafyniaksacha:
As the issue raised in the pr #226 : 
> To reproduce the bug, just create a new user with admin profile, update it with anonymous profile. Then get it and see that the profile is switched to default.

Did you can write a test to ensure the resolution of the bug ?

j33f:
@stafyniaksacha I did can write tests, yep :)

xbill82:
:+1:

scottinet:
+1

PR #277, opened 2016-05-20
[Beta] bugfix#263 - Hash filter's value to avoid '.' in it 

jenow:
Hash filter's values to avoid having a '.' character which causes a bug


stafyniaksacha:
+1

PR #281, opened 2016-05-25
KUZ-560 Add swagger support

j33f:
Add 2 routes to /api
- /api/swagger.json
- /api/swagger.yml

Both routes send swagger files from the routes definitions with minimal informations.
Should be completed with valid and real routes informations in the future.

Both routes get the data from files generated bu the CLI command ```kuzzle swagger``` which generates 2 files at the Kuzzle project root : kuzzle-swagger.json and kuzzle-swagger.yml

The ```kuzzle swagger``` CLI command generates basic swagger sheme for routes which have been poorly described, and uses the given description for routes which have one. 

A route is is described properly in /lib/config/httpRoutes.js file.


AnthonySendra:
Still don't understand why this file is in Kuzzle?
We have a specific repo for the documentation why do we want to put this file in Kuzzle application?

j33f:
in fact it could be everywhere... is everybody ok to put it in the documentation repo ?

swagger.yaml
```diff
@@ -0,0 +1,2651 @@
...
+  /_login:
+    post:
+      description: 'Controller: auth. Action: login.'
+      produces:
+        - application/json
+      responses:
+        '200':
+          description: ''
+          schema: {}
+  '/_login/:strategy':
```
ballinette: missing also "get" method for route '/_login/:strategy' => auth/login

swagger.yaml
```diff
@@ -0,0 +1,2651 @@
...
+  /profiles/_search:
+    post:
+      description: 'Controller: security. Action: searchProfiles.'
+      produces:
+        - application/json
+      responses:
+        '200':
+          description: ''
+          schema: {}
+  '/roles/:id':
```
ballinette: missing also methods for route '/roles/:id':
*  "get" => security/getRole
* "post" => security/updateRole
* "delete" => security/deleteRole

swagger.yaml
```diff
@@ -0,0 +1,2651 @@
...
+  /ms/_zunionstore:
+    post:
+      description: 'Controller: ms. Action: zunionstore.'
+      produces:
+        - application/json
+      responses:
+        '200':
+          description: ''
+          schema: {}
+  '/profiles/:id':
```
ballinette: missing also methods for route '/profiles/:id':
*  "get" => security/getProfile
* "post" => security/updateProfile
* "delete" => security/deleteProfile

swagger.yaml
```diff
@@ -0,0 +1,2651 @@
...
+  /roles/_search:
+    post:
+      description: 'Controller: security. Action: searchRoles.'
+      produces:
+        - application/json
+      responses:
+        '200':
+          description: ''
+          schema: {}
+  '/users/:id':
```
ballinette: missing also methods for route '/users/:id':
* "get" => security/getUser
* "post" => security/updateUser
* "delete" => security/deleteUser

swagger.yaml
```diff
@@ -0,0 +1,2651 @@
...
+          name: collection
+          required: 'true'
+          type: string
+      produces:
+        - application/json
+      responses:
+        '200':
+          description: ''
+          schema: {}
+  '/:index/:collection/_mapping':
```
ballinette: missing also "get" method for route '/:index/:collection/_mapping' => admin/getMapping

swagger.yaml
```diff
@@ -0,0 +1,2651 @@
...
+          name: collection
+          required: 'true'
+          type: string
+      produces:
+        - application/json
+      responses:
+        '200':
+          description: ''
+          schema: {}
+  '/:index/_autoRefresh':
```
ballinette: missing also "get" method for route '/:index/_autoRefresh' => admin/getAutorefresh

swagger.yaml
```diff
@@ -0,0 +1,2651 @@
...
+          name: collection
+          required: 'true'
+          type: string
+      produces:
+        - application/json
+      responses:
+        '200':
+          description: ''
+          schema: {}
+  '/:index/:collection/:id':
```
ballinette: missing also methods for route '/:index/:collection/:id':
*  "get" => read/get
* "delete" => write/delete
* "delete" => security/deleteRole

swagger.yaml
```diff
@@ -0,0 +1,2651 @@
...
+          name: index
+          required: 'true'
+          type: string
+      produces:
+        - application/json
+      responses:
+        '200':
+          description: ''
+          schema: {}
+  '/:index/:collection':
```
ballinette: missing also "post" method for route '/:index/:collection' => write/publish

ballinette:
I don't care that we put it in the documentation repo or the core repo, both are OK for me.

anyway, +1 for me as long as the missing routes are added (maybe a bug with your script who generates the file, which does not merge correctly while the same URL is used for multiple HTTP methods ?)

benoitvidis:
@AnthonySendra: It actually makes sense to me this file definition is included inside Kuzzle as it is tightly tied to the REST router implementation.
Further, I  would even expose an http route that would serve the swagger.json file so that the clients can directly consume it from Kuzzle.

j33f:
good suggestion about the route @benoitvidis !

j33f:
don't know if it's *that* a good idea to create routes for this ?

lib/api/controllers/remoteActions/swagger.js
```diff
@@ -0,0 +1,137 @@
...
+
+generateSwagger = (routes, apiVersion) => {
+  var swagger;
+
+  swagger = {
+    basePath: '/api',
+    consumes: [
+      'application/json'
+    ],
+    host: 'sandbox.kuzzle.io:7511',
```
jenow: Why sandbox ?

j33f: why not ? :) the goal of this url is to allow people to test the routes, so a sandbox looks a good idea to me

benoitvidis:
+1

lib/api/controllers/remoteActions/swagger.js
```diff
@@ -0,0 +1,145 @@
...
+      reg = /:([^\/]*)/g, 
+      m;
+
+    if (swagger.paths[route.url] === undefined) {
+      swagger.paths[route.url] = {};
+    }
+
+    if (swagger.paths[route.url][route.verb] === undefined) {
+      if (route.infos !== undefined) {
+        route.infos.description += '\nController: ' + route.controller + '. Action: ' + route.action + '.';
```
ballinette: if route.infos is defined, but route.infos.description is not the result of this will be:
```
undefined
Controller: foo Action: bar
```

we should add, just before this line:
```
if (route.infos.description === undefined) {
  route.infos.description = '';
}
```

Or guess that the "description" attribute is mandatory ? but how to check this ?

lib/api/controllers/remoteActions/swagger.js
```diff
@@ -0,0 +1,145 @@
...
+                status: 'integer'
+              }
+            }
+          },
+          parameters: []
+        };
+      }
+
+      while ((m = reg.exec(route.url)) !== null) {
+        if (m.index === reg.lastIndex) {
```
ballinette: Why is it necessary ?
If I understand the regexp specifications well:
* m.index will be the index of the first character of the matched pattern
* reg.lastIndex will be the index just after the last character of the patched pattern.
To technically, we always have reg.lastIndex > m.index.

Is there a case when it not works as expected ?

lib/api/controllers/remoteActions/swagger.js
```diff
@@ -0,0 +1,145 @@
...
+
+  routes.push({verb: 'get', url: '/_serverInfo', controller: 'read', action: 'serverInfo'});
+
+  kuzzle.pluginsManager.routes.forEach(_route => {
+    var route = _.assign({}, _route);
+    route.url = '/' + kuzzle.config.apiVersion + '/_plugin' + route.url;
+    routes.push(route);
+  });
+
+  if (this.kuzzle.isServer) {
```
ballinette: Maybe we should move the condition in top of the method ?
Something like:
```
module.exports = function GenerateSwaggerFiles (kuzzle) {
  var routes = [];

  if (! this.kuzzle.isServer) {
    return q({isWorker: true});
  }

  this.kuzzle = kuzzle;

// and the rest of the code....

}
```

lib/api/controllers/remoteActions/swagger.js
j33f: in the swager format, it is mandatory. I do prefer to inform people that they have to do things right, instead of putting unnecessary (imho) tests, to test if things have been done the right way...

lib/api/controllers/remoteActions/swagger.js
j33f: good one, yep, you are right

lib/api/controllers/remoteActions/swagger.js
j33f: because it is the way it works : lastIndex can be seen as "the last captured ocurence" or "last captured match" when using the /g (greedy) flag.

ballinette: It's not what I understand from the JS reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex

> lastIndex is set to **the next position following** the most recent match.

j33f: from: https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/RegExp/exec
> re 	lastIndex	The index at which to start the next match. When "g" is absent, this will remain as 0.

ballinette: yes "at wich to start the next match", that means "after the current match"
Example:
```
> var re = /\d\d/g;
> console.log(re.exec('1234567890'));
[ '12', index: 0, input: '1234567890' ]
> console.log(re.lastIndex);
2
> console.log(re.exec('1234567890'));
[ '34', index: 2, input: '1234567890' ]
> console.log(re.lastIndex);
4
> console.log(re.exec('1234567890'));
[ '56', index: 4, input: '1234567890' ]
```

lib/api/controllers/remoteActions/swagger.js
```diff
@@ -82,9 +82,7 @@ generateSwagger = (routes, apiVersion) => {
       }
 
       while ((m = reg.exec(route.url)) !== null) {
-        if (m.index === reg.lastIndex) {
-          reg.lastIndex++;
-        }
+        reg.lastIndex++;
```
ballinette: again, no need to increment reg.lastIndex here, the cursor is already placed at the good index to continue the searh.

lib/api/controllers/remoteActions/swagger.js
ballinette: mhh, I could agree with your approach if we piped directly the route.infos to swagger without altering it (in that case, swagger would claim the missing attributes).

But here, we alter the routes.infos.description attribute, adding controller and action.
Then, if the user do not give this attribute, he won't have an explicit error, but just generate a bad description content, and may not be notified about that.

stafyniaksacha:
LGTM

bin/commands/swagger.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable no-console */
```
scottinet: This eslint command should be removed

ballinette:
+1

bin/kuzzle
```diff
@@ -77,6 +77,12 @@ program
   .description('Empty all the Kuzzle storage and make it look like a virgin, touched for the very first time')
   .action(require('./commands/reset'));  
 
+// $ kuzzle likeAvirgin
```
scottinet: This comment is unuseful and wrong

lib/api/controllers/remoteActions/swagger.js
```diff
@@ -0,0 +1,148 @@
...
+                state: 'string',
+                status: 'integer'
+              }
+            }
+          },
+          parameters: []
+        };
+      }
+
+      while ((m = reg.exec(route.url)) !== null) {
```
scottinet: I have a hard time figuring out what you're filtering using this regular expression.
Could you add a comment with some example to make this a bit clearer?

lib/api/controllers/remoteActions/swagger.js
```diff
@@ -0,0 +1,148 @@
...
+        reg.lastIndex++;
+        swagger.paths[route.url][route.verb].parameters.push({
+          description: 'TODO',
```
scottinet: Why a "TODO" description?

lib/api/controllers/remoteActions/swagger.js
```diff
@@ -0,0 +1,148 @@
...
+  });
+
+  return q(swagger);
+};
+
+writeFiles = swagger => {
+  var baseDir = path.join(__dirname, '..', '..', '..', '..');
+
+  try {
+    fs.writeFileSync(path.join(baseDir, 'kuzzle-swagger.json'), stringify(swagger));
```
scottinet: Nitpicking: is there a need for the JSON output to be sorted? If not, could this be replaced with a basic `JSON.stringify` to lower the number of dependancies?

lib/api/controllers/remoteActions/swagger.js
```diff
@@ -0,0 +1,148 @@
...
+  }
+  catch (e) {
+    q.reject(new InternalError('Unable to write the kuzzle-swagger.json file'));
```
scottinet: This rejected promise must be returned. Otherwise, you're only creating an unused promise, and the final result will be `return q(swagger)` even if this fails.

lib/api/controllers/remoteActions/swagger.js
```diff
@@ -0,0 +1,148 @@
...
+  }
+  try {
+    fs.writeFileSync(path.join(baseDir, 'kuzzle-swagger.yml'), jsonToYaml.stringify(swagger));
+  }
+  catch (e) {
+    q.reject(new InternalError('Unable to write the kuzzle-swagger.yml file'));
```
scottinet: This rejected promise must be returned. Otherwise, you're only creating an unused promise, and the final result will be `return q(swagger)` even if this fails.

lib/api/controllers/routerController.js
```diff
@@ -418,15 +215,44 @@ module.exports = function RouterController (kuzzle) {
...
       response.end(stringify({status: 200, error: null, result: {message: 'Available routes for this API version by verb.', routes: _routes}}));
     });
 
+    apiBase.get('/swagger.json', (request, response) => {
+      var content;
+      try {
+        content = fs.readFileSync(__dirname + '/../../../kuzzle-swagger.json');
+      }
+      catch (e) {
+        content = stringify({error: 'The swagger file has not been generated. Use the `kuzzle swagger` CLI command to generate it.', status: 500});
```
scottinet: I'm not fond of this solution, because this means that an administrator **must** execute this CLI command for the Swagger route to work.

Since the CLI is using remote actions anyway, wouldn't be a better approach to write the swagger generation code in a core module, returning a promise resolving to a JSON object?
This way, the CLi would still allow generating JSON and YAML files, using this piece of code to generate the swagger object, and this HTTP route would simply call it and return the result.

What do you think?

lib/api/controllers/routerController.js
```diff
@@ -418,15 +215,44 @@ module.exports = function RouterController (kuzzle) {
...
       response.setHeader('Access-Control-Allow-Origin', '*');
       response.setHeader('Access-Control-Allow-Headers', 'X-Requested-With');
       response.writeHead(200, {'Content-Type': 'application/json'});
       response.end(stringify({status: 200, error: null, result: {message: 'Available routes for this API version by verb.', routes: _routes}}));
     });
 
+    apiBase.get('/swagger.json', (request, response) => {
+      var content;
+      try {
+        content = fs.readFileSync(__dirname + '/../../../kuzzle-swagger.json');
```
scottinet: Exposing a `sync` FS operation to a HTTP API route is literally asking for problems: anyone executing this route will **freeze** NodeJS' callstack and Kuzzle during the time the file is read.

This PR can't go live until this has been replaced with an async fie read.

scottinet:
Sorry but I must give a -1 to this one, at least until my comment about the "readFileSync" is treated

lib/api/controllers/routerController.js
```diff
@@ -418,15 +215,44 @@ module.exports = function RouterController (kuzzle) {
...
+      response.setHeader('Access-Control-Allow-Origin', '*');
+      response.setHeader('Access-Control-Allow-Headers', 'X-Requested-With');
+      response.writeHead(200, {'Content-Type': 'application/json'});
+      response.end(content);
+    });
+
+    apiBase.get('/swagger.yml', (request, response) => {
+      var content;
+      try {
+        content = fs.readFileSync(__dirname + '/../../../kuzzle-swagger.yml');
```
scottinet: :warning: same here, please replace this with an async version

lib/api/controllers/routerController.js
ballinette: I agree with this suggestion.

test/api/controllers/remoteActionsController/swagger.test.js
```diff
@@ -0,0 +1,113 @@
...
+
+    swagger(kuzzle)
+      .then((response) => {
+        should(fs.writeFileSync.called).be.false();
+        should(response.isWorker).be.true();
+      });
+  });
+
+  it ('should generate default swagger infos for poorly described routes', () => {
+    sandbox.stub(fs, 'writeFileSync', () => {
```
scottinet: You can simply write: `sandbox.stub(fs, 'writeFileSync').returns(true)` :smile:

test/api/controllers/remoteActionsController/swagger.test.js
```diff
@@ -0,0 +1,113 @@
...
+        should(result.paths['/1.0/_plugin/myplugin/bar/:name']).be.an.Object();
+      });
+  });
+
+  it ('should fail if something goes wrong with the swagger files writing', () => {
+    sandbox.stub(fs, 'writeFileSync', () => {
+      throw 'error';
+    });
+
+    should(swagger(kuzzle)).be.rejectedWith(InternalError);
```
scottinet: This test doesn't work (meaning it always succeeds no matter what): should require promises to be returned for `fulfilled...` and `rejected...` to work.

This should be replaced with:
`return should(swagger(kuzzle)).be.rejectedWith(InternalError);`

test/api/controllers/remoteActionsController/swagger.test.js
```diff
@@ -0,0 +1,113 @@
...
+    sandbox.stub(fs, 'writeFileSync', () => {
+      return true;
+    });
+
+    sandbox.stub(kuzzle.pluginsManager, 'routes', [
+      {verb: 'post', url: '/myplugin/bar', controller: 'myplugin/foo', action: 'bar'},
+      {verb: 'get', url: '/myplugin/bar/:name', controller: 'myplugin/foo', action: 'bar'}
+    ]);
+
+    swagger(kuzzle)
```
scottinet: When testing promises with should, you either need a `done` callback or, better yet, to return that promise, ensuring that `should` will wait for it to be fulfilled/rejected before marking the test as a success/failure

test/api/controllers/remoteActionsController/swagger.test.js
```diff
@@ -0,0 +1,113 @@
...
+
+  it ('should do nothing if Kuzzle is a worker', () => {
+    sandbox.stub(fs, 'writeFileSync', () => {
+      return true;
+    });
+
+    sandbox.stub(kuzzle, 'isServer', false);
+    sandbox.stub(kuzzle, 'isWorker', true);
+
+    swagger(kuzzle)
```
scottinet: same here, the promise must be returned

test/api/controllers/remoteActionsController/swagger.test.js
```diff
@@ -0,0 +1,113 @@
...
+        should(response.isWorker).be.true();
+      });
+  });
+
+  it ('should generate default swagger infos for poorly described routes', () => {
+    sandbox.stub(fs, 'writeFileSync', () => {
+      return true;
+    });
+
+    swagger(kuzzle)
```
scottinet: same here, the promise must be returned

test/api/controllers/remoteActionsController/swagger.test.js
```diff
@@ -0,0 +1,113 @@
...
+        should(response.paths['/' + kuzzle.config.apiVersion + '/basic'].description).be.eql('Controller: basic. Action: basic.');
+      });
+  });
+
+  it ('should the swagger infos described for routes which have them', () => {
+    sandbox.stub(fs, 'writeFileSync', () => {
+      return true;
+    });
+
+    swagger(kuzzle)
```
scottinet: same here, the promise must be returned

lib/api/controllers/remoteActions/swagger.js
j33f: absolutely not and yes JSON.stringify is ok

test/api/controllers/routerController/initRouterHttp.test.js
```diff
@@ -29,11 +30,22 @@ function parseHttpResponse(response) {
...
 
-    try {
-      result = JSON.parse(data);
-    }
-    catch (e) {
-      deferred.reject(e);
+    if (yaml === true) {
+      yamlToJson(data, (err, res) => {
+        if (err) {
+          deferred.reject(err);
```
scottinet: since the execution flow isn't stopped, the deferred promise will eventually be resolved instead
you should directly returns this with `return deferred.reject(err);`

test/api/controllers/routerController/initRouterHttp.test.js
```diff
@@ -29,11 +30,22 @@ function parseHttpResponse(response) {
...
+
+        result = res;
+      });
+    } 
+    else {
+      try {
+        result = JSON.parse(data);
+      }
+      catch (e) {
+        deferred.reject(e);
```
scottinet: since the execution flow isn't stopped, the deferred promise will eventually be resolved instead
you should directly returns this with `return deferred.reject(err);`

lib/api/controllers/remoteActions/swagger.js
j33f: because the description must exists, and TODO is here to say that we will complete the description, one day

lib/api/controllers/remoteActions/swagger.js
j33f: yep I'll do

lib/api/controllers/remoteActions/swagger.js
j33f: :+1:

lib/api/controllers/remoteActions/swagger.js
scottinet: Thanks!

Tell me when it's done so I can review this code part (still can't figure out what it really does :smile: )

lib/api/controllers/routerController.js
j33f: well, the goal here, was to respond to those: 
- should the swagger file include routes declared by plugins ? (we ( @benoitvidis and I) answered yes)
- should the swagger file be generated on demand (via http get) ? (we answered no)
- should the swagger files be distributed with Kuzzle ? (we answered no again)

The idea behind the CLI way of generating the swagger files was to allow anyone to generate a  file reflecting its very own Kuzzle environment (with plugins path) and to generate it when it is useful.

Since not anyone will use those swagger files at the moment, and since it is really useful to generate them when an administrator (de)activates plugins, this way looked a good way to do this.

Another option may be to add a specific route for admins only in order to generate the swagger files (from the BO for instance).

lib/api/controllers/routerController.js
j33f: yep my fault... please don't hurt me

test/api/controllers/remoteActionsController/swagger.test.js
j33f: fixed

test/api/controllers/remoteActionsController/swagger.test.js
j33f: fixed

j33f:
referring to [this comment](https://github.com/kuzzleio/kuzzle/pull/281#discussion_r66093167)

Another option could be to generate the swagger files during the Kuzzle start process

What do you think about it ? @scottinet ?

lib/api/controllers/routerController.js
scottinet: The solution I propose wouldn't generate swagger files. What I propose is to generate *swagger documentation* dynamically and, if called by the CLI, to store this documentation into files. This would answer to the points you stated.

The problem I see with your solution is a practical one: like you said, swagger routes should be used scarcely. But when it's needed, the probability that the files haven't been generated, or that they are outdated, is very high. 
And, of course, there is the problem of having an enabled HTTP route basically answering that an administrator action is needed first. That's not very practical either.

Here is an example: let's assume that you install Kuzzle and generate those files. Then you run it for a long time, patch it several times and also install a bunch of controller plugins, without ever generating the swagger files. When you suddenly need the documentation using HTTP, you get a dangerously outdated one. And you don't even know it.

One last matter: I didn't see where you inject controller plugins swagger documentation in the files. Can you point me to it?

scottinet:
Well, I'm not fond of performing potentially useless tasks too. I mean, generating these files if they aren't needed will only slower Kuzzle start sequence for nothing.

Moreover, when we'll make plugin installation and configuration dynamic, the swagger documentation won't be updated until restart. And if you're installing a controller plugin, you may want to check the swagger documentation right away to check if it has succeeded, leading you to think that the plugin install failed, when actually everything went fine.

lib/api/controllers/routerController.js
```diff
@@ -226,31 +226,33 @@ module.exports = function RouterController (kuzzle) {
...
-      }
       response.setHeader('Access-Control-Allow-Origin', '*');
       response.setHeader('Access-Control-Allow-Headers', 'X-Requested-With');
-      response.writeHead(200, {'Content-Type': 'application/json'});
-      response.end(content);
+
+      fs.readFile(__dirname + '/../../../kuzzle-swagger.json', (err, data) => {
+        if (err) {
+          response.writeHead(500, {'Content-Type': 'application/json'});
+          reponse.end(JSON.stringify({error: 'The swagger file has not been generated. Use the `kuzzle swagger` CLI command to generate it.', status: 500}))
```
scottinet: you need to return from the function after sending the error response, otherwise you'll send the data one too (with undefined data might I add)

lib/api/controllers/routerController.js
```diff
@@ -226,31 +226,33 @@ module.exports = function RouterController (kuzzle) {
...
-      }
       response.setHeader('Access-Control-Allow-Origin', '*');
       response.setHeader('Access-Control-Allow-Headers', 'X-Requested-With');
-      response.writeHead(200, {'Content-Type': 'application/yaml'});
-      response.end(content);
+
+      fs.readFile(__dirname + '/../../../kuzzle-swagger.yml', (err, data) => {
+        if (err) {
+          response.writeHead(500, {'Content-Type': 'application/yaml'});
+          reponse.end(JSON.stringify({error: 'The swagger file has not been generated. Use the `kuzzle swagger` CLI command to generate it.', status: 500}))
```
scottinet: you need to return from the function after sending the error response, otherwise you'll send the data one too (with undefined data might I add)

lib/api/controllers/remoteActions/swagger.js
```diff
@@ -44,58 +43,69 @@ generateSwagger = (routes, apiVersion) => {
 
   routes.forEach(route => {
     var 
-      reg = /:([^\/]*)/g, 
+      reg = /:([^\/]*)/g, // since url parameters got the following form "/:parameterName/:otherParameter" this reg captures, all chars following : which are not a / char
+                          // the first and las / are reg delimiters
```
scottinet: nitpicking: well... the first comment line is **very** helpful, but the others are only explaining regular expressions, and can/should be removed

lib/api/controllers/remoteActions/swagger.js
```diff
@@ -0,0 +1,158 @@
...
+      'http'
+    ],
+    swagger: '2.0',    
+    paths: {}
+  };
+
+  routes.forEach(route => {
+    var 
+      reg = /:([^\/]*)/g, // since url parameters got the following form "/:parameterName/:otherParameter" this reg captures, all chars following : which are not a / char
+                          // the first and las / are reg delimiters
```
scottinet: nitpicking: well... the first comment line is **very** helpful, but the other ones are only explaining regular expressions, and can/should be removed

lib/api/controllers/routerController.js
j33f: since the routes are taken directly from the Kuzzle object, the plugins routes are automagically included :)

j33f:
ready for a last +1 @scottinet :)

scottinet:
+:100:

PR #284, opened 2016-05-26
Kuz 552 Communication between LB and Kuzzle Server

AnthonySendra:
> :warning: Must be merged along with [this LB PR](https://github.com/kuzzleio/kuzzle-load-balancer/pull/4)

**Remove kuzzle/lib/api/core/servers.js**
Because we don't have only "server" where the data can comes from, the initialization has been moved to `kuzzle/lib/api/core/entryPoints`.
This new script will run the HTTP server (because the LB proxyfy all request to this server), the MQ server (will be removed once protocol plugin MQ is ready) and the LB communication. All those 3 communications way are the new entry points for the Kuzzle server.

**Move plugin-protocol-multiplexer to Kuzzle Core**
The old code in the plugin protocol multiplexer is now in Kuzzle core, in `entryPoints/lb.js` for LB communication and in the new file `api/core/cluster/ClusterManager` for the `httpPort` sharing.

**Cluster**
A new file `core/cluster/ClusterManager` was created for manage the cluster: communication between Kuzzle Server and all the logic about Cluster (scalability, data dump, ...) must be in this file (or, at least, in this folder)
 
**Dummy: true**
Fix service initialization when we provide an empty whitelist.
Before, when we provide an empty whitelist, all services was initialized instead of no one. Now, when dummy is true, we don't init any service.
Fix unit test for use mock instead of service.


lib/api/core/entryPoints/lb.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  this.kuzzle.services.list.lbBroker.send('broadcast', data);
+};
+
+function onRequest (data) {
+  var requestObject;
+
+  if (data.request && data.context && data.context.connection && data.context.connection.type) {
+    requestObject = new RequestObject(data.request, {}, data.context.connection.type);
```
benoitvidis: If I am not mistaken, the data body will be lost using this.

    requestObject = new RequestObject(data.request , data.request.data, data.context.connection.type);

lib/api/core/entryPoints/lb.js
```diff
@@ -0,0 +1,71 @@
...
+  this.kuzzle.services.list.lbBroker.send('broadcast', data);
+};
+
+function onRequest (data) {
+  var requestObject;
+
+  if (data.request && data.context && data.context.connection && data.context.connection.type) {
+    requestObject = new RequestObject(data.request, data.request.data, data.context.connection.type);
+
+    this.kuzzle.funnel.execute(requestObject, data.context, (error, responseObject) => {
```
ballinette: why do you call directly funnel.execute() instead of calling routerController.execute() method (cf. https://github.com/kuzzleio/kuzzle/blob/KUZ-552-lb-communication/lib/api/controllers/routerController.js#L78 ) ?

Are the checks implemented in the routerController.execute() method unnecessary ? if so, I guess we should simply remove the routerController.execute() method, as plugin protocols are now moved to LB...

benoitvidis: +1 on removing the routerController.execute method, which is plugin specific.

AnthonySendra: Yep the function`routerController.execute` must be removed but this is not the aim of this PR. We have an issue for removing plugin protocol: KUZ-550

benoitvidis: true.

benoitvidis:
+1

lib/api/core/entryPoints/lb.js
ballinette: OK

ballinette:
+1

j33f:
+1

PR #285, opened 2016-05-26
Kuz 555 split internal broker into client and server

benoitvidis:
This PR splits the Internal broker into a client and a server component. The purpose is to be able to re-use these components for the communication between Kuzzle and its LB on one hand, and between Kuzzle slave and master nodes on the other hand.

Changes:
* The `broadcast` method now sends the given message to all clients **but** the current one.
* Removed method: `listenOnce`
* New method: `unsubscribe`
* Tests: Previous tests were using and exposing a real Websocket server. The Websocket components are now mocked.
* Not directly related change: The DSL currified methods are now built in a minimal form using eval (I know...). The purpose is to later on be able to serialize and share them between the Kuzzle cluster nodes.


bin/commands/start.js
```diff
@@ -132,7 +135,7 @@ module.exports = function () {
       }
     })
     .catch(err => {
-      console.error(err);
+      console.log(err.stack);
```
scottinet: Since we're exiting the program with an error code, shouldn't the error stack be printed on stderr?

lib/api/dsl/methods.js
```diff
@@ -856,11 +858,12 @@ function buildCurriedFunction(index, collection, field, operatorName, value, cur
   }
 
   if (!methods.dsl.filtersTree[index][collection].fields[field][curriedFunctionName]) {
-    curriedFunction = _.curry(operators[operatorName]);
-    curriedFunction = _.curry(curriedFunction(field, value));
-    if (not) {
-      curriedFunction = _.negate(curriedFunction);
-    }
+    // we need the ugly eval code to make sure the function can later on be serialized and
```
scottinet: I disagree with the word "ugly" (but I'm fine with your comment): `eval` should be used with extreme care and only if no other option is possible, as is the case here, so... I see no problem (it reminds me of old debates about the `goto` instruction :stuck_out_tongue: )

scottinet:
+1

lib/api/dsl/methods.js
benoitvidis: I cannot agree with you on this one. eval is evil, especially using a direct call like we are doing here. But you're right, I cannot figure out a better alternative for the time being.

scottinet: For the record, here is what we decided Benoit and I about this matter: we'll try to get rid of this `eval` call by only storing operators arguments instead of a currified function. 
This should be better performance-wise (because we would get rid of eval + 2 x JSON.stringify), but we don't know how it would perform memory-wise. Should be way better than the situation before this pull request.

The operators arguments will be as easily shareable across all kuzzle instances than what Benoit is proposing in this pull request.

I'll check if that's feasible in a reasonable amount of time (according to Benoit, it should be), in another branch starting from this one.

lib/services/broker/wsBrokerServer.js
```diff
@@ -0,0 +1,275 @@
...
+      msg = JSON.parse(msg);
+
+      switch (msg.action) {
+        // the listen action is called by the client to inform the server it
+        // wants to be notified on the room activity
+        case 'listen':
+          if (this.rooms[msg.room] === undefined) {
+            this.rooms[msg.room] = new CircularList([]);
+          }
+          if (this.rooms[msg.room].getArray().indexOf(clientSocket) === -1) {
```
scottinet: `getArray` needs the `easy-circular-list` module to be upgrade to v1.0.13

AnthonySendra:
+1

lib/services/broker/wsBrokerServer.js
```diff
@@ -0,0 +1,275 @@
...
+      }
+    },
+    err => {
+      if (err) {
+        clients = -1;
+        _kuzzle.pluginsManager.trigger('log:error', err);
+      }
+    });
+
+  return clients;
```
stafyniaksacha: The `clients` variable may not be filled with the real number of clients due to `async.each` usage.

I think we must use the third argument of `async.each(coll, iteratee, [callback])` to resolve the `clients` in a promise 
Or use `Array.forEach` if we have to keep the `broadcast` in a synchrone way.

benoitvidis: Good catch!

Switched the broadcast to a synchone version back. 
NB: Additional change: The broadcast does not stop anymore if one or more clients do not receive the message.

stafyniaksacha:
+1

PR #288, opened 2016-05-31
add cli plugins --list option

stafyniaksacha:


bin/commands/plugins.js
```diff
@@ -88,8 +106,10 @@ module.exports = function pluginsManager (plugin, options) {
       .then(() => {
         release();
 
-        if (!plugin) {
-          console.log(clcOk('███ kuzzle-plugins: Plugins installed'));
+        if (options.install) {
```
ballinette: avoid nested if-then-else if not needed.
Here, we can just write 
```
if (options.install && !plugin) {
// (...)
}
```

Or: 
```
if (options.install) {
  if (plugin) {
    console.log(clcOk('███ kuzzle-plugins: Plugin ' + plugin ' installed'));
  }
  else {
    console.log(clcOk('███ kuzzle-plugins: Plugins installed')); 
  }
}
```

bin/commands/plugins.js
```diff
@@ -427,27 +443,29 @@ function checkOptions(plugin, options) {
     process.exit(1);
   }
 
-  // --install is the only option working without specifying a plugin name
-  if (!plugin && !options.install) {
-    console.error(clcError('A plugin [name] is required for this operation'));
-    process.exit(1);
-  }
+  // --install and --list are the only options working without specifying a plugin name
+  if (!options.list) {
```
ballinette: This code should be improved : 
* --install and --list are mutually exclusive (what is not checked yet)
* the condition "if (!plugin && !options.install && !options.list)" should not be nested.

Improvement proposal : 
```
if (!plugin && !options.install && !options.list) {
  console.error(clcError('A plugin [name] is required for this operation'));
  process.exit(1);
}

if (options.install && options.list) {
  console.error(clcError('Options --install and --list are mutually exclusive));
  process.exit(1);
}

// then put the rest of the code, unchanged
```

ballinette:
despite my nitpicking remarks, +1

bin/commands/plugins.js
stafyniaksacha: Done

scottinet:
+1

PR #289, opened 2016-05-31
KUZ-580 Replace eval with arguments storage

scottinet:
Following up #285, where DSL functions currified with lodash have been replaced with "eval" to allow sharing across Kuzzle instances, it has been decided with @benoitvidis that this solution could still be upgraded.

The proposed solution is to simply store operators arguments instead of currified functions.
These arguments can easily be shared between instances, and instead of calling the currified function, the correct operator will be invoked with the stored arguments instead.

The memory imprint might be a bit larger than with the "eval" solution (to be proved), but performances should be better with the removal of "eval" and of 2 x "JSON.stringify" calls.

# CHANGELOG

* removed all occurences of the word "currified" in the code
* replaced operator function evaluation with operator arguments storage in the DSL
* adapted unit tests
* split some geolocation unit tests between `dsl/methods` and `dsl/operators` tests, as these test files were testing both parts of the code
* replaced some uses of `q.defer()` with `q()` or `q.reject()`

### DIFF
--- a/lib/api/dsl/methods.js
+++ b/lib/api/dsl/methods.js
@@ -1,4 +1,3 @@
-
 var
   _ = require('lodash'),
   async = require('async'),
@@ -75,15 +74,15 @@ module.exports = methods = {
     async.each(Object.keys(filter[field]), function (fn, callback) {
       var
         value = filter[field][fn],
-        curriedFunctionName = '',
+        encodedFunctionName = '',
         result;
 
       if (not) {
-        curriedFunctionName += 'not';
+        encodedFunctionName += 'not';
       }
-      curriedFunctionName += 'range' + field + fn + value;
+      encodedFunctionName += 'range' + field + fn + value;
 
-      result = buildCurriedFunction(index, collection, field, fn, value, curriedFunctionName, roomId, not);
+      result = addToFiltersTree(index, collection, field, fn, value, encodedFunctionName, roomId, not);
       if (util.isError(result)) {
         callback(result);
         return false;
@@ -301,7 +300,7 @@ module.exports = methods = {
       deferred = q.defer(),
       fieldName,
       formattedFilters,
-      curriedFunctionName = '',
+      encodedFunctionName = '',
       inGlobals = false,
       result;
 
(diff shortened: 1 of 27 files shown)


lib/api/dsl/methods.js
```diff
@@ -75,15 +74,15 @@ module.exports = methods = {
...
 
       if (not) {
-        curriedFunctionName += 'not';
+        encodedFunctionName += 'not';
       }
-      curriedFunctionName += 'range' + field + fn + value;
+      encodedFunctionName += 'range' + field + fn + value;
 
-      result = buildCurriedFunction(index, collection, field, fn, value, curriedFunctionName, roomId, not);
+      result = addToFiltersTree(index, collection, field, fn, value, encodedFunctionName, roomId, not);
```
stafyniaksacha: What is the `fn` variable here ?

stafyniaksacha:
+1

lib/api/dsl/methods.js
scottinet: It's the range operator.

Here is a typical range filter:

```json
{
  "range": {
    "field": {
      "gt": 10,
      "lte": 123
    }
  }
}
```

The `fn` variable here refers to `gt` and `lte`. As it is not very clear in the code, I'll change this variable name to `rangeOperator`

benoitvidis:
nice ! :+1:

lib/api/dsl/methods.js
```diff
@@ -857,22 +852,15 @@ function buildCurriedFunction(index, collection, field, operatorName, value, cur
...
-    fnSource = 'd => ' + (not ? '!' : '') + `operators['${operatorName}'](${JSON.stringify(field)}, ${JSON.stringify(value)}, d);`;
-    curriedFunction = eval(fnSource);
-    /* eslint-enable no-eval */
-
-    methods.dsl.filtersTree[index][collection].fields[field][curriedFunctionName] = {
+  if (!methods.dsl.filtersTree[index][collection].fields[field][hashedFunctionName]) {
+    methods.dsl.filtersTree[index][collection].fields[field][hashedFunctionName] = {
       rooms: [],
-      fn: curriedFunction
+      args: {operator, not, field, value}
```
benoitvidis: did not know we could do this. This is neat.

jenow:
+1

PR #290, opened 2016-06-01
Fix: use getPathPlugin instead of require.resolve

scottinet:
Invoking `kuzzle plugins --remove <plugin name>` will make the CLI use `require.resolve` to get the directory path of the plugin, in order to delete it.
As seen with @dbengsch, using `require.resolve` is a bad idea as the entry file of the plugin is not necessarily located in its root directory.

This PR replaces `require.resolve` with the already existing `getPathPlugin` function.


dbengsch:
+1

jenow:
+1

PR #291, opened 2016-06-01
KUZ-484: Share DSL to plugins

scottinet:
# Purpose

This PR transforms the Kuzzle DSL into an independant internal module, and shares it to plugins via the plugins context. 
This way, plugins would be allowed to have their own DSL instance, independant from the one used by Kuzzle, and they would be free to use it as they please.

# Changelog

* made `dsl/methods` instantiable to ensure the filtersTree link is unique for each DSL instance
* renamed all "room..." variables into "filter..." in the DSL, to make these variables agnostic. Only the `hotelClerk` core module should manage rooms.
*  created a `dsl/filters` submodule, handling the filters tree and all associated operations. This will be the only mutable part in the DSL module
* the DSL now only publishes the following methods: 
  * `createFilterId(index, collection, filters)`: creates an unique ID corresponding to the provided arguments
  * `register(filterId, index, collection, filters)`: subscribes a filter to the DSL 
  * `remove(filterId)`: removes a filter from the DSL
  * `test(index, collection, data, [documentId])`: tests data against stored DSL filters
  * `exists(index, collection)`: checks if filters are registered on an index-collection pair
  * `getFilterIds(index, collection)`: gets all filter IDs registered on an index-collection pair
*  decoupled the DSL from the `hotelClerk` core module. The `hotelClerk` is now an ordinary DSL user
* The `hotelClerk` does not maintain encoded filters anymore. These are now managed by the new `dsl/filters` module instead
* The plugins context now contains a DSL constructor, allowing plugins to instanciate their own DSL
* Updated the DSL `README.md` documentation

# Boyscout changes

* replaced `promise().then(() => callback).catch(err => callback(err))` with `promise.nodeify(callback)`
* replaced `if (err) { deferred.reject(err) } else { deferred.resolve(...)}` with `deferred.makeNodeResolver()`


scottinet:
This PR can be reviewed. Obviously I still have some work to do on code coverage, so please don't merge it yet. :-)

scottinet:
Codecov seem to not update comments anymore: coverage is now about 92.9%

benoitvidis:
:+1

lib/api/core/hotelClerk.js
```diff
@@ -410,66 +336,39 @@ function removeListRoomsInCollection (index, collection, rooms) {
...
-    filters = {};
-  }
-
-  Object.keys(filters).forEach(key => sortedFilters.push([key, filters[key]]));
-  stringifiedFilters = stringify(sortedFilters.sort((a, b) => {
-    return String(a[0]).localeCompare(b[0]);
-  }));
-
-  stringifiedObject = stringify({index: index, collection: collection, filters: stringifiedFilters});
+  roomId = this.kuzzle.dsl.createFilterId(index, collection, filters);
```
benoitvidis: Is there some kind of canonical transformation left? I could not see it in the dsl::createFilter method?

lib/api/dsl/index.js
```diff
@@ -1,117 +1,67 @@
...
-    }
-
-    privateFilterName = _.camelCase(filterName);
-
-    if (!methods[privateFilterName]) {
-      return q.reject(new NotFoundError('Unknown filter with name '+ privateFilterName));
-    }
-    
-    return this.methods[privateFilterName](roomId, index, collection, filters[filterName]);
+    return crypto.createHash('md5').update(stringify(idObject)).digest('hex');
```
ballinette: In the former version, we had this kind of code to sort filters and be sure that filters with different orders (for example: `{and: [{term: {city: 'NYC'}},{term: {hobby: 'computer'}}]}` and `{and: [{term: {hobby: 'computer'}},{term: {city: 'NYC'}}]}` ) will produce the same hash:
```
idObject.filters = stringify(idObject.filters.sort((a, b) => {
  return String(a[0]).localeCompare(b[0]);
}));
```

Is it unnecessary now ?

lib/api/core/hotelClerk.js
scottinet: This code has been removed because it's redundant with the `json-stable-stringify` module.

I'm ashamed to say that it was me who have added these transformations a few months ago without reading `json-stable-stringify` documentation. This module does exactly what we did here: sort the JSON terms to ensure we'll get the same hash when encoding differently sorted identical filters.

lib/api/dsl/index.js
scottinet: Same answer than the one I gave to @benoitvidis : it's redundant with `json-stable-stringify` :smiley:

ballinette: ok ;)

lib/api/core/hotelClerk.js
```diff
@@ -561,26 +451,23 @@ function cleanUpRooms (roomId) {
      */
     this.rooms[roomId].destroyed = true;
 
-    this.kuzzle.dsl.removeRoom(this.rooms[roomId])
+    return this.kuzzle.dsl.remove(roomId)
       .then(() => {
         this.kuzzle.pluginsManager.trigger('room:remove', roomId);
-        deferred.resolve(roomId);
+        return roomId;
```
ballinette: shouldn't we return `q(roomId)` instead ?

scottinet: It leads to the same result: using [propagation](https://github.com/kriskowal/q#propagation), a promise chain resolves promises until it gets a value, or until it encounters a throw/promise rejection.

So returning a new promise resolving to `roomId` will only add a promise to the chain, postponing the result with no gain.

ballinette:
great job! :+1:

stafyniaksacha:
Yeah + :100:

PR #292, opened 2016-06-02
KUZ-588 Enabling Kuzzle modules

benoitvidis:
The proposed implementation is pretty straight-forward:

* Modules are plugins. The plugin configuration specifies a privileged = true option
* Modules are handled exactly just like a regular plugin, with the exception that the given context is ~~the Kuzzle instance~~ a new privilegedContext object: .

```javascript
    privilegedContext = {
      kuzzle,
      contructors: {
        services: {
          broker: {
            WsBrokerClient: require('../../../services/broker/wsBrokerClient'),
            WsBrokerServer: require('../../../services/broker/wsBrokerServer')
          }
        }
      }
    };
```


j33f:
+1

stafyniaksacha:
+1

ballinette:
I still disagree with the idea to manage modules exactly like standard plugins, but if it is really what the team wants, why the privilegedContext has a structure totally different as the non-privileged context ?

We have 2 choices:

1. Either we prefer really distinguish privileged modules and non privileges plugins (it is my opition), because, for example, injecting the whole kuzzle object to a plugin could enable it to break the kernel.
  In that case:
  * modules should not be installed, enabled or disabled on the fly by the CLI
  * They should only be installed via dependencies in the package.json, configured within .kuzzlerc like the kernel's configuration, and launched only at startup.

2.  Or, we decide to not care about separating the notions of plugins and modules, and let the integrators be responsible about which privileged an non-privileged plugins he installs, and the plugin's developper be responsible about what he does with the context...
  ... but in that case:
  *  why not just inject the whole privileged context to all plugins ?
  * if we keep a dinstinction btw. privileged/non privileged, I suggest at least to keep the same structure, I mean, let the privilegedContext inherit to the non privileged one, just adding privileged objects.

ballinette:
need @scottinet 's advice about that.

scottinet:
I have mixed feelings about this issue.

I personally prefer a clean separation between plugins and modules:
* I find it far easier to document and to explain: plugins allow adding features and behaviors easily and quickly, while modules allow to modify Kuzzle at a much lower level.
* the API should soon support installation, configuration and removal of plugins without having to restart Kuzzle, and modules should be left out of this feature, for obvious reasons. It's more difficult to explain why "privileged **plugins**" cannot be configured without a restart
* I find it far easier to administrate Kuzzle if plugins and modules are kept well separated

Now that I have said that, I understand @benoitvidis' take on this one. For one, it can be difficult to distinguish the "non-lowlevelness" of a plugin, for instance with protocol plugins.
And moreover, modules and plugins are disturbingly similar, separating the two will inevitably lead to repetition of code.

Anyway, my take on this is: as I stated yesterday, I personally prefer not treating modules like plugins. But I don't have stronger arguments than what I stated, so for now I'm fine with this PR (didn't review it yet, though).

About the suggestions @ballinette proposed in his part 2:
* I think it's still better to draw a clear line between non-privileged and privileged plugins, if only to assure people that some plugin is potentially harmless (or the opposite with privileged plugins). 
* I agree with the second suggestion: even if the objects exposed in the plugin context can also be accessed through the kuzzle instance directly, I think it makes it easier for plugin developers to have direct access to them

benoitvidis:
I don't see the issue in dealing with privileges plugins the same way as we would do with regular ones.
We actually do this every day on our systems when running `sudo` or `docker run --privileged`.

@ballinette to answer your questions on the second hypothesis:
> why not just inject the whole privileged context to all plugins ?

To me, this is the exact same reason why we do not run Linux as a root user.
> Privileged context should inherit from the regular one

Done, but I am not happy with it. I modified the PluginContext to use a real constructor but its current implementation is a flat structured object, without any kind of organisation. I put some additional structure in the PrivilegedContext, which brings some inconsistency.

Still on this subject, if we decide we need to deal with it with some extra code, maybe it would be worth thinking about restructuring the plugin context. If so, the sooner we do, the less painful it will be but will come with some costs anyway.

Anyway, we have some current developments that rely on the ability to access some core components from Kuzzle. This proposal is not perfect but has the benefit of requiring a very low amount of work.

If we have some disagreements on how to implement such a need, I would suggest to discuss it IRL.

scottinet:
We definitely need a workshop on this one. Can you organize one, @benoitvidis ?

xbill82:
**Nitpicking**: the `pluginsContext.js` file name does not match the `PluginContext` class name. Maybe this is because the file contains both `PluginContext` and `PrivilegedPluginContext`, so the intent of the naming was "here go the contexts of the plugins" => which would mean the file should be called `pluginContexts` (with the `s` at the end).

This is the worst nitpicking I've ever made.

By the way, I would call the file `pluginContext.js`.

scottinet:
Following our yesterday's discussion, I guess the only change on this PR would be to put the `kuzzle` accessor in an `accessors` sub-object.
That way you can work on your branch, and I'll harmonize everything in an upcoming PR, without moving any of this PR newly added objects. WDYT?

lib/api/core/plugins/pluginsContext.js
```diff
@@ -1,52 +1,61 @@
...
+
+function PluginContext () {
+  this.RequestObject = RequestObject;
+  this.RequestObject = ResponseObject;
+  this.NotificationObject = NotificationObject;
+  this.Dsl = require('../../dsl');
+  this.httpPort = _kuzzle.config.httpPort;
+
+  Object.keys(errors).forEach(errorConstructor => {
+    this[errorConstructor] = errors[errorConstructor];
```
scottinet: For compatibility reasons, `this[errorConstructor]` must have its first letter capitalized: `this[_.upperFirst(errorConstructor)]` = ....`

lib/api/core/plugins/pluginsContext.js
```diff
@@ -1,52 +1,61 @@
...
+  errors = require('kuzzle-common-objects').Errors,
+  RequestObject = require('kuzzle-common-objects').Models.requestObject,
+  ResponseObject = require('kuzzle-common-objects').Models.responseObject,
+  NotificationObject = require('../models/notificationObject'),
+  _kuzzle;
+
+
+function PluginContext () {
+  this.RequestObject = RequestObject;
+  this.RequestObject = ResponseObject;
```
scottinet: should be `this.ResponseObject` instead

benoitvidis:
closing this PR. Handled by #316

PR #294, opened 2016-06-02
[BETA] update passport local

jenow:
Updated kuzzle-plugin-auth-passport-local.
See kuzzleio/kuzzle-plugin-auth-passport-local#10


ballinette:
+1

stafyniaksacha:
+1

PR #300, opened 2016-06-06
Rename loadbalancer as proxy in all docker compose and base it on new image names

dbengsch:


stafyniaksacha:
+1

ballinette:
+1
Just to be sure: do we also plan to rename LB "entry point" and LB broker ?

dbengsch:
@ballinette : I think so, but I must admit I only cared about the docker images consistency in these changes.

jenow:
+1

PR #303, opened 2016-06-08
Add Errors objects into PluginContext

stafyniaksacha:
Add Errors objects into PluginContext

### DIFF
--- a/lib/api/core/plugins/pluginsContext.js
+++ b/lib/api/core/plugins/pluginsContext.js
@@ -1,11 +1,9 @@
 module.exports = function PluginContext(kuzzle) {
   var
-    context = {
-      RequestObject: require('kuzzle-common-objects').Models.requestObject,
-      ResponseObject: require('kuzzle-common-objects').Models.responseObject,
-      Errors: require('kuzzle-common-objects').Errors,
-      NotificationObject: require('../models/notificationObject')
-    };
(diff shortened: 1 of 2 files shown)


scottinet:
uh... errors are shared, too. They are simply dynamically searched. Or at least they were, but since these objects have been moved into the `kuzzle-common-objects` module, the filesystem search fails and there is no error objects to be exposed in the context.

So I suggest we fix the dynamic error objects exposure instead of removing `RequestObject` and `ResponseObject` objects.

ballinette:
PR title should be updated, because now, we do not remove RequestObject & ResponseObject, but we add Error objects ;)

But +1 for this new version ;)

scottinet:
+1!

benoitvidis:
Hum... If I am not mistaken, the current implementation breaks the previous implementation where the Error objects were put at the root level.
I agree that we may need to update the plugin context structure, as discussed in #292. 

I suppose we first need to decide if we change the plugin context structure and if so, agree on the new formatting before we update all Kuzzle plugins.

scottinet:
Ah, didn't think of that. I agree with @benoitvidis, let's hold this one until our (little) workshop about the plugins context structure.

scottinet:
Closing this PR, I'll prepare another branch with a new plugin context structure.

PR #312, opened 2016-06-09
[release-1.0.0-RC3] Fix repositories

stafyniaksacha:
Temporary fix for profile & role hydratation.

also add new actions from authController into defaults roles


xbill82:
+1

ballinette:
+1

benoitvidis:
+1

ballinette:
hmm your latest change has broken the ProfileRepository.prototype.loadProfile() method.
As if the cache engine would not be correctly set.

I'm looking at this

ballinette:
@benoitvidis @xbill82 is it still OK for you after my changes ?

.kuzzlerc
```diff
@@ -118,9 +118,11 @@
       "controllers": {
         "auth": {
           "actions": {
-            "login": true,
+            "logout": true,
```
scottinet: I think that allowing the `logout` action for `anonymous` users is meaningless, because this role is for unautenticated users...

lib/api/core/models/repositories/profileRepository.js
```diff
@@ -2,6 +2,7 @@ module.exports = function (kuzzle) {
   var
     _ = require('lodash'),
     q = require('q'),
+    extend = require('util')._extend,
```
scottinet: I can't find any `_extend` method in the NodeJS v4 `util` (API documentation)[https://nodejs.org/dist/latest-v4.x/docs/api/util.html].
I see it in the (v6 API documentation)[https://nodejs.org/api/util.html#util_util_extend_obj], and it is already marked as **deprecated**.
Why don't you use lodash's `extend` method?

lib/api/core/models/repositories/userRepository.js
```diff
@@ -2,6 +2,7 @@ module.exports = function (kuzzle) {
   var
     _ = require('lodash'),
     q = require('q'),
+    extend = require('util')._extend,
```
scottinet: Same remark: I think we should use lodash extend instead

.kuzzlerc
ballinette: I discussed about that with @stafyniaksacha and had at the beginning the same advice as you...
... but a user that has an expired token is automatically mapped to anonymous, and he may want to logout while his token is still expired... In that case, it'd be better to allow the action (that just do nothing) instead of reponse by an UnauthorizedError.

scottinet: That doesn't seem logical. The problem here is that if you attempt a `logout` action (or any action actually) with an expired token, you shouldn't get an `UnauthorizedError` but a new `TokenExpiredError` instead.

ballinette: well, "doing nothing" is not exactly the good words.
The use case, from the client's point of view, is: 

> If I call the `logout` action, then I will be logged out, either I was logged in or not

Given this, I agree that calling the `logout` action for an anonymous user doesn't make sense, however, it's not an issue to allow it.

But anyway, It's also OK for me to remove this permission... we just need to check with Sacha when he comes back if he had a real specific use case that needed this.

benoitvidis:
The dicussion topic is not harmful and can be debated after the merge is done.
+1

scottinet:
I agree. I was holding my approval because of my other 2 remarks, but since the issue I raised has been fixed: +1 :)

PR #315, opened 2016-06-14
Kuz 558 import plugins config

jenow:
- Moved plugin managment in remoteAction
- Added importConfig

/!\ This PR moved a lot of untested code from the cli to the remoteActionsController, only the code related to the import config for plugin is covered /!\


stafyniaksacha:
+1

lib/api/controllers/remoteActions/managePlugins.js
```diff
@@ -0,0 +1,591 @@
...
+        kuzzleConfiguration.pluginsManager.dataCollection,
+        pluginsList);
+    });
+}
+
+/**
+ * Import a configuration from a given file for a given plugin
+ * @param db
+ * @param plugin
+ * @param filePath
```
scottinet: missing `@param {Object} kuzzleConfiguration`

lib/api/controllers/remoteActions/managePlugins.js
```diff
@@ -0,0 +1,591 @@
...
+ */
+function importPluginConfiguration(db, plugin, filePath, kuzzleConfiguration) {
+  var
+    content;
+
+  try {
+    content = fs.readFileSync(filePath, 'UTF-8');
+  } catch (err) {
+    return q.reject(new Error('File not found: ' + filePath));
```
scottinet: Not necessarily `file not found`. An error can occur because its access has been denied, or it's locked, or too big, or it's a directory, or there is not enough memory on the system, etc.

lib/api/controllers/remoteActions/managePlugins.js
```diff
@@ -0,0 +1,591 @@
...
+    }
+
+    initializeInternalIndex(dbService, kuzzleConfiguration)
+      .then(() => {
+        if (options.list) {
+          return getPluginsList(dbService, kuzzleConfiguration)
+            .then(pluginsInfos => {
+              var pluginsList = [];
+              Object.keys(pluginsInfos).forEach(name => {
+                // todo return promises instead of console.log
```
ballinette: I guess the `console.log` are moved, so this comment is no longer necessary ?

lib/api/controllers/remoteActions/managePlugins.js
```diff
@@ -0,0 +1,591 @@
...
+                  pluginsList.push(clcOk(name + ' (activated)'));
+                }
+                else if (needInstall(pluginsInfos[name], name)) {
+                  pluginsList.push(clcNotice(name + ' (not installed)'));
+                }
+                else {
+                  pluginsList.push(clcNotice(name + ' (disabled)'));
+                }
+              });
+              deferred.resolve(pluginsList);
```
ballinette: Shouldn't we have rather `return q(pluginList)` ?

lib/api/controllers/remoteActions/managePlugins.js
```diff
@@ -0,0 +1,591 @@
...
+    content = fs.readFileSync(filePath, 'UTF-8');
+  } catch (err) {
+    return q.reject(new Error('File not found: ' + filePath));
+  }
+
+  try {
+    content = JSON.parse(content);
+  }
+  catch (err) {
+    return q.reject(new Error('Unable to parse ' + content + '. Expected: JSON Object'));
```
scottinet: If the file content is large enough, returning it may make life difficult for administrators trying to guess why parsing it failed.
I think returning the file name and the error message would be better.

lib/api/controllers/remoteActions/managePlugins.js
```diff
@@ -0,0 +1,591 @@
...
+
+        if (options.install) {
+          if (plugin) {
+            console.log(clcOk('███ kuzzle-plugins: Plugin ' + plugin + ' installed'));
+          }
+          else {
+            console.log(clcOk('███ kuzzle-plugins: Plugins installed'));
+          }
+        }
+        deferred.resolve(res);
```
ballinette: same remark as above: prefer `return q(res);`

lib/api/controllers/remoteActions/managePlugins.js
```diff
@@ -0,0 +1,591 @@
...
+
+        /*
+         If elasticsearch returns with a "Service temporary unavailable" error,
+         we retry until it's ready
+         */
+        if (error.status === 503) {
+          console.log('The database seems to not be ready yet. Retrying in 5s...');
+          return setTimeout(() => pluginsManager(plugin, options), 5000);
+        }
+        deferred.reject(error);
```
ballinette: in the same way, we should better have: `return q.reject(error)`

lib/api/controllers/remoteActions/managePlugins.js
```diff
@@ -0,0 +1,591 @@
...
+      });
+  });
+  return deferred.promise;
```
ballinette: If you follow the precedent recommandation, there is no need to use deferred, just have something like:

```javascript
return initializeInternalIndex(dbService, kuzzleConfiguration)
  .then(() => {
    if (options.list) {
      return getPluginsList(dbService, kuzzleConfiguration)
        .then(pluginsInfos => {
          // (...)
          return q(pluginList);
        })
    }
    // (...)
  })
  .then(res => {
    release();
    // (...)
    return q(res);
  })
  .catch(error => {
    release();
    // (...)
    return q.reject(error);
  });
```

bin/commands/plugins.js
```diff
@@ -1,612 +1,101 @@
...
     })
     .catch(err => {
-      // ignoring error if it's raised because the index already exists
-      if (err.status === 400) {
-        return q();
-      }
-
-      return q.reject(err);
+      console.error(clcError(err.message));
+      process.exit(1);
```
ballinette: why not exit with `err.status` like before ?

lib/api/controllers/remoteActions/managePlugins.js
```diff
@@ -0,0 +1,591 @@
...
+
+  childProcess.exec('npm install ' + name + (plugin.npmVersion ? ('@' + plugin.npmVersion) : ''), (err, stdout, stderr) => {
+    if (err) {
+      console.error(`Plugin download error. Full log:\n${stderr}`);
+      return deferred.reject(err);
+    }
+
+    console.log('███ kuzzle-plugins: Plugin', name, 'downloaded');
+    if (plugin.gitUrl) {
+      createLock(getPathPlugin(plugin, name));
```
ballinette: Just to be sure... I don't see where you release the lock while the plugin installation is done ?

lib/api/controllers/remoteActions/managePlugins.js
ballinette: OK, forgot my remarks, I did not notice that we were inside a callback function :/

lib/api/controllers/remoteActions/managePlugins.js
ballinette: OK, I just see it, within needInstall method ;)

bin/commands/plugins.js
jenow: I try to do the exits as much as possible in the cli itself (bin/commands/plugins.js)

lib/api/controllers/remoteActions/managePlugins.js
jenow: In the then() of initializeInternalIndex. Which make me think I don't release before the exit in acquirePlugins() !

lib/api/controllers/remoteActions/managePlugins.js
```diff
@@ -0,0 +1,591 @@
...
+ * @param {string} plugin name
+ * @param {object} db - kuzzle database handler
+ * @param {object} cfg - kuzzle configuration
+ * @returns {Promise}
+ */
+function getPluginConfiguration(plugin, db, cfg) {
+  return db
+    .get(cfg.pluginsManager.dataCollection, plugin)
+    .then(result => {
+      console.dir(result._source, {depth: null});
```
ballinette: is this `console.dir` necessary or is it an old debug stuff ?

lib/api/controllers/remoteActions/managePlugins.js
```diff
@@ -0,0 +1,591 @@
...
+
+  try {
+    jsonProperty = JSON.parse(property);
+  }
+  catch (err) {
+    return q.reject(new Error('Unable to parse ' + property + '. Expected: JSON Object\n' + err));
+  }
+
+  return db
+    .get(cfg.pluginsManager.dataCollection, plugin)
```
ballinette: do we need to fetch the document if we just call the "update" method then (which requests a partial update, needing only the new/changed attributes) ?

bin/commands/plugins.js
ballinette: hmm I don't understand :/
Actually, we are in bin/commands/plugins.js here, and you do the exit ;)
What I ask, is, why you do `process.exit(1)` instead of `process.exit(err.status)` ?

jenow: oh right!! >.<

lib/api/controllers/remoteActions/managePlugins.js
jenow: Well if update merges the actual document with the new changes, then no we don't have to do it manually with _.extend

ballinette:
+1

lib/api/controllers/remoteActions/managePlugins.js
jenow: I fixed this but github still shows your comment, no clue why

scottinet:
+1

PR #316, opened 2016-06-16
KUZ-599: plugin context refactor

scottinet:
:warning:  
This PR depends on the following ones:
*  Enabling Kuzzle Modules: #292
* Made plugin Auth Local compatible with the new context structure: https://github.com/kuzzleio/kuzzle-plugin-auth-passport-local/pull/14
* Made plugin socketio protocol compatible with the new context structure: https://github.com/kuzzleio/kuzzle-plugin-socketio/pull/8
* Made Kuzzle Proxy compatible with the new context structure: https://github.com/kuzzleio/kuzzle-proxy/pull/11

:warning: 

# Context

The current plugin context returns an object with a bunch of constructors and getters.

This poses the following problems:

* the current implementation returns an object, preventing any use of the "instanceOf" operator. This is problematic when trying to unit test this class, and a module returning a POJO is considered bad practice (already fixed by @benoitvidis in #292 )
* privileged plugins need more exposed objects, and without any kind of structure the plugin context will become messy
* even without considering privileged plugins, it can be assumed that more objects will be exposed in the context. A cleaner structure will be easier to document and plugin developpers should have an easier time understanding what's at their disposal in the context

Here is an example of the new structure:

```js
var context = new PluginContext(kuzzle);

console.dir(context, {depth: null});
/*
  {
    accessors: {
      kuzzle,    // <= privileged plugins only
      remoteActions,
      repositories,
      router    // <= renamed from "getRouter"
    },
    config: {
      // kuzzle configuration, need to be propagated to worker threads
    },
    constructors: {
      Dsl,      
      RequestObject,
      ResponseObject
      services: { // <= privileged plugins only
      }
    },
    errors: {
      BadRequestError,
      NotFoundError,
      ...
    }
  }
 */
```

# Changes

* renamed `core/plugins/pluginsContext.js` to `core/plugins/pluginContext.js` (comment from @xbill82 )
* extracted the `PrivilegedPluginContext` constructor from `core/plugins/pluginContext.js` and created a new `core/plugins/privilegedPluginContext.js` file
* implemented the above structure


ballinette:
LGTM

stafyniaksacha:
+1

PR #317, opened 2016-06-16
Enhance developer experience

dbengsch:
- Refactor some code to be more comprehenive by IDEs
- Add JSDoc comments in order to educate the IDE
- Remove JSHint and replace it by ESLint everywhere


.eslintrc
```diff
@@ -6,7 +6,6 @@
     "eqeqeq": 2,
     "guard-for-in": 2,
     "indent": [2, 2, {"SwitchCase": 1}],
-    "keyword-spacing": 2,
```
benoitvidis: As far as I can see, this option is not part of eslint:recommended.
Shouldn't we keep it?

test/api/controllers/memoryStorageController.test.js
```diff
@@ -430,20 +428,20 @@ describe('Test: memoryStore controller', function () {
...
+      request = new RequestObject({
         body: {
           weights: 'unvalid'
         }
       });
-      should(extractArgumentsFromRequestObjectForZInterstore.bind(null, requestObject)).throw(BadRequestError);
+      should(extractArgumentsFromRequestObjectForZInterstore.bind(null, request)).throw(BadRequestError);
 
     });
   });
```
benoitvidis: Just curious: Is there a reason to rename all requestObject variables to request?

test/api/controllers/readController.test.js
```diff
@@ -155,9 +155,9 @@ describe('Test: read controller', function () {
           should(realtime).be.true();
           should(stored).be.true();
           should(result.data.body.type).be.exactly('all');
-          should(result.data.body.collections).not.be.undefined().and.be.an.Object();
-          should(result.data.body.collections.stored).not.be.undefined().and.be.an.Array();
-          should(result.data.body.collections.realtime).not.be.undefined().and.be.an.Array();
+          should(result.data.body.collections).not.be.eql(undefined).and.be.an.Object();
+          should(result.data.body.collections.stored).not.be.eql(undefined).and.be.an.Array();
+          should(result.data.body.collections.realtime).not.be.eql(undefined).and.be.an.Array();
           should(result.data.body.collections.stored.sort()).match(['foo']);
```
benoitvidis: Why? [Undefined is a valid should assertion](https://shouldjs.github.io/#assertion-undefined)

test/api/controllers/routerController/executeFromRest.test.js
```diff
@@ -74,12 +74,12 @@ describe('Test: routerController.executeFromRest', function () {
     executeFromRest.call(kuzzle, callParams, {headers: {'content-type': 'application/json'}, params: { collection: 'foobar', index: '%test'}}, mockupResponse);
 
     should(mockupResponse.statusCode).be.exactly(400);
-    should(mockupResponse.header['Content-Type']).not.be.undefined();
+    should(mockupResponse.header['Content-Type']).not.be.eql(undefined);
     should(mockupResponse.header['Content-Type']).be.exactly('application/json');
-    should(mockupResponse.response.result).be.null();
+    should(mockupResponse.response.result).be.eql(null);
```
benoitvidis: same remark as for undefined.

test/api/core/notifier/notifyDocumentDelete.test.js
```diff
@@ -16,6 +16,7 @@ var mockupCacheService = {
   id: undefined,
 
   remove: function (id) {
+    console.log('called', id);
```
benoitvidis: should be removed.

benoitvidis:
I guess it is a matter of taste but I really dislike the `.eql(undefined)` notation. I really hope there is not a blocking reason to use them.

test/api/controllers/memoryStorageController.test.js
benoitvidis: Sorry, my bad, I had not seen it was previously defined in the global scope.

.eslintrc
ballinette: agree with that

features/step_definitions/role.js
```diff
@@ -116,7 +116,7 @@ var apiSteps = function () {
       return callback('Fixture for role ' + role + ' does not exist');
     }
 
-    index = Object.keys(this.roles[role].controllers)[0];
+    // todo : This test seams to have been wrongly adapted, controller variable is undefined
```
ballinette: My bad, I did not refactored properly this test after the role refactoring.
Actually, we should keep the deleted line, just replacing `index =` by `controller =`:
```
controller = Object.keys(this.roles[role].controllers)[0]; 
```

ballinette: hmm well, this test looks ugly anyway... I'm looking at it and try to fix it in another PR ;)

features/step_definitions/subscription.js
```diff
@@ -143,6 +143,8 @@ var apiSteps = function () {
   });
 
   this.Then(/^In my list there is a collection "([^"]*)" with ([\d]*) room and ([\d]*) subscriber$/, function(collection, countRooms, countSubscribers, callback) {
+    var rooms = Object.keys(this.result[this.fakeIndex][collection]);
+    var count = 0;
```
ballinette: should it be nicer to write 
```
var
  rooms = Object.keys(this.result[this.fakeIndex][collection]),
  count = 0;
```
?

lib/api/core/models/repositories/tokenRepository.js
```diff
@@ -140,7 +140,7 @@ module.exports = function (kuzzle) {
     try {
       jwt.verify(token, kuzzle.config.jsonWebToken.secret);
 
-      this.load(token)
+      Repository.prototype.load.call(this, token)
```
ballinette: why id `this.load()` not convenient here ?

lib/api/remoteActions/cleanAndPrepare.js
```diff
@@ -39,7 +46,6 @@ timeOutCB = () => {
   console.log(error('Can\'t contact Kuzzle'));
   process.exit(1);
 };
-
```
ballinette: nitpicking: shouldn't it be cleaner to keep an empty line before `module.exports...` affectation ?

lib/api/remoteActions/prepareDb.js
```diff
@@ -42,7 +42,6 @@ timeOutCB = () => {
   console.log('Can\'t contact Kuzzle');
   process.exit(1);
 };
-
```
ballinette: see comment above

test/api/controllers/remoteActionsController/prepareDb.test.js
```diff
@@ -306,12 +304,10 @@ describe('Test: Prepare database', function () {
...
       workerCalled,
       mappingCreated,
-      workerPromise,
-      prepareDb;
+      workerPromise;
 
     before(function () {
-      prepareDb = rewire('../../../../lib/api/controllers/remoteActions/prepareDb');
-      importMapping = prepareDb.__get__('importMapping');
+      importMapping = rewire('../../../../lib/api/controllers/remoteActions/prepareDb').__get__('importMapping');
```
ballinette: why not sharing the same 'prepareDb' in the whole test file ? (or at least use a global variable for the filename ?)

test/api/dsl/filters/testGlobalsFilter.test.js
```diff
@@ -1,4 +1,5 @@
 var
+
```
ballinette: unnecessary new line

.eslintrc
dbengsch: The default behavior seams to be :
- before: true
- after : true

src: http://eslint.org/docs/rules/keyword-spacing

but anyway, 2 is not a valid value for this rule

lib/api/core/models/repositories/tokenRepository.js
dbengsch: everywhere else we are using Repository.prototype notation

test/api/controllers/readController.test.js
dbengsch: not recognized by IDE ...

test/api/controllers/routerController/executeFromRest.test.js
dbengsch: same response

lib/api/controllers/adminController.js
```diff
@@ -116,7 +120,7 @@ module.exports = function AdminController (kuzzle) {
    * Reset all indexes
    *
    * @param {RequestObject} requestObject
-   * @param {Context} context
+   * @param {*} context
```
scottinet: shouldn't this param type be `{Object}`?

lib/api/controllers/remoteActions/cleanDb.js
```diff
@@ -3,6 +3,12 @@ var
   BadRequestError = require('kuzzle-common-objects').Errors.badRequestError,
   q = require('q');
 
+/**
+ * Removes all indexes from the database
+ *
+ * @param {Kuzzle} kuzzle
+ * @returns {*}
```
scottinet: should be `@returns {Promise}`

lib/api/controllers/remoteActions/enableServices.js
```diff
@@ -2,6 +2,11 @@ var
   BadRequestError = require('kuzzle-common-objects').Errors.badRequestError,
   q = require('q');
 
+/**
+ * @param {Kuzzle} kuzzle
+ * @param {Request} request
+ * @returns {*}
```
scottinet: same here: `@returns {Promise}`

dbengsch:
@benoitvidis (about the comment for eql(null) ); in fact it was a matter of IDE recognition here. I reverted all of them but it will be yellow in every test

lib/api/controllers/remoteActions/prepareDb.js
```diff
@@ -3,15 +3,27 @@ var
...
+ * @typedef {{kuzzle: Kuzzle, defaultRoleDefinition: *, files: {fixtures: *, mapping: *}, data: *}} PrepareDb
+ */
+
+/**
+ * This is not a real constructor
+ *
+ * @this {PrepareDb}
+ * @param {Kuzzle} kuzzle
+ * @param {Request} request
+ * @returns {*}
```
scottinet: `@returns {Promise}`

lib/api/controllers/securityController.js
```diff
@@ -607,14 +607,28 @@ module.exports = function SecurityController (kuzzle) {
...
 
   return this.repositories.role.validateAndSaveRole(role, context, opts);
 }
 
+/**
+ * @this Kuzzle
+ * @param {RequestObject} requestObject
+ * @param context
+ * @param opts
+ * @returns {*}
```
scottinet: `@returns {Promise}`

lib/api/core/auth/passportResponse.js
```diff
@@ -4,7 +4,11 @@ HTTP Response Mockup to emulate response objects for Passport Authentication
 (see https://nodejs.org/api/http.html#http_class_http_serverresponse)
 */
 
-module.exports = function PassportResponse (deferred) {
+/**
+ * @param deferred
```
scottinet: `@param {Promise} deferred - passport authentication promise`

lib/api/core/hotelClerk.js
```diff
@@ -289,6 +293,13 @@ function removeAllRoomsInCollection (index, collection) {
   return deferred.promise;
 }
 
+/**
+ * @this HotelClerk
+ * @param {String} index
+ * @param {String} collection
+ * @param {String} rooms
```
scottinet: the `rooms` argument is an array of strings: 

`@param {string[]} rooms`

.eslintrc
ballinette: ok

lib/api/core/models/repositories/tokenRepository.js
ballinette: nope, only when we call the parent's method under the child's method, like there:
https://github.com/kuzzleio/kuzzle/blob/develop/lib/api/core/models/repositories/userRepository.js#L76
```
UserRepository.prototype.hydrate = function (user, data) {
  // (...)
  return Repository.prototype.hydrate(user, data)
  // (...)
}
```


But here, we want to call the `load` method of the current prototype (which inherits from the parent one, I know, but could be overriden), like there:
https://github.com/kuzzleio/kuzzle/blob/develop/lib/api/core/models/repositories/profileRepository.js#L38
```
ProfileRepository.prototype.loadProfile = function (profile) {
  // (...)
  return this.load(profileId)
  // (...)
}
```

lib/api/dsl/operators.js
```diff
@@ -177,7 +177,7 @@ module.exports = operators = {
    * Return true if the document field is into the given bounding box
    *
    * @param {String} field the field that we have to check
-   * @param {String} value containing the bounding box coords
+   * @param {String|{lat: *, lon: *}|{from: *, to: *}} value containing the bounding box coords
```
scottinet: If I read the code correctly, then `value` can never be of type `string`, it was a JSDoc error.
So I guess the correct typing description should be: 
`@param {{lat: *, lon: *}|{from: *, to: *}} value containing the bounding box coords`

lib/api/dsl/operators.js
```diff
@@ -207,7 +207,7 @@ module.exports = operators = {
    * Return true only if the point in field is in a specific distance from a geo point
    *
    * @param {String} field the field that we have to check
-   * @param {String} value containing the coordinates from which to calculate the distance and the distance
+   * @param {String|{lat: *, lon: *}|{from: *, to: *}} value containing the coordinates from which to calculate the distance and the distance
```
scottinet: If I read the code correctly, then `value` can never be of type `string`, it was a JSDoc error.
So I guess the correct typing description should be: 
`@param {{lat: *, lon: *}|{from: *, to: *}} value containing the bounding box coords`

lib/api/dsl/operators.js
```diff
@@ -236,7 +236,7 @@ module.exports = operators = {
    * Return true only if the point in field is in a specific distance included in a range
    *
    * @param {String} field the field that we have to check
-   * @param {String} value containing the coordinates from which to calculate the distance and the distance
+   * @param {String|{lat: *, lon: *}|{from: *, to: *}} value containing the coordinates from which to calculate the distance and the distance
```
scottinet: If I read the code correctly, then `value` can never be of type `string`, it was a JSDoc error.
So I guess the correct typing description should be: 
`@param {{lat: *, lon: *}|{from: *, to: *}} value containing the bounding box coords`

test/.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
+  "rules": {
+    "no-invalid-this": 0,
+    "no-console": 0,
```
scottinet: I think that unit tests should not be allowed to use the console, but should instead use `should` error system, because of the way `should` handle errors: during tests it displays only a `test failed` line, and after running every tests it can, it displays all necessary informations on the console, including a stack trace.
So printing an error on the console would display it with the single `test failed` line, and it would easily go unnoticed if one use only the full error display at the end of the test sequence.

test/api/controllers/remoteActionsController/prepareDb.test.js
dbengsch: The only thing I did is to remove the set `prepareDb` once every `before`, as it is already done in `beforeEach`, it seamed to do twice the job without any need. I think the complete test should be redone with `sinon`, but as the scope of this PR is already a bit too large, I didn't do it.

test/.eslintrc
dbengsch: I removed all console calls from tests and made a test that always fails instead.

test/api/controllers/remoteActionsController/prepareDb.test.js
ballinette: OK, agree with that

stafyniaksacha:
+1

ballinette:
+1

scottinet:
+1

PR #321, opened 2016-06-17
kuz-610 - Implement regexp in dsl

benoitvidis:


jenow:
+1

lib/api/dsl/methods.js
```diff
@@ -670,8 +670,53 @@ module.exports = function Methods(filtersObject) {
...
+      flags;
+
+    if (typeof filter !== 'object') {
+      return q.reject(new BadRequestError('Regexp argument must be an object'));
+    }
+
+    for (k in filter) {
+      if (filter.hasOwnProperty(k)) {
+        field = k;
+        value = filter[k];
```
scottinet: If I understand correctly, this means that if there are multiple regular expressions defined, we silently take and process the last one.
Shouldn't it be safer to detect multiple regexp defined and throw a `BadRequestError`?

scottinet:
+1

dbengsch:
+1

PR #323, opened 2016-06-20
fix: plugins workers without context

scottinet:
:warning: Must be merged after #316 :warning: 

Workers plugins don't have a plugin context provided to their `init` function, causing 2 problems:

- a discrepancy between worker plugins and every other plugins in their `init` method prototype
- the inhability for worker plugins to use a plugin context

This PR solves the issue by providing worker plugins with a properly instantiated plugin context. 
With one exception: there is no `accessors` in the provided context, since worker plugins run in a different process and are isolated from kuzzle.


docker-compose/debug.yml
```diff
@@ -10,6 +10,7 @@ kuzzle:
   command: /run-debug.sh
   volumes:
     - "..:/var/app"
+    - "../kuzzle-enterprise-probe:/probes"
```
ballinette: something to do with workers plugin ?

ballinette:
despite my question about docker-compose file, +1

benoitvidis:
:+1:

stafyniaksacha:
+1

PR #324, opened 2016-06-20
issue #322 - fix broker client init + add a timeout for services init

### LINKED ISSUE #322: broker client does not init when an error occurs on first try.
The current implement of the broker client kills the service promise in case of error.
The promise that is retrieved during the services initialization stays pending forever.

* implement a timeout on the services initialization
* reuse the original promise for the broker client.

benoitvidis:
Resolves #322 

## Changelog

* fix ws broker client killing its init promise on first connection error
* adds a timeout on the whole services chain init.


.kuzzlerc
```diff
@@ -7,6 +7,12 @@
   */
   "serviceSettingsCollection": "services",
 
+  "default": {
```
scottinet: I'm not fond of adding sublevels here, unless we start grouping services configuration together: after reading the file several times, I think this added structure won't make much sense to newcomers.

What do you think of `serviceSettingsTimeout`, at the same level than `serviceSettingsCollection`?
Not fan of it, but it would be homogeneous.

We may have to better structure this file  sooner or later.

scottinet:
I let you decide what to do with my comment above. :)

For my part: +1

.kuzzlerc
benoitvidis: I agree, this is not clean.
I am sorry but on my side, I don't like having some parameters at the root of the rc file.

What about restructuring things a bit?:

```json
{
  "core": {
    "server": {
      "rest": {
        "port": 7511
      }
    },
    "requests": {
      "maxConcurrentRequests": 50,
      "maxRetainedRequests": 200
    }
  },
  "services": {
    "common": {
      "collection": "services",
      "initTimeout": 10000,
      "retryInterval": 1000
    },
    "service1": {
      "host": "api",
      "port": 7331
    }
  },
  "plugins": {
    "common": {
      "pipeWarnTime": 40,
      "pipeTimeout": 250,
      "dataCollection": "plugins"
    },
    "kuzzle-plugin-logger": {},
    "kuzzle-plugin-auth-passport-local": {}
  }
}
```

scottinet: Looks a lot cleaner, I like your proposal a lot. :+1:

benoitvidis:
@scottinet I agree that we should soon or later restructure the rc file and the way it is loaded by Kuzzle.
For the time being, even if I don't like it, I am going to move the timeout setting down to the root level.

Refactoring the rc is too-much work for this bug-fix.

I also agree, KUZ-606 is a good canditate.
I propose we all sit and agree on the new structure to use (the one above is just a proposal). I updated the [KUZ-606](https://jira.kaliop.net/browse/KUZ-606) ticket to link to this discussion.

stafyniaksacha:
+1

ballinette:
+1

PR #327, opened 2016-06-22
add http rest information about each routes into read:serverInfo action

stafyniaksacha:
Change format of serverInfo response from : 
```javascript
"serverInfo": {
    "kuzzle": {
        "api": {
          "routes": {
            "admin": [
              "createIndex",
              "..."
            ]
         }
      }
   }
}
```

To : 
```javascript
"serverInfo": {
    "kuzzle": {
        "api": {
          "routes": {
            "admin": {
              "createIndex": {
                "method": "put",
                "name": "createIndex",
                "route": "/api/1.0/:index"
              },
              "...": {}
            ]
         }
      }
   }
}
```


j33f:
what about the routes listing via http://localhost:7511/ or the swagger representation one ?
and what is it in the read controller and not the router one ?

stafyniaksacha:
In fact, I need a way to link http path and method from controller & actions informations;
The swagger representation isn't well formatted to be parsed from theses infos.

And about the default route, to my mind it will be better to link it directly to the serverInfo action (now we have a custom code that generate a piece of information into the routerController, and I think it's ugly)

j33f:
maybe we could replace the actual logic of the default / route by the one you've just created ?

benoitvidis:
@j33f , I agree that having 3 routes to display informations about routes is a bit too much.

+1 on @stafyniaksacha proposal to remap the / route to _serverInfo, in which case, I am also OK for this PR.

+1

ballinette:
If I understand the discussion here, the remaping of "/" route to serverInfo action should also be done in this PR (in routerController) ?

ballinette:
+1

scottinet:
+1

dbengsch:
+1

PR #329, opened 2016-06-22
Kuz-536 Internal state updates hooks

benoitvidis:
This PR adds some hooks to Kuzzle's core to allow plugins to hook on its internal state updates.

## Changelog

Kuzzle exposes some new hooks, triggered on internal state changes:

* `core:kuzzleStart`: emitted when Kuzzle is started
* `core:hotelClerk:addSubscription`: sends a diff containing the filters and internal hotelClerk updates.
* `core:hotelClerk:join`: sends hotelClerk diff when a room is joined.
* `core:hotelClerk:removeRoomForCustomer`: sends the room unsubscription information if it changed.
* `core:indexCache:add`: triggered if some data were actually added to Kuzzle's index cache.
* `core:indexCache:remove`: triggered if some data were actually removed from Kuzzle's index cache.
* `core:indexCache:reset`: triggered if the indexCache is resett.


lib/api/dsl/index.js
```diff
@@ -53,6 +53,14 @@ function Dsl () {
   /**
    * Test data against filters in the filters tree to get the matching
    * filters ID, if any
+    }
+
+    return this.filters.addSubscription(filterId, index, collection, filters);
```
scottinet: nitpicking: code comment mishap

lib/api/dsl/methods.js
```diff
@@ -435,7 +423,7 @@ function Methods(filtersObject) {
 
     formattedFilters[result.path] = result.filter;
 
-    return q(formattedFilters);
+    return q({ diff: result.diff ? [ result.diff ] : [], filer: formattedFilters });
```
scottinet: `filter` instead of `filer`

lib/config/index.js
```diff
@@ -24,5 +26,43 @@ module.exports = function (params) {
...
+        cfg[k] = false;
+      }
+      else if (/^[0-9]+$/.test(cfg[k])) {
+        cfg[k] = parseInt(cfg[k]);
+      }
+      else if (/^[0-9\.]+$/.test(cfg[k])) {
+        cfg[k] = parseFloat(cfg[k]);
+      }
+    }
+    else if (cfg[k] instanceof Object) {
```
scottinet: As arrays are instances of Object, if `cfg[k]` is an array, this test will return true.

scottinet:
About documenting the new hooks, since I'm currently refactoring the plugins documentation in the kuzzle guide page, I'll directly add these new hooks in it.

lib/config/index.js
benoitvidis: Yes, the same process can be used both for plain objects and arrays:
```
> Object.keys(['a', 'b'])
[ '0', '1' ]
```

scottinet: Didn't know that, nice!
Thanks for the explanation. :-)

scottinet:
+1

lib/api/dsl/methods.js
```diff
@@ -61,44 +61,45 @@ function Methods(filtersObject) {
...
         value = filter[field][rangeOperator],
-        encodedFunctionName = '',
+        encodedFunctionName,
         result;
 
-      if (not) {
-        encodedFunctionName += 'not';
-      }
-      encodedFunctionName += 'range' + field + rangeOperator + value;
+      encodedFunctionName = `${not ? 'not' : ''}range${field}${rangeOperator}${value}`;
```
xbill82: badass :sunglasses:

.kuzzlerc
```diff
@@ -10,6 +10,27 @@
   "servicesDefaultInitTimeout": 10000,
 
   /*
+   Cluster settings.
```
ballinette: does this has anything to do with the actual PR ?

xbill82:
:+1: !

lib/services/broker/wsBrokerClient.js
```diff
@@ -20,13 +21,19 @@ function WSBrokerClient (brokerType, options, pluginsManager, notifyOnListen) {
...
-    retryInterval: options.retryInterval
+    retryInterval: options.retryInterval || 1000
   };
   this.server = {
     host: options.host,
     port: options.port
   };
+  
   this.handlers = {};
+  this.onConnectHandlers = [];
```
ballinette: Can we have a example of how these properties are used ?

lib/api/start.js
```diff
@@ -16,7 +16,7 @@ var
   RouterController = require('./controllers/routerController'),
   RemoteActionsController = require('./controllers/remoteActionsController'),
   Dsl = require('./dsl'),
-  // Load all configuration files (database, brokers...)
+// Load all configuration files (database, brokers...)
```
dbengsch: Why ?

lib/services/broker/wsBrokerClient.js
benoitvidis: sure, [here you have](https://github.com/kuzzleio/kuzzle-plugin-cluster/blob/master/lib/cluster/slaveNode.js#L55).

.kuzzlerc
benoitvidis: Well, yes and no. The main motivation for this PR is to enable some cluster mode for Kuzzle and these settings are required for the current implementation to work.
But you're right, it can be removed from Kuzzle and injected by the cluster.

lib/api/start.js
benoitvidis: hum.. looks a typo... reverted.

lib/services/broker/wsBrokerClient.js
ballinette: thx

ballinette:
+1

PR #330, opened 2016-06-23
remove debug console output

### LINKED ISSUE #334: bin/kuzzle plugins --get returns [Object] for embeded properties
```
kuzzle # ./bin/kuzzle plugins --get kuzzle-plugin-logger                                                                                                                        /var/app (kuz-536-internal-state-updates-hooks) 8d257f306019
{ npmVersion: '2.0.2',
  activated: true,
  config: 
   { services: { file: [Object], stdout: [Object] },
     loadedBy: 'all' } }
```

We should get the full plugin configuration.

benoitvidis:
## Changelog

* options `--noColors` added for all Kuzzle commands in case the end user terminal does not support ANSI colors
* Plugin configurations are now displayed with an infinite depth (fixes #334)

### DIFF
--- a/bin/commands/createFirstAdmin.js
+++ b/bin/commands/createFirstAdmin.js
@@ -3,16 +3,16 @@
 var
   readlineSync = require('readline-sync'),
   clc = require('cli-color'),
-  error = clc.red,
-  question = clc.whiteBright,
-  notice = clc.cyanBright,
-  ok = clc.green.bold,
   request = require('request-promise'),
   rc = require('rc'),
   params = rc('kuzzle'),
   name,
   password,
   resetRoles = false,
+  error,
+  question,
+  notice,
+  ok,
   step = 0,
   // steps definitions
   steps = [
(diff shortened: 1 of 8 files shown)


scottinet:
+1

jenow:
Maybe we should console.dir instead of console.log because in case of a --get we only see [Object]

benoitvidis:
@jenow I agree but it should be handled in a new PR.
I opened the issue #334, which is under progress.

benoitvidis:
@jenow My bad, I had not realized this console output was also used by `--list` and `--get` options.

xbill82:
:+1:

bin/commands/createFirstAdmin.js
```diff
@@ -230,11 +230,22 @@ var checkIfFistAdminNeeded = () => {
 
 process.stdin.setEncoding('utf8');
 
-module.exports = {
-  check: checkIfFistAdminNeeded,
-  action: () => {
-
-    checkIfFistAdminNeeded()
+module.exports = function FirstAdmin (options, run) {
+  if (!(this instanceof FirstAdmin)) {
```
ballinette: I don't understand the usage of this statement?

lib/api/controllers/remoteActions/managePlugins.js
```diff
@@ -9,21 +9,22 @@ var
...
+  clcError = clc.red;
 
 module.exports = function pluginsManager(plugin, options) {
   var
     kuzzleConfiguration = require('../../../config')(defaultConfig),
     dbService = new DatabaseService({config: kuzzleConfiguration}),
     deferred = q.defer();
-
+  
+  if (options.parent && options.parent.noColors) {
```
ballinette: Why not writing colors options as you made in `bin/commands/*` scripts ?
```
var
  clcOk = string => options.parent.noColors ? string : clc.geen.bold(string),
  clcNotice = string => options.parent.noColors ? string : clc.cyan(string),
  clcError = string => options.parent.noColors ? string : clc.red(string);
```

benoitvidis: These functions are used outside of module.exports, where the options variable does not exist.
It was much easier, notably regarding the tests we have, to keep the default behavior and override the methods only when asked.

bin/commands/createFirstAdmin.js
ballinette: ok

lib/api/controllers/remoteActions/managePlugins.js
ballinette: ok

ballinette:
+1

PR #332, opened 2016-06-23
Kuz 569 repositories implement lazy loading

j33f:
Implement the lazy loadding of repositories objects.
- user:
    - a `getProfile` method allow to retrieve the profile
    - an `isActionAllowed` method allow to know if an action is allowed to the user (via its profile then roles) it is no longer required to load the user profile prior to check allowed actions
    - the `hydrate` method of the userRepository is now a real hydratation method + checks if the user profile exists
- profile:
    - same as above
- token:
    - a implementation problem in the tokenRepository was returning an object when the user was "anonymous" and a string otherwise: fixed it to "always a string"
    - another implementation problem was not taking care of the `null` value passed as a `readEngine` forcing tokens to be searchable in cache only have been fixed too.

The repository class does not provide a default hydratation method anymore. If a specific repository need hydratation, it have to be implemented in it only.

This PR is doing some code refactor due to the changes to the hydratation concept.

Other changes: 
- correct a misbehavior on repository instantiation and business properties affectations
- fix the way of handling redis responses which was inefficient
- rename the `user` member of 'token' business object to `userId` and fixing it into a string
- rename the `profile` member of 'user' business object to `profileId` and fixing it into a string
- rename the `roles` member of 'profile' business object to `policies` since it is an array of objects and not an  array of roles ids


lib/api/core/models/repositories/repository.js
```diff
@@ -140,10 +136,10 @@ Repository.prototype.loadMultiFromDatabase = function (ids, hydrate) {
  *
  * @returns {promise}
  */
-Repository.prototype.search = function (filter, from, size, hydrate) {
+Repository.prototype.search = function (filter, from, size) {
   var
     deferred = q.defer(),
-    hydratedObjects = [],
+    /*hydratedObjects = [],*/
```
dbengsch: Can be removed ?

lib/api/core/models/repositories/repository.js
```diff
@@ -217,13 +190,15 @@ Repository.prototype.loadFromCache = function (id, opts) {
...
+      var object;
 
       if (response === null) {
         return q(null);
       }
-      
-      return this.hydrate(object, JSON.parse(response));
+
+      object = new this.ObjectConstructor();
+      //return this.hydrate(object, JSON.parse(response));
```
dbengsch: Can be removed ?

lib/api/core/models/security/user.js
```diff
@@ -1,24 +1,38 @@
 var
   q = require('q');
 
-module.exports = function (kuzzle) {
+function User () {
```
dbengsch: For dev XP

lib/api/core/models/repositories/repository.js
j33f: :+1:

lib/api/core/models/repositories/repository.js
j33f: :+1:

lib/api/core/models/security/user.js
j33f: :+1:

lib/api/controllers/funnelController.js
```diff
@@ -119,19 +119,17 @@ function processRequest(kuzzle, controllers, requestObject, context) {
       return requestObject.checkInformation();
     })
     .then(() => {
-      if (!controllers[requestObject.controller] ||
-        !controllers[requestObject.controller][requestObject.action] ||
+      if (!controllers[requestObject.controller] || !controllers[requestObject.controller][requestObject.action] ||
```
ballinette: really little nitpicking, but as we have 3 conditions that are too long to be kept in one single line, I find it more readable to let 1 condition per line.

lib/api/core/models/repositories/repository.js
```diff
@@ -2,8 +2,8 @@ var
   _ = require('lodash'),
   q = require('q'),
   InternalError = require('kuzzle-common-objects').Errors.internalError,
-  RequestObject = require('kuzzle-common-objects').Models.requestObject,
-  async = require('async');
+  RequestObject = require('kuzzle-common-objects').Models.requestObject/*,
+  async = require('async')*/;
```
ballinette: so we can remove this line ?

lib/api/controllers/funnelController.js
j33f: :+1:

lib/api/core/models/repositories/repository.js
j33f: :+1:

lib/api/core/models/repositories/repository.js
```diff
@@ -99,31 +113,13 @@ Repository.prototype.loadMultiFromDatabase = function (ids, hydrate) {
...
-            object,
-            data;
+          var object = new this.ObjectConstructor();
 
-          object = new this.ObjectConstructor();
-          data = _.extend({}, document._source);
-          data._id = document._id;
-
-          return this.hydrate(object, data);
+          //return this.hydrate(object, data);
```
ballinette: line to remove ?

benoitvidis:
I would find it a clearer if the own properties of the business objects were renamed as xxxId.

```
Token.user => Token.userId
User.profile => User.profileId
etc.
```

j33f:
ok I implement it

lib/api/core/models/repositories/roleRepository.js
```diff
@@ -29,30 +29,39 @@ module.exports = function (kuzzle) {
       deferred = q.defer(),
       keys = [],
       buffer = {},
-      result = [];
+      result = [],
+      internalIndexAllowedIds = {},
+      restrictions = {};
 
-    roleKeys.forEach((roleKey) => {
+    roleKeys.forEach(roleKey => {
```
ballinette: Be careful, roleKeys should be an array of role Ids (strings) and not an array of role objects (sample usage here: https://github.com/kuzzleio/kuzzle/blob/53d744514c1f4e1eeebf023a68c77df06340c38f/lib/api/core/models/repositories/profileRepository.js#L134 )

as well, the notions of "internalIndexAllowed" and "restrictions" should not be handled at role repository's side, but rather at profile busniess object's side (those properties are dependant of the mapping btw. profile and role, not of the role himself).

Difficult to explain in a good english what I mean, but I can try to make a PR into your branch to be more clear ;)

stafyniaksacha:
LGTM (just need to rename properties like @benoitvidis said)

lib/api/core/models/repositories/roleRepository.js
ballinette: take a look at my proposal: https://github.com/kuzzleio/kuzzle/commit/3d22511d4632eb235ed1cac5936dcceba8821db4

If it's OK for you, you can merge directly my branch to yours ;)

ballinette:
apart from my comment and my suggestion above, really geat job!!!! :100:

lib/api/core/auth/formatProcessing.js
```diff
@@ -31,25 +20,11 @@ module.exports = {
...
-        response._source.roles = profile.roles.map((role) => {
-          return this.formatRoleForSerialization(role);
-        });
-      }
-      else if (key !== '_id') {
-        response._source[key] = profile[key];
-      }
-    });
-
+    response = _.assignIn({}, {_id: profile._id, _source: profile});
```
scottinet: What's the difference with `response = {_id: profile.id, _source: profile};`?

lib/api/core/auth/formatProcessing.js
```diff
@@ -61,34 +36,11 @@ module.exports = {
...
-          Object.keys(triggeredUser).forEach((key) => {
-            if (key === 'profile') {
-              response._source.profile = this.formatProfileForSerialization(triggeredUser.profile, hydrate);
-            }
-            else if (key !== '_id') {
-              response._source[key] = triggeredUser[key];
-            }
-          });
-        }
+        var response = _.assignIn({}, {_id: triggeredUser._id, _source: triggeredUser});
```
scottinet: Same than above: wouldn't it be simpler to write `var response =  {_id: triggeredUser._id, _source: triggeredUser};`?

lib/api/core/hotelClerk.js
```diff
@@ -180,10 +180,12 @@ function HotelClerk (kuzzle) {
     async.each(Object.keys(this.rooms), (roomId, callback) => {
       var room = this.rooms[roomId];
 
-      requestObjectCollection.index = room.index;
-      requestObjectCollection.collection = room.collection;
+      kuzzle.repositories.user.load(context.token.user)
+        .then(user => {
+          var requestObject = _.assign(requestObjectCollection, {index: room.index, collection: room.collection});
```
scottinet: Nitpicking: since `_.assign` mutates the object, it's simpler to get rid of the `requestObject` variable and simply use the updated `requestObjectCollection` object in subsequent calls:
```js
_.assign(requestObjectCollection, {index: room.index, collection: room.collection});
return user.isActionAllowed(requestObjectCollection, context, kuzzle);
```

lib/api/core/models/repositories/profileRepository.js
```diff
@@ -129,14 +128,13 @@ module.exports = function (kuzzle) {
       profile.roles = [ {_id: 'default'} ];
     }
 
-    Repository.prototype.hydrate.call(this, profile, data)
+    q(_.assignIn(profile, data))
```
scottinet: This line prevents changing the old code, but in the future I think it'll seem weird to anyone reading this.

I think it would be better to :
* get rid of the `result` variable entirely
* remove the `o` variable and use `profile` instead
* and to rewrite line 131 and the next 4-5 lines like that:

```js
_.assignIn(profile, data);
kuzzle.repositories.role.loadRoles(extractRoleIds(profile.roles))
  .then(...
```

And if you do follow these suggestions, it would be a good idea to remove the use of `q.defer()` and simply return the call to `loadRoles` to the caller.

What do you think?

lib/api/core/models/repositories/profileRepository.js
```diff
@@ -205,28 +188,7 @@ module.exports = function (kuzzle) {
...
-      }
-      if (!_.isEmpty(role.restrictedTo)) {
-        serializedRole.restrictedTo = role.restrictedTo;
-      }
-      if (role.allowInternalIndex) {
-        serializedRole.allowInternalIndex = role.allowInternalIndex;
-      }
-      result.roles.push(serializedRole);
-    });
+    var result = _.assignIn({}, profile);
```
scottinet: nitpicking: following our discussion: could you add a small comment explaining that this line is used to avoid the mutation of the `profile` object?

lib/api/core/models/repositories/tokenRepository.js
```diff
@@ -96,12 +96,12 @@ module.exports = function (kuzzle) {
     try {
       expiresIn = parseTimespan(options.expiresIn);
       encodedToken = jwt.sign({_id: user._id}, kuzzle.config.jsonWebToken.secret, options);
-      this.hydrate(token, {
+      q(_.assignIn(token, {
```
scottinet: This too can be simplified by simply modifying `token` with `_.assignIn()` and then start the promise chain with `persistToCache`:

```js
_.assignIn(token, {
        _id: encodedToken,
        user: user._id,
        ttl: expiresIn,
        expiresAt: Date.now() + expiresIn
});

return this.persistToCache(token)
  .then(() => {
    kuzzle.tokenManager.add(token, context);
    deferred.resolve(token);
  })
  .catch(err => { ...
```

lib/api/core/models/repositories/tokenRepository.js
```diff
@@ -128,15 +128,13 @@ module.exports = function (kuzzle) {
 
   TokenRepository.prototype.verifyToken = function (token) {
     var
-      deferred,
+      deferred = q.defer(),
```
scottinet: This change should be reverted: if `token` is null, then a deferred promise has been created for nothing. So creating _after_ the test is a good thing.

lib/api/core/models/repositories/tokenRepository.js
```diff
@@ -184,59 +183,28 @@ module.exports = function (kuzzle) {
...
-    var
-      deferred = q.defer(),
-      result;
-
     if (!_.isObject(data)) {
       return q(userToken);
     }
 
-    Repository.prototype.hydrate.call(this, userToken, data)
+    return q(_.assignIn(userToken, data))
```
scottinet: Same here: there is no need of creating a promise on an `_.assignIn()` call

lib/api/core/auth/formatProcessing.js
j33f: none, you're right :)

lib/api/core/models/repositories/userRepository.js
```diff
@@ -73,22 +79,22 @@ module.exports = function (kuzzle) {
       Object.assign(data, source);
     }
 
-    return Repository.prototype.hydrate(user, data)
+    return q(_.assignIn(user, data))
```
scottinet: Same here: you can simplify the code a lot by removing the `u` and `result` variables, calling `_.assignIn()` directly without creating a promise, and use the `user` variable thenafter.

lib/api/core/models/security/profile.js
```diff
@@ -10,6 +10,23 @@ function Profile () {
...
+ *
+ * @return {Promise}
+ */
+
+Profile.prototype.getRoles = function (kuzzle) {
+  if (!this._roles) {
+    return kuzzle.repositories.role.loadRoles(this.roles)
+      .then(roles => {
+        this._roles = roles;
+        return q(this._roles);
```
scottinet: You don't have to return a promise to make the whole chain resolve `this._roles`. Simply returning the variable will end the chain with this variable as its result.

lib/api/core/models/security/profile.js
```diff
@@ -10,6 +10,23 @@ function Profile () {
...
 
+/**
+ * @param {Kuzzle} kuzzle
+ *
+ * @return {Promise}
+ */
+
+Profile.prototype.getRoles = function (kuzzle) {
+  if (!this._roles) {
+    return kuzzle.repositories.role.loadRoles(this.roles)
```
scottinet: What's the difference between `this.roles` and `this._roles`?

lib/api/core/models/security/profile.js
```diff
@@ -21,23 +38,33 @@ function Profile () {
 Profile.prototype.isActionAllowed = function (requestObject, context, kuzzle) {
   var
     deferred = q.defer();
```
scottinet: Should be instantiated after the test returning `q(false)`. Or better yet: return the `this.getRoles` promise chain directly :-)

lib/api/core/models/security/profile.js
```diff
@@ -21,23 +38,33 @@ function Profile () {
...
   }
 
-  async.some(this.roles, (role, callback) => {
-    role.isActionAllowed(requestObject, context, kuzzle)
-      .then(isAllowed => {
-        callback(isAllowed);
+  this.getRoles(kuzzle)
+    .then(roles => {
+      async.some(roles, (role, callback) => {
+        role.isActionAllowed(requestObject, context, kuzzle)
```
scottinet: Should catch an error if, for some reason, `isActionAllowed` fails.

Easy way to do that with Q:

```js
role.isActionAllowed(requestObject, context, kuzzle).nodeify(callback);
```

lib/api/core/models/security/profile.js
```diff
@@ -21,23 +38,33 @@ function Profile () {
...
+          .then(isAllowed => {
+            callback(isAllowed);
+          });
+      }, result => {
```
scottinet: Since `async` takes NodeJS-style callbacks, here you are processing an errored callback, and I believe this is not exactly what you wanted to code. :-)

Easy way to pass the result of a NodeJS callback to deferred promises:

```js
}, deferred.makeNodeResolver());
```

lib/api/core/models/security/profile.js
j33f: nope, just a naming problem.
getRoles use the this.roles var (which stores the roles list as an object like ```{_id:'roleId', restrictedTo:{}...}```) retrieves them via the loadRoles method, then keep them in their full business representation into this._roles.

I am currently renaming the this.roles into this.policies

lib/api/core/models/security/profile.js
```diff
@@ -55,46 +82,48 @@ Profile.prototype.validateDefinition = function () {
...
-              rightsKey = md5(JSON.stringify(rightsItem));
-
-            rightsItem.value = actionRights;
-            rightsObject[rightsKey] = rightsItem;
-            profileRights = _.assignWith(profileRights, rightsObject, mergeRights);
+  return this.getRoles(kuzzle)
+    .then(roles => {
+      roles.forEach(role => {
+        Object.keys(role.controllers).forEach(controller => {
+          var restrictedTo = _.cloneDeep(role.restrictedTo);
```
scottinet: I gave a long hard look to this line, and after some time I came to the conclusion that we don't need to perform this `_.cloneDeep` more than once per role, instead of once per role controller like it's done here.

Should be in the first `roles.forEach()` loop instead.

lib/api/core/models/security/profile.js
```diff
@@ -55,46 +82,48 @@ Profile.prototype.validateDefinition = function () {
...
+
+          Object.keys(role.controllers[controller].actions).forEach(action => {
+            var actionRights = role.controllers[controller].actions[action];
+
+            if (restrictedTo === undefined || restrictedTo.length === 0) {
```
scottinet: Just like above, this operation can be done only once, inside the `roles.forEach()` loop.

lib/api/core/models/security/profile.js
```diff
@@ -55,46 +82,48 @@ Profile.prototype.validateDefinition = function () {
...
+            });
           });
         });
       });
-    });
-  });
-
-  return q(profileRights);
+    })
+    .then(() => q(profileRights));
```
scottinet: Can be simplified with `.then(() => profileRights);`

lib/api/core/models/security/profile.js
```diff
@@ -55,46 +82,48 @@ Profile.prototype.validateDefinition = function () {
...
+                    controller: controller,
+                    action: action,
+                    index: restriction.index,
+                    collection: collection
+                  },
+                  rightsKey = md5(JSON.stringify(rightsItem));
+
+                rightsItem.value = actionRights;
+                rightsObject[rightsKey] = rightsItem;
+                profileRights = _.assignWith(profileRights, rightsObject, mergeRights);
```
scottinet: No need to assign the result to the same variable, as `_.assignWith` mutates the target object:
`_.assignWith(profileRights, rightsObject, mergeRights);`

lib/api/core/models/security/user.js
```diff
@@ -1,24 +1,41 @@
...
+ * @param {Kuzzle} kuzzle
+ *
+ * @return {Promise}
+ */
+User.prototype.getProfile = function (kuzzle) {
+  if (this._profile === undefined) {
+    return kuzzle.repositories.profile.loadProfile(this.profile)
+      .then(profile => {
+        this._profile = profile;
+        return q(profile);
```
scottinet: You can simply return profile: `return profile;`

lib/api/core/models/security/profile.js
j33f: just cannot make it work right now... dunno why...

ballinette: We still user async v1.5.2. 
And callback arguments for async.some have changed in 2.0.0, with the node-js style: 
https://www.npmjs.com/package/async#some 

> Callback arguments changed in 2.0

In our actual version, it's not possible to use nodeify. We should probably upgrade to v2.0.0-rc6 (but we have to check if we have no regression)... or just add the error catching here.

lib/api/core/models/repositories/roleRepository.js
j33f: @ballinette if you could adapt your PR to the actual code it would be deeply appreciated

ballinette:
+ :100: :100: :100: :100: :100: :100: :100: :100:

lib/api/core/models/security/profile.js
j33f: nope, async.some does not support this

stafyniaksacha: +1 @j33f 

[like said in the documentation](https://github.com/caolan/async/blob/v1.5.2/README.md#somearr-iterator-callback): 
**Note: the callbacks do not take an error as their first argument.**

stafyniaksacha:
+1

stafyniaksacha:
+1

lib/api/controllers/authController.js
```diff
@@ -75,7 +75,7 @@ function AuthController (kuzzle) {
    * @returns {Promise}
    */
   this.getCurrentUser = function (requestObject, context) {
-    requestObject.data._id = context.token.user;
+    requestObject.data.body._id = context.token.userId;
```
scottinet: Seems weird that we need to put the user id in the body part of the request object, when this field is directly in the data part for every other API requests. Can you take a look at it?

lib/api/controllers/remoteActions/prepareDb.js
```diff
@@ -357,7 +357,7 @@ function createProfileCollection(kuzzle) {
     collection: 'profiles',
     body: {
       properties: {
-        roles: {
+        properties: {
```
scottinet: Didn't you mean `policies`?

lib/api/controllers/securityController.js
```diff
@@ -294,19 +290,23 @@ function SecurityController (kuzzle) {
     return kuzzle.pluginsManager.trigger('security:beforeGetUser', requestObject)
       .then(newRequestObject => {
         modifiedRequestObject = newRequestObject;
-
-        if (!modifiedRequestObject.data._id) {
+        if (!modifiedRequestObject.data._id && !modifiedRequestObject.data.body._id) {
```
scottinet: Isn't this test a consequence of putting the `_id` field in the body part of the request object in `authController.getCurrentUser()`?
If so, can't we just test if `modifiedRequestObject.data._id` is falsey and keep this field exclusively in the data part of a request object?

lib/api/controllers/securityController.js
```diff
@@ -294,19 +290,23 @@ function SecurityController (kuzzle) {
...
           return q.reject(new BadRequestError('No user id given'));
         }
+        // TODO find where the _id is passed directly into data and not data.body
```
scottinet: Found it! :-)

lib/api/controllers/securityController.js
```diff
@@ -294,19 +290,23 @@ function SecurityController (kuzzle) {
...
+        // quick fix
+        if (modifiedRequestObject.data._id && !modifiedRequestObject.data.body._id) {
```
scottinet: Same as above: we should only need to test `modifiedRequestObject.data._id`

lib/api/controllers/securityController.js
```diff
@@ -294,19 +290,23 @@ function SecurityController (kuzzle) {
...
+          modifiedRequestObject.data.body._id = modifiedRequestObject.data._id;
```
scottinet: Why in the request object's body?

lib/api/core/models/repositories/profileRepository.js
```diff
@@ -124,30 +119,25 @@ module.exports = function (kuzzle) {
...
+          return q.reject(new NotFoundError(`Unable to hydrate the profile ${data._id}. The following roles don't exist: ${rolesNotFound}`));
         }
 
-        deferred.resolve(o);
+        return profile;
       })
-      .catch(error => deferred.reject(error));
-
-    return deferred.promise;
+      .catch(error => q.reject(error));
```
scottinet: nitpicking: this line is superfluous, since the whole promise chain is returned to the caller anyway, including rejections.

lib/api/core/models/repositories/repository.js
```diff
@@ -217,40 +227,34 @@ Repository.prototype.loadFromCache = function (id, opts) {
...
           .then(objectFromDatabase => {
             if (objectFromDatabase !== null) {
               this.persistToCache(objectFromDatabase);
             }
-            deferred.resolve(objectFromDatabase);
+            return objectFromDatabase;
           })
           .catch(err => {
-            deferred.reject(err);
+            return q.reject(err);
```
scottinet: nitpicking: this `catch` can be rejected, as it's strictly equivalent to letting the whole chain being rejected.

lib/api/core/models/repositories/repository.js
```diff
@@ -217,40 +227,34 @@ Repository.prototype.loadFromCache = function (id, opts) {
...
-      else {
-        this.refreshCacheTTL(object);
-        deferred.resolve(object);
-      }
+      this.refreshCacheTTL(object);
+      return object;
     })
     .catch(err => {
-      deferred.reject(err);
+      return q.reject(err);
```
scottinet: nitpicking: same here

lib/api/core/models/repositories/repository.js
```diff
@@ -334,24 +338,28 @@ Repository.prototype.persistToCache = function (object, opts) {
...
     ttl = options.ttl;
   }
   if (ttl === false) {
     return this.cacheEngine.set(key, JSON.stringify(this.serializeToCache(object)))
-      .then(doThen);
+      .then(() => {
+        return object;
+      })
+      .catch(error => {
+        return error;
```
scottinet: Is it intended to resolve the promise with an error as its resolved value, instead of rejecting it?

lib/api/core/models/repositories/repository.js
```diff
@@ -334,24 +338,28 @@ Repository.prototype.persistToCache = function (object, opts) {
...
+      });
   }
 
   return this.cacheEngine.volatileSet(key, JSON.stringify(this.serializeToCache(object)), ttl)
-    .then(doThen);
+    .then(() => {
+      return object;
+    })
+    .catch(error => {
+      return error;
```
scottinet: ditto

scottinet:
+1

j33f:
:tada:

PR #333, opened 2016-06-24
KUZ-613: repositories in plugincontext

scottinet:
Instead of the full `repositories` object been exposed in the plugin context, we now expose only a few methods that can be used by plugin developers.
This makes plugin development easier to grasp and to document.

# Changelog 
* replaced `PluginContext.accessors.repositories` with `PluginContext.accessors.users`
* instead of exposing the entire repository system, the new `users` accessor now only exposes the following methods:
  * `load(userId)`: returns a promise resolving to a user object
  * `create({user: 'informations'})`: returns a promise resolving to the updated user object
* added unit tests to the plugin context object
* moved `authController.passport` into `kuzzle.passport`
* plugin context now exposes the `passport.use` method in its accessors
* removed the special `auth:loadStrategies` hook (this hook does not behave like other hooks)
* replaced event `auth:loadScope`: auth plugins now require a `this.scope` variable to manage scope instead

### DIFF
--- a/.kuzzlerc
+++ b/.kuzzlerc
@@ -179,7 +179,7 @@
         "activated": true
       },
       "kuzzle-plugin-auth-passport-local": {
-        "npmVersion": "2.0.1",
+        "npmVersion": "2.0.2",
         "activated": true
       }
     }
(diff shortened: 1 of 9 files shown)


ballinette:
+1

dbengsch:
+1

PR #336, opened 2016-06-27
fix #335: init indexCache synchronously

### LINKED ISSUE #335: synchronicity issue at kuzzle start-up
When we start kuzzle, sometimes, it tries to create the internalIndex at "PREPAREDB" step, while it already exists.
Consequently, elastic launches an `IndexAlreadyExistsException` and kuzzle start-up crashes, and tries to restart again and again...

Most often, the start-up script works as expected, yet.
It's probably due to a bad synchronicity at kuzzle start-up, btw. `IndexCache init` (that list the already created indexes in elasticsearch) and `prepareDB` (that creates the internalIndex only if it still not exists).

ballinette:


ballinette:
NB: I wanted to add a unit test for this bug, but it would really difficult to have a pertinent unit test with actual code (indeed, unit tests start kuzzle as dummy, but the internalCache is loaded only while not dummy mode), and I would need to rewrite deeply sthe start.js just to be able to run unit tests (not a good idea for such a little fix, I guess).

benoitvidis:
:+1:

scottinet:
:+1:

dbengsch:
+1

PR #340, opened 2016-07-01
fix #339: Unsubscription from a subfilter clean all related filters

### LINKED ISSUE #339: Unsubscription from a subfilter clean all related filters
Let's say you subscribe to an entire index/collection pair and, on another connection, you subscribe to that index/collection with some DSL filter.

Then you unsubscribe that subfilter, and the subscription on the global index/collection pair does not receive any notification from that point on.

Wild guess: I think that when cleaning field filters, we don't check if there are global filters remaining on a collection before wiping it clean.

scottinet:
Fixes #339 

# Bug

When the last field filter on a data collection leaves, then the whole DSL filters tree is cleaned up even if there is still subscriptions on global filters.

This causes subscriptions on a data collection to stop working on certain conditions.

# How to reproduce

* on a connection, subscribe to a collection using global filters (whole collection or not exists filter)
* on another connection, subscribe to some field filters (e.g. a "term" filter)
* disconnect the second connection
* the first one does not receive any notification anymore

# Fixes

* the DSL `removeFilterPath` method now checks that there is no global filters left on a collection before removing it from the filters tree
* added a unit test to ensure the bug has been fixed

# Other changes

* simplified a bit how the DSL `index.js` file calls the `filters.js` one to remove filters
* in `removeGlobalFilters`, check if a data collection or a data index can be removed from the filters tree _if and only if_ a global filter has been removed from it

### DIFF
--- a/lib/api/dsl/filters.js
+++ b/lib/api/dsl/filters.js
@@ -642,10 +642,7 @@ function removeFilterPath(filterId, filterPath) {
     return false;
   }
 
-  /*
-   If there is no another filter using this filter part, we can remove it,
-   unless there is still global filters in it
-   */
+  // If there is no another filter using this filter part, we can remove it
   if (parent.fields) {
     delete parent.fields[subPath];
   }
--- a/lib/api/dsl/filters.js
+++ b/lib/api/dsl/filters.js
@@ -349,61 +349,15 @@ function Filters () {
   };
 
   /**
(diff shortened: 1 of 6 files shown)


benoitvidis:
:+1:

xbill82:
:+1:

j33f:
:+1:

PR #341, opened 2016-07-04
Kuz 577 lb master election

benoitvidis:
Allows Kuzzle broker client to not try to reconnect on `close` or `error` events.

Typical use case is when the server is down and a new one was popped up.


scottinet:
+1

dbengsch:
+1

stafyniaksacha:
+1

PR #344, opened 2016-07-12
KUZ-490 Multi Profile Implementation

j33f:
Adds the ability to use more than one profile per user
Changes : 
     

- the ```profileId``` property of the ```user``` business object is now named ```profilesIds``` and is an array of string represening the profiles IDs
- the ```_id``` property of the ```profile```'s ```policies``` entries is now named ```roleId``` to be more clear

### DIFF
--- a/lib/api/core/models/repositories/profileRepository.js
+++ b/lib/api/core/models/repositories/profileRepository.js
@@ -8,13 +8,12 @@ module.exports = function (kuzzle) {
     NotFoundError = require('kuzzle-common-objects').Errors.notFoundError,
     Repository = require('./repository');
 
-  var extractRoleIds = function (policies) {
-    return policies.map(policy => {
-      if (typeof policy === 'string') {
-        return policy;
+  var extractRoleIds = policiesOrRoles => {
+    return policiesOrRoles.map(element => {
+      if (element.roleId) {
+        return element.roleId;
       }
-
-      return policy._id;
+      return element._id;
     });
   };
 
@@ -58,12 +57,12 @@ module.exports = function (kuzzle) {
   /**
    * Loads a Profile object given its id.
    *
-   * @param {array} profilesId
+   * @param {array} Array of profiles ids
    * @returns {Promise} Resolves to the matching Profile object if found, null if not.
    */
   ProfileRepository.prototype.loadProfiles = function (profilesIds) {
     var promises = [];
-    
+
     if (!profilesIds) {
       return q.reject(new BadRequestError('Missing profilesIds'));
     }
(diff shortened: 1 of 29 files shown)


lib/api/core/models/security/user.js
```diff
@@ -1,31 +1,70 @@
...
+      var promises = profiles.map(profile => {
+        return profile.getRights(kuzzle);
+      });
+      
+      return q.all(promises)
+        .then(results => {
+          var rights = {};
+
+          results.forEach(right => {
+            _.assign(rights, right);
```
ballinette: be careful, if a first profile returns a right item with value "allowed" for a controller/action/index/collection tuple, and a second profile returns a right item with value "conditional", here, you will get the value "conditional", while we want the value "allowed".

For each right item, we want to be sure to have following values:
* "allowed" if at least a profile right has "allowed" value
* else,  "conditional" if at least a profile right has "conditional" value
* else, "denied"

For that, this should be replace by something similar to `_.assignWith(rights, right, mergeRights)`, like in Profile prototype (see https://github.com/kuzzleio/kuzzle/blob/develop/lib/api/core/models/security/profile.js#L123 ), using the `mergeRights` method defined there: https://github.com/kuzzleio/kuzzle/blob/develop/lib/api/core/models/security/profile.js#L142 (that should be moved away from Profile prototype to a shared piece of code, I guess).

ballinette:
apart to my remark, it looks a priori good ;)

jenow:
+1

lib/api/core/models/repositories/profileRepository.js
```diff
@@ -8,13 +8,12 @@ module.exports = function (kuzzle) {
     NotFoundError = require('kuzzle-common-objects').Errors.notFoundError,
     Repository = require('./repository');
 
-  var extractRoleIds = function (policies) {
-    return policies.map(policy => {
-      if (typeof policy === 'string') {
-        return policy;
+  var extractRoleIds = policiesOrRoles => {
```
stafyniaksacha: Why did we have this kind of name ? Normally we only have to extract roles ids from policies

j33f: just a relic

j33f: nope... it is just because this function is used both to extract roles Ids from policies or from roles objects :)

stafyniaksacha:
+1

lib/api/core/models/security/user.js
```diff
@@ -1,31 +1,70 @@
...
+  
+  if (this.profilesIds === undefined || this.profilesIds.length === 0) {
+    return q(false);
+  }
+
+  deferred = q.defer();
+
+  this.getProfiles(kuzzle)
+    .then(profiles => {
+      async.some(profiles, (profile, callback) => {
```
benoitvidis: Why using async? Wouldn't Array.prototype.some be good enough?

scottinet:
+1

lib/api/core/models/security/user.js
scottinet: Oops didn't see your comment. I'll remove async when merging to #346 's branch

PR #345, opened 2016-07-13
Return the right total on searchUser

AnthonySendra:
Fix:
On http://kuzzle:7511/api/1.0/users/_search with from/size, the total returned is the total in the _hits_ array instead of the real total


jenow:
+1

j33f:
:+1:

stafyniaksacha:
+1

PR #350, opened 2016-07-21
fix #349

### LINKED ISSUE #349: user.getRights do not merge policies correctly for multi-profile users
Example:
A user with 2 profiles with following policies:
* profile1: 
```
[
  {
    controller: 'read', action: 'get', index: 'foo', collection: 'bar',
    value: 'allowed'
  },
  { ... }
]
```

* profile2: 
```
[
  {
    controller: 'read', action: 'get', index: 'foo', collection: 'bar',
    value: 'conditional'
  },
  { ... }
]
```

The userRights should merge this with the value ```allowed``` for policy ```controller: 'read', action: 'get', index: 'foo', collection: 'bar'```.
But with the actual code, the returned value is ```conditional``` (for the same policy, value for last profile override the existing value).

ballinette:
Note for reviewers: I'm not sure about where to put the ```mergeRights()``` method, so I'm open to your advice about it ;)

### DIFF
--- a/lib/api/core/models/security/mergeRights.js
+++ b/lib/api/core/models/security/mergeRights.js
@@ -1,0 +1,54 @@
+var
+  _ = require('lodash');
+
+/**
+ * Merge function for rights
+ * @param {{value: String}} prev existing rights object
+ * @param {{value: String}} cur new rights object to merge
+ *
+ * @return {{value: String}} the merged rights object
+ */
+module.exports = function mergeRights (prev, cur) {
+  if (cur.value === 'allowed') {
+    return cur;
+  }
+  if (cur.value === true) {
+    cur.value = 'allowed';
+    return cur;
+  }
+
+  if (prev !== undefined) {
+    if (prev.value === 'allowed') {
+      cur.value = 'allowed';
+      return cur;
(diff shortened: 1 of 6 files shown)


benoitvidis:
:+1: About where to put the mergeRights method: Functionaly, this methods applies to a User instance properties. 
I would put it as a method of the User (but it is not blocking to me).

ballinette:
@benoitvidis actually, it can apply to both a User instance and a Profile instance properties, that's why I have no clear idea on where to put it (before that, it applied only to Profiles)

scottinet:
Well, firstly I think it's really `mergePolicies` instead of `mergeRights`. And that leads me to think that this function could be put as a static `merge` function inside a `policies.js` file.
Thus we would have: `Policies.merge(rights1, rights2)`

I'm playing with that idea and there is a thing or two we could do with it.

That said, like Benoit, I think this is a minor question, and this PR gets my :+1:

ballinette:
Thx for your suggestion, Scott ;) 
Indeed, it's more graceful inside a `policies.js` file

dbengsch:
+1

PR #353, opened 2016-07-25
Fix roles format : missing _source entries in results

stafyniaksacha:
* Added `formatProcessing.formatRoleForSerialization` call on roles that are fetched though `security.mGetRoles` & `security.searchRoles`
* Removed role's `allowInternalIndex` & `restrictedTo` data from responses 
* Fixed a bug in `ProfileRepository`: internal cache wasn't updated when a profile was updated
* Renamed `kuzzle.services.list.userCache` to `kuzzle.services.list.securityCache` because it was also used by profiles & roles repositories 
* Removed references to old hydrate options


j33f:
👍

benoitvidis:
+1

ballinette:
+1

PR #354, opened 2016-07-26
kuz-600 kuz-623 - create first admin & remote actions refactor

benoitvidis:
I apologize, this PR was initially meant to cover the migration of the create first admin command to use some remote actions.
It ended up implementing some of the refactors from kuz-623 and some other changes that may need to be discussed :confused:.

# Changelog

## createFirstAdmin

* Moved to a remote action
* Subsequently, the `pid` parameter now becomes mandatory.
* Removed all constraints regarding login and password choice. The only constraint left is that a login cannot be empty (password can be).

## remoteActions "clients" (/lib/api/remoteActions/*)

* Removed the need to crete one file per action to call.

## remoteActionsController

* Fixed a  very nasty situation where some actions where using `this` while not having been invoked using `new`. As a result, the context was populated with some internal properties of the actions.
* All actions now take as a RequestObject object for input, like any other controller actions.
* Removed all `console.log` calls. I believe the message displayed to the end user should be done by the cli component.

### cleanDb

* Now deletes Kuzzle internal index only. The business data if any are left intact.


bin/commands/install.js
```diff
@@ -1,12 +1,29 @@
...
-      process.exit(code);
+  var 
+    kuzzle = new Kuzzle(),
+    requestObject = new RequestObject({
+      body: {install: true}
+    });
+
+  console.log('███ kuzzle-plugins: Starting plugins installation...');
+  
+  kuzzle.start(params, {dummy: true})
```
ballinette: why starting kuzzle here ?
isn't it sufficient to just load the internalEngine ?

ballinette: (or call `kuzzle.remoteActions.do('managePlugins')` instead of `kuzzle.remoteActionsController.actions.managePlugins` ?)

benoitvidis: Yep, you're absolutely right.

benoitvidis: Actually, went a bit too fast. I reverted my changes.

The thing is that this command is run *before* kuzzle actually starts. It is triggered from the run*.sh scripts: 
```
node bin/kuzzle install && pm2 start /config/processes-dev.json
```

So, we need the controllers, the pluginsManager and the InternalEngine but **not** the internalbroker client.
The dummy start just does what we need.

Because we don't have a running Kuzzle to talk to, we cannot actually use the remoteActions.do method either.

ballinette: I agree that is is run before kuzzle starts, but I do not understand why it needs kuzzle to be started yet.

If I look at the "managePlugins" remote action, you just need the internalEngine and the pluginManager, so it would be better to just init these services explicitly.

Especially as I'm currently refactoring the startup scripts, and removing this ugly dummy mode...

benoitvidis: The `remoteActions.do` method's job is to forward the request to a running Kuzzle instance using the InternalBroker transport layer. We cannot use it during the install as we don't have any target Kuzzle running yet.

Regarding the Kuzzle start, OK, I will manually init the required components and remove the start.

bin/commands/disable.js
```diff
@@ -1,12 +1,31 @@
+/* eslint-disable no-console */
```
scottinet: This may be the right time to implement KUZ-632 (removal of service enabling/disabling features).
What do you think?

lib/services/internalEngine.js
```diff
@@ -249,12 +255,35 @@ function InternalEngine (kuzzle) {
...
+   */
+  this.listIndexes = function () {
+    return this.client.indices.getMapping()
+      .then(response => Object.keys(response));
+  };
+
+  /**
    * Create the internal index
    *
    * @returns promise}
```
scottinet: Nitpicking (old one): `@returns {Promise}`

package.json
```diff
@@ -43,7 +43,7 @@
     "json-stable-stringify": "1.0.1",
     "json2yaml": "^1.1.0",
     "jsonwebtoken": "^7.0.1",
-    "kuzzle-common-objects": "^1.0.0",
+    "kuzzle-common-objects": "^1.0.1",
```
scottinet: It's 1.0.2 now :-)

scottinet:
:+1:

dbengsch:
+1

ballinette:
+1

PR #355, opened 2016-07-27
Multichannel notifications

scottinet:
Instead of sending 1 message per notification and per channel, now sends 1 message per notification.
The message format also changed slightly: instead of a `channel` parameter, it now contains a `channels` one, pointing to an array of channels to send the notification to.

This change improves performances greatly.

Other changes:
* replaced `KuzzleProxy.notify` and `KuzzleProxy.broadcast` with `KuzzleProxy.dispatch`, to avoid repetition of code
* merged `notifier.notify` and `notifier.send`
* rewrote `notifier.notify` unit tests entirely
* Made "overload" messages and events less frequent by adding a new "warnRetainedRequestsLimit" parameter in .kuzzlerc


docker-compose.yml
```diff
@@ -1,6 +1,9 @@
 proxy:
-  image: kuzzleio/proxy:alpine
+############# UPDATE TO kuzzleio/proxy:alpine
+############# WHEN RELEASING RC5
+  image: kuzzleio/proxy:develop
   ports:
+    - "5713:5713"
```
ballinette: what is listening this port ?

dbengsch: the websocket plugin

ballinette:
+1

docker-compose.yml
dbengsch: It might be better to align this port to map 7513 instead of 5713; it can be confusing and error pron with the other non-standard ports (7511 & 7512)

scottinet: Done.

dbengsch:
Despite my comment +1

stafyniaksacha:
+1

PR #356, opened 2016-07-28
Kuz 627 separate kuzzle start server worker

ballinette:
:warning:  As this PR introduce a lot of changes, it could be safer to wait after the next release to merge it ;) :warning: 

This PR aims to improve start-up scripts, with a better separation of Kuzzle Server and Kuzzle Worker objects. (each has now its own `start` method).

Along with this, the "dummy" mode has been removed from `start` method.
Kuzzle instance has now all its needed components ready (but not initialised with the external services) within the constructor, so that we can test each component without needing to start kuzzle.
Each internal component has now a `init` method, launched at start-up, to connect it to external services.

This PR also implied to refactor all the Unit Tests, by removing the `kuzzle.start` calls, and adding stubs for internal components.

Next steps:
* move `cleanDB` and `prepareDB` from `start` to `install` commands (to be discussed later)
* remove the Kuzzle Worker (also need some benchmarks to validate that the worker is finally useless for performances)


lib/api/core/models/repositories/profileRepository.js
```diff
@@ -7,37 +7,35 @@ var
...
-        return element.roleId;
-      }
-      return element._id;
+function ProfileRepository (kuzzle) {
+  Repository.apply(this, arguments);
+  this.collection = 'profiles';
+  this.ObjectConstructor = Profile;
+  this.profiles = {};
+
+  ProfileRepository.prototype.init = function() {
```
benoitvidis: Overriding the prototype within the constructor is disturbing me.
 
In this particular case, we have a kind-of singleton and we could say that this is not that bad but still, conceptually, I am really not comfortable with designing class-level properties during the creation of an instance.

lib/api/core/models/repositories/profileRepository.js
```diff
@@ -7,37 +7,35 @@ var
...
+    Repository.prototype.init.call(this, {
+      cacheEngine: kuzzle.services.list.userCache
```
benoitvidis: The PR #353 renamed the cache service to `securityCache` if I am not mistaken.

lib/api/core/models/repositories/roleRepository.js
```diff
@@ -6,26 +6,22 @@ var
...
-    ObjectConstructor: Role,
-    cacheEngine: kuzzle.services.list.securityCache
-  });
+function RoleRepository(kuzzle) {
+  Repository.apply(this, arguments);
+  this.collection = 'roles';
+  this.ObjectConstructor = Role;
+  this.roles = {};
+
+  RoleRepository.prototype.init = function() {
```
benoitvidis: same remark as for the ProfileRepository

lib/api/core/models/repositories/profileRepository.js
ballinette: good catch ;)

lib/api/core/models/repositories/profileRepository.js
ballinette: Do you mean I should add an object `this.kuzzle` within the prototype, and use `this.kuzzle` instead of `kuzzle` for prototypes method ?

benoitvidis: Yep, thank you :)

benoitvidis:
+1

stafyniaksacha:
LGTM

scottinet:
Since #362 embeds these changes, I guess we'll have to merge this one before releasing RC5 :-)

bin/commands/start.js
```diff
@@ -4,136 +4,131 @@ var
   fs = require('fs'),
   rc = require('rc'),
   params = rc('kuzzle'),
-  kuzzle = require('../../lib'),
+  //kuzzle = require('../../lib'),
```
scottinet: Nitpicking: can we delete this line?

lib/api/core/models/repositories/repository.js
```diff
@@ -6,28 +6,39 @@ var
...
  * @constructor
  */
-function Repository (kuzzle, options) {
-  this.collection = options.collection;
-  this.index = options.index;
-  this.ttl = options.ttl || kuzzle.config.repositories.cacheTTL;
-  this.ObjectConstructor = options.ObjectConstructor;
-  this.writeLayer = options.writeLayer;
+function Repository (kuzzle) {
+  if (arguments.length === 0) { return; }
```
scottinet: Nitpicking: This does not comply with NPM coding rules. Moreover, is this check necessary, since `Repository` objects are only used internally (risk of wrong arguments list is low)

scottinet:
Wow, huge amount of work. +1!

This can be merged as soon as the 2 small issues I raised are solved.

lib/api/core/models/repositories/repository.js
ballinette: I'll check again, but I remember I had some strange errors within unit tests, without this line.

ballinette: OK, found it, I have a bad case on prototype inheritance.
with `RoleRepository.prototype = new Repository();`, nodejs calls the constructor without arguments when we declare the prototype.
Fixed with `RoleRepository.prototype = Object.create(Repository.prototype);` instead.

PR #357, opened 2016-07-28
kuz-563 mount run scripts as volumes

benoitvidis:
* Better way to wait for ES (now waits for a yellow shard status instead of just pinging it)
* init scripts mounted as volumes from the project => will later on able to reuse docker images between kuzzle projects.


scottinet:
+1

ballinette:
+1

stafyniaksacha:
+1

PR #358, opened 2016-07-29
small wording fixes for Kuzzle commands

benoitvidis:

PR #359, opened 2016-07-29
Remove swagger command

benoitvidis:
:warning: Do not merge before #358  is.

From PR #358 , it appeared that the swagger command creates some files that are never used by the swagger controller.


ballinette:
+1 (but need to check with @j33f the initial aim of the swagger command, and if it is still needed or not)

j33f:
@ballinette already checked and its 👍

stafyniaksacha:
+1

PR #362, opened 2016-08-02
Kuz 642 remove worker

benoitvidis:
:warning: This branch is taken from the ones referenced in PRs #356  & #359. It should be merged only after the 2 ones are.

This PR removes the worker from Kuzzle' stack.

# Changelog

## Worker

* Removed worker in docker-compose files + pm2 init script (only one node left)
* Removed workerlistener
* Updated controllers to immediately send their queries to the db engine.
* NB: The internal broker is still in and init as it is used by the CLI and the cluster layer.
* Now that we only have one node running, removed the `pid` argument from all CLI commands.

## Misc

* Removed the Cluster Manager
* Fixed a small bug in the Plugin Context where the multiple profiles were not implemented yet.
* Unit tests: A new Kuzzle mock has been added with some stubs already set up that can be used instead of instantiating a real Kuzzle object.


bin/commands/disable.js
```diff
@@ -3,13 +3,13 @@
 var
   rc = require('rc'),
   params = rc('kuzzle'),
-  Kuzzle = require('../../lib/api');
+  KuzzleServer = require('../../lib/api/kuzzle');
```
ballinette: should we rather reuse "Kuzzle" term instead of "KuzzleServer" ?
(same remark for all other commands)

bin/commands/start.js
```diff
@@ -4,136 +4,110 @@ var
   fs = require('fs'),
   rc = require('rc'),
   params = rc('kuzzle'),
-  kuzzle = require('../../lib'),
+  //kuzzle = require('../../lib'),
```
ballinette: commented line to delete

bin/commands/start.js
```diff
@@ -4,136 +4,110 @@ var
...
-      console.log(`
- ████████████████████████████████████
- ██     KUZZLE ` + (kuzzle.isServer ? 'SERVER' : 'WORKER') + ' STARTED      ██');
-
-      if (kuzzle.isServer) {
-        process.title = 'KuzzleServer';
-        console.log(` ██   ...WAITING FOR WORKERS...    ██
- ████████████████████████████████████`);
-      } else {
+  if (options.worker) {
```
ballinette: this block is to be removed also, it was just there to start the worker

bin/commands/start.js
```diff
@@ -4,136 +4,110 @@ var
...
     ok = string => options.parent.noColors ? string : clc.green.bold(string),
     kuz = string => options.parent.noColors ? string : clc.greenBright.bold(string);
-  
+
   if (process.env.FEATURE_COVERAGE === '1' || process.env.FEATURE_COVERAGE === 1) {
     coverage = require('istanbul-middleware');
     console.log(warn('Hook loader for coverage - ensure this is not production!'));
     coverage.hookLoader(__dirname+'/../lib');
   }
+  console.log(kuz('Starting Kuzzle'), (options.server ? notice('Server') : warn('Worker')));
```
ballinette: to be changed to
`console.log(kuz('Starting Kuzzle'), notice('Server'));`
or just `console.log(kuz('Starting Kuzzle'));`

bin/commands/start.js
```diff
@@ -4,136 +4,110 @@ var
...
 
-      return Promise.resolve();
+      request = new RequestObject({controller: 'remoteActions', action: 'prepareDb', body: data});
+      return kuzzle.remoteActionsController.actions.prepareDb(request)
+        .catch(() => Promise.resolve());
     })
     .then(() => {
-      if (kuzzle.isServer) {
-        console.log(`
+      console.log(`
```
ballinette: nitpicking: maybe we could add `kuzzleLogo` in this log message?

bin/kuzzle
```diff
@@ -9,83 +9,78 @@
...
-// $ kuzzle swagger
+// $ kuzzle start
 program
-  .command('swagger')
-  .description('Generate the kuzzle-swagger.json and the kuzzle-swagger.yml files according to your current routes')
-  .action(require('./commands/swagger'));  
+  .command('start')
+  .description('start a Kuzzle server|worker node')
+  .option('-p, --port <port>', 'Kuzzle port number', parseInt)
+  .option('-s, --server', 'start Kuzzle in server mode (default)')
```
ballinette: options `--server` and `--worker` not needed anymore

lib/api/controllers/adminController.js
```diff
@@ -42,8 +43,7 @@ function AdminController (kuzzle) {
     return kuzzle.pluginsManager.trigger('data:beforeGetMapping', requestObject)
       .then(newRequestObject => {
         modifiedRequestObject = newRequestObject;
-
-        return kuzzle.services.list.readEngine.getMapping(modifiedRequestObject);
+        return engine.getMapping(modifiedRequestObject);
```
ballinette: why replacing `readEngine` by `writeEngine` here? (I know we are discussing about merging them, but actually, it's not done yet... or, if we replace it here, why not replacing it anywhere in this controller?)

lib/api/kuzzle.js
```diff
@@ -0,0 +1,159 @@
...
+ *     _id: String
+ *   }
+ * }} Params
+ */
+
+/**
+ * @this {KuzzleServer}
+ * @constructor
+ */
+function KuzzleServer () {
```
ballinette: should be `function Kuzzle ()` again

lib/api/kuzzle.js
```diff
@@ -0,0 +1,159 @@
...
+   *   - Hooks emitters
+   *
+   * @this {KuzzleServer}
+   * @param {Object} params command line and/or configuration file arguments
+   *                        overrides the 'feature' argument
+   * @param {Object} [feature={}] allow to programatically tune what part of Kuzzle you want to run
+   */
+  this.start = function (params, feature) {
+
+    if (feature !== undefined && feature.dummy) {
```
ballinette: to be checked: I guess `feature.dummy` is not used anymore here?

ballinette:
should also remove `lib/workers/` folder

lib/api/kuzzle.js
```diff
@@ -0,0 +1,159 @@
...
+   *   - Internal services
+   *   - Controllers
+   *   - Hooks emitters
+   *
+   * @this {KuzzleServer}
+   * @param {Object} params command line and/or configuration file arguments
+   *                        overrides the 'feature' argument
+   * @param {Object} [feature={}] allow to programatically tune what part of Kuzzle you want to run
+   */
+  this.start = function (params, feature) {
```
ballinette: NB: I've just updated PR #356 with some clean-ups in `KuzzleServer.start()` method (see https://github.com/kuzzleio/kuzzle/pull/356/commits/fab5b90e46974c3bc6af6d4df4ae459bd01a863e )
These changes should be reported here also ;)

test/api/controllers/adminController.test.js
```diff
@@ -1,595 +1,510 @@
...
   Promise = require('bluebird'),
+  rewire = require('rewire'),
+  should = require('should'),
   sinon = require('sinon'),
-  params = require('rc')('kuzzle'),
-  Kuzzle = require.main.require('lib/api/Kuzzle'),
-  User = require.main.require('lib/api/core/models/security/user'),
-  Profile = require.main.require('lib/api/core/models/security/profile'),
-  Role = require.main.require('lib/api/core/models/security/role'),
+  AdminController = rewire('../../../lib/api/controllers/adminController'),
```
ballinette: No need to use rewire, we should just load the constructor with `require.main.require`

test/api/controllers/adminController.test.js
```diff
@@ -1,595 +1,510 @@
...
 
-  before(() => {
-    kuzzle = new Kuzzle();
-
-    return kuzzle.start(params, {dummy: true});
-  });
-
   beforeEach(() => {
-    sandbox = sinon.sandbox.create();
+    kuzzle = {
```
ballinette: open question: maybe we could create a mock for Kuzzle instance within `test/mocks/`, with a default feature for each method of each internal component... and just override the methods we need within each test file ?

test/api/controllers/bulkController.test.js
```diff
@@ -1,79 +1,71 @@
 var
-  should = require('should'),
-  params = require('rc')('kuzzle'),
-  Kuzzle = require.main.require('lib/api/Kuzzle'),
   Promise = require('bluebird'),
+  should = require('should'),
+  sinon = require('sinon'),
+  BulkController = require('../../../lib/api/controllers/bulkController'),
```
ballinette: `BulkController = require.main.require('lib/api/controllers/bulkController')`

test/api/controllers/writeController.test.js
```diff
@@ -1,447 +1,369 @@
 var
   should = require('should'),
   Promise = require('bluebird'),
-  params = require('rc')('kuzzle'),
-  Kuzzle = require.main.require('lib/api/Kuzzle'),
+  sinon = require('sinon'),
   RequestObject = require.main.require('kuzzle-common-objects').Models.requestObject,
-  ResponseObject = require.main.require('kuzzle-common-objects').Models.responseObject;
+  ResponseObject = require.main.require('kuzzle-common-objects').Models.responseObject,
+  WriteController = require('../../../lib/api/controllers/writeController');
```
ballinette: `WriteController = require.main.require('lib/api/controllers/writeController')`

test/api/kuzzle.test.js
```diff
@@ -1,9 +1,10 @@
 var
   rc = require('rc'),
-  Kuzzle = require.main.require('lib/api/Kuzzle'),
-  should = require('should');
+  rewire = require('rewire'),
+  should = require('should'),
+  Kuzzle = rewire('../../lib/api/kuzzle');
```
ballinette: no need to use `rewire`.
Kuzzle contructor can be loaded by `Kuzzle = require.main.require('lib/api/kuzzle')`

test/api/controllers/bulkController.test.js
benoitvidis: why?

ballinette: I find it more elegant, within tests files, to require files from the app root, instead of those ugly relative paths (if we move the test file, it will be broken)

test/api/controllers/adminController.test.js
benoitvidis: Good idea. 
Done and implemented as a POC in bulk, admin and write controllers. Seems to work just fine.

ballinette:
:+1:

lib/api/controllers/adminController.js
```diff
@@ -10,7 +9,9 @@ var
  * @param {Kuzzle} kuzzle
  * @constructor
  */
-function AdminController (kuzzle) {
+function AdminController(kuzzle) {
+  var engine = kuzzle.services.list.writeEngine;
```
scottinet: Is there still a use to the read/write engines distinction?

lib/api/core/plugins/pluginContext.js
```diff
@@ -21,44 +21,38 @@ function PluginContext (kuzzle) {
...
-  if (kuzzle.isServer) {
-    Object.defineProperty(this.accessors, 'passport', {
-      enumerable: true,
-      get: function () {
-        return {
-          use: kuzzle.passport.use
-        };
-      }
-    });
+  Object.defineProperty(this.accessors, 'passport', {
```
scottinet: We still need to only inject accessors if the context is built by a Kuzzle instance. Because there is still the matter of threaded plugins, and these injections will make their initialization crash because they can't provide a complete Kuzzle object to this constructor (see lib/core/plugins/workerReady.js)

scottinet: I also wanted to add a documentation link to my comment, but it seems I forgot to mention this point about worker plugins when I rewrote the documentation. :fearful: 

I'll submit a PR on the Kuzzle Guide documentation ASAP.

lib/api/controllers/adminController.js
benoitvidis: To me, no, but I'd rather tackle it in a separate PR.

lib/api/core/plugins/pluginContext.js
benoitvidis: Good catch.

ballinette: mhh well, we don't have the `isServer` attribute anymore... 
I suggest to replace the condition by `if (kuzzle instanceof Kuzzle)`

lib/api/controllers/adminController.js
scottinet: I agree :-)

lib/api/core/plugins/pluginContext.js
scottinet: It should work.

j33f:
the bin/README.md file should be modified too (to reflect the pid option disparition

> **Note:** I let this file here because we do not have a man entry for the kuzzle command and it is always cool to have a quick reference directly accessible from the console

Otherwise 💯 !!!!! damn good job !!!

scottinet:
+1

PR #363, opened 2016-08-04
upgrade default plugins version

ballinette:
following:
https://github.com/kuzzleio/kuzzle-plugin-auth-passport-local/pull/17
https://github.com/kuzzleio/kuzzle-plugin-logger/pull/9


scottinet:
+1

PR #365, opened 2016-08-08
add default config into kuzzle image

stafyniaksacha:
Simplify the default files needed to run a kuzzle stack

- Added default pm2 config file into kuzzle image
- Added default run file into kuzzle image
- Removed default volume from base docker-compose file
- Remove environment entry from base docker-compose file


ballinette:
+1

dbengsch:
+1

PR #367, opened 2016-08-10
Api create first admin

AnthonySendra:
* Expose adminExists in API
* Expose createFirstAdmin in API
* Remove kuzzle worker node-inspector in `run-dev.sh`

PR #368, opened 2016-08-10
kuz-606 & kuz-607 - kuzzlerc refactor

benoitvidis:
* remove .kuzzlerc file => replaced by a default.config.js, which can be overriden by the user using rc
* .kuzzlerc.sample file added
* Side changes: the services init script has been slightly modified to allow defining a backend for a service and/or some aliases for the same service.


dbengsch:
+1

lib/api/controllers/remoteActionsController.js
```diff
@@ -20,7 +20,7 @@ function RemoteActionsController (kuzzle) {
   };
 
   this.init = () => {
-    this.kuzzle.services.list.broker.listen(this.kuzzle.config.queues.remoteActionsQueue, this.onListenCB);
+    this.kuzzle.services.list.broker.listen(kuzzle.config.queues.remoteActionsQueue, this.onListenCB);
```
ballinette: why?
I mean: here, `kuzzle` and `this.kuzzle` point to the same object, but why replacing `this.kuzzle` by `kuzzle` only here and not anywhere else in the controller ?

lib/services/internalEngine.js
```diff
@@ -43,8 +42,8 @@ function InternalEngine (kuzzle) {
   this.init = function () {
     if (!this.client) {
       this.client = new Elasticsearch.Client({
-        host: kuzzle.config[this.engineType].hosts,
-        apiVersion: kuzzle.config[this.engineType].apiVersion
+        host: kuzzle.config.services.db.host + ':' + kuzzle.config.services.db.port,
```
ballinette: nitpicking: If I understand correctly the refactor of `lib/services/index.js`, we could also use `config` as a constructor's parameter (`function InternalEngine(kuzzle, options, config) {...}`), and use here:
```
this.client = new Elasticsearch.Client({
  host: config.host + ':' + config.port,
  apiVersion: config.apiVersion
});
```

(to be consistent with other services implementation)

ballinette:
+1

lib/services/internalEngine.js
benoitvidis: the thing is that despite the fact the internalEngine source is located in the services directory, it is actually not initialized by the service layer but directly from Kuzzle start process.

I agree this would make the service look more consistent but the sole purpose of passing the service configuration from `lib/services/index.js` is to deal with mutliple backend/aliases cases where it could be tricky to get the proper config from the constructor.

In this specific case, this logic is not needed.

.kuzzlerc.sample
```diff
@@ -0,0 +1,285 @@
...
+        "admin": {
+          "controllers": {
+            "*": {
+              "actions": {
+                "*": true
+              }
+            }
+          }
+        },
+        "default": {
```
stafyniaksacha: missing `{controller:'admin', action:'adminExists'}` action (added in #367)

.kuzzlerc.sample
```diff
@@ -0,0 +1,285 @@
...
+        },
+        "anonymous": {
+          "controllers": {
+            "auth": {
+              "actions": {
+                "checkToken": true,
+                "getCurrentUser": true,
+                "getMyRights": true,
+                "login": true,
+                "logout": true
```
stafyniaksacha: Same as above

default.config.js
```diff
@@ -0,0 +1,176 @@
...
+          controllers: {
+            '*': {
+              actions: {
+                '*': true
+              }
+            }
+          }
+        },
+        default: {
+          controllers: {
```
stafyniaksacha: missing `{controller:'admin', action:'adminExists'}` action (added in #367)

default.config.js
```diff
@@ -0,0 +1,176 @@
...
+            },
+            read: {
+              actions: {
+                serverInfo: true
+              }
+            }
+          }
+        },
+        anonymous: {
+          controllers: {
```
stafyniaksacha: Same as above

stafyniaksacha:
Despite my remark, LGTM

.kuzzlerc.sample
ballinette: not agree.
The "adminExists" action should only be available for admin users (role "admin" as permissions for all actions of all controllers, see line 120), and for the temporary permissions that are used before the admin is created (that is set ad line 89).

It's useless for "default" and "anonymous" roles.

stafyniaksacha: Ok, it's required to work with the backoffice. We have to handle it defferently so.

ballinette: Maybe the BO use case gives a needs that I didn't think about... this can be discussed ;)

PR #371, opened 2016-08-12
[HOTFIX] Listen end response for Passport Authentication

Filsh:
In current master auth via passportjs not working, this commit fix it


jenow:
Hi @Filsh. Thanks for your contribution! Could you explain what does it fix and give a use case when the auth passport does not work?

Filsh:
PassportResponse is mockup object for native response, so when passportjs call end() function instead of return an answer to the client, nothing happens and the Promise does not receive the user object.

jenow:
Thanks for the help!

PR #374, opened 2016-08-22
prepareDb createIndex bug when using mapping & fixture on same index

stafyniaksacha:
+ output message instead of object on prepareDb when mapping is wrong


jenow:
+1

ballinette:
+1

j33f:
+1

PR #376, opened 2016-08-26
profilesIds as a string

AnthonySendra:
In Elasticsearch mapping there is no type `array`. A type `string` can contains an array and profilesIds field has mapping like
```
"profilesIds": {
  "type": "string"
}
```

In BO we can't know if this is a `string` or an `array`. So, currently we have a different behaviour: Elasticsearch accept `string` and convert to an `array` but not Kuzzle.


ballinette:
+1

jenow:
+1

PR #375, opened 2016-08-26
[HOTFIX] #326

Filsh:

PR #377, opened 2016-08-29
fix #326

### LINKED ISSUE #326: Plugins configuration are reimported after restarting Kuzzle
Kuzzle shouldn't reimport the default configuration of plugins after a restart if there's already one.

jenow:
Fix the issue #326 
From #375


j33f:
+1

stafyniaksacha:
+1

PR #378, opened 2016-08-30
[BREAKING] Replace profilesIds to profileIds

jenow:
Replaced `profilesIds` to `profileIds` because some of the code is using `profilesIds` and some other `profileIds` so we decided to change them all to `profileIds`


j33f:
👍

ballinette:
+1

stafyniaksacha:
+1

PR #382, opened 2016-09-05
Fix #381: Startup message about administrator status is wrong

### LINKED ISSUE #381: Startup message about administrator status is wrong
When Kuzzle starts, it should print either "Administrator account configured" or "Entering no-administrator mode..."

Now it seems only the latter message appears, even on a fresh install, so there may be something wrong with the way Kuzzle detects if an administrator has been configured.

scottinet:
The startup script uses the `adminExists` method of the admin controller, but it interprets the result as a boolean instead of a response object.

This PR fixes this problem, and also changes the response of the `adminExists` method.

Instead of returning this:

```json
{
  "data": {
    "body": true
  }
}
```

It now returns this:

```json
{
  "data": {
    "body": {
      "exists": true
    }
  }
}
```

### DIFF
--- a/bin/commands/start.js
+++ b/bin/commands/start.js
@@ -88,7 +88,7 @@ module.exports = function (options) {
  ████████████████████████████████████`);
       return kuzzle.remoteActionsController.actions.adminExists()
         .then((res) => {
-          if (res) {
+          if (res.data.body.exists) {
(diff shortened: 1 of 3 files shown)


j33f:
👍

jenow:
+1

PR #387, opened 2016-09-08
Fix #385 reformate bulk data

### LINKED ISSUE #385: requestObject.data.body should always be an object
for almost every controllers, the `requestObject.data.body` attribute is a JSON object (that can be optionally empty, for example for a search with empty filters).... except for the bulk controller, where it is an array.
see the doc here:
http://kuzzle.io/api-reference/#import
```
{
  index: "<data index>",
  collection: "<data collection>",
  controller: "bulk",
  action: "import",

  // Data mapping using ElasticSearch bulk syntax.
  body: [
    {create: {}},
    { a: "document", with: "any", number: "of fields" },
    {create: {}},
    { another: "document" },
    {create: {}},
    { and: { another: "one"} },
    ...
  ]
}
```

To be consistent, we should use also a JSON object here, with following format:
```
body: {
  bulkData: [
    {create: {}},
    { a: "document", with: "any", number: "of fields" },
    {create: {}},
    { another: "document" },
    {create: {}},
    { and: { another: "one"} },
    ...
  ]
}
```

ballinette:
fix #385

### DIFF
--- a/lib/services/elasticsearch.js
+++ b/lib/services/elasticsearch.js
@@ -355,15 +355,30 @@ function ElasticSearch (kuzzle, options, config) {
   this.import = function (requestObject) {
     var
       nameActions = ['index', 'create', 'update', 'delete'],
+      optionalAttributes = ['consistency', 'refresh', 'routing', 'timeout', 'fields'],
       data = cleanData.call(this, requestObject),
+      bulkData,
       error = null;
 
     if (!data.body) {
       return Promise.reject(new BadRequestError('Bulk import: Parse error: document <body> is missing'));
     }
 
+    if (!data.body.bulkData) {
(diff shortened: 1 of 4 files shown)


lib/services/elasticsearch.js
```diff
@@ -355,15 +355,30 @@ function ElasticSearch (kuzzle, options, config) {
...
     }
 
+    if (!data.body.bulkData) {
+      return Promise.reject(new BadRequestError('Bulk import: Parse error: input paramter <bulkData> is missing'));
+    }
+
+    bulkData = {
+      body: data.body.bulkData
+    };
+    optionalAttributes.forEach(attr => {
```
ballinette: Inject also optional bulk parameters (see https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/api-reference.html#api-bulk )

scottinet:
+1

j33f:
+1

PR #389, opened 2016-09-09
Fix #388: allow Kuzzle to reconnect after losing connection to proxy

### LINKED ISSUE #388: Kuzzle fails to reconnect if it lose connection to proxy
If Kuzzle loses connection to Kuzzle Proxy (either because the proxy restarts or whatever the reason may be), it tries to reconnect. 
But when it finally succeeds, an error occurs and the kuzzle instance is in an unstable state: a socket is opened with the proxy, but all requests sent to this instance are lost.

scottinet:
Fixed a bug where if Kuzzle was connected to a proxy, and then lose the connection, it was able to reconnect successfully but an error prevented it to accept requests afterward.


ballinette:
+1

AnthonySendra:
+1

j33f:
+1

PR #392, opened 2016-09-20
KUZ-682: rename DSL keywords

scottinet:
:warning: Breaking change :warning: 

To prevent confusion with Elasticsearch's DSL, we have to rename the following keywords from our real-time DSL:

- "term" => "equals"
- "terms" => "in"


scottinet:
Oops, I got carried away when replacing term/terms with equals/in. I also replaced them in ES searches in the security modules. :fearful:

scottinet:
Revert changes in security modules. This PR can be reviewed now.

features/step_definitions/readDocument.js
```diff
@@ -102,8 +102,7 @@ var apiSteps = function () {
   });
 
   this.Then(/^I ?(don't)* find a document with "([^"]*)"(?: in field "([^"]*)")?(?: in index "([^"]*)")?$/, function (dont, value, field, index) {
-    var filters = {filter: { term: {}}};
-    filters.filter.term[field] = value;
+    var filters = {filter: { term: { [field]: value }}};
```
ballinette: shouldn't it be `{ equals: { [field]: value }` instead of "term" ?

scottinet: This test read documents using Elasticsearch, while we replaced "term" with "equals" only for our real-time engine.

PR #397, opened 2016-09-23
add elasticsearch error translation dict

stafyniaksacha:


lib/services/elasticsearch.js
```diff
@@ -703,17 +703,90 @@ function refreshIndexIfNeeded(data, response) {
...
+    formattedMessages,
+    message,
+    kuzzleError;
+
+  formattedMessages = {
+    // [illegal_argument_exception] object mapping [titi] can\'t be changed from nested to non-nested
+    '^\\[illegal_argument_exception\\] object mapping \\[([^\\]]+)\\] can\'t be changed from nested to non-nested$': 'Can not change mapping for field "$1" from nested to another type',
+
+    // [illegal_argument_exception] object mapping [baz] can\'t be changed from non-nested to nested
+    '^\\[illegal_argument_exception\\] object mapping \\[([^\\]]+)\\] can\'t be changed from non-nested to nested$': 'Can not change mapping for field "$1" from object to another type',
```
benoitvidis: when this error occurs, it is more likely that the origin field type is a scalar than an object.
The (non)nested terminology sounds understanble and more suitable to me.

lib/services/elasticsearch.js
```diff
@@ -703,17 +703,90 @@ function refreshIndexIfNeeded(data, response) {
 }
 
 function formatESError(error) {
-  var kuzzleError = new Error(error.message);
+  var
+    formattedMessages,
+    message,
+    kuzzleError;
+
+  formattedMessages = {
```
benoitvidis: Having the regexes expressed as strings and evaluated on each function likely forces the code to compile the regexes on each call.
Would be slightly more efficient using an array of some pre-compiled regexes on the module level.

Something like:

```js
function Elasticsearch (kuzzle, options, config) {
  this.errorMessagesMapping = [
    {   regex: /^\[illegal_argument_exception\] object mapping \[(.*?)\] can't be changed from nested to non-nested$/,
        replacement: 'Cannot change mapping for field "$1" from nested to another type'   },
    {   regex: /^\[illegal_argument_exception\] object mapping \[(.*?)\] can't be changed from non-nested to nested$/,
        replacement: 'Cannot change mapping for field "$1" from non-nested to nested' },
    [..]
  ];
}

Elasticsearch.prototype.formatESError (error) {
  var message = error.message;

  this.errorMessagesMapping.some(mapping => {
    message = message.replace(mapping.regex, mapping.replacement);
    return (message !== error.message);
  });

  [..]
}

module.exports = Elasticsearch;

File: lib/services/elasticsearch.js (hunk 1 above)
Comment (stafyniaksacha) [inline lib/services/elasticsearch.js]: This error occurs even if the user change a field from `object` to `nested`

File: lib/services/elasticsearch.js (hunk 2 above)
Comment (stafyniaksacha) [inline lib/services/elasticsearch.js]: Ok, implemented

File: lib/services/elasticsearch.js (hunk 1 above)
Comment (benoitvidis) [inline lib/services/elasticsearch.js]: Still, it can also occur when updating a scalar type to a nested one, in which case the original type is not `object`.

Comment (stafyniaksacha) [inline lib/services/elasticsearch.js]: The corresponding error has been added

File: lib/services/elasticsearch.js [hunk 3]
```diff
@@ -38,6 +38,60 @@ function ElasticSearch (kuzzle, options, config) {
     }
   });
 
+  this.errorMessagesMapping = [
+    {
+      // [illegal_argument_exception] object mapping [titi] can't be changed from nested to non-nested
+      regex: /^\[illegal_argument_exception\] object mapping \[(.*?)\] can't be changed from nested to non-nested$/,
+      replacement: 'Can not change mapping for field "$1" from nested to another type'
```
ballinette: why don't you just keep the "non-nested" term ? its more explicit that "another type"

lib/services/elasticsearch.js
```diff
@@ -38,6 +38,60 @@ function ElasticSearch (kuzzle, options, config) {
...
+    },
+    {
+      // [illegal_argument_exception] object mapping [baz] can't be changed from non-nested to nested
+      regex: /^\[illegal_argument_exception\] object mapping \[(.*?)\] can\'t be changed from non-nested to nested$/,
+      replacement: 'Can not change mapping for field "$1" from object to another type'
```
ballinette: Same remark: "another type" is quite vague.
Why not `replacement: 'Can not change mapping for field "$1" from a non-nested to a nested object'` ?

lib/services/elasticsearch.js
stafyniaksacha: Because one day I asked me if a type `object` is a nested or non-nested type and in fact no.

ballinette: hmmm 
for me, "nested" is not a "type", but rather a property of an object.
an object can be either nested or non-nested.
a scalar can only be non-nested

So "from nested to another type" doesn't make sense

stafyniaksacha: Nested is a real type and it's not acting like Object type: 
https://www.elastic.co/guide/en/elasticsearch/reference/2.2/nested.html
https://www.elastic.co/guide/en/elasticsearch/reference/2.2/object.html

lib/services/elasticsearch.js
stafyniaksacha: Same as above, we can't speak about `nested` when we are on `object` type.

But if it's not clear we need to find other words

lib/services/elasticsearch.js
benoitvidis: ok

lib/services/elasticsearch.js
ballinette: ok

lib/services/elasticsearch.js
ballinette: OK, i did not notice that "nested" was considered as a specific type by ES.
But as I understand the original ES error message, "from non-nested to nested", it means that this error can occur when we try to convert an `object` to a `nested`, but also when we try to convert a scalar type to a `nested`
So "from object to another type" is not exactly what it means... better say "from object or a scalar type to nested", or something like this?

ballinette:
despite my remarks, that are very pernickety, consider that I approve this PR ;)

benoitvidis:
This PR breaks the develop branch once merged.
Reopen branch to fix the tests.

PR #399, opened 2016-09-23
Fix 396: remove redis flushdb from init script

### LINKED ISSUE #396: Redis service should not flush DBs at init
A flushDB command is sent to Redis each time the service is initialized (that means, at least each time a Kuzzle instance is started).
This should be avoided, to keep cache fulfilled when a new instance is started in a cluster.

Additionnaly, add a CLI command to flush the database by hand when needed.

ballinette:
fix #396 :
* remove redis flushdb from init script
* add CLI commands to flush DB by hands, with following behaviour:
  * if no database name is provided, all volatile databases are flushed (that means: all except "memoryStorage")
  * if a database name is provided, only this database is flushed (with a strong confirmation needed for "memoryStorage")
* flush also "memoryStorage" within "cleanDB" command

### DIFF
--- a/bin/commands/clearCache.js
+++ b/bin/commands/clearCache.js
@@ -36,9 +36,9 @@ module.exports = function (database, options) {
     console.log(warn('[ℹ] This operation cannot be undone.\n'));
     userIsSure = params.noint || readlineSync.question('[❓] Are you sure? If so, please type "I am sure" (if not just press [Enter]): ') === 'I am sure';
   } else if (database) {
-    userIsSure = readlineSync.keyInYN(question('[❓] Do you want t clear Kuzzle internal cache for "'.concat(database, '".')));
+    userIsSure = readlineSync.keyInYN(question('[❓] Do you want to clear Kuzzle internal cache for "'.concat(database, '".')));
   } else {
-    userIsSure = readlineSync.keyInYN(question('[❓] Do you want t clear all Kuzzle internal cache'));
+    userIsSure = readlineSync.keyInYN(question('[❓] Do you want to clear all Kuzzle internal cache'));
   }
 
   if (userIsSure) {
--- a/bin/commands/clearCache.js
+++ b/bin/commands/clearCache.js
@@ -1,0 +1,58 @@
+/* eslint-disable no-console */
+
+var
+  rc = require('rc'),
+  params = rc('kuzzle'),
+  Kuzzle = require('../../lib/api/kuzzle'),
+  readlineSync = require('readline-sync'),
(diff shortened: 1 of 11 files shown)


bin/commands/clearCache.js
```diff
@@ -0,0 +1,58 @@
...
+  notice = string => options.parent.noColors ? string : clc.cyanBright(string);
+  ok = string => options.parent.noColors ? string: clc.green.bold(string);
+  question = string => options.parent.noColors ? string : clc.whiteBright(string);
+
+  if (database === 'memoryStorage') {
+    console.log(warn('[ℹ] You are about to clear Kuzzle memoryStorage database.'));
+    console.log(warn('[ℹ] This operation cannot be undone.\n'));
+    userIsSure = params.noint || readlineSync.question('[❓] Are you sure? If so, please type "I am sure" (if not just press [Enter]): ') === 'I am sure';
+  } else if (database) {
+    userIsSure = readlineSync.keyInYN(question('[❓] Do you want t clear Kuzzle internal cache for "'.concat(database, '".')));
```
dbengsch: Do you want **to** clear (twice) ?

ballinette: fixed

j33f:
why not add an option to likeAvirgin command to flush cache alltogether ?

lib/services/redis.js
```diff
@@ -47,18 +47,7 @@ function Redis (kuzzle, options, config) {
...
-          }
-
-          this.client.flushdb(err => {
-            if (err) {
-              return reject(err);
-            }
-
-            resolve(this);
-          });
+          return resolve(this);
```
benoitvidis: nitpicking: Would be nice to have notice log in the 'ready' event to catch the case where a network configuration may force the client to reconnect repeatedly, like we've seen lately (i.e. client timeout set in haproxy) .

ballinette:
@j33f not sure about adding an "option".

In fact, I actually clear the MemoryStorage data with `likeAvirgin` command (without needing any specific option)... but I should also flush all the cache at the same time, yes

lib/services/redis.js
ballinette: Sure, I'll add this... what message text do you suggest ?

j33f:
well, the option is not a good idea, you are right, but yep, flushing the cache should come along with likeAvirgin imho

ballinette:
OK, it's now done (see https://github.com/kuzzleio/kuzzle/pull/399/commits/d452cfac1fe07db0eddad624fc757aa2cde5aa5f )

PR #402, opened 2016-09-27
Let IP be an IP

AnthonySendra:
If the host is an IP the file `lib/config/index` will transform it into a float. This can produce something like

```
export kuzzle_services__proxyBroker__host=10.0.2.251
export kuzzle_services__proxyBroker__port=7331
pm2 restart all
```

```
[proxyBroker] Error while trying to connect to ws://10:7331
```


scottinet:
+1

Issue #400, opened 2016-09-26
Default value for redis and elastic search


jenow:
Today the host default configuration value of redis and elastic search are "redis" and "elasticsearch". Would be better to set it to "localhost" by default

PR #404, opened 2016-09-27
Kuz 692 redis cluster

ballinette:
Add support to Redis Cluster (cf. http://redis.io/topics/cluster-spec)
* merge all volatile cache DB into one single "internalCache" (+ add a key prefix for each usage)
* split settings of "internalCache" and "memoryStorage" to let use separate redis instance for each one (to keep descendent compatibility, it still uses the same Redis by default, with DB "0" for "internalCache", "5" for "memoryStorage" )
* add .kuzzlerc sample settings 
  * for Master/Slave with Sentinels
  * for Cluster mode


.kuzzlerc.sample
```diff
@@ -204,38 +204,71 @@
...
+    // 1. using a single Redis database:
+    //   node:
+    //     * host: The host on which Redis can be reached.
+    //       Can take an IP address, an URI or a hostname
+    //     * port: The port on which Redis is running
+    //   database: (optional) the ID of the redis database (default: 0)
+    "internalCache": {
+      "backend": "redis",
+      "node": {
+        "host": "redis",
```
scottinet: Isn't this configuration re-introducing the problem detailed in #400 (and solved in #406)?

default.config.js
```diff
@@ -125,24 +125,16 @@ module.exports = {
...
-      databases: [
-        'notificationCache',
-        'statsCache',
-        'securityCache',
-        'internalCache',
-        'tokenCache',
-        'memoryStorage'
-      ],
+      node: {
+        host: 'redis',
```
scottinet: Same as my comment above

default.config.js
```diff
@@ -125,24 +125,16 @@ module.exports = {
...
+        port: 6379
+      }
+    },
+    memoryStorage: {
+      backend: 'redis',
```
scottinet: Ditto

.kuzzlerc.sample
ballinette: you're asbolutly right ;)

ballinette: (but #406 isn't merged yet :P )

scottinet: So first merged PR wins? :stuck_out_tongue:

PR #406, opened 2016-09-27
Changed default value for host configuration 

### LINKED ISSUE #400: Default value for redis and elastic search

jenow:
fix #400


ballinette:
+1

PR #417, opened 2016-10-04
Kuz 698 auto http port

benoitvidis:
Kuzzle http port is configured by default to use 7511. This is the same port which is used by the proxy for its REST interface.
On a single host installation, this creates a port conflict.

As Kuzzle declares its http server port to the proxy, we can let it pick one random. This also ease the process of running several Kuzzle instances on a single host.

### DIFF
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -171,7 +171,10 @@
   // Configuration for Kuzzle server
   /*
   "server": {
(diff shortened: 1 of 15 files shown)


scottinet:
As discussed yesterday, I think that it would be a good idea to be able to set up a port range, to allow configuring firewalls and forward ports between kuzzle instances.
It can be done in another pull request, though. If so, this one gets my approval.

PR #418, opened 2016-10-04
Kuz 696 Implementation of the validation mechanism

dbengsch:
- Adds a new property in the `kuzzle` object : `validation`.
- Validation exposes methods to treat the validation specification (aka: the schema), check its conformity and validates the documents against it.
- a call to validation is done after data:beforePublish, data:beforeCreate, data:beforeCreateOrReplace, data:beforeUpdate a,d data:beforeReplace
- **Plugin context** : a new constructor and new accessors have been introduced to interaction with the validation: 
 - context.constructors.BaseValidationType : This constructor must be extended by the validation types in order to be used in the validation mechanism.
 - context.accessors.validation.addType : Adds a new type to the validation (must be a constructor that extends `BaseValidationType`).
 - context.accessors.validation.validate : Takes a `requestObject` in argument and validate it against the specification; returns true when it succeeds, false otherwise.


lib/api/core/validation/types/boolean.js
```diff
@@ -0,0 +1,36 @@
+var
+  BaseConstructor = require('../baseType');
+
+/**
+ * @constructor
+ */
+function MyNewType () {
```
j33f: MyNewType ? should not be BooleanType instead ?

lib/api/core/validation/types/anything.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * @constructor
+ */
+function AnythingType () {
+  this.typeName = 'anything';
+  this.allowChildren = false;
+  this.allowedTypeOptions = [];
+}
+
+AnythingType.prototype = new BaseConstructor();
```
benoitvidis: nitpicking: `Anything.constructor` is not set to `Anything`. Also applies to all other types.

```js
var o = new AnythingType(); 
console.log(o.constructor);
[Function: BaseType]
```

Whe can use [`util.inherits`](https://nodejs.org/docs/latest/api/util.html#util_util_inherits_constructor_superconstructor) here.

lib/api/core/validation/types/anything.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @returns {boolean}
+ */
+AnythingType.prototype.validate = function () {
```
benoitvidis: Why reimplement the same method as the prototype one?

lib/api/core/validation/types/anything.js
```diff
@@ -0,0 +1,29 @@
...
+  return true;
+};
+
+/**
+ * @return {boolean}
+ */
+AnythingType.prototype.validateFieldSpecification = function () {
```
benoitvidis: same as above

lib/api/core/validation/types/boolean.js
```diff
@@ -0,0 +1,36 @@
...
+    return false;
+  }
+
+  return true;
+};
+
+/**
+ * @return {boolean}
+ */
+BooleanType.prototype.validateFieldSpecification = function () {
```
benoitvidis: Same as above.

lib/api/core/validation/types/date.js
```diff
@@ -0,0 +1,211 @@
...
+
+  if (typeOptions.hasOwnProperty('formats')) {
+    if (!Array.isArray(typeOptions.formats) || typeOptions.formats.length === 0) {
+      return false;
+    }
+
+    typeOptions.formats.forEach(format => {
+      if (Object.keys(formatMap).indexOf(format) === -1) {
+        result = false;
+        return false;
```
benoitvidis: same as in validation/index.js. Looks like an invalid attempt to escape forEach.

lib/api/core/validation/types/enum.js
```diff
@@ -0,0 +1,63 @@
...
+ */
+EnumType.prototype.validate = function (typeOptions, fieldValue, errorMessages) {
+  var result = false;
+
+  if (typeof fieldValue !== 'string') {
+    errorMessages.push('The field must be a string.');
+    return false;
+  }
+
+  typeOptions.values.forEach(value => {
```
benoitvidis: why use a forEach?
```js
result = typeOptions.values.indexOf(value) !== -1;
```

lib/api/core/validation/types/enum.js
```diff
@@ -0,0 +1,63 @@
...
+ * @throws InternalError
+ */
+EnumType.prototype.validateFieldSpecification = function (typeOptions) {
+  var result = true;
+
+  if (!typeOptions.hasOwnProperty('values') || !Array.isArray(typeOptions.values) || typeOptions.values.length === 0) {
+    return false;
+  }
+
+  typeOptions.values.forEach(value => {
```
benoitvidis: nitpicking: better use Array.prototype.every

lib/api/core/validation/types/geoShape.js
```diff
@@ -0,0 +1,282 @@
...
+      }
+      break;
+    case 'geometrycollection':
+      coordinateValidation = () => true;
+      if (!Array.isArray(shape.geometries)) {
+        errorMessages.push('The geometries property must be an array.');
+        return false;
+      }
+
+      for (i = 0; i < shape.geometries.length; i++) {
```
benoitvidis: Array.prototype.every

lib/api/core/validation/types/geoShape.js
```diff
@@ -0,0 +1,282 @@
...
+ * @param {string[]} allowedShapes
+ * @param {GeoShape} shape
+ * @param {string[]} errorMessages
+ * @returns {boolean}
+ */
+GeoShapeType.prototype.checkStructure = function (allowedShapes, shape, errorMessages) {
+  var result = true;
+  if (!this.checkAllowedProperties(shape, allowedShapeProperties)) {
+    errorMessages.push('The shape object has a not allowed property.');
+    result = false;
```
benoitvidis: Why not return on failure?

lib/api/core/validation/types/geoShape.js
```diff
@@ -0,0 +1,282 @@
...
+/**
+ * @param {GeoShapePointCoordinates} point
+ * @returns {boolean}
+ */
+function isPoint (point) {
+  if (!Array.isArray(point) || point.length !== 2) {
+    return false;
+  }
+
+  return !(point[0] < -180 || point[0] > 180 || point[1] < -90 || point[1] > 90);
```
benoitvidis: So a point is [longitude, latitude]. Is it standard? lat, lon seems to be more common to me.

lib/api/core/validation/types/string.js
```diff
@@ -0,0 +1,61 @@
...
+    errorMessages.push('The field must be a string.');
+    return false;
+  }
+
+  if (!typeOptions.length) {
+    return true;
+  }
+
+  if (typeOptions.length.hasOwnProperty('min') && fieldValue.length < typeOptions.length.min) {
+    errorMessages.push('The string is not long enough.');
```
benoitvidis: Invalid string length. Expected min: ${min} max: ${max}. Received: ${fieldValue.lengh} "${fieldValue}"

Can then also be merged with the over max case below.

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+  multivaluedProperties = ['value', 'minCount', 'maxCount'],
+  Promise = require('bluebird'),
+  Dsl = require('../../dsl'),
+  defaultFieldSpecification = {
+    'mandatory': false,
+    'multivalued': {
+      value: false
+    }
+  },
+  defaultTypesFiles = [
```
dbengsch: removed all unnecessary vars

lib/api/core/validation/baseType.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * @constructor
+ */
+function BaseType () {
+
+}
+
+/**
+ * Must be overrided in order for the validation type to be useful
```
scottinet: nitpicking: remove "in order"

lib/api/core/validation/baseType.js
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * @param {*} object
+ * @param {string[]} allowedProperties
+ * @returns {boolean}
+ */
+BaseType.prototype.checkAllowedProperties = function (object, allowedProperties) {
+  var result = true;
+
+  if (typeof object !== 'object') {
```
scottinet: Beware: `typeof null` and `typeof <array>` also return `object`

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+ * @param {string} index
+ * @param {string} collection
+ * @param {string} controller
+ * @param {string} action
+ * @param {string} id
+ * @param {*} documentCandidate
+ * @param {string[]|*} structuredError
+ * @returns {Promise.<{documentBody: *, errorMessages:string[], validation: boolean}>}
+ */
+Validation.prototype.validationPromise = function (index, collection, controller, action, id, documentCandidate, structuredError) {
```
dbengsch: No, because the same function will be used to make the document validation route where the requestObject would be drastically different.

lib/api/core/validation/baseType.js
```diff
@@ -0,0 +1,59 @@
...
+    return false;
+  }
+
+  Object.keys(object).forEach(propertyName => {
```
scottinet: This would be better written usinng Array.every:

```js
return Object.keys(object).every(p => allowedProperties.indexOf(p) !== -1);
```

Array.every() stops at the first falsey callback, while Array.forEach() executes it for all array elements, regardless of its return value.

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+
+      if (collectionSpec.fields) {
+        try {
+          validation = validation && this.recurseFieldValidation(documentBody, updateBodyClone || null, collectionSpec.fields.children, collectionSpec.strict, errorMessages, structuredError);
+        }
+        catch (error) {
+          if (error.message === 'strictness') {
+            // The strictness message can be received here only if it happens at the validation of the document's root
+            addErrorMessage('document', errorMessages, 'The document validation is strict; it can not add unspecified sub-fields.', structuredError);
+            strictnessError = true;
```
dbengsch: I had 2 choices, stop at the first error, or report all errors. I chose to report all errors. And yes, there is a high change the DSL error can come from malformed or missing fields, but we can't tell ourselves so ...

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+ * @param {string} action
+ * @param {string} id
+ * @param {*} documentCandidate
+ * @param {string[]|*} structuredError
+ * @returns {Promise.<{documentBody: *, errorMessages:string[], validation: boolean}>}
+ */
+Validation.prototype.validationPromise = function (index, collection, controller, action, id, documentCandidate, structuredError) {
+  var
+    // Avoid side effects on the requestObject during the validation
+    requestBodyClone = JSON.parse(JSON.stringify(documentCandidate)),
```
benoitvidis: and? is it a problem?
My concern is that the serialization process is costly and if it can be avoided, I'd rather to.

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+          addErrorMessage('document', errorMessages, 'The document is not valid due to field validation.', structuredError);
+        }
+      }
+
+      if (collectionSpec.validators) {
+        return this.dsl.test(index, collection, documentBody, id)
+          .then(filters => {
+            var result = false;
+
+            filters.forEach(filter => {
```
dbengsch: done

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+
+/**
+ * @param {*} documentSubset
+ * @param {*} originalBodySubset
+ * @param {StructuredFieldSpecification} collectionSpecSubset
+ * @param {boolean} strictness
+ * @param {string[]} errorMessages
+ * @param {boolean} structuredError
+ */
+Validation.prototype.recurseFieldValidation = function (documentSubset, originalBodySubset, collectionSpecSubset, strictness, errorMessages, structuredError) {
```
dbengsch: The problem comes from default values as we want to get them updated in the requestObject, but we don't want to get the complete document to be updated but just the part that was originally in the requestObject

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+  if (strictness) {
+    if (!checkAllowedProperties(documentSubset, Object.keys(collectionSpecSubset))) {
+      // We use a throw to be able to provide information about the field or the document in whole
+      throw new BadRequestError('strictness');
+    }
+  }
+
+  return Object.keys(collectionSpecSubset).reduce((reductionResult, fieldName) => {
```
scottinet: Just a reminder, after we debate about fail-fast vs. get all errors: if the former is chosen, then Array.every should be used instead of Array.reduce.

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+Validation.prototype.curateSpecification = function () {
+  var
+    promises = [],
+    specification = {};
+
+  return getValidationConfiguration(this.kuzzle)
+    .then(validation => {
+      this.rawConfiguration = validation;
+
+      _.each(this.rawConfiguration, (indexSpec, indexName) => {
```
dbengsch: done

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+              documentSubset[fieldName],
+              originalBodySubset.hasOwnProperty(fieldValues) ? originalBodySubset[fieldValues] : null,
+              field.children,
+              nestedStrictness,
+              errorMessages
+            );
+          }
+          catch (error) {
+            if (error.message === 'strictness') {
+              addErrorMessage(field.path, errorMessages, 'The field validation is strict; it can not add unspecified sub-fields.', structuredError);
```
dbengsch: done

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+
+    if (documentSubset.hasOwnProperty(fieldName) && typeof documentSubset[fieldName] !== 'undefined') {
+      if (field.multivalued.value) {
+        if (!Array.isArray(documentSubset[fieldName])) {
+          addErrorMessage(field.path, errorMessages, 'The field must be multivalued, unary value provided.', structuredError);
+          return false;
+        }
+
+        if (field.multivalued.hasOwnProperty('minCount') && documentSubset[fieldName].length < field.multivalued.minCount) {
+          addErrorMessage(field.path, errorMessages, 'The multivalued field has not enough values.', structuredError);
```
dbengsch: done

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+};
+
+/**
+ * @param {string} indexName
+ * @param {string} collectionName
+ * @param {CollectionSpecification} collectionSpec
+ */
+Validation.prototype.validateSpecification = function (indexName, collectionName, collectionSpec) {
+  // We make a deep clone to avoid side effects
+  var specification = JSON.parse(JSON.stringify(collectionSpec));
```
scottinet: Should be avoided.

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+
+                specification[indexName][collectionName] = curatedSpec;
+
+                this.kuzzle.pluginsManager.trigger('log:info', `Validation specification for ${indexName} / ${collectionName} has been loaded.`);
+                return resolve({});
+              })
+              .catch(error => {
+                this.kuzzle.pluginsManager.trigger('log:error', `Specification for the collection ${collectionName} triggered an error`);
+                this.kuzzle.pluginsManager.trigger('log:error', `Error: ${error.message}`);
+                return resolve({});
```
dbengsch: We resolve to avoid Promise.all to reject (line 335), and doesn't prevent Kuzzle to start.

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+Validation.prototype.structureCollectionValidation = function (collectionSpec) {
+  var
+    fields = {},
+    maxDepth = 0;
+
+  Object.keys(collectionSpec.fields).forEach(fieldName => {
+    var
+      field,
+      // We deep clone the field because we will modify it
+      fieldSpecClone = JSON.parse(JSON.stringify(collectionSpec.fields[fieldName]));
```
scottinet: +serialization to be removed

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+  }
+
+  returnedTypeOptions = this.types[fieldSpec.type].validateFieldSpecification(fieldSpec.typeOptions);
+
+  if (typeof returnedTypeOptions === 'boolean') {
+    if (returnedTypeOptions) {
+      return fieldSpec;
+    }
+
+    throw new InternalError(`Field of type ${fieldSpec.type} is not specified properly`);
```
dbengsch: Jeff is working on it.

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+  var specification = JSON.parse(JSON.stringify(collectionSpec));
+
+  return new Promise (resolve => {
+    this
+      .curateCollectionSpecification(indexName, collectionName, specification, true)
+      .then(() => {
+        return resolve(true);
+      })
+      .catch(() => {
+        return resolve(false);
```
dbengsch: function renamed to `isValidSpecification`

lib/api/core/validation/types/boolean.js
dbengsch: done

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+
+/**
+ * @param {*} object
+ * @param {string[]} allowedProperties
+ * @returns {boolean}
+ */
+function checkAllowedProperties(object, allowedProperties) {
+  var result = true;
+
+  if (typeof object !== 'object') {
```
scottinet: Should also check if object is not null and not an array, as `typeof` returns `object` on both these cases

lib/api/core/validation/types/geoPoint.js
```diff
@@ -0,0 +1,40 @@
+var
+  BaseConstructor = require('../baseType'),
+  geoUtil = require('../../../dsl/geoutil');
```
scottinet: :warning: This won't work with the new DSL.
We should see eachother to settle a way to modify both the current DSL and the future one to properly expose these utility methods.
If different modules use them, then we may have to store them elsewhere in the core architecture.

lib/api/core/validation/types/geoShape.js
```diff
@@ -0,0 +1,282 @@
+var
+  BaseConstructor = require('../baseType'),
+  allowedShapeProperties = ['type', 'coordinates', 'radius', 'orientation', 'geometries'],
+  allowedOrientations = ['right', 'ccw', 'counterclockwise', 'left', 'cw', 'clockwise'],
+  mutliTypes = ['multipoint', 'multilinestring', 'multipolygon'],
+  geoUtils = require('../../../dsl/geoutil'),
```
scottinet: Same as above.

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+};
+
+/**
+ * @param {FieldSpecification} fieldSpec
+ * @returns {boolean}
+ * @throws InternalError
+ */
+Validation.prototype.curateFieldSpecificationFormat = function (fieldSpec) {
+  if (!checkAllowedProperties(fieldSpec, fieldSpecProperties)) {
+    throw new InternalError('The field specification has invalid properties.');
```
dbengsch: Jeff is working on it

lib/api/core/validation/types/anything.js
dbengsch: Because I check if each validation types overrides the 2 functions (and because I'm lazy).

lib/api/core/validation/types/enum.js
dbengsch: it was late, I had a drink ... Sorry dear it won't happen again ;).

PS: done

lib/api/core/validation/types/enum.js
dbengsch: done

lib/api/core/validation/types/geoShape.js
dbengsch: done

lib/api/core/validation/types/geoShape.js
dbengsch: It is the geoJSON spec (as well as the one of ES)

lib/api/core/validation/baseType.js
dbengsch: done

lib/api/core/validation/baseType.js
dbengsch: done

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+ * @returns {boolean}
+ */
+function checkAllowedProperties(object, allowedProperties) {
+  var result = true;
+
+  if (typeof object !== 'object') {
+    return false;
+  }
+
+  Object.keys(object).forEach(propertyName => {
```
dbengsch: done

lib/api/core/validation/types/boolean.js
dbengsch: Because I check if each validation types overrides the 2 functions (and because I'm lazy).

lib/api/core/validation/types/geoShape.js
dbengsch: To be able to report all errors

lib/api/core/validation/index.js
dbengsch: done

lib/api/core/validation/types/geoPoint.js
dbengsch: To be discussed

lib/api/core/validation/index.js
dbengsch: Done only once at startup

dbengsch:
- Fail fast when structured false during document validation
- Fail slow when structured true during document validation
- Split the document validation and the default value settings: avoids JSON.parse & JSON.stringify and the requestObject is modified only if validation succeeds (won't avoid a partial copy during an update).
- Discuss usage of geoUtils with new DSL

lib/api/core/validation/index.js
```diff
@@ -0,0 +1,788 @@
...
+ * Retrieve the plugins list from the database and returns it,
+ * along with their configuration
+ *
+ * @param kuzzle
+ * @returns {Promise}
+ */
+function getValidationConfiguration(kuzzle) {
+
+  return kuzzle.internalEngine
+    .search('validation', false, 0, 1000)
```
scottinet: Why don't use handle multi-paged results?

lib/api/core/validation/index.js
scottinet: Then why cloning it at all?

xbill82:
Fixes #483

lib/api/core/validation/types/geoShape.js
scottinet: What Benoit meant is that you're testing for a point with its first coordinate between ]-180,180[ and its second one between ]-90, 90[
This looks like a "lon, lat" point, not a "lat, lon" one.

j33f:
@scottinet can you validate the changes done and merge ?

scottinet:
I'm a bit late, but :+1:

PR #419, opened 2016-10-07
fix getConfig behavior with LB plugin

j33f:
Safer plugins config cleanup avoiding redundant and circular references

### DIFF
--- a/lib/api/controllers/adminController.js
+++ b/lib/api/controllers/adminController.js
@@ -121,15 +121,15 @@ function AdminController(kuzzle) {
 
         _.forEach(kuzzle.pluginsManager.plugins, (object, id) => {
           pluginsConfig[id] = JSON.parse(stringifySafe(object));
-          if(pluginsConfig[id].object.context && pluginsConfig[id].object.context.config) {
+          if (pluginsConfig[id].object.context && pluginsConfig[id].object.context.config) {
             // remove the kuzzle configuration which can be included into plugin context
             delete pluginsConfig[id].object.context.config;
           }
-          if(pluginsConfig[id].object.kuzzle) {
+          if (pluginsConfig[id].object.kuzzle) {
             // remove the kuzzle configuration which can be included into plugin object
             delete pluginsConfig[id].object.kuzzle;
           }
-          if(id === 'kuzzle-plugin-cluster') {
+          if (id === 'kuzzle-plugin-cluster') {
             // remove the kuzzle configuration which can be included into plugin object node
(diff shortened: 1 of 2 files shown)


lib/api/controllers/adminController.js
```diff
@@ -119,11 +120,34 @@ function AdminController(kuzzle) {
         modifiedRequestObject = newRequestObject;
 
         _.forEach(kuzzle.pluginsManager.plugins, (object, id) => {
-          pluginsConfig[id] = _.assignIn({}, object);
-          if(pluginsConfig[id].object.context && pluginsConfig[id].object.context.config) {
+          pluginsConfig[id] = JSON.parse(stringifySafe(object));
```
benoitvidis: If I understand correctly, the returned value will be a dump of the complete plugin. Shouldn't we rather return the configuration stored in `%kuzzle/plugins`?

j33f: It was my first idea but doing this way (via %kuzzle/config) you loose the eventually computed / generated configuration (for example, kuzzle-plugin-cluster have a lot of things generated in it) then you loose the context too. 
Since this route aims to give some informations for troubleshoot purposes, displaying only what is inside %kuzzle/config is not enough in my opinion.

benoitvidis: I am not very conmfortable with dumping some objects we cannot have control on.
The debug use case is already handled by the core dumps and getting the plugins configuration from the db is already a valuable information to me.

benoitvidis: I am not very conmfortable with dumping some objects we cannot have control on.
The debug use case is already handled by the core dumps and getting the plugins configuration from the db is already a valuable information to me.

j33f:
@benoitvidis all done

lib/api/controllers/adminController.js
```diff
@@ -114,27 +114,28 @@ function AdminController(kuzzle) {
...
 
         modifiedRequestObject = newRequestObject;
 
         _.forEach(kuzzle.pluginsManager.plugins, (object, id) => {
-          pluginsConfig[id] = _.assignIn({}, object);
-          if(pluginsConfig[id].object.context && pluginsConfig[id].object.context.config) {
-            // remove the kuzzle configuration which can be included into plugin context
-            delete pluginsConfig[id].object.context.config;
-          }
+          promises.push(kuzzle.internalEngine.get('plugins', id));
```
scottinet: nitpicking: `_.forEach` is not needed anymore since you're only using the object key.

I would rewrite this using a mere `Object.keys` call.

PR #422, opened 2016-10-10
Implement diagnostic tools

stafyniaksacha:
## Changes done for dump implementation:
- Add a CLI command `./bin/kuzzle dump` to generate a dump on the fly
- Add `signals` process watcher to generate dump when kuzzle crash
- Add `handledRejection` & `uncaughtException` handlers to generate dump when occurs
- Add `pendingRequests` & `requestHistory` variables into kuzzle core to have metrics on core dump
- Add a parameter to activate dumps on handled errors, with defaults:
  - `Error`
  - `RangeError`
  - `TypeError`
  - `KuzzleError`
  - `InternalError`
  - `PluginImplementationError`

### A dump contains:
- Kuzzle configuration (config.json)
- Plugins configuration (plugins.json)
- NodeJS configuration (nodejs.json)
- System configuration (os.json)
- PM2 logs stdout & stderr, which contains output for all pm2 process, even worker plugins (logs/)
- core-dump
- NodeJS binary (used to read core-dump)
- Kuzzle node stats (statistics.json)

## Changes done to improve error reporting
- Add an error handler on `redis` service: throw a `ServiceUnavailableError` when redis is not connected
- Improved error handling in `elasticsearch` & `internalEngine` services: throw a `ServiceUnavailableError` when elasticsearch is not connected

## Changes done relatively to the plugin system
- Encapsulate plugins hooks and pipes execution errors into `PluginImplementationError` object
- Set `workerPrefix` name into a setting (can be overridden by env vars)
- Improve CLI output of `kuzzle plugins --install` (more verbose)
- Improve package install output detection: now works with npm 2 & npm 3 (node 4 & node 6)
- Pakage install is always fulfilled: kuzzle will start even if a plugin installation / initialisation fail 

## Other changes:
- Implement changes due to Nodejs 6.9.1 LTS
- Update `kuzzle-plugin-logger` to v 2.0.5 (worker plugin)
- Migrate `codecov.io` package (deprecated) to `codecov`


lib/api/controllers/funnelController.js
```diff
@@ -70,13 +76,52 @@ function FunnelController (kuzzle) {
...
+                errorMessage = err.message;
+                if (errorMessage.indexOf('\n') > -1) {
+                  errorMessage = errorMessage.split('\n')[0];
+                }
+                errorMessage = errorMessage.toLowerCase().replace(/[:,'"\.\\/\s]/g, '-').replace(/[-]+/g, '-').split('-');
+                errorMessage = errorMessage.filter(value => value !== '').join('-');
+
+                request.data.body.sufix = 'handled-'.concat(errorType.toLowerCase(), '-', errorMessage);
+
+                if (!this.lastDumpedErrors[request.data.body.sufix] || this.lastDumpedErrors[request.data.body.sufix] < lastSeen - 100000) {
```
scottinet: nitpicking: can we store this `100000` number in a constant to make the code more readable, something like "MIN_TIME_BEFORE_DUMP"?

lib/api/controllers/remoteActions/dump.js
```diff
@@ -0,0 +1,108 @@
...
+    cpus: os.cpus(),
+    mem: {
+      total: os.totalmem(),
+      free: os.freemem()
+    },
+    networkInterfaces: os.networkInterfaces()
+  }, null, ' ').concat('\n'));
+
+  // dumping std err/out
+  console.log('> dumping std err/out');
```
scottinet: can't we have both stderr and stdout logs?

lib/api/kuzzle.js
```diff
@@ -157,6 +161,71 @@ Kuzzle.prototype = new EventEmitter({
...
+    // SIGBUS: 10, can not be handled
+    SIGTERM: 15,
+  };
+  var request = new RequestObject({
+    controller: 'actions',
+    action: 'dump',
+    body: {}
+  });
+
+  process.removeAllListeners('unhandledRejection');
```
scottinet: Why do we remove all these listeners before registering our own? Are listeners pre-registered before Kuzzle starts?

(this question applies to all `removeAllListeners` of this method)

lib/services/internalEngine/index.js
```diff
@@ -83,6 +84,13 @@ function InternalEngine (kuzzle) {
         }
 
         return result;
+      })
+      .catch(error => {
+        if (error instanceof Elasticsearch.errors.NoConnections) {
```
scottinet: Since this code portion is repeated for all ES commands invoked, wouldn't it be better to store it in a function and calling it for each catch?

Something like:

```js
  .catch(error => handleError(error));
```

With:

```js
function handleError(error) {
  if (error instanceof Elasticsearch.errors.NoConnections) {
    return Promise.reject(new ServiceUnavailableError('Elasticsearch service is not connected'));
  }

  return Promise.reject(error);
}
```

lib/api/controllers/remoteActions/dump.js
scottinet: As seen together, this comment can be dismissed :-)

j33f: I eared the discussion, I agree :)

lib/api/kuzzle.js
stafyniaksacha: This is due to PM2 which will register theses listeners to do some works, and then exit the process, so we are unable to do anything.

Removing theses listeners does not affect PM2 restarting nor restart count values

scottinet: Thanks, it's clearer now. Could you add a comment explaining this?

stafyniaksacha: Yup, done

lib/api/controllers/remoteActions/dump.js
```diff
@@ -0,0 +1,108 @@
...
+function dump (request) {
+  var sufix = request && request.data.body.sufix ? '-' + request.data.body.sufix: '';
+  var dumpPath = path.join(
+    path.normalize(_kuzzle.config.dump.path),
+    moment().format(_kuzzle.config.dump.dateFormat).concat(sufix)
+  );
+
+  console.log('===========================================================================');
+  console.log('Generating dump in '.concat(dumpPath));
+  mkdirp.sync(dumpPath);
```
benoitvidis: We need to document the fact the user running Kuzzle needs some write permissions on this directory.

stafyniaksacha:
Blocked: MDB V8 module doesn't support V8 5.x yet

PR #423, opened 2016-10-10
Add _kuzzle_info to document

jenow:
Add a _kuzzle_info property to all document which contains an active attribute to handle a trash mode
See https://confluence.kaliop.net/display/LABS/2016-02-26+-+Delete+notifications

:warning: This PR does not change the active attribute to false, it still deletes the document. This is part of the second task (see https://jira.kaliop.net/browse/KUZ-639?jql=project%20%3D%20KUZ%20AND%20text%20~%20metadata for more)


lib/services/elasticsearch.js
```diff
@@ -184,6 +196,18 @@ function ElasticSearch (kuzzle, options, config) {
     }
 
     return this.client.get(data)
+      .then(result => {
+        if (result._source._kuzzle_info && !result._source._kuzzle_info.active) {
+          // todo Feedback how to get it from the 'trash' once it is implemented
+          return Promise.reject(new NotFoundError('Document has been deleted'));
```
stafyniaksacha: could be nice if we have the document id in the error message !

lib/services/elasticsearch.js
```diff
@@ -251,11 +277,51 @@ function ElasticSearch (kuzzle, options, config) {
     var
       data = cleanData.call(this, requestObject);
 
-    // extends the response with the source from requestObject
-    // When we write in ES, the response from it doesn't contain the initial document content
+    // Add metadata
+    data.body._kuzzle_info = {
+      author: requestObject.userId ? String(requestObject.userId) : null,
```
stafyniaksacha: Why did you force the cast of ```requestObject.userId``` ?

lib/services/elasticsearch.js
```diff
@@ -251,11 +277,51 @@ function ElasticSearch (kuzzle, options, config) {
...
+      createdAt: new Date(),
```
stafyniaksacha: ```new Date()``` will return a date formatted with local time (ex: ```Mon Oct 17 2016 14:34:29 GMT+0000 (UTC)```) 

I'll suggest to use ```Date.now()``` function instead.
Or define mapping to recognize a specific date format and format date field with momentjs: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html 
http://momentjs.com/docs/#/displaying/format/

lib/services/elasticsearch.js
```diff
@@ -268,8 +334,15 @@ function ElasticSearch (kuzzle, options, config) {
     var
       data = cleanData.call(this, requestObject);
 
-    // extends the response with the source from requestObject
-    // When we write in ES, the response from it doesn't contain the initial document content
+    // Add metadata
+    data.body._kuzzle_info = {
+      author: requestObject.userId ? String(requestObject.userId) : null,
```
stafyniaksacha: Same as above:

Why did you force the cast of ```requestObject.userId``` ?

lib/services/elasticsearch.js
```diff
@@ -286,6 +359,12 @@ function ElasticSearch (kuzzle, options, config) {
     var
       data = cleanData.call(this, requestObject);
 
+    // Add metadata
+    data.body._kuzzle_info = {
+      updatedAt: new Date(),
+      updater: requestObject.userId ? String(requestObject.userId) : null
```
stafyniaksacha: Same as above: 

Why did you force the cast of ```requestObject.userId``` ?

lib/services/elasticsearch.js
```diff
@@ -268,8 +334,15 @@ function ElasticSearch (kuzzle, options, config) {
     var
       data = cleanData.call(this, requestObject);
 
-    // extends the response with the source from requestObject
-    // When we write in ES, the response from it doesn't contain the initial document content
+    // Add metadata
+    data.body._kuzzle_info = {
+      author: requestObject.userId ? String(requestObject.userId) : null,
+      createdAt: new Date(),
```
stafyniaksacha: Same as above: 

```new Date()``` will return a date formatted with local time (ex: ```Mon Oct 17 2016 14:34:29 GMT+0000 (UTC)```) 

I'll suggest to use ```Date.now()``` function instead.
Or define mapping to recognize a specific date format and format date field with momentjs: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html 
http://momentjs.com/docs/#/displaying/format/

File: lib/services/elasticsearch.js [hunk 7]
```diff
@@ -286,6 +359,12 @@ function ElasticSearch (kuzzle, options, config) {
     var
       data = cleanData.call(this, requestObject);
 
+    // Add metadata
+    data.body._kuzzle_info = {
+      updatedAt: new Date(),
```
stafyniaksacha: Same as above: 

```new Date()``` will return a date formatted with local time (ex: ```Mon Oct 17 2016 14:34:29 GMT+0000 (UTC)```) 

I'll suggest to use ```Date.now()``` function instead.
Or define mapping to recognize a specific date format and format date field with momentjs: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html 
http://momentjs.com/docs/#/displaying/format/

File: lib/services/elasticsearch.js [hunk 8]
```diff
@@ -308,6 +387,14 @@ function ElasticSearch (kuzzle, options, config) {
         id: data.id
       };
 
+    // Add metadata
+    data.body._kuzzle_info = {
+      author: requestObject.userId ? String(requestObject.userId) : null,
+      createdAt: new Date(),
```
stafyniaksacha: Same as above: 

```new Date()``` will return a date formatted with local time (ex: ```Mon Oct 17 2016 14:34:29 GMT+0000 (UTC)```) 

I'll suggest to use ```Date.now()``` function instead.
Or define mapping to recognize a specific date format and format date field with momentjs: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html 
http://momentjs.com/docs/#/displaying/format/

File: lib/services/elasticsearch.js [hunk 9]
```diff
@@ -308,6 +387,14 @@ function ElasticSearch (kuzzle, options, config) {
         id: data.id
       };
 
+    // Add metadata
+    data.body._kuzzle_info = {
+      author: requestObject.userId ? String(requestObject.userId) : null,
```
stafyniaksacha: Same as above: 

Why did you force the cast of ```requestObject.userId``` ?

lib/services/elasticsearch.js
jenow: Because if the first time we put an integer we wouldn't be able to put a string next time so I prefer to normalize it with a string

xbill82:
This PR references a document from our internal Confluence space, which is not public. This flow looks "broken" to me... IMHO we should make these documents public. GitHub offers a per-repo Wiki space we could use to host this content. Just suggesting :arrow_right: we may discuss this item...

test/services/implementations/elasticsearch.test.js
```diff
@@ -806,13 +930,17 @@ describe('Test: ElasticSearch service', () => {
...
+      };
 
-      sandbox.stub(elasticsearch.client.indices, 'getMapping').resolves(mappings);
+      result[index] = {mappings};
+
+      sandbox.stub(elasticsearch.client.indices, 'getMapping').resolves(result);
 
       return should(elasticsearch.getMapping(requestObject)
-        .then(result => {
+        .then(() => {
```
dbengsch: The result of the function is not tested anymore

lib/services/elasticsearch.js
```diff
@@ -782,6 +879,42 @@ function getAllIdsFromQuery(data) {
 }
 
 /**
+ * Add filter to get only the active documents
+ * @param data
+ */
+function addActiveFilter(data) {
+  var inputFilter = (data.body ? data.body.query : null);
+
+  data.body = {
```
stafyniaksacha: Changed `must` term to `filter`

test/services/implementations/elasticsearch.test.js
stafyniaksacha: Done

PR #427, opened 2016-10-17
Kuz 694 install on first start

benoitvidis:
This PR removes the need to first launch the `kuzzle install` command to bootstrap Kuzzle.

Due to some architectural changes that were not reported to the install script, Kuzzle's internal organisation has been modified in a few places.

## Remote actions

Historically, the install command was passing through the remote actions controller. Now that we are triggering the installation from Kuzzle main process, the whole install process can be run from Kuzzle itself.

### cleanDb & cleanAndPrepare

`cleanAndPrepare` action has been removed. `cleanDb` now triggers a new bootstrap call after it erases the data.

### prepareDb

The `prepareDb` action is not in charge of bootstrapping Kuzzle anymore. It only deals with importing some fixtures and/or some mappings.

As such, it is renamed to `data`.

### ManagePlugins

Now only performs actions triggered from the CLI. Relies heavily on the new `pluginsManager.packages` component (cf below).

### services

Remaining unused remote actions to manage services are removed.

### Internal architecture

A new `packages` component has been added to Kuzzle `pluginsManager`. The `packages` components is in charge of the plugins installation / removal and their configuration updates.

### Functional changes

* A plugin marked a 'deleted' in the database but configured in .kuzzlerc will now not be installed during bootstrap.
* The installation now uses npm for all installation modes. This leads to the following changes:
  * plugins installed from a local path will now be copied into the `node_modules` directory
  * :warning: `npmVersion` and `gitUrl` are deprecated (but still compatible), in favor or `version` and `url`. This allows to set a git commit/tag/branch and to install a plugin from a remote tarball url.
  * The option to install `all` plugins has been removed.
  * The plugins remote action timeout is increased to 5mn to leave npm some time to install modules. (may need to be even increased). A basic progress bar is now displayed on the CLI during plugins operations.

## Elasticsearch bootstrap

A new `bootstrap` component has been added to Kuzzle `internalEngine`. It is responsible of creating Kuzzle internal index and collections if needed.


benoitvidis:
wip: plugin install from local path should set the plugin name to the on defined in packages.json

benoitvidis:
wip: install command does not accept `version` and `url` options

test/api/remoteActions/action.test.js
```diff
@@ -133,16 +133,25 @@ describe('Tests: remoteActionController action client', () => {
 
     it('should start the timer', () => {
       return Action.__with__({
-        setTimeout: sinon.spy()
+        setTimeout: sinon.spy(),
+        'Function.prototype.bind': sinon.spy(Function.prototype.bind)
```
benoitvidis:
```javascript
let fn = function () {};
let o = {}
let f1 = fn.bind(o);
let f2 = fn.bind(o);

assert(f1 === f2) 
// false
```

bin/kuzzle
```diff
@@ -55,8 +49,8 @@ program
   .option('    --set <JSONObject>', 'merge current plugin configuration with JSONObject')
   .option('    --unset <property>', 'unset property from the plugin configuration')
   .option('    --replace <JSONObject>', 'erase the plugin configuration and apply JSONObject instead')
-  .option('-v, --npmVersion <version>', 'install plugin <version> from npm')
-  .option('-u, --gitUrl <url>', 'install plugin from a git repository')
+  .option('-v, --packageVersion <version>', 'install plugin <version> from npm')
```
benoitvidis: note: cannot use `version` here as it is swallowed by the main program.

benoitvidis:
wip: plugin set needs to be fixed

bin/kuzzle
```diff
@@ -46,7 +40,7 @@ program
   .arguments('[name]')
   .description('manage plugins')
   .option('    --list', 'list currently installed plugins')
-  .option('    --install', 'if plugin [name] is provided, install it from --npmVersion, --gitUrl or --path, otherwise, (re-)install all listed plugins')
+  .option('    --install', 'if plugin [name] is provided, install it from --version, --url or --path, otherwise, (re-)install all listed plugins')
```
dbengsch: packageVersion ?

xbill82:
@benoitvidis shall we add the "doc required" label?

PR #428, opened 2016-10-17
Kuz 700 Validation routes

j33f:
Implements 
- GET:/:index/:collection/_specifications to get a collection validation specifications
- PUT:/_specifications to set/update specifications
- POST:/_validateSpecifications to validate specifications format
- POST:/:index/:collection/_validate to validate a document against the current collection validation specifications
- DELETE:/:index/:collection/_specifications to delete the current collection validation specifications


j33f:
Pending questions : 
- I am not sure about the routes names and syntax
- Do we need to test each types and variants in functional tests ?

lib/api/controllers/adminController.js
```diff
@@ -51,6 +53,107 @@ function AdminController(kuzzle) {
   };
 
   /**
+   * Get the collection validation specifications
+   * @param {RequestObject} requestObject
+   * @returns {Promise}
+   */
+  this.getSpecifications = requestObject => {
```
scottinet: Doesn't this break the new eslint policy about unnamed function?
If not, can we easily retrieve the corresponding code from a coredump if this function causes it (or is in the stack trace), or is it listed as an "anonymous function"?

@stafyniaksacha > what do you think about it?

lib/api/controllers/adminController.js
```diff
@@ -51,6 +53,107 @@ function AdminController(kuzzle) {
...
+      .then(newRequestObject => {
+        modifiedRequestObject = newRequestObject;
+
+        return prepareSpecificationValidation(modifiedRequestObject, kuzzle);
+      })
+      .then(response => {
+        var promises = [];
+        if (response.error) {
+          kuzzle.pluginsManager.trigger('log:error', `Validation specification has not been updated because of an error: ${requestObject.data.body}.`);
+          return Promise.resolve(response.responseObject);
```
scottinet: Shouldn't this be `Promise.reject(...)`?

lib/api/controllers/adminController.js
```diff
@@ -51,6 +53,107 @@ function AdminController(kuzzle) {
...
+        }
+        response.specifications.forEach(specification => {
+          var specLogName = specification._id.split('#').join(' / ');
+          promises.push(internalEngine.createOrReplace('validations', specification._id, specification._source));
+          kuzzle.pluginsManager.trigger('log:info', `Validation specification for ${specLogName} is about to be stored.`);
+        });
+
+        return Promise.all(promises)
```
scottinet: Nitpicking: can be flattened 

```js
  ...
  return Promise.all(promises);
})
.then(() => kuzzle.internalEngine.refresh())
.then(() => kuzzle.validation.curateSpecification())
.then(() => ...
```

Also, I don't see the point of:
`.then(() => modifiedRequestObject.data.body)`
(well, I understand why you done this like that, but I think that we should reject the promise if an error occurs)

lib/api/controllers/adminController.js
```diff
@@ -51,6 +53,107 @@ function AdminController(kuzzle) {
...
+
+    return kuzzle.pluginsManager.trigger('data:beforeValidateSpecifications', requestObject)
+      .then(newRequestObject => {
+        modifiedRequestObject = newRequestObject;
+
+        return prepareSpecificationValidation(modifiedRequestObject, kuzzle);
+      })
+      .then(response => {
+        if (response.error) {
+          return Promise.resolve(response.responseObject);
```
scottinet: Same here: why don't we reject the promise in case of error?

lib/api/controllers/adminController.js
```diff
@@ -398,4 +501,68 @@ function resetProfiles () {
...
+      var responseObject;
+
+      if (response === true) {
+        return Promise.resolve({error: false, specifications: specifications});
+      }
+
+      responseObject = new ResponseObject(requestObject, response);
+      responseObject.error = 'Some errors with provided specifications.';
+      responseObject.status = 400;
+      return Promise.resolve({error: true, responseObject: responseObject});
```
scottinet: Or maybe reject the promise here so we don't have to check for an error in code using this method.

lib/api/controllers/writeController.js
```diff
@@ -205,6 +205,33 @@ function WriteController (kuzzle) {
...
+        return kuzzle.validation.validationPromise(modifiedRequestObject, true);
+      })
+      .then(response => {
+        var responseObject = new ResponseObject(modifiedRequestObject, response);
+
+        if (!response.validation) {
+          responseObject.status = 400;
+          responseObject.error = response.errorMessages;
+        }
+        kuzzle.pluginsManager.trigger('log:error', `The document does not comply with the ${requestObject.index} / ${requestObject.collection} : ${JSON.stringify(requestObject.data.body)}`);
```
scottinet: If there is a lot of activity, this can seriously hurt performances, along with clogging up log files. I think this should be silently rejected (except, of course, for a response to the user), and maybe we can add an event if someone want to listen to it (and configure the logger if one wants to)

lib/api/core/validation/index.js
```diff
@@ -426,12 +440,24 @@ Validation.prototype.curateCollectionSpecification = function validationCurateCo
...
+        return resolve({isValid: false, errors: [errorMessage]});
+      }
+      return reject(new InternalError(errorMessage));
     }
 
     if (collectionSpec.fields) {
       try {
-        treatedSpecification.fields = this.structureCollectionValidation(collectionSpec);
+        result = this.structureCollectionValidation(collectionSpec, indexName, collectionName, verboseErrors);
+        if (verboseErrors) {
```
scottinet: can be simplified:

```js
if (verboseErrors && result.isValid === false) {
  return reject(result);
}

treatedSpecification.fields = result;
```

lib/api/controllers/adminController.js
j33f: ... and you understood it :) (the reason why I did not reject the promise)

You may be right

lib/api/core/validation/index.js
j33f: yep, forgot to refactor this after debugging...

lib/api/controllers/adminController.js
scottinet: Which one?

lib/api/controllers/writeController.js
scottinet: I think that rejecting documents because they are not valid is not an error nor a warning, it's just normal activity to me.
If you insist on logging it, then I wouldn't use anything higher than `log:silly`, as this log can be very cost-heavy on large applications.

j33f:
you're right @benoitvidis I'm not a big fan of "specifications" word too, but I'm lacking of imagination to rename it...

lib/api/kuzzle.js
```diff
@@ -22,6 +22,31 @@ var
   TokenManager = require('./core/auth/tokenManager'),
   Validation = require('./core/validation');
 
+/** @typedef {{
```
dbengsch: This typedef is outdated since the config file refactor.

lib/api/kuzzle.js
```diff
@@ -110,9 +133,9 @@ function Kuzzle () {
     return this.internalEngine.init()
       .then(() => this.internalEngine.bootstrap.all())
       .then(() => this.pluginsManager.packages.bootstrap())
-      .then(() => this.validation.init())
```
dbengsch: validation.init should be done before the the pluginManager.init as plugins can add types and pre-defined types should be loaded first.

PR #429, opened 2016-10-17
Remove all anonymous functions

stafyniaksacha:
* Remove all anonymous functions (needed to inspect stack traces)
* Add ```func-name``` eslint rule


lib/api/controllers/routerController.js
```diff
@@ -145,7 +144,7 @@ function RouterController (kuzzle) {
      * @param request
      * @param response
      */
-    handleRoute = function(request, response) {
+    function handleRoute (request, response) {
```
scottinet: Function declarations inside blocks are invalid according to ECMA specifications.
Consider assigning this one to a variable or, better yet, promote this snippet of code as a real function of its own.

lib/api/controllers/routerController.js
```diff
@@ -276,7 +275,7 @@ function executeFromRest(params, request, response) {
     delete request.params.index;
   }
 
-  _.forEach(request.params, function (value, param) {
+  _.forEach(request.params, function bindParams (value, param) {
```
scottinet: Is it necessary to name this one ? Shouldn't we convert it to an arrow function instead?

Moreover, it may be the right time to get rid of this unnecessary call to lodash, replacing it with something like this:

```js
Object.assign(request.body, request.params);
```

lib/api/controllers/writeController.js
```diff
@@ -88,11 +88,11 @@ function WriteController (kuzzle) {
    * @param {RequestObject} requestObject
    * @returns {Promise}
    */
-  this.update = function (requestObject) {
+  this.update = function writeUpdate (requestObject) {
     var modifiedRequestObject = null;
 
     return requestObject.isValid()
-      .then(function () {
+      .then(() => {
```
scottinet: nitpicking: can be simplified using implicit return:

```js
  .then(() => kuzzle.pluginsManager.trigger('data:beforeUpdate', requestObject))
```

lib/api/controllers/writeController.js
```diff
@@ -110,11 +110,11 @@ function WriteController (kuzzle) {
    * @param {RequestObject} requestObject
    * @returns {Promise}
    */
-  this.replace = function (requestObject) {
+  this.replace = function writeReplace (requestObject) {
     var modifiedRequestObject = null;
 
     return requestObject.isValid()
-      .then(function () {
+      .then(() => {
```
scottinet: nitpicking: can be simplified using implicit return

PR #488, opened 2016-10-21
Kuz-641 Document garbage collector

jenow:


lib/services/elasticsearch.js
```diff
@@ -561,6 +564,41 @@ function ElasticSearch(kuzzle, options, config) {
...
+          });
+        });
+      });
+  };
+
+  /**
+   * Delete all document that match the given filter from the trash
+   * @param requestObject
+   */
+  this.deleteByQueryFromTrash = function elasticsearchDeleteByQueryFromTrash (requestObject) {
```
xbill82: Shouldn't this function ensure that only documents from the trash (i.e. with `active` meta-attr set to `false`) are matched?

jenow: This is used in the `GarbageCollector` which add the filter active to the RequestObject.
Maybe we will have to refactor it once we add a route or something to manually delete document from the trash

xbill82: :thinking: So, you mean that the GarbageCollector is responsible of ensuring that only the documents from the trash are deleted, which is :ok: to me. But, in this case, I think this function should be named `deleteByQuery` rather that `deleteByQueryFromTrash` (since the current name suggests that this particular function is in charge of deleting documents from the trash, which is not the case).

PR #489, opened 2016-10-24
fix #427 Options wording

benoitvidis:
cd @dbengsch comment in #427


bin/kuzzle
```diff
@@ -40,16 +40,16 @@ program
   .arguments('[name]')
   .description('manage plugins')
   .option('    --list', 'list currently installed plugins')
-  .option('    --install', 'if plugin [name] is provided, install it from --version, --url or --path, otherwise, (re-)install all listed plugins')
-  .option('    --remove', 'removes plugin [name] from Kuzzle')
-  .option('    --activate', 'mark the plugin as "activated" (Kuzzle ignores deactivated plugins)')
-  .option('    --deactivate', 'mark the plugin as "deactivated" (Kuzzle ignores deactivated plugins)')
-  .option('    --importConfig <file>', 'import a configuration from a file for a given plugin')
+  .option('    --install', 'install a new plugin package from npm repository if [name] if provided or from --path or --url options')
```
dbengsch: typo ?

dbengsch: misses --packageVersion ?

xbill82: As per https://github.com/kuzzleio/kuzzle/blob/develop/lib/api/core/plugins/packages/pluginPackage.js#L21 I don't see any `packageVersion` option available. AFAIK, only `version` is allowed, but not mandatory.

xbill82: Wow, ok, sorry, I'm wrong. The option (for the CLI command) is actually `packageVersion` (translated to `version` [here](https://github.com/kuzzleio/kuzzle/blob/develop/bin/commands/plugins.js#L28)). But, still, I don't think it's mandatory.

benoitvidis: It is not. 

* It cannot be used from a static source (directory or tarball).
* When installing from npm repository, if `--packageVersion` is not given, the last version is installed
* When installing from git, the `master` branch is taken as default.

bin/kuzzle
```diff
@@ -40,16 +40,16 @@ program
...
+  .option('    --remove', 'remove plugin [name] from Kuzzle')
+  .option('    --activate', 'mark the plugin as "activated"')
+  .option('    --deactivate', 'mark the plugin as "deactivated"')
+  .option('    --importConfig <file>', 'import plugin [name] configuration from <file>')
   .option('    --get', 'get plugin [name] configuration stored in Kuzzle')
   .option('    --set <JSONObject>', 'merge current plugin configuration with JSONObject')
```
stafyniaksacha: should be:
`--set <JSONObject> [name]`

benoitvidis: Indeed. Fixed.

Issue #434, opened 2016-10-18
Add some https / wss support in Kuzzle docker-compose sample stack


benoitvidis:
There are numerous cases where we would benefit from having Kuzzle run in https mode.

One of these is when the back office is hosted on some https server. If so, it won't be allowed by the browser to access a Kuzzle running in http only.

The docker-compose could be improved by adding a haproxy server that would be in charge of the SSL termination.

PR #492, opened 2016-10-24
Feature 434 : Add a docker-compose example with SSL support

dbengsch:


dbengsch:
fix #434

docker-compose/haproxy/haproxy.cfg
```diff
@@ -0,0 +1,62 @@
...
+    timeout tunnel      2h
+
+frontend kuzzle_https
+    bind *:7511 ssl crt /etc/ssl/private/snakeoil.pem
+    default_backend https
+    mode        http
+    option      httpclose
+    option      forwardfor
+    tcp-request inspect-delay 500ms
+    tcp-request content accept if HTTP
```
ballinette: Not really needed for Kuzzle implementation, but the protocol convention "requires" to add a header "X-Forwarded-Proto" when the client's request is in HTTPS:
`   reqadd X-Forwarded-Proto:\ https`

PR #494, opened 2016-10-24
Fix #443 disable _routing parameter in creation request

### LINKED ISSUE #443: Prevent the usage of _routing of ES in Kuzzle
The _routing parameter in Elasticsearch allows to force the usage of only one shard. This behavior could lead to security issues. We have to check if it can be injected somehow from a client request, and if it is, disallow it.

dbengsch:
fix #443


ballinette:
:warning: do you really want to merge this PR on `master` branch, and not `develop` ?

dbengsch:
no sorry, bad branch

PR #496, opened 2016-10-24
Fix #443 disable _routing parameter in creation request

### LINKED ISSUE #443: Prevent the usage of _routing of ES in Kuzzle

dbengsch:
fix #443


dbengsch:
@xbill82 : it doesn't represent a consistency issue. The user can specify _routing in a search in any portion of a filter, it would be difficult to dig it out...

scottinet:
Is Kuzzle's controllers the right place for such a check?
Since it seems to be ES specific, wouldn't it be better to change ES service instead?

dbengsch:
@scottinet @benoitvidis : agreed, transferred the _routing checks in the ES service

PR #500, opened 2016-10-25
Fix #499 - Disallow multi-index search

### LINKED ISSUE #499: Disallow multiple index search to prevent indirect search on internal index
If multiple indexes are specified are specified separated by a comma, it allows to search toward multiple indexes. A policy with no restriction on search would allow to search on the internal index

dbengsch:
fix #499

### DIFF
--- a/lib/api/controllers/readController.js
+++ b/lib/api/controllers/readController.js
@@ -15,12 +15,13 @@ function ReadController (kuzzle) {
   this.search = function readSearch (requestObject) {
     var modifiedRequestObject = null;
 
-    return requestObject.isValid()
-      .then(() => {
-        if (requestObject.index && typeof requestObject.index === 'string' && requestObject.index.split(',').length > 1) {
-          throw new BadRequestError('Search on multiple indexes is not available.');
-        }
-      })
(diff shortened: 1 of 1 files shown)


dbengsch:
waiting #502 to be merged properly

lib/api/controllers/readController.js
```diff
@@ -15,7 +15,16 @@ function ReadController (kuzzle) {
   this.search = function readSearch (requestObject) {
     var modifiedRequestObject = null;
 
-    return kuzzle.pluginsManager.trigger('data:beforeSearch', requestObject)
+    return new Promise((resolve, reject) => {
+      if (requestObject.index && typeof requestObject.index === 'string' && requestObject.index.split(',').length > 1) {
+        reject(new BadRequestError('Search on multiple indexes is not available.'));
```
scottinet: Since rejecting does not involve breaking the code flow, then in that case, the promise is rejected AND fulfilled.

Should be:
`return reject(new BadRequestError('Search on multiple indexes is not available.'));`

dbengsch: Right !

PR #502, opened 2016-10-26
New dependency version breaks unit tests

dbengsch:
- New version 0.7.2 of ms breaks unit tests.
- Fixing package version to 0.7.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "jsonwebtoken": "^7.1.7",
     "kuzzle-common-objects": "^1.0.2",
     "lodash": "4.16.3",
-    "ms": "^0.7.1",
+    "ms": "0.7.1",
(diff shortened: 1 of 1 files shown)


scottinet:
wouldn't it be better to upgrade our code along with the version of `ms`?

xbill82:
Oh, I fixed the problem with `ms` in this commit https://github.com/kuzzleio/kuzzle/pull/498/commits/23735918da6c716c3e19c7c286f844f3d7eae732, part of the #498 

Please, let me know if this solution is suitable. I would rather agree with @scottinet about keeping our code safe with regard of the minor evolutions of out dependencies.

dbengsch:
As discussed I would rather use shrinkwrap to fix the dependencies but this PR is indeed useless. I close it.

PR #508, opened 2016-10-28
Use the stub Dsl.validate in validation to prepare Dsl change

dbengsch:


scottinet:
(thanks, this will make merging #510 easier!)

PR #516, opened 2016-11-02
Implement elasticsearch scroll api method

### LINKED ISSUE #433: Expose ES scroll method in read controller
For the time being, Kuzzle only exposes ES `search` method.

Users that need a consistent full data set have the only choice of send a `search` call with some very high size.
This leads to some high network & memory usage and is not reliable.

Implementing the `scroll` method should tackle this use case.

stafyniaksacha:
fix #433


dbengsch:
did merge, needs doc

lib/api/controllers/readController.js
```diff
@@ -22,16 +22,32 @@ function ReadController (kuzzle) {
...
         modifiedRequestObject = newRequestObject;
         return engine.search(modifiedRequestObject);
       })
       .then(response => kuzzle.pluginsManager.trigger('data:afterSearch', new ResponseObject(modifiedRequestObject, response)));
   };
 
+  this.scroll = function readScroll (requestObject) {
+    var modifiedRequestObject = null;
+
+    return new Promise((resolve, reject) => {
```
scottinet: Why creating a promise? Wouldn't it be simpler to write something like this?

```js
  if (requestObject.index && typeof requestObject.index === 'string' && requestObject.index.split(',').length > 1) {
     return Promise.reject(new BadRequestError('Scroll on multiple indexes is not available.'));
  }

  return kuzzle.pluginsManager.trigger('data:beforeScroll', requestObject))
    .then(newRequestObject => ...
```

PR #518, opened 2016-11-04
Add from and size parameters in the listCollections route

### LINKED ISSUE #440: Add pagination to read::listCollections
Add "size" & "from" options to listCollections

dbengsch:
fix #440 

Response format has changed.
Previously: 
```javascript
{
  collections: {
    realtime: [
      'realtimeCollection',
      '...'
    ],
    stored: {
      'storedCollection',
      '...'
    }
  }
  type: 'all'
}
```
Now:
```javascript
{
  collections: [
    {name: 'realtimeCollection', type: 'realtime'},
    {name: 'storedCollection', type: 'stored'},
    {...}
  }
  type: 'all'
  from: 0,
  size: 12
}
```

Collections are sorted alphanumerically. `from` and `size` appear in the response only if they have been used in the request.


lib/api/controllers/readController.js
```diff
@@ -83,16 +84,21 @@ function ReadController (kuzzle) {
           .map(collection => collection.name);
 
         if (type === 'realtime') {
-          return kuzzle.pluginsManager.trigger('data:afterListCollections', new ResponseObject(modifiedRequestObject, {type, collections: {realtime: realtimeCollections}}));
+          let realtimeResponse = {type, collections: {realtime: realtimeCollections}};
```
benoitvidis: Have we already all agreed node 6 is a strong requirement?

dbengsch: @benoitvidis I think we did already.

dbengsch: @benoitvidis added a 'use strict'

PR #519, opened 2016-11-04
[develop] Fix router plugins routes

j33f:
Plugins routes were not usable since the router have a route of type `/:index/:collection/:id` which is evaluated before all the `/_plugin` prefixed routes


j33f:
looks like it fixes the problem with get routes, but introduce a bug with post ones...

ScreamZ:
Can you please ping me on release (on develop / master),

Thanks

j33f:
close this PR when @scottinet has finished his work on http router

PR #520, opened 2016-11-04
[Hotfix] [Master] allow plugins routes to be used by the router

j33f:
Plugins routes were not usable since the router have a route of type /:index/:collection/:id which is evaluated before all the /_plugin prefixed routes


j33f:
looks like it fixes the problem with get routes, but introduce a bug with post ones...

j33f:
useless since @scottinet's work and future PR

PR #521, opened 2016-11-06
Repair plugin routes loader

ScreamZ:
Hi,

just spotted a bug while loading routes with GET verb. It was leading to an error instead of searching in plugins routes...

This repaired on my instance, and I don't expect any border effect ?

Regards


ScreamZ:
I've found an issue, it makes GET work again, but can't send JSON body anymore with POST verb... If I swap again line, I can send JSON, but I can't use get (Everything goes well with websocket API, it's only related to REST)

stafyniaksacha:
closed due to duplication with #519

PR #523, opened 2016-11-07
fixes #452 - index & collection exist routes

### LINKED ISSUE #452: Index & collection 'exists' route
This will be useful for the Backoffice, which currently has no way to efficiently determine whether the collection/indes exists once the user landed on it via the URL.

**TODO** Add the methods to the SDKs

benoitvidis:


lib/api/controllers/readController.js
```diff
@@ -97,20 +108,31 @@ function ReadController (kuzzle) {
       });
   };
 
-  this.now = function readNow (requestObject) {
-    return kuzzle.pluginsManager.trigger('data:beforeNow', requestObject)
-      .then(newRequestObject => kuzzle.pluginsManager.trigger('data:afterNow', new ResponseObject(newRequestObject, {now: Date.now()})));
+  this.collectionExists = requestObject => {
```
j33f: should use the function keyword with a proper name in order to have it in debug

lib/api/controllers/readController.js
```diff
@@ -97,20 +108,31 @@ function ReadController (kuzzle) {
...
+      .then(newRequestObject => {
+        modifieRequestObject = newRequestObject;
+        return engine.collectionExists(modifieRequestObject);
+      })
+      .then(response => kuzzle.pluginsManager.trigger('data:afterCollectionExists', new ResponseObject(modifieRequestObject, response)));
   };
 
-  this.listIndexes = function readListIndexes (requestObject) {
-    var modifiedRequestObject = null;
+  this.indexExists = requestObject => {
```
j33f: ditto

lib/api/controllers/readController.js
stafyniaksacha: +1

PR #524, opened 2016-11-08
update tests to pass on both node 6.x and node 4.x

stafyniaksacha:


scottinet:
I need this PR now. Considering the change, I'm merging it right away. :-)

PR #530, opened 2016-11-10
set NODE_ENV to development in dev & test images / production in others images

stafyniaksacha:
Improve stacktraces on development & tests images

A lot of warning is raised now when doing bad things with promises :sunglasses:

Improve performance on standards images due to env var NODE_ENV=production


docker-compose/test.yml
```diff
@@ -22,6 +22,7 @@ services:
       - kuzzle_services__internalCache__node__host=redis
       - kuzzle_services__memoryStorage__node__host=redis
       - kuzzle_services__proxyBroker__host=proxy
+      - NODE_ENV=development
```
dbengsch: Wouldn't it be better for travis to be in production mode ?

lib/api/controllers/adminController.js
```diff
@@ -116,6 +116,8 @@ function AdminController(kuzzle) {
       })
       .then(() => {
         kuzzle.pluginsManager.trigger('log:info', `Validation specification for ${modifiedRequestObject.index}#${modifiedRequestObject.collection} has been deleted.`);
+
+        return null;
```
stafyniaksacha: As of bluebird/promises specifications, if we dont care about the result of a promise, we need to use `return null` to let promises run in background

In this case, we don't care if the `log:info` event fail.
So yes, `return null` is better here

j33f: ok ! thx

PR #541, opened 2016-11-18
HTTP Performances

scottinet:
:warning: Needs https://github.com/kuzzleio/kuzzle-proxy/pull/31 to be merged before automated tests on this PR can pass

# HTTP Performances 

* Remove HTTP server from Kuzzle server instance:
  * Kuzzle Server now listens to `httpRequest` internal broker room
  * the request size check is now performed by Kuzzle Proxy
  * HTTP configuration removed from Kuzzle server configuration files
  * Does not respond directly to clients anymore, but instead sends a `HttpResponse` object to the proxy, using `httpResponse` internal broker room
* Replace Express Router with a custom and much faster one:
  * index routes using a tree instead of regular expressions
  * now throws an `InternalError` when attempting to add a route conflicting with another one
  * route insertion order does not matter anymore

# HTTP Routes

* fix HTTP route /ms/geodist (was improperly named /ms/goedist)
* removed duplicate `/ms/_hdel/` route
* when defining HTTP routes, `id` is no longer silently converted to `_id`. 
* due to the new router no longer using regular expressions, the PUT route `/:index/:collection/:_id/_:action` has been replaced by explicit route definitions

# Other changes 

* Remove functional testing coverage tests, as `istanbul-middleware` explicitly needs an Express router. Since noone seems to use this, I thought it was better to remove this functionality altogether instead of adapting the middleware to the new router
* Fix a bug in the token garbage collector, when attempting to remove an `undefined` token on an empty token repository led to an out-of-control runtimer/checkTokensValidity calls (thousands of calls performed for nothing). Also rewrote parts of this garbage collector to optimize it, using sorted arrays structures


lib/api/core/httpRouter/index.js
```diff
@@ -0,0 +1,187 @@
...
+    requestObject;
+
+  if (!this.routes[request.method]) {
+    requestObject = new RequestObject(base, parsed.query, 'rest');
+    return replyWithError(cb, requestObject, new BadRequestError(`Unrecognized HTTP method ${request.method}`));
+  }
+
+  handler = this.routes[request.method].getHandler(parsed.pathname.split('/'), parsed.query);
+
+  // Do not duplicate those informations to the data.body part of the request object
```
ballinette: I don't understand the need for following lines:
* Is there a use-case where "index" or "collection" can be present within the queryParameters ?
* if it's the case, the RequestObject constructor still merge data from  `parsed.query` to `base` object.
(see https://github.com/kuzzleio/kuzzle-common-objects/blob/master/models/requestObject.js#L43 )

ballinette: Ok, found it for my 1st question: index/collections are populated into parsed.query by getHandler method ?

scottinet: Exactly. And this create errors in some particular cases, like with the following route:

    /api/1.0/:index/:collection/_search

Here, with an ES query as the body content of the request, ES receives both the query and the index and collection parameters, which result in an error.

That's why I have to clean up these arguments. We probably need to refactor `RequestObject` but I wouldn't risk it here.

ballinette:
Not very fan of the side effects on arguments passed to getHandler() method, but I have nothing better to suggest.
And + :100:  :100:  :100:  :100:    for the whole work <3 ;)

lib/api/core/httpRouter/index.js
scottinet: That work was previously performed by `RouterController.executeFromRest`, where we analyzed the request params and picked the index/collection only once. I wanted something less tedious, but I ended up with this, which is almost exactly what was done before :-(

scottinet:
You make a point. It might be cleaner to make `getHandler` return a promise with some kind of `Route` object containing the following properties: 

```
{
  hander: <function>,
  args: {
    ... parametric arguments
  }
}
```

What's your take on this?

lib/api/core/httpRouter/index.js
ballinette: Ok, let's keep it like that for the moment
I'm surprised that the duplication is not already done by RequestObject (it should, according to the code, or I miss something), but I agree we can wait later to fix ;)

scottinet: It's not, because of this line: https://github.com/kuzzleio/kuzzle-common-objects/blob/master/models/requestObject.js#L71

scottinet:
Done!

scottinet:
@ballinette, @dbengsch  > I wasn't satisfied with the way different parts of my code were entangled together so I updated it with better separation of concerns. You might want to review it again.

dbengsch:
+1 again

Issue #535, opened 2016-11-15
Enable connections on the internal broker using a Unix socket


benoitvidis:
The internalBroker component is used to connect to the cli and in cluster mode.
In a single host setup, using a Unix socket for messaging transport would allow to avoid exposing yet another tcp port for Kuzzle.

ref: https://github.com/kuzzleio/kuzzle-proxy/issues/26

Additionally, when using a TCP transport, Kuzzle should allow to bind its internalBroker to a dedicated host.

PR #543, opened 2016-11-21
Unix socket and host binding support

benoitvidis:
Fixes: #535

This PR adds some network configuration options to the internal broker, which is now used by the cli only in stand-alone mode.

* Unix socket are supported and are set as the default transport
* `host` binding option added to the TCP transport option.
* option to disable the internal broker completely 

:warning: @tbd: For the time being, the same configuration is used by the server and the client to connect to the internal broker.
When using a TPC connection, this may problems with special ip addresses (i.e. `0.0.0.0`).

## Additional un-related fixes:

* fix a plugin serialization issue which was causing the logger to fail to start on first try.
* code coverage is back


benoitvidis:
update: 

* the plugin serialization first fix induced another situation where the dump itself would crash
* on some situation, functional tests could start before kuzzle was started => moved the entrypoint init to last action of kuzzle start process and fix related tests.

.kuzzlerc.sample
```diff
@@ -177,6 +177,11 @@
   // Configuration for Kuzzle server
   /*
   "server": {
+    // * http disabled: Defaults to false. Set true to disable http support in Kuzzle
```
scottinet: HTTP support will be removed from Kuzzle when the following PR is merged: https://github.com/kuzzleio/kuzzle/pull/541

lib/api/core/entryPoints/index.js
```diff
@@ -15,6 +16,7 @@ EntryPoints.prototype.init = function entryPointsInit () {
   return this.http.init()
     .then(() => {
       this.proxy.init();
+      return Promise.resolve();
```
scottinet: Potential conflict with https://github.com/kuzzleio/kuzzle/pull/541 since there is no HTTP server anymore

lib/api/controllers/cliController.js
```diff
@@ -21,7 +21,7 @@ function CliController (kuzzle) {
     };
 
     kuzzle.services.list.broker.listen(kuzzle.config.queues.cliQueue, this.onListenCB);
-    kuzzle.pluginsManager.trigger('log:info', 'CLI controller initialized');
+    return kuzzle.pluginsManager.trigger('log:info', 'CLI controller initialized');
```
stafyniaksacha: Returning the response of `log:info` event make it available to pipe plugins.

I think we should not return on this promise to let it run in background.
So I suggest to only `return null` here

PR #545, opened 2016-11-21
Fix unit tests

scottinet:
* fix coverage calculation by restoring `process.emit` when testing kuzzle start process (@benoitvidis)
* remove `sinon-as-promised`, which modifies `sinon` prototype globally and also hid some test errors and promise warnings
* fix some broken tests thanks to promise warnings in dev mode
* add `--no-exit` to mocha options, preventing it to call `process.exit` when the last test "ends"
  * force unit tests to clean up the event loop after running, isolating them further
  * prevents test errors to be hidden if occurring after unit tests are done
  * fix unit tests staying in the event loop so that they clean up after them


ballinette:
I notice that unit tests on "ElasticSearch service" are broken. (indeed, they where still broken before this PR, but silently forgotten).
I mean: if we run unit tests without starting an ElasticSearch instance, the tests fail with this error:
```
Elasticsearch ERROR: 2016-11-23T11:01:27Z
  Error: Request error, retrying
  GET http://localhost:9200/ => connect ECONNREFUSED 127.0.0.1:9200
```

We need to mock the elasticsearch service, as it is done by Redis client.

scottinet:
Good catch. And the client stays on the event loop, trying to connect to ES, meaning that these unit tests never ends with the `--no-exit` mocha option.

I'll mock the ES client right away.

scottinet:
@ballinette > done!
Tested without an elasticsearch instance running, all warnings and silent errors have been removed

PR #547, opened 2016-11-23
fix error when getting document that didn't exist

stafyniaksacha:


lib/services/elasticsearch.js
```diff
@@ -226,13 +226,15 @@ function ElasticSearch(kuzzle, options, config) {
...
-          // todo Feedback how to get it from the 'trash' once it is implemented
-          return Promise.reject(new NotFoundError('Document ' + result._id + ' has been deleted'));
-        }
-        if (result._source._kuzzle_info) {
-          result._kuzzle_info = result._source._kuzzle_info;
-          delete result._source._kuzzle_info;
+        if (result._source) {
+          if (result._source._kuzzle_info && !result._source._kuzzle_info.active) {
+            // todo Feedback how to get it from the 'trash' once it is implemented
+            return Promise.reject(new NotFoundError('Document ' + result._id + ' has been deleted'));
```
scottinet: I know it was already there, but I find it a bit disturbing to get a `NotFoundError` with a `document deleted` message.

Maybe we can change this message slightly:

    `Document ${result._id} has already been deleted`

PR #549, opened 2016-11-24
Feature 548 Add execute in PluginContext

### LINKED ISSUE #548: Add an execute accessor in PluginContext
Add an `execute` accessor in the `PluginContext`.
Make the `userContext` available in each before and after `data:*` events

dbengsch:
- Align all before and after 'data:*' events
- Adapt funnel
- Add execute in PluginContext

fix #548 

:warning: auth plugin's version 3.0.0 must be published on npm before this PR can work :warning:


scottinet:
I wonder if, instead of using on-the-fly objects such as `{requestObject, userContext}`, we could add a new `eventData` object in the `kuzzle-common-objects` library.
This would be easier to document and it would help plugin developers to distinguish between an event providing this kind of object, with other kind of events.

scottinet:
As you can see, this also leads me to a potential new object: `UserContext`, also in `kuzzle-common-objects`

Issue #420, opened 2016-10-08
Trigger event from plugin


farwayer:
I want to throw error during plugin initialization with `log:error` event. But look like plugin must be privileged to get access to `pluginManager`. I do not need access to any internal functionality nor any internal data. Only throw event.

I think throwing events it's very useful functionality for plugins and it should be available to all modules. Some plugins may throw events and another may catch them. It's very valuable case. I propose to add `trigger()` method to plugin context. 

Also it will be nice to have `logger` helper object inside context with `warn()`, `info()`, `debug()`,  `verbose()` and `error()` methods. This functions should throw `log:*` events. For convenience only.

xbill82:
Hi @farwayer sorry for the late reply.
As you can see, plugins will be soon able to send messages to the logger, which is a very important need that you pointed out, and we thank you for that.
On the other side, we are thinking about exposing the `trigger()` method, so that plugins may trigger events on their own (even custom events, in this case). We didn't take the time to thoroughly talk about this feature with all the team but we'll do it soon (but we won't implement this feature in a rush since it may have consequences on the whole plugin architecture).

Again, thanks for your contributions and stay tuned!

xbill82:
My personal thoughts on this subject involve the use in Kuzzle of a Mailer plugin within a shopping cart checkout process. Imagine that

* the customer checks out her cart and pays it;
* the Checkout plugin manages the whole transaction with the third-party payment service;
* now a mail should be sent to the user, confirming that the purchase process is successful.

What would be the best way to do it?

* The Mailer plugin would be of type "listener", and would listen to some sort of `send-mail` event, which would contain the _whole_ mail (sender, receipients, Cc, Bcc and body) and just send it. Doesn't look very neat (and implies the use of `trigger` within the Checkout plugin).
* The Mailer plugin would be of type "controller", thus expose a new API method like
```json
{
  "controller": "mailer",
  "method": "send"
}
```
so that the Checkout plugin (assuming that it has access to the Kuzzle API) would call it to send the mail. Looks better and doesn't require the use of `trigger`.

Any thoughts are welcome.

farwayer:
Thank you guys for your attention!
Event system has more power. Let's discuss your example.

Think a bit what is really want to do *Checkout* controller. Does he want send email? No at all. Actually *Checkout* controller should **notify user** about purchase status. From this point of view things become more interesting.

[...]

PR #557, opened 2016-11-28
Develop 420 log trigger for plugins

benoitvidis:
Expose Kuzzle logger in Plugin context.
cf #420 .


ballinette:
Just created an issue on kuzzle-guide to add the documentation about this new context accessor: 
https://github.com/kuzzleio/kuzzle-guide/issues/74

xbill82:
@stafyniaksacha I'm not against your proposal but I think it deserves a deeper team-wise reflection.

PR #559, opened 2016-11-30
Harmonization of the Request and result handling

### LINKED ISSUE #556: Controller harmonization
- Harmonize controllers to use the new Request common object
- Trigger before and after events in the funnel

dbengsch:
- Every controllers have been refactored;
- [Breaking changes] `before` and `after` events are now triggered by the funnel with an argument `request` of type `Request`;
- Most methods that were requiring a `RequestObject` now require a `Request`;
- Methods that were requiring a `context` now require either a `RequestContext` or a `Request`;
- [Breaking changes] `['from', 'size', 'scroll', 'scrollId', 'refresh', 'type']` previously integrated to the `body` have to be provided as arguments (either in the root of the incoming request or in `query_string` / `headers` for Http).
- Added basic checks on the `Request` structure to all controllers;
- The `funnel`, the `routeController` and the `routeHandler` have been refactored accordingly.

**Unrelated changes**

* Rename the CLI command `likeAvirgin` to `reset` and modify the interaction messages accordingly
* Remove `logout` from the default `anonymous` role definition
* Optimize the overload mode handled by the funnel controller (avoid v8 deopt and better overall performances)

:warning: Proxy modifications must be merged with its plugin changes in on proxy's develop branch to allow this branch to work properly

fix #556


lib/api/cli/action.js
```diff
@@ -40,13 +42,16 @@ Action.prototype.onSuccess = function actionOnSuccess (response) {
 };
 
 Action.prototype.onListenCB = function actionOnListenCB (response) {
+  var request;
   clearTimeout(this.timeoutTimer);
+
+  request = new Request(response.data, response.options);
```
ballinette: Don't understand why we create here a `Request` object with an input `response` data.
Probably the `response` variable should be renamed to avoid confusion, but I need some more explanation on how the Cli actions actually work...

lib/api/controllers/funnelController.js
```diff
@@ -117,14 +120,18 @@ function FunnelController (kuzzle) {
 
     if (this.cachedRequests >= kuzzle.config.server.maxRetainedRequests) {
       error = new ServiceUnavailableError('Request discarded: Kuzzle Server is temporarily overloaded');
+
       kuzzle.pluginsManager.trigger('log:error', error);
-      callback(error, new ResponseObject(requestObject, error));
+
+      request.setError(error);
+
+      callback(error, request.response);
```
ballinette: isn't it `callback(error, request)` ?

dbengsch: Totally right

benoitvidis:
\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/\o/

lib/api/controllers/writeController.js
```diff
@@ -13,295 +18,219 @@ function WriteController (kuzzle) {
...
       })
       .then(response => {
-        kuzzle.notifier.notifyDocumentCreate(modifiedData.requestObject, response);
+        kuzzle.notifier.notifyDocumentCreate(modifiedRequest, response);
 
-        return kuzzle.pluginsManager.trigger('data:afterCreate', {
-          responseObject: new ResponseObject(modifiedData.requestObject, response),
-          userContext: modifiedData.userContext
-        });
+        return Promise.resolve(response);
```
xbill82: This kind of return statement is quite systematic in Kuzzle but I don't think it is useful. AFAIK, within the **last** `then` callback, returning a value is equivalent to returning a Promise resolving a value. Please, tell me if I'm wrong, cuz I need to know if I'm totally next to the plaque...

bin/commands/reset.js
```diff
@@ -40,11 +40,11 @@ function commandReset (options) {
     }
   }
 
-  console.log(warn('[ℹ] You are about to reset Kuzzle and make it like a virgin.'));
+  console.log(warn('[ℹ] You are about to reset Kuzzle configuration and users'));
```
stafyniaksacha: nitpicking: it will also clean stored documents !

lib/api/controllers/adminController.js
```diff
@@ -1,543 +1,338 @@
...
       .then(() => kuzzle.validation.curateSpecification())
-      .then(() => kuzzle.pluginsManager.trigger('data:afterUpdateSpecifications', {
-        responseObject: new ResponseObject(modifiedData.requestObject, modifiedData.requestObject.data.body),
-        userContext: modifiedData.userContext
-      }));
+      .then(() => Promise.resolve(request.input.body));
   };
 
   /**
    * Add a specification collections
```
scottinet: Indeed.

lib/api/controllers/authController.js
```diff
@@ -14,144 +20,104 @@ function AuthController (kuzzle) {
...
-
-        return kuzzle.repositories.token.expire(modifiedData.userContext.token);
-      })
-      .then(() => kuzzle.pluginsManager.trigger('auth:afterLogout', {
-        responseObject: new ResponseObject(modifiedData.requestObject, {}),
-        userContext: modifiedData.userContext
-      }));
+  this.logout = function authLogout (request) {
+    return kuzzle.repositories.token.expire(request.context.token)
+      .then(() => Promise.resolve({}))
```
scottinet: Nope: it forces resolving an empty object instead of `undefined`, which would be rejected by `Request.setResult` in the funnel controller.

lib/api/controllers/funnelController.js
```diff
@@ -80,51 +86,65 @@ function FunnelController (kuzzle) {
...
+    if (!this.overloaded) {
+      this.overloaded = true;
+
+      /*
+       /!\ Invoking this function here with setTimeout() leads to V8 deoptimizing
+       the entire funnelExecute method (as of node.js 6.9.1),
+       because of an "out of bounds" heuristic error (caused by node's
+       setTimeout code written in JS? this needs further investigations)
+
+       We get better performances executing this function once for nothing each
```
scottinet: That's because I've mixed up the beginning and the end of this sentence when I wrote it. I fixed it.

lib/api/controllers/funnelController.js
```diff
@@ -153,11 +173,11 @@ function FunnelController (kuzzle) {
           errorMessage = errorMessage.toLowerCase().replace(/[^a-zA-Z0-9-_]/g, '-').replace(/[-]+/g, '-').split('-');
           errorMessage = errorMessage.filter(value => value !== '').join('-');
 
-          request.data.body.sufix = 'handled-'.concat(errorType.toLowerCase(), '-', errorMessage);
+          request.data.body.suffix = 'handled-'.concat(errorType.toLowerCase(), '-', errorMessage);
```
scottinet: Object.seal() does not cause exception throws. It only fails silently.
Fixed.

lib/api/controllers/routerController.js
```diff
@@ -159,38 +87,25 @@ function RouterController (kuzzle) {
...
-      },
-      token: null
-    };
-
-  requestObject.controller = params.controller;
-  requestObject.action = params.action;
+function executeFromHttp(funnel, params, request, cb) {
+  request.input.controller = params.controller;
+  request.input.action = params.action;
+  request.context.protocol = 'rest';
```
scottinet: Fixed.

lib/api/controllers/writeController.js
scottinet: You're absolutely right, this creates unnecessary promises. Fixed in all controllers.

PR #560, opened 2016-12-01
add admin:updeteUserMapping & admin:getUserMapping actions

stafyniaksacha:


lib/api/controllers/adminController.js
```diff
@@ -64,6 +88,28 @@ function AdminController(kuzzle) {
...
+   * @param {Object} userContext
+   * @returns {Promise}
+   */
+  this.getUserMapping = function adminGetUserMapping (requestObject, userContext) {
+    var
+      modifiedData = null;
+
+    return kuzzle.pluginsManager.trigger('data:beforeGetUserMapping', {requestObject, userContext})
+      .then(data => {
+        modifiedData = data;
```
xbill82: what do you need the `modifiedData` object for?

benoitvidis: To build the response object later on.
But this code will eventually have to be entirely rewritten following #559

ballinette:
agree with @benoitvidis , but we need these actions right now for a customer's project.
So I suggest to validate this right now, and then, refactor these actions within #559

PR #563, opened 2016-12-05
Refactor: controller plugins definition

scottinet:
The current way of defining a controller plugin is a bit complicated and hard to document. 

To define a controller plugin, we need to expose a `controllers` object, detailing the controllers to add to the API, each one of them exposing a function, creating an object, containing actions.

Moreover, this architecture makes it difficult to understand  what the `action` of a HTTP route refers to.

In order to simplify the way controller plugins are defined, this PR proposes the following: 

* the plugin needs to expose a `controllers` object. This object simply describes the controllers to add using the following format:

```js
this.controllers = {
  controllerToAdd: {
    actionName: exposedFunctionName
  }
};
```

* `exposedFunctionName` function implementations must simply be exposed to Kuzzle

* no change done to `this.routes`: `route.controller` and `route.action` simply refer to the controller's description in `this.controllers`

This should bring controller plugins in line with the way other plugin types are defined (e.g. listener or pipe plugins).

This PR also adds a lot more tests to check a controller plugin validity, thus limiting potential crashes because of an invalid controller plugin.

### DIFF
--- a/lib/api/core/plugins/pluginsManager.js
+++ b/lib/api/core/plugins/pluginsManager.js
@@ -403,7 +403,7 @@ function initControllers (controllers, routes, plugin, pluginName) {
           return false;
         }
 
-        if (typeof route[key] !== 'string' || route[key].length === 0) {
+        if (typeof route[key] !== 'string' || (route[key].length === 0 && key !== 'url')) {
           // eslint-disable-next-line no-console
           console.error(`[Plugin Manager] Error initializing route ${route.url}: ${key} must be a non-empty string`);
           return false;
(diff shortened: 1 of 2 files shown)


lib/api/core/plugins/pluginsManager.js
```diff
@@ -358,17 +358,83 @@ function initHooks (plugin, kuzzle) {
...
+      var valid = Object.keys(route).every(key => {
+        if (['verb', 'url', 'controller', 'action'].indexOf(key) === -1) {
+          // eslint-disable-next-line no-console
+          console.error(`[Plugin Manager] Error initializing route ${route.url}: unknown route definition ${key}`);
+          return false;
+        }
+
+        if (typeof route[key] !== 'string' || route[key].length === 0) {
+          // eslint-disable-next-line no-console
+          console.error(`[Plugin Manager] Error initializing route ${route.url}: ${key} must be a non-empty string`);
```
xbill82: what if `route.url` is empty? :)

PR #585, opened 2016-12-16
Set Request's context.user with the user loaded in the funnel

### LINKED ISSUE #424: Save user to context
`funnelController` requests user info from database to check user permissions. But after using this data his [discards it](https://github.com/kuzzleio/kuzzle/blob/master/lib/api/controllers/funnelController.js#L132). Maybe it will be better to keep user object inside context so any other part of application or plugins can access user without repeat request to database (or cache).

For now `auth`, `admin` controllers and `hotelClerk` can recall `repositories.user.load()`.

And what is about passing `context` to all `event` handlers? For ex. context is not passing to `data:*` handlers so it's harder to access current `userId` (there is need to parse authorization header and call `token.verifyToken()` again).

dbengsch:
fix #424 

:warning: Only requests going through the `Funnel Controller` will have the user set (or when the request is built with a complete context).

### DIFF
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -21,7 +21,7 @@ function AuthController (kuzzle) {
    * Logs the current user out
    *
    * @param {Request} request
-   * @returns {Promise<Object>}
+   * @returns {Promise<object>}
    */
   this.logout = function authLogout (request) {
     return kuzzle.repositories.token.expire(request.context.token)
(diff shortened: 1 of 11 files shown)


ballinette:
maybe also change all instances of `author: request.context.token.userId ? String(request.context.token.userId)` to `author: request.context.user.userId ? String(request.context.user.userId)` in services/elasticsearch, even if it has no performance impact there (just for code consistency)

and within `funnelController/processRequest`:
```
        // logged-in user with insufficient permissions => we throw a 403 (Forbidden) error
        return Promise.reject(new ForbiddenError(
          `Forbidden action [${input.index}/${input.collection}/${input.controller}/${input.action}] for user ${request.context.user.userId}`)
        );
```

dbengsch:
@ballinette Done

PR #588, opened 2016-12-19
change overload event name

scottinet:
With the recent controllers normalization, the `server:overload` event seemed out of place, among other `server:xxx` events dedicated to the new `server` controller.

To make our API more consistent, I propose to rename this event to `core:overload`, so that it appears with other `core` events.


ballinette:
added an issue on kuzzle-guide to update the documentation:
https://github.com/kuzzleio/kuzzle-guide/issues/94

scottinet:
Documentation already updated here: https://github.com/kuzzleio/kuzzle-guide/pull/93

PR #605, opened 2017-01-03
Logger one processs

benoitvidis:
Following https://github.com/kuzzleio/kuzzle-plugin-logger/pull/13/files

This PR sets by default one process for the logger, which should be good enough in most cases and limits memory usage.


ballinette:
why merge into master branch ?

benoitvidis:
good catch, was a mistake, sorry..

PR #607, opened 2017-01-03
Re-organize http routes to comply more with method meanings

dbengsch:
* Consider `createOrReplace` as the default action of `PUT` method (and remove them from url);
* Consider `POST` method has no default action (and add `_create` action when necessary);
* Change HTTP method depending on idem-potency (update in `PUT`, index creation in `POST`...);
* Change `createOrReplace` to always require an `_id`

### DIFF
--- a/lib/config/httpRoutes.js
+++ b/lib/config/httpRoutes.js
@@ -1,5 +1,5 @@
 module.exports = [
-  // GET
+  // GET (idempotent)
   {verb: 'get', url: '/users/_me', controller: 'auth', action: 'getCurrentUser'},
   {verb: 'get', url: '/users/_me/_rights', controller: 'auth', action: 'getMyRights'},
   {verb: 'get', url: '/_login/:strategy', controller: 'auth', action: 'login'},
@@ -14,8 +14,8 @@ module.exports = [
 
   {verb: 'get', url: '/:index/:collection/:_id', controller: 'document', action: 'get'},
 
-  {verb: 'get', url: '/:index/_autoRefresh', controller: 'index', action: 'getAutoRefresh'},
   {verb: 'get', url: '/:index/_exists', controller: 'index', action: 'exists'},
+  {verb: 'get', url: '/:index/_autoRefresh', controller: 'index', action: 'getAutoRefresh'},
   {verb: 'get', url: '/_list', controller: 'index', action: 'list'},
   {verb: 'get', url: '/_refreshInternal', controller: 'index', action: 'refreshInternal'},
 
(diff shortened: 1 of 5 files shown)


lib/config/httpRoutes.js
```diff
@@ -209,47 +208,45 @@ module.exports = [
...
   {verb: 'delete', url: '/:index', controller: 'index', action: 'delete'},
+  {verb: 'delete', url: '/_mdelete', controller: 'index', action: 'mDelete'},
 
-  {verb: 'delete', url: '/roles/:_id', controller: 'security', action: 'deleteRole'},
   {verb: 'delete', url: '/profiles/:_id', controller: 'security', action: 'deleteProfile'},
+  {verb: 'delete', url: '/roles/:_id', controller: 'security', action: 'deleteRole'},
   {verb: 'delete', url: '/users/:_id', controller: 'security', action: 'deleteUser'},
 
+  {verb: 'delete', url: '/ms/:_id', controller: 'ms', action: 'del'},
+  {verb: 'delete', url: '/ms', controller: 'ms', action: 'del'},
```
scottinet: Doesn't the Redis' `DEL` action require a key to delete? What does this route do?

lib/config/httpRoutes.js
```diff
@@ -209,47 +208,45 @@ module.exports = [
...
   {verb: 'delete', url: '/ms/_zremrangebyscore/:_id', controller: 'ms', action: 'zremrangebyscore'},
-  {verb: 'delete', url: '/ms/:_id', controller: 'ms', action: 'del'},
-  {verb: 'delete', url: '/ms', controller: 'ms', action: 'del'},
 
 
-  // PUT
+  // PUT (idempotent)
   {verb: 'put', url: '/_updateSelf', controller: 'auth', action: 'updateSelf'},
 
   {verb: 'put', url: '/:index/:collection', controller: 'collection', action: 'create'},
```
scottinet: Since you moved `index:create` to `POST` actions, shouldn't `collection:create` be moved there too?

lib/config/httpRoutes.js
dbengsch: I don't know, only moved it for the sake of ordering. I'll take a look

dbengsch: the `del` can delete multiple elements at a time. _id can be provided but also the argument `keys`. So the user can use a route like DELETE /ms/?keys[]=foo&keys[]=bar

dbengsch:
@scottinet comments replied

lib/config/httpRoutes.js
scottinet: :+1:

lib/config/httpRoutes.js
scottinet: WAT

scottinet: Well... ok then.

PR #616, opened 2017-01-06
Implement multi actions mGet, mCreate, mReplace, mUpdate, mCreateOrReplace and mDelete

dbengsch:
**Implemented actions:**

* `document:mGet`
* `document:mCreate`
* `document:mReplace`
* `document:mUpdate`
* `document:mCreateOrReplace`
* `document:mDelete`

**Breaking changes:**

* Aligned http routes for `security:mGetProfiles` and `security:mGetRoles` to be more aligned with route conventions (camel case).
(:warning: Regenerate PHP routes).

**Boy-scout:**

* Simplify url generation in `ApiHttp` in functional tests.
* Repair elasticsearch `create` when it was actually replacing a trashed document => it never provided the document back.
* Repair request status when status is not 102 (in `FunnelController`).
* Small changes (see comments)

fix #436 and #450

PR #623, opened 2017-01-11
mDeleteXX on security components

### LINKED ISSUE #619: Delete by query Users/Roles/Profiles
Hello,

Because we can't perform a `.deleteDocument` on the index `%kuzzle` we can't delete users/roles/profiles with a query (or simply with an array of IDs).

We need to have a way to delete mutli user/profile/role in Kuzzle and in SDKs.

benoitvidis:
Fixes #619

This PR exposes the following methods:

* `mDeleteProfile`
* `mDeleteRole`
* `mDeleteUser`

in the `security` controller.

# Boyscouting - breaking change

In order to align actions naming between #616 and because the `m` prefix already specifies the plurality, `security:mGetProfiles` and `security:mGetRoles` are respectively renamed to `mGetProfile` and `mGetRole`


benoitvidis:
Just realized http routes were missing... sorry -_-;

lib/config/httpRoutes.js
```diff
@@ -128,6 +128,9 @@ module.exports = [
   {verb: 'post', url: '/users/:_id/_createRestricted', controller: 'security', action: 'createRestrictedUser'},
   {verb: 'post', url: '/users/_create', controller: 'security', action: 'createUser'},
   {verb: 'post', url: '/users/:_id/_create', controller: 'security', action: 'createUser'},
+  {verb: 'post', url: '/profiles/_mdelete', controller: 'security', action: 'mDeleteProfile'},
```
benoitvidis: The http RFC is not clear about allowing a body in a DELETE request. To be safe, _mdelete calls are set as POST
http://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request

features/step_definitions/profiles.js
```diff
@@ -247,7 +247,7 @@ var apiSteps = function () {
 
     main = function (callbackAsync) {
       setTimeout(() => {
-        this.api.mGetProfiles(body)
+        this.api.mGetProfile(body)
```
benoitvidis: boycouting - align action names with #616

lib/api/controllers/securityController.js
```diff
@@ -550,3 +564,57 @@ function resetProfiles () {
...
+ * @param {Request} request
+ * @returns {Promise.<*>}
+ */
+function mDelete (kuzzle, type, request) {
+  const
+    action = _.upperFirst(type);
+
+  assertHasBody(request, `security:${action}`);
+  assertBodyHasAttribute(request, 'ids', `security:${action}`);
+  if (!Array.isArray(request.input.body.ids)) {
```
ballinette: nitpicking: replace it by `assertHasIndexAndCollection` introduced here: https://github.com/kuzzleio/kuzzle/pull/616/files/eb4baa72454df15e01cc11bdc246029bbd5e52e2..98450af8137849d5e3b0942b949ab64e248c765b#diff-91a156dc5237febe996201236bc18f66R44

(implies a dependency with #616)

dbengsch: read `AssertBodyAttributeType`

lib/api/controllers/securityController.js
```diff
@@ -550,3 +564,57 @@ function resetProfiles () {
...
+ * @param {Kuzzle} kuzzle
+ * @param {string.<profile|role|user>} type
+ * @param {Request} request
+ * @returns {Promise.<*>}
+ */
+function mDelete (kuzzle, type, request) {
+  const
+    action = _.upperFirst(type);
+
+  assertHasBody(request, `security:${action}`);
```
dbengsch: shouldn't it be `security:delete${action}` ? (if depends on #616 as suggested below, it becomes useless)

lib/config/httpRoutes.js
```diff
@@ -128,8 +128,11 @@ module.exports = [
   {verb: 'post', url: '/users/:_id/_createRestricted', controller: 'security', action: 'createRestrictedUser'},
   {verb: 'post', url: '/users/_create', controller: 'security', action: 'createUser'},
   {verb: 'post', url: '/users/:_id/_create', controller: 'security', action: 'createUser'},
-  {verb: 'post', url: '/profiles/_mget', controller: 'security', action: 'mGetProfiles'},
-  {verb: 'post', url: '/roles/_mget', controller: 'security', action: 'mGetRoles'},
+  {verb: 'post', url: '/profiles/_mdelete', controller: 'security', action: 'mDeleteProfile'},
```
dbengsch: I just aligned all `m` action in camel case in #616 ...

xbill82:
It makes me sad, being a very effective tool to express plurality available in the english language (kind of adding an "s" at the end of a word), that we simply do the other way around (putting an "m" at the beginning of a singular word).

Would `getUsers`, `getRoles` and `getProfiles` be so ugly?

PR #630, opened 2017-01-18
Fix scroll action to get scroll and scroll_id from the body

dbengsch:


benoitvidis:
provided the tests are fixed, lgtm

lib/api/controllers/documentController.js
```diff
@@ -39,6 +39,13 @@ function DocumentController (kuzzle) {
    * @returns {Promise<Object>}
    */
   this.scroll = function documentScroll (request) {
+    if (!request.input.args.scroll) {
```
ballinette: why not adding an `assertHasArg` method ?

PR #633, opened 2017-01-19
Plugin installation refactoring

### LINKED ISSUE #609: Plugin management refactor
Question: While this issue addresses several points at once, I believe they most of them are dependent on each-other and cannot be easily be split into several different issues. This is of course open to discussion.

# Situation

For the time being, plugins are installed during Kuzzle startup, which downloads the packages and their dependencies and writes their configuration in the database.
This leads to several issues, including:

* First startup is delayed, possibly generating some useless timeout errors
* Kuzzle needs an active internet connection on start
* Plugins are not easy to migrate from an instance to another, mostly because of the database settings
* Plugins versions are difficult to handle during a migration process.

The aim of this proposal is to simplify the installation of plugins.

# Proposal

Main principles:

* rely on the file system only to know which plugins should be loaded/activated
* configuration is in one place only (kuzzle one)

## Installation

Installing a plugin should be as simple as dropping a folder in a dedicated location.
Similar to Apache, here is the proposed plugin directory organization:

```
kuzzle
└── plugins
    ├── available
    │   ├── plugin1
    │   └── plugin2
    └── enabled
        └── plugin1 -> ../available/plugin1
```

* Each plugin is a folder inside `plugins/enabled`. 
* In order to easily activate or deactivate, the default setup proposes to symlink the plugin folder from the `available` directory
* During startup, Kuzzle lists the plugins in the `enabled` folder and requires them.

## Default plugins

Default plugins can directly be bundled in Kuzzle installation.
Extra core team plugins (i.e. oauth authentication strategy) can be bundled too and left unactivated.

## Plugin development

* Plugins do not need to be a module anymore. They just need to be requireable.
* From the above point, we should not enforce plugins to ship a default configuration in the package.json anymore, which was quite dirty IMO anyway.
Plugins should be responsible to set their default configuration and use the overriden one given during init from Kuzzle.

xbill82:
Fixes #609

Please refer to the mentioned issue for a complete description of the refactoring.

I'm not quite sure of the implementation... I kinda suspect it cannot be so easy so, please, be super critic. I need your help making this rock-solid.

I've taken a decision about what we do consider a "require-able" plugin (vs a valid Node.js module, which is based on the existence of a well-formed `package.json`). I consider a plugin requireable if doing `require('directory-path')` works. And it will only work if an `index.js` file exists. No further searching for files in `lib` or stuff. Do you agree? We'll have to document this.

I used the `fs::existsSync` function which is not recommended by the NodeJS documentation, as it may introduce a race condition (it doesn't lock the file for reading). I estimate this is unlikely to happen in our case. I'd like to have your opinions on this.

Breaking change: Plugins no more declare themselves as Workers, as the `threads` config parameter can only be specified in Kuzzle configuration.


xbill82:
BTW, behavioral tests will fail until plugins are not adapted to embed a default configuration. Also, the docker image must provide the local authentication plugin enabled by default. How may we proceed? I'll surely sumbit PRs on each plugin to adapt them to this new installation system.

scottinet:
`package.json` should be updated to remove the `rimraf` dependency, as it was only used to remove plugins from Kuzzle

lib/api/core/plugins/pluginsManager.js
```diff
@@ -552,16 +537,59 @@ function triggerWorkers(event, data) {
  *
  * @param plugins - list of installed plugins to load
  */
-function loadPlugins(plugins) {
-  _.forEach(plugins, (plugin, name) => {
+function loadPlugins(config) {
+  const pluginsDir = path.resolve('/var/app/plugins/enabled');
```
xbill82: This is total crap. Gonna change it to relative.

scottinet:
About your other questions: I agree with your choices:

* I think using `*sync` methods is the right course of action, as blocking the event loop while loading plugins seems to be a good idea to me.
* I agree with your choice of allowing plugins to be loaded with a mere `require` command. Though adding a check to verify that a `init` method is exposed might be a good idea, to exclude plugins without one.

lib/api/core/plugins/pluginsManager.js
benoitvidis: This may help: https://github.com/kuzzleio/kuzzle/blob/develop/lib/api/kuzzle.js#L35

xbill82:
@scottinet thanks for the feedback. However, I don't consider really useful to explicitly check the existence of the `init` function. The `run` function will throw whenever something goes wrong, either `init` is not a function, either something (anything) goes wrong with it. This way, the plugin is "automatically" excluded (and the error message stays meaningful).

Do you see any issues with this?

scottinet:
@xbill82 > no, no issue with this. As long as we cleanly exclude a malformed plugin, then I'm fine with it. :+1:

PR #644, opened 2017-01-30
Proxy ping/pong keepalive

stafyniaksacha:
Add ping / pong system to keep alive connection between kuzzle proxy and kuzzle server

## New .kuzzlerc options :
```javascript
{
  "proxyBroker": {
     "host": "localhost",
     "port": 7331,
     "retryInterval": 1000,
     "pingTimeout": 50, // Timeout between ping request and pong response
     "pingInterval": 60000 // Interval between a successful pong response and another ping request
   }
}
```

## Boyscout

* Upgrade Elasticsearch minor version from 5.0 to 5.1
* Fixed collection & index information in forbidden error message (funnel controller)
* Fixed document garbage collector, context was lost during 2nd run (kuiz crash)


.kuzzlerc.sample
```diff
@@ -283,7 +283,9 @@
       "aliases": ["broker"],
       "host": "localhost",
       "port": 7911,
-      "retryInterval": 1000
+      "retryInterval": 1000,
+      "pingTimeout": 50,
```
scottinet: These two new options must be documented in the sample file, using the comments immediately above the `internalBroker` and `proxyBroker` configuration sections.

lib/services/garbageCollector.js
```diff
@@ -40,13 +40,13 @@ function GarbageCollector(kuzzle) {
 
     if (kuzzle.funnel.overloaded) {
       // Reschedule the task to run in 1 hour
-      this.timer = setTimeout(this.run, OneHour);
+      this.timer = setTimeout(this.run.bind(this), OneHour);
```
scottinet: Since run is a property of the `GarbageCollector` object, I'm not sure binding the context is necessary.

scottinet: After our discussion: I stand corrected. :-)

.kuzzlerc.sample
stafyniaksacha: Done

PR #647, opened 2017-01-31
Elasticsearch emits unnecessary log events

### LINKED ISSUE #646: Storage service should not trigger the logger
The storage engine currently systematically triggers the logger on each error.

This service is always called in a context where the errors are already handled (mainly by the funnel).

This leads to several problems, including readability and some possible unneeded I/O stress.
The special `NotFoundError` case is also quite painful.

dbengsch:
fixes #646

### DIFF
--- a/lib/services/elasticsearch.js
+++ b/lib/services/elasticsearch.js
@@ -920,8 +920,6 @@ function ElasticSearch (kuzzle, options, config) {
       case 'BadRequest':
         if (!messageReplaced) {
           message = error.body.error.root_cause ? error.body.error.root_cause[0].reason : error.body.error.reason;
-
-          this.kuzzle.pluginsManager.trigger('log:warn', '[warning] unhandled elasticsearch error:\n' + error.message);
         }
 
         kuzzleError = new BadRequestError(message);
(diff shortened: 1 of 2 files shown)


scottinet:
I wonder if that's enough to solve #646 
To me, I think the ES client should also be configured so that its logs are separated from kuzzle ones, for instance by restricting logs to errors and warnings and by putting log messages into a specific file.
See https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/logging.html

What do you think of this? (/cc @benoitvidis)

stafyniaksacha:
@scottinet These logs are ones that we haven't mapped to our [elasticsearch errors messages mapping](https://github.com/kuzzleio/kuzzle/blob/3fdf3d8a219f9718c312628617ac9caba0c6049a/lib/services/elasticsearch.js#L46) as I proposed to @dbengsch, I think we must not warn them with log system but use the new debug system introduced in #643 
What do you think about ?

scottinet:
@stafyniaksacha > interesting idea, I like it. But as discussed, what I propose is quite different :)

benoitvidis:
@scottinet The related issue is not about ES client logs but in general, I am not in favor of, by default, outputting things to files. IMO, there is too much risk in forgetting about it and ending up filling up a disk in production.

That being said, offering the option to customize ES client configuration (sticking to the default logger or/and using Kuzzle's one sounds like a good idea to me), but I believe this should be addressed in a separate issue.

@sstafyniak :+1:  on your proposal to use debug instead of the logger.

PR #653, opened 2017-02-02
Fix #635: multiactions routes bypass overload

### LINKED ISSUE #635: `document:m*` routes bypass overload protection
Simple recipe to overload Kuzzle's event loop and freeze a node without effort: submit a mCreate query with thousands of documents to process.

The `document` controller will parse the query content and submit 1 `processRequest` action from the funnel controller, without going through the overload protection mechanism nor imposing a process count limit.

Simple fix: the `doMultipleActions` function, in the `document` controller, should invoke the `execute` function in the funnel controller, instead of the `processRequest` one.
Pro: fixes the problem
Cons:
* impose a new rights check for each subrequest, which is less than ideal performance-wise
* impose users to have the same rights for `m*` routes than for their single document action counterpart

Less simple but better fix: separate and expose the funnel's overload mechanism from its `execute` function, by exposing a new `start(request, callback)` function (function name to be discussed), resolving the callback only when the event loop charge allows it.
Make the `execute` function invokes it, as well as the `m*` routes for each document to process.

scottinet:
Fixes #635 

Add configurable limits on how many documents can be either fetched or written in a single API request, and make multidocuments API routes go through the overload-protection system.

# Configuration changes

* rename the `config.server` section into `config.limits`, to better reflect the purpose of the included settings.
* rename existing `config.limits` parameters, as these are now contextualized as server limits. This allows for better parameter names
* add `limits.documentsFetchCount` and `limits.documentsWriteCount` parameters, limiting the number of impacted documents by a single API read or write request
* use a better documentation format and rewrite most of the documentation in the `.kuzzlerc.sample` file
* fix multiple errors in `.kuzzlerc.sample` file to make it usable directly by users
* add `.kuzzlerc` in `.gitignore`
* enable diagnostic tools by default

# Documents limitation changes

* isolate the overload-protection mechanism in a `getRequestSlot` method in the funnel controller
* make document:m* routes call `getRequestSlot` prior to executing sub-action requests
* make security:m* routes call `getRequestSlot` prior to executing sub-action requests
* make `document:m*`, `security:m*` and all search API calls enforce the new `limits.documents[Fetch|Write]Count` configuration limits
* add a configuration check, to make sure that even if a user put incoherent values in the `limits` section of the configuration file, Kuzzle correctly fallbacks to acceptable values

# Boyscout

* remove unnecessary `Promise.resolve()` calls
* add missing jsdoc in the funnel controller


lib/api/controllers/documentController.js
```diff
@@ -72,11 +73,14 @@ function DocumentController (kuzzle) {
     assertBodyAttributeType(request, 'ids', 'array');
     assertHasIndexAndCollection(request);
 
+    if (request.input.body.ids.length > kuzzle.config.server.maxMultiActionsCount) {
```
dbengsch: mGet is a multi-action but should only count as one as it only sends one request to ES.

scottinet: Well... technically, yes. But it occured to me that if I was a Kuzzle administrator, I may also want to limit the number of fetched documents using `mGet`, for performances and bandwidth reasons. And I thought the best way to do that was to apply `maxMultiActionsCount` to this API route too.

What do you think?

benoitvidis:
:+1: for the config check

lib/api/controllers/documentController.js
stafyniaksacha: @scottinet I agree, if someone need to retrieve a list of documents, the best way is to use search API with `ids` filters

dbengsch: First, I would say mGet is faster than search to get a list of documents by their ids.
Second, I don't really see the point of associating the `maxMultiActionsCount` with `mGet`; I thought the purpose of the settings was to avoid a `m*` to stuck Kuzzle completely because it was not going through the overload system properly.
At this point we should limit the size parameter of search requests as well...

scottinet: Search requests are already paginated by ES.

xbill82: I have tendency to agree with @dbengsch, but I find @scottinet 's remark pertinent.

I mean, it's interesting to be able to limit the size of a response from Kuzzle to the client. If a guy makes a naive `m*` request that results in a 2Mb response because of the huge volome of documents in ES, well... He may never receive the response.
Yet, `maxMultiActionsCount` doesn't mean `maxResponseSize`. It means `maxMultiActionsCount`. Which is very good to avoid overheading Kuzzle in terms of memory and computation.

IMHO, you're both right, but we shoudn't over-responsabilize the `maxMultiActionsCount` parameter. I propose we introduce another `maxResponseSize` parameter, probably interesting from a networking point of view, the implementation of which is not to be discussed here immediately, especially since I have no clue about how to do it.

scottinet: Well... `mGet` is probably one of the only API route capable of generating such large responses, because there is currently no check on how many documents can be retrieved using it.
All other API routes I know of are either intrinsically limited in size, or paginated. The only exception I see are controllers plugin routes, where your proposal would be beneficial.

And the response size of a `mGet` call depends directly on the number of documents to fetch. 

Another argument I have is API harmonization. I agree, the `mGet` implementation benefits from a dedicated ES API route, but I think that's circumstantial. I would find it weird if all multi-documents routes  had to comply to this option, either in the `document` or in the `security` controllers, except for this one route.

dbengsch: You can also make a search request with a from / size resulting to a batch of 9 999 documents (perhaps 10 000).

lib/api/controllers/documentController.js
```diff
@@ -307,15 +315,16 @@ function DocumentController (kuzzle) {
...
-          })
-          .catch(error => {
-            deleteRequest.setError(error);
+          }
 
-            return resolve(deleteRequest);
-          });
+          kuzzle.funnel.processRequest(deleteRequest)
+            .catch(error => deleteRequest.setError(error))
+            .then(() => resolve(deleteRequest));
```
ballinette: shouldn't it by rather `finally` ?

scottinet:
@dbengsch, @ballinette, @xbill82 > following our discussion, I adopted a middle ground by renaming and splitting the new `maxMultiActionsCount` into `documentsFetchCount` and `documentsWriteCount`

Since I wasn't satisfied with some of the configuration parameter names, I also refactored them.

@sstafyniak, @benoitvidis > another review may be needed to confirm your initial approval.

lib/api/controllers/documentController.js
```diff
@@ -1,13 +1,15 @@
...
+let
   BadRequestError = require('kuzzle-common-objects').errors.BadRequestError,
+  SizeLimitError = require('kuzzle-common-objects').errors.SizeLimitError,
   assertHasBody = require('./util/requestAssertions').assertHasBody,
   assertHasId = require('./util/requestAssertions').assertHasId,
   assertBodyHasAttribute = require('./util/requestAssertions').assertBodyHasAttribute,
   assertHasIndexAndCollection = require('./util/requestAssertions').assertHasIndexAndCollection,
   assertBodyAttributeType = require('./util/requestAssertions').assertBodyAttributeType,
   Request = require('kuzzle-common-objects').Request,
+  Promise = require('bluebird'),
```
stafyniaksacha: would be nice to think about not override default javascript object (`Promise` is a build-in object)

lib/api/controllers/funnelController.js
```diff
@@ -216,19 +241,21 @@ function FunnelController (kuzzle) {
           // anonymous user => we throw a 401 (Unauthorised) error
           if (request.context.token.userId === -1) {
             return Promise.reject(new UnauthorizedError(
-              `Unauthorized action [${request.input.resource.index}/${request.input.resource.collection}/${request.input.controller}/${request.input.action}] for anonymous user`)
+              `Unauthorized action [${request.input.index}/${request.input.collection}/${request.input.controller}/${request.input.action}] for anonymous user`)
```
stafyniaksacha: hum, `index` and `collection` has moved from `resource` ?

lib/api/controllers/documentController.js
scottinet: Indeed. But as discussed, this will have to be tackled in a separate PR.

lib/api/controllers/funnelController.js
scottinet: I don't remember changing this o_O

According to git blame, this has not changed since 2 months. Maybe a merge problem, I'm fixing that right away.

dbengsch: auto-merge problem ?

PR #654, opened 2017-02-02
Improve separation of concern on roleRepository::searchRole() method

ballinette:
`RoleRepository.prototype.searchRole` method used a full `Request` object as input parameter, while it only needs the list of controllers associated to the roles to search.

Now, we use only the needed parameters.

**Boyscout** : also add strict mode within files where it was missing (that broke the linter)


lib/api/core/models/repositories/roleRepository.js
```diff
@@ -129,28 +129,27 @@ RoleRepository.prototype.loadRole = function roleRepositoryLoadRole (roleId) {
...
  *
- * @param {Request} request
+ * @param {Array} controllers
+ * @param {Object} args
  */
-RoleRepository.prototype.searchRole = function roleRepositorySearchRole (request) {
-  var query = {query: {}};
-
-  request.input.body = request.input.body || {};
+RoleRepository.prototype.searchRole = function roleRepositorySearchRole (controllers, args) {
```
dbengsch: At this point of atomification, I would also extract `size` and `from` from `args` in the function signature.

ballinette: Done

PR #658, opened 2017-02-06
enable hot reload on plugins

stafyniaksacha:


docker-compose/config/pm2-dev.json
```diff
@@ -3,8 +3,8 @@
     "name"         : "KuzzleServer",
     "script"       : "bin/kuzzle",
     "args"         : ["start"],
-    "watch"        : ["lib", "config", "bin", "default.config.js"],
-    "ignore_watch" : ["node_modules"],
-    "node_args": "--inspect"
+    "watch"        : ["lib", "config", "bin", "default.config.js", "plugins/enabled"],
```
benoitvidis: nitpicking: If we are going to improve this section:

* is it worth watching the bin directory?
* shouldn't we, rather than having an ignore_watch, watch js files only (`["lib/**/*.js", "default.config.js", "plugins/enabled/**.js"]`)?

xbill82: not sure that we only want to watch `.js` files. Like, imagine a plugin that loads, for whatever reason, a `.json` config file -or another format (e.g. `.yml`) for parsing, it wouldn't be watched.
I know it doesn't really make sense, but who knows? Future use-cases are always mysterious (and sometimes completely stupid) but I'm not sure it's worth restraining the scope for the sake of minimalism.

xbill82:
Waiting for the discussion with @benoitvidis to be over before merging.

docker-compose/config/pm2-dev.json
scottinet: Hm. But what happens if a plugin use the filesystem for, say, lock a resource, open a unix socket, or log things in a file?

I'm more in favor of watching only .js files

scottinet: @benoitvidis, @xbill82, @stafyniaksacha > so, what do we decide?

PR #661, opened 2017-02-06
Protect against multi-index and multi-collection search requests

dbengsch:


lib/api/controllers/documentController.js
```diff
@@ -24,10 +24,10 @@ function DocumentController (kuzzle) {
   this.search = function documentSearch (request) {
     assertHasIndexAndCollection(request);
 
-    if (request.input.resource.index.split(',').length > 1) {
+    if (['_all', ',', '*', '+'].some(searchValue => request.input.resource.index.indexOf(searchValue) !== -1)) {
```
scottinet: With this method, I can't search in an index named, for instance: `mario_all_stars`.

I think this check cannot be done in 1 test, we need to perform 2:

`if (request.input.resource.index.match(/[,*+]/) !== null || request.input.resource.index === '_all')`

lib/api/controllers/documentController.js
```diff
@@ -24,10 +24,10 @@ function DocumentController (kuzzle) {
...
       throw new BadRequestError('document:search on multiple indexes is not available.');
     }
-    if (request.input.resource.collection.split(',').length > 1) {
+    if (['_all', ',', '*', '+'].some(searchValue => request.input.resource.collection.indexOf(searchValue) !== -1)) {
```
scottinet: Same remark as above.

dbengsch:
@scottinet done

PR #665, opened 2017-02-08
Dependencies update

scottinet:
The main purpose of this PR is to update Kuzzle so that it uses v1.0.3 of the `boost-geospatial-index` module (https://github.com/kuzzleio/boost-geospatial-index/releases/tag/1.0.3), fixing compilation problems on MacOS systems.

Since I was updating dependencies anyway, I thought I would do an overall upgrade. Little did I know that so many breaking changes would appear :-)

# Modules changes

* the new version of the `node-interval-tree` module can now only be imported using `import ... from ...`, which is not compatible with CommonJS. The `babel-polyfill` module, which ensured compatibility until now, is not sufficient anymore, and has been removed. The `reify` module has been added to add support for ES6 modules to Node.JS (if NodeJS ever supports ES6 modules, removing reify is a matter of a single line of code to delete)
* `istanbul`, used for code coverage calculation, is somewhat deprecated. This is not official (yet), but support and upgrades are now made on the new `istanbuljs` repo, and the old istanbul CLI has been replaced with the new `istanbuljs/nyc`. Replacing istanbul with nyc is mandatory to add support for ES6 modules, since istanbul parses code by itself, making `reify` useless in this case
* `eslint` now needs to be configured to scan sources as if they were for a `module`, instead of the default `script` configuration, to add support for ES6 modules
* remove the unused `fs-extra` and `portfinder` modules

# Fixes

* the new `eslint` version catches more problems. These have been fixed, especially the `invalid "this" usage` errors, forcing a minor refactor of the `WSBrokerClient` and `PluginsManager` and `security/roles` files
* fix an event leak in the internal broker client

# Boyscout

* add strict mode on updated files
* remove dead code related to node-inspector in closure sandboxes
* update `security/role` unit tests to our current standards, using the `KuzzleMock` object and replacing `rewire` with `mock-require` to allow the tested code to store modules in ES6 constants (/cc @dbengsch )


lib/api/dsl/storage/storeOperands.js
```diff
@@ -1,13 +1,10 @@
 'use strict';
 
-// Required to load node-interval-tree until Node implements the "import module" syntax
-require('babel-polyfill');
+import IntervalTree from 'node-interval-tree';
 
-// eslint-disable-next-line vars-on-top
 var
```
stafyniaksacha: `var` still here

lib/api/core/models/security/role.js
```diff
@@ -1,12 +1,17 @@
...
   Promise = require('bluebird'),
   async = require('async'),
   vm = require('vm'),
   BadRequestError = require('kuzzle-common-objects').errors.BadRequestError,
   InternalError = require('kuzzle-common-objects').errors.InternalError,
   ParseError = require('kuzzle-common-objects').errors.ParseError,
-  Request = require('kuzzle-common-objects').Request,
+  Request = require('kuzzle-common-objects').Request;
+
+// needs to be a variable so that unit tests can stub it
```
dbengsch: Nitpicking: Make the source code test compliant makes me feel sad.
We could use mock-require instead of rewire (https://www.npmjs.com/package/mock-require); It would at least limit these kind of cases (same for other files in the same case)

scottinet: Sure. I'll take a look at it.

PR #671, opened 2017-02-10
Protect document creation to avoid collision with http routes

dbengsch:
For instance, if a document is created with the _id , it would conflict with the _create route.
Underscores can't be the first character of an _id.

### DIFF
--- a/lib/api/controllers/util/requestAssertions.js
+++ b/lib/api/controllers/util/requestAssertions.js
@@ -80,7 +80,7 @@ module.exports = {
   /**
    * @param {Request} request
    */
-  assertIdStartsNotUnderscore: function requestAssertHasId (request) {
+  assertIdStartsNotUnderscore: function requestAssertIdStartsNotUnderscore (request) {
     if (request.input.resource._id && request.input.resource._id.charAt(0) === '_') {
       throw new BadRequestError(`${request.input.controller}:${request.input.action} must not specify an _id that starts with an underscore (_).`);
     }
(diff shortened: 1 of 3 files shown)


lib/api/controllers/util/requestAssertions.js
```diff
@@ -80,6 +80,15 @@ module.exports = {
   /**
    * @param {Request} request
    */
+  assertIdStartsNotUnderscore: function requestAssertHasId (request) {
```
scottinet: Well... Technically, an undefined or null ID does not start with an underscore.
So I personally would modify this method to return without throwing if the id is not defined, allowing to get rid of all the `if (request.input.resource.id) {` lines embedding this assertion call.

lib/api/controllers/util/requestAssertions.js
```diff
@@ -80,6 +80,15 @@ module.exports = {
...
+    if (request.input.resource._id.indexOf('_') === 0) {
```
scottinet: I propose a more efficient way to perform this check:

`if (request.input.resource.id.charAt(0) === '_') {`

lib/api/controllers/util/requestAssertions.js
ballinette: and rename `function requestAssertHasId ` (which is copied/pasted from the previous function) by `function requestAssertIdStartsNotUnderscore`

dbengsch:
@scottinet @ballinette done

PR #680, opened 2017-02-17
Standardization of the memory storage API

scottinet:
# Description

Some memory storage routes are either missing, have erroneous HTTP routes or have inconsistent parameters definition.

### Routes fixes

* `bitcount`: 
  * make `start` and `end` arguments optional, as stated by the Redis documentation, and moved them out of the `body` part of the request to make it more alike other routes like `bitpos`
  * the missing HTTP route has been added as a GET route
* `bitop`: remove erroneous HTTP route definition and changed the remaining one to make it more consistent with other defined routes
* `bitpos`: made the `bit` argument explicit for the HTTP route
* `del`: remove single key removal API route, as it was a duplicate of the multi-keys one
* `exists`: delete an erroneous mapping definition overwrite
* `geoadd`: make the key parameter required instead of optional, and rework the route to make it consistent with other multi-data routes
* `geohash`: change this route definition to 1/ make it work and 2/ make it consistent with other defined routes
* `geopos`: apply the same changes than for `geohash`
* `georadius`: better optional parameters handling
* `georadiusbymember`: better optional parameters handling
* `getbit`: make the `offset` argument explicit for the HTTP route
* `getrange`: make the `start` and `end` arguments explicit for the HTTP route
* `hdel`: remove the single field removal API route and keep only the multi-fields removal one
* `hexists`: add missing HTTP route
* `hmget`: better fields parameters handling
* `hmset`: make the creation of field-value pairs explicit and easier to understand
* `hscan`: add missing route
* `hstrlen`: fix invalid number of handled parameters
* `lrange`: make the `start` and `stop` arguments explicit for the HTTP route
* `lset`: rename the `idx` argument to `index`
* `mget`: fix HTTP route inconsistency
* `mset`: make the creation of key-value pairs explicit and easier to understand
* `object`: refactor the HTTP route definition to make it consistent with other ones
* `pfadd`: remove the differenciation between 1 and multiple elements to add
* `pfcount`: remove the differenciation between 1 and multiple elements to count
* `pfmerge`: harmonize this route with the rest of the API
* `rpushx`: add missing route
* `sadd`: remove the differenciation between 1 and multiple elements to add
* `scan`: add missing route
* `sdiff`: separate the key being compared with the list of keys to compare with for more clarity
* `sdiffstore`: make the source and comparing keys mandatory. Also remove the HTTP route without an explicit source key argument
* `set`: 
  * make the `value` argument required
  * throw a BadRequestError error if EX and PX are set, or if NX and XX are set, as these options are mutually exclusive
  * fix PX option handling (typo in the argument passed to redis)
  * add an explicit `_set` prefix to the HTTP route to prevent accidental keys creation
* `setnx`: add missing definition
* `sinter`: fix invalid HTTP route
* `sinterstore`: fix HTTP route and definition
* `smove`: remove erroneous HTTP route with no key identifier
* `sort`: 
  * remove the defaulting of the `offset` option to 0, as the documentation states that it should be explicitly set
  * group `count` and `offset` option in a new `limit` object, as these options can only be set together. If one of those options is set and not the other one, then a `BadRequestError` is returned
  * the `get` option now accepts multiple patterns, as stated by the redis documentation
* `sscan`: add missing route
* `sunion`: change the HTTP route from `POST` to `GET` and harmonize the definition with the rest of the API
* `touch`: add missing definition + HTTP route
* `zadd`:
  * throw an error if both `XX` and `NX` options are set, as they are mutually exclusive
  * throw an error if no score/member pair is provided
  * throw an error if more than 1 score/member pair is provided with the `incr` option set
  * booleans options are set if they are truthy, and not only if they are equal to boolean `true`
  * take multiple score/member pairs, as specified by the redis documentation
  * make the score/member pairs explicit, to prevent incomplete pairs to be provided
* `zcount`: harmonize the HTTP route definition with the rest of the API
* `zinterstore`:
  * the `_id` parameter is now the destination key, instead of a possible source key. Also remove the `destination` parameter in the body of the request
  * the `keys` parameter is now mandatory and must hold at least 1 key
  * remove the `weight` parameter, leaving only the `weights` one to define weights
  * skip the `weights` parameter if the input array is empty
  * the `aggregate` parameter now only accepts the following values (case insensitive): SUM, MIN, MAX
* `zlexcount`: harmonize the HTTP route definition with the rest of the API
* `zrange`: harmonize the HTTP route definition with the rest of the API
* `zrangebylex`: 
  * harmonize the HTTP route definition with the rest of the API
  * enforce the presence of `offset` if `count` is provided, and vice versa
* `zrevrangebylex`: apply the same changes than for `zrangebylex`
* `zrem`: rename `member` to `members` and allow the route to take multiple values, to comply with redis' documentation
* `zremrangebyrank`: add missing route
* `zrevrange`: apply the same modifications than for `zrange`
* `zscan`: add missing route

### Routes removals

The following routes have been removed, as these are considered either dangerous or incompatible with the way Kuzzle expose features to its clients:

* `bgrewriteaof` (dangerous: database administration)
* `bgsave` (dangerous: database administration)
* `blpop` (incompatible: blocking command)
* `brpush` (incompatible: blocking command)
* `brpoplpush` (incompatible: blocking command)
* `discard` (incompatible: transaction function)
* `dump` (incompatible: needs development to forward raw data from redis to client without altering the content)
* `exec` (incompatible: transaction function)
* `info` (too many information returned and redundant with `server:info`)
* `lastsave` (useless for client API)
* `multi` (incompatible: transaction function)
* `publish` (pubsub features deactivated)
* `restore` (incompatible until the `dump` command is fixed)
* `save` (dangerous: database administration + blocks clients)
* `setbit` (dangerous: reallocate potentially large amounts of memory, unless we implement safeguards)
* `setrange` (dangerous: reallocate potentially large amounts of memory, unless we implement safeguards)
* `unwatch` (incompatible: transaction function)
* `wait` (incompatible: blocking command)
* `watch` (incompatible: transaction function)

### Controller changes

* Remove the blacklist variable maintaining the list of non-exposed Redis commands, and replace it with a whitelist strategy, where only explicitly mapped routes are exposed
* If a parameter mapping hasn't an explicit `skip: true` option and a value is undefined, then a `BadRequestError` is thrown instead of succeeding silently and passing potentially erroneous parameters to redis (potentially leading to semantically correct commands but with unintended behaviors)

# Other changes

* the `assertBodyAttributeType` assertion erroneously accepted arrays as 'object' type
* unit tests on this controller now use the Kuzzle and Redis mock objects instead of the real ones
* add missing functional tests on geospatial capabilities

# Related Issues

* documentation: https://github.com/kuzzleio/documentation/pull/108


lib/api/controllers/memoryStorageController.js
```diff
@@ -666,40 +836,71 @@ function extractArgumentsFromRequestForZAdd (request) {
...
 
-    keys = keys.concat(request.input.body.keys);
+  if (request.input.body.keys.length === 0) {
+    throw new BadRequestError(`ms:${request.input.action} At least 1 source key must be provided`);
   }
 
-  args.push(keys.length);
-  args = args.concat(keys);
+  args.push(request.input.body.keys.length);
+  args.push(...request.input.body.keys);
```
dbengsch: This could lead to an unexpected error if the user input `keys` is big enough:
```
let
  a = [],
  b = [];

for (let i = 0; i < 1000000; i++) {
  a.push(i);
}

b.push(...a);

RangeError: Maximum call stack size exceeded
    at repl:1:3
    at sigintHandlersWrap (vm.js:22:35)
    at sigintHandlersWrap (vm.js:96:12)
    at ContextifyScript.Script.runInThisContext (vm.js:21:12)
    at REPLServer.defaultEval (repl.js:313:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.<anonymous> (repl.js:503:10)
    at emitOne (events.js:101:20)
    at REPLServer.emit (events.js:188:7)
```

scottinet: Should be prevented by request size limit. Even if it isn't, the exception will be catched and a proper error 500 will be returned to the client.

If we want to catch this, an arbitrary limit should be added. And not only to this API: every route taking an array should be checked. 

I don't feel like fixing this small issue right now :)

Issue #679, opened 2017-02-15
security:searchRole ignore roles using wildcarded controllers


scottinet:
The `security:searchRole` API route searches all roles related to a provided list of controllers.
As such, roles applied to all controllers (using a wildcard) should always be returned. Currently, they are ignored by the API route.

ballinette:
OK, it's not only the wildcarded controllers, but ALL roles that cannot be searched.
Because within the `role` mapping in the internal index, the `controllers` attribute is not searchable.

ballinette:
I've found a workaround with the elasticsearch mapping for roles, but really not fan of it... let's discuss about it tomorrow

ballinette:
Finally fixed with another approach : we fetch all roles from ElasticSearch, and then apply filters on the list (see #684 )

PR #684, opened 2017-02-17
Fix security:searchRole API route

ballinette:
# Related issue
#679 

# Description

The security:searchRole API route should search all roles related to a provided list of controllers.

But, since `role` collection's mapping is configured to not index role's controllers description, it was impossible to filter roles by controller's name with a simple ElasticSearch query.

Role controllers are not indexed in ES because, in their description, the field `action` can either be a boolean value, or a complex object representing a closure, so we cannot set a mapping for that.

This PR changes the way Kuzzle filters roles for `searchRole` action: now, Kuzzle fetches all roles from internal Engine (we expect that it is not a good practice to have more than a few dozen roles), and then filters them according to the controllers list given within the request.


features/HTTP.feature
```diff
@@ -236,19 +236,21 @@ Feature: Test HTTP API
 
   @usingHttp @cleanSecurity
   Scenario: Create/get/search/update/delete role
-    When I create a new role "role1" with id "test"
+    When I create a new role "foo" with id "test"
     Then I'm able to find a role with id "test"
-    And I update the role with id "test" with role "role2"
-    Then I'm able to find a role with id "test" with role "role2"
-    Then I'm able to find "1" role by searching controller corresponding to role "role2"
+    And I update the role with id "test" with role "bar"
```
xbill82: What does this sentence mean? My guess is "I apply all the properties of the role with id='bar' to the role with id='test'" is that right?

features/HTTP.feature
```diff
@@ -236,19 +236,21 @@ Feature: Test HTTP API
...
+    Then I'm able to find a role with id "test" with role "bar"
```
xbill82: I don't understand this sentence.

features/HTTP.feature
ballinette: Yes, it's exaclty that.
This sentence was still there, I just changed the role's name.

features/HTTP.feature
ballinette: It means "I'm able to find a role with id `test`, and the properties of this role are the same as role `bar` defined in `world.js`" (see https://github.com/kuzzleio/kuzzle/blob/fix-679-searchRole-ignore-wildcards/features/support/world.js#L134)

features/HTTP.feature
xbill82: :thinking: 
I understand now. I think we can do better. I'll eventually come up with a proposition.

features/HTTP.feature
xbill82: What about
"I am able to find a role with id "test" equivalent to role "bar""?

features/HTTP.feature
ballinette: what about `I update the role with id "test" with definition from role "bar"` ?

lib/api/core/models/repositories/roleRepository.js
```diff
@@ -130,27 +130,20 @@ RoleRepository.prototype.loadRole = function roleRepositoryLoadRole (roleId) {
...
  */
-RoleRepository.prototype.searchRole = function roleRepositorySearchRole (controllers, from, size) {
-  var query = {};
-
-  if (controllers && Array.isArray(controllers)) {
-    query.bool = {should: []};
-
-    controllers.forEach(controller => {
-      query.bool.should.push({exists: {field : 'controllers.' + controller}});
+RoleRepository.prototype.searchRole = function roleRepositorySearchRole (controllers) {
```
stafyniaksacha: Why we remove the `from` / `size` options here ? may still needed !

ballinette: Is it relevant to have more than 20-30 roles ? So, is it really necessary to implement pagination on it ?

stafyniaksacha: Ok, well see it if we need it later

features/HTTP.feature
scottinet: Proposal: `I update the role "test" with the test content "bar"`

ballinette: It sounds good for me.
@xbill82 , your advice ?

lib/api/core/models/repositories/roleRepository.js
scottinet: I have counter-arguments: 
* the default is set to 10 IIRC, and it seems quite a low limit to me
* it is implemented, it works, and it may be needed: what's the need behind this feature removal?

ballinette: In any case, we have to fetch all roles from ES, so if we want pagination, we need to sort the results and then return an subset according to pagination arguments.

In term of performances, returning a paginated content will not be faster than returning all the contents.

But I'm ok to re-add it.

ballinette: done

benoitvidis:
Hum.. I am not comfortable with this implementation.

If we are in a hurry, provided the tests are OK, we can apply it as a hot-fix but I would suggest a more robust solution.

I did not dig into the subject but at a first glance, it should be possible to index role description and not the action one.

If not, Kuzzle repositories provide all the tools we need to decouple the Javascript objects from their storage representation. It should be quite easy in any case to index the field we need.

features/HTTP.feature
xbill82: ok for @scottinet proposal

PR #686, opened 2017-02-20
Add extraParams argument to getElasticsearchRequest

### LINKED ISSUE #625: Error DeleteDocument with refresh
```
kuzzle
  .dataCollectionFactory('collection', 'index')
  .deleteDocumentPromise({query: {ids: {values: ids}}}, {refresh: 'wait_for'})
```

The error is `legal_argument_exception] request [/mycollection/otherCollection/_search] contains unrecognized parameter: [refresh]`
This is because we do a search before the delete and we put option with `refresh` and this is not supported by search method.

### LINKED ISSUE #682: getElasticsearchRequest in elasticsearch service should only return useful argument for the current request action
When it is called with extra parameters, they are all transmitted to the elasticsearch request, even if they are not useful (and potentially make the request fail). `['from', 'size', 'scroll', 'scrollId', 'refresh']` should only be transmitted to the request when they are useful.

xbill82:
# Description
So far, `getElasticsearchRequest` just checked for a hardcoded list of extra-params in the Request to clone to the resulting ES request. Yet, some of these params were not good for certain ES action, resulting in the action to fail.
The `getElasticsearchRequest` function now accepts an optional `extraParams` argument that specifies which of these parameters are actually desired in the resulting ES request. This way, the caller is in charge of whitelisting the desired parameters based on the ES action to be called with the resulting ES request.

If the `extraParams` argument is not specified or left empty, no extra param will be added to the resulting ES request.

# Related Issues
Fixes #682 
Fixes #625

### DIFF
--- a/lib/services/elasticsearch.js
+++ b/lib/services/elasticsearch.js
@@ -272,7 +272,7 @@ function ElasticSearch (kuzzle, options, config) {
    * @returns {Promise} resolve an object that contains _id
    */
   this.create = function elasticsearchCreate (request) {
-    var esRequest = getElasticsearchRequest(request, kuzzle);
+    var esRequest = getElasticsearchRequest(request, kuzzle, ['refresh']);
 
     if (esRequest.body._routing) {
       return Promise.reject(new BadRequestError('Kuzzle does not support "_routing" in create action.'));
@@ -347,7 +347,7 @@ function ElasticSearch (kuzzle, options, config) {
    * @returns {Promise} resolve an object that contains _id
    */
   this.createOrReplace = function elasticsearchCreateOrReplace (request) {
-    var esRequest = getElasticsearchRequest(request, kuzzle);
+    var esRequest = getElasticsearchRequest(request, kuzzle, ['refresh']);
 
     if (esRequest.body._routing) {
       return Promise.reject(new BadRequestError('Kuzzle does not support "_routing" in createOrReplace action.'));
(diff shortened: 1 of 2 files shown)


xbill82:
Er... Did I go on-fire with labels or are they pertinent? I can't figure out if this really introduces breaking changes.

lib/services/elasticsearch.js
```diff
@@ -982,7 +986,8 @@ function getElasticsearchRequest(request, kuzzle) {
   }
 
   ['from', 'size', 'scroll', 'scrollId', 'refresh'].forEach(argumentName => {
-    if (typeof request.input.args[argumentName] !== 'undefined') {
+    if (typeof request.input.args[argumentName] !== 'undefined' &&
+        extraParams.includes(argumentName)) {
```
ballinette: should it be simply replaced by 
```
extraParams.forEach(argumentName => {
  if (typeof request.input.args[argumentName] !== 'undefined') {
    data[argumentName] = request.input.args[argumentName];
  }
});
```

scottinet:
And you indeed went a little overboard with labels: I don't think any documentation is required, and there is no breaking change on this one.

lib/services/elasticsearch.js
xbill82: :thinking:  Interesting point.
Your solution gives full responsibility on the consistency of the extra params to the caller. Which does make sense since we are in the Elasticsearch Service, which is supposed to be aware of the ES API.
Moreover, the hardcoded list is implicit, which means that some extra parameters specified by the caller may be stripped out without this being reflected by the contract specified in the function signature. Which is crappy.

I like your proposition.

xbill82: But it will break my unit tests so I hate you. But your solution is good. But it will break my unit tests... (put this into a `while (1)`)

scottinet: You can still apply @ballinette's solution without breaking unit tests, simply test if `extraParams` is defined prior to iterate over it.

PR #538, opened 2016-11-15
protect internal index in elasticsearch service

stafyniaksacha:
Protect usage of `storageEngine` to manipulate what happen in `internalIndex`

## Breaking changes
* Actions like `write:deleteFromQuery`, `read:search`, `admin:refreshIndex` with `index: %kuzzle` are now discarded with a `BadRequestError`

## Changes
* Added a route `admin:searchSpecifications` to avoid using `read:search`
* Create/Update/Delete actions on `internalEngine` are now send with `refresh=true` to ES
* Rewrite functional tests (`cleanSecurity` & `cleanValidation`)
* Changed default mapping for internal collections (users / profiles / roles / specifications) to match with ES 5.x

### Todo (other PR)
* Remove `allowInternalIndex` from policies (now it's obsolete)
* Add a way to delete at least users (maybe also validation specification, profiles and roles) from a query

PR #689, opened 2017-02-22
remove usage of `allowInternalIndex` within policies

### LINKED ISSUE #687: Relicate of allowInternalIndex in default configuration
In the file `default.config.js`, there is a remaining of the `allowInternalIndex` mechanism.

ballinette:
fix #687 
`allowInternalIndex`property in role's description is no longer needed. 
All actions to the internal Engine are now be denied (see #538)

### DIFF
--- a/lib/api/core/models/security/role.js
+++ b/lib/api/core/models/security/role.js
@@ -17,9 +17,8 @@ const
 function Role () {
   this.controllers = {};
 
-  // closures, allowInternalIndex and restrictedTo are computed for internal use only.
+  // closures and restrictedTo are computed for internal use only.
   this.closures = {};
-  this.allowInternalIndex = false;
 
   // Injected by Profile.getRoles, contains a profile's policies
   this.restrictedTo = [];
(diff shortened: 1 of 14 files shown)


lib/api/core/models/security/role.js
```diff
@@ -45,11 +44,6 @@ Role.prototype.isActionAllowed = function roleIsActionAllowed (request, kuzzle)
     promises = [],
     path = [];
 
-  // explicit actions on internalIndex must be denied if no explicit restriction on internalIndex is set:
-  if (request.input.resource.index === kuzzle.internalEngine.index && !this.allowInternalIndex) {
```
dbengsch: wasn't this check useful to disallow the internalEngine index to be accessed. I think this check is not done anywhere else.

ballinette: Not needed anymore here, as this check is implemented within `storageEngine` since #538 (see https://github.com/kuzzleio/kuzzle/pull/538/files#diff-8927ec75654a199501c9220ef47f3b5cR982 )

dbengsch: OK, sorry, I did recall this one. Good to me then

PR #706, opened 2017-03-07
Add support for HTTP HEAD requests

scottinet:
# Description

Adds a default HTTP route on URL `/` with the HEAD verb. This route answers with default headers.

### DIFF
--- a/lib/api/core/httpRouter/index.js
+++ b/lib/api/core/httpRouter/index.js
@@ -24,9 +24,9 @@ function Router(kuzzle) {
   this.kuzzle = kuzzle;
 
   this.defaultHeaders = {
-    'Content-Type': 'application/json',
+    'content-type': 'application/json',
     'Access-Control-Allow-Origin': kuzzle.config.http.accessControlAllowOrigin || '*',
-    'Access-Control-Allow-Methods': 'GET,POST,PUT,DELETE,OPTIONS',
+    'Access-Control-Allow-Methods': 'GET,POST,PUT,DELETE,OPTIONS,HEAD',
     'Access-Control-Allow-Headers': 'Content-Type, Access-Control-Allow-Headers, Authorization, X-Requested-With'
   };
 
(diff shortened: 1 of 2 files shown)


lib/api/core/httpRouter/index.js
```diff
@@ -24,19 +24,27 @@ function Router(kuzzle) {
   this.kuzzle = kuzzle;
 
   this.defaultHeaders = {
-    'Content-Type': 'application/json',
+    'content-type': 'application/json',
```
ballinette: why `content-type` in lower case ?

ballinette:
Why is it considered as a "bug" ?

Anyway, despite my little question about the lowercase of header name, it's OK for me.

lib/api/core/httpRouter/index.js
scottinet: It has no impact since headers are case insensitive. I lowercased it in the HTTP router default headers so that it can be directly compared with headers handled by the `RequestResponse` object, greatly simplifying unit tests.

scottinet:
@benoitvidis and I decided to use the `bug` label on this one because we should have handled `HEAD` requests in the first place. So, it's really more a fix than an enhancement.

lib/api/core/httpRouter/index.js
ballinette: OK

ballinette:
hmm I still don't understand...  As far as I know, `HEAD` requests are not mandatory by HTTP protocol.
But the choice of using or not the`bug` label is not as important, of course ;)

scottinet:
This fixes this awkward situation:

```
$ curl -I http://localhost:7512
HTTP/1.1 400 Bad Request
content-type: application/json
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET,POST,PUT,DELETE,OPTIONS
Access-Control-Allow-Headers: Content-Type, Access-Control-Allow-Headers, Authorization, X-Requested-With
Date: Wed, 08 Mar 2017 11:16:18 GMT
Connection: keep-alive
```

;-)

xbill82:
Can we merge this?

Issue #707, opened 2017-03-08
Cannot read property _kuzzle_info of undefined


scottinet:
When using the `document:search` route with the following document body:

```json
{
  "stored_fields": []
}
```

Kuzzle responds with the following internal error: 

```
 "Error: Cannot read property '_kuzzle_info' of undefined
    at ElasticSearch.elasticsearchFormatESError [as formatESError] (/var/app/lib/services/elasticsearch.js:948:23)
    at client.search.then.catch.error (/var/app/lib/services/elasticsearch.js:194:43)
    at tryCallOne (/var/app/node_modules/promise/lib/core.js:37:12) 
    at /var/app/node_modules/promise/lib/core.js:123:15
    at flush (/var/app/node_modules/asap/raw.js:50:29)
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickDomainCallback (internal/process/next_tick.js:122:9)"
```

scottinet:
Also occurs while passing the following property in the search body: `"_source": false` (works fine with any other `_source` value)

PR #708, opened 2017-03-08
[HOTFIX] Support search results with no _source entries

scottinet:
# Description

When passing options to Elasticsearch disabling the `_source` retrieval while performing search queries (e.g. with `stored_fields: []` or with `_source: false`), Kuzzle responds with an `InternalError` error, because it tries to fetch information from `_source._kuzzle_info`.

This PR solves this problem by handling Kuzzle's metadata in search results only if there is a `_source` property present.

# Solved issue

#707


scottinet:
Not so much: the only workaround available to exclude the `_source` part of searched documents would be to add a `_source: ['notExistingField']` parameter to the search request, and this looks dirty to me.

PR #709, opened 2017-03-08
[HOTFIX] Fix document.scroll method

scottinet:
# Description

The `document.scroll` method is now a HTTP GET method, with the following format:

`GET http://<server>:<port>/_scroll/<scrollId>[?scroll=<scroll ttl>]`

The `scroll` parameter is now optional, and is defaulted to a value taken in Kuzzle configuration (path: `services.db.defaults.scrollTTL`)
The default has been set to 15s.

# Boyscout

* Rephrase the functional test case for the scroll API route
* Convert the websocket and HTTP functional tests APIs to ES6

### DIFF
--- a/lib/api/controllers/documentController.js
+++ b/lib/api/controllers/documentController.js
@@ -49,7 +49,7 @@ function DocumentController (kuzzle) {
    * @returns {Promise<Object>}
    */
   this.scroll = function documentScroll (request) {
-    if (!request.input.args.scroll) {
+    if (!request.input.body || !request.input.body.scroll) {
       throw new BadRequestError('document:scroll must specify an argument "scroll"');
     }
     if (!request.input.args.scrollId) {
--- a/lib/api/controllers/documentController.js
+++ b/lib/api/controllers/documentController.js
@@ -17,7 +17,7 @@ let
  * @constructor
  */
 function DocumentController (kuzzle) {
-  var engine = kuzzle.services.list.storageEngine;
+  const engine = kuzzle.services.list.storageEngine;
 
   /**
    * @param {Request} request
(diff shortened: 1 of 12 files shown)


lib/api/controllers/documentController.js
```diff
@@ -49,7 +49,7 @@ function DocumentController (kuzzle) {
    * @returns {Promise<Object>}
    */
   this.scroll = function documentScroll (request) {
-    if (!request.input.args.scroll) {
+    if (!request.input.body || !request.input.body.scroll) {
       throw new BadRequestError('document:scroll must specify an argument "scroll"');
     }
     if (!request.input.args.scrollId) {
```
ballinette: If I follow the documentation, we should also replace `request.input.args.scrollId` by `request.input.body.scrollId` ?

ballinette: Or to follow the action's documentation for HTTP protocol, this should be 
`request.input.resource.scrollId`

and we should fix the API documentation for other protocols with following:
```javascript
{
  "controller": "document",
  "action": "scroll",
  "scrollId": "<scrollId>"
  "body": {
    // The scroll parameter tells Elasticsearch to keep the search context open for another 1m.
    "scroll": "1m"
  }
}

```

scottinet:
After heated discussions about the place of `scroll` parameter in HTTP requests, it has been decided that we'll transform this API route into a HTTP GET one, with the `scroll` parameter being an optional argument. If not present, Kuzzle will set it with a default value taken from its configuration.

scottinet:
Done. I'll synchronize with @dbengsch to check how I can update the documentation without impacting the changes he made.

/cc @ballinette, @benoitvidis

PR #721, opened 2017-03-16
Replace MD5 with HighwayHash

scottinet:
# Description

Replace hash-generated values from MD5 to [HighwayHash](https://github.com/google/highwayhash), which has 2 advantages over MD5:

* protect Kuzzle from hash attacks, where crafted real-time filters might generate the same hashes than existing ones (very unlikely as provided filters are completely rewritten as minimized versions using Espresso)
* more than 10x faster than MD5

As HighwayHash requires a 256 bits key, a new `hashKey` property has been added to the kuzzle main object, and a new process has been added to the start sequence:

* generate a new random hash key, and set it in redis only if one does not already exist
* if redis answers with a "key already exists" return value, return the redis stored hash key
* otherwise, use the newly generated hash key

This allows Kuzzle cluster nodes to share the same hash values for their objects. 
(note: using redlock might be a good idea for this one, to make key sync more robust on a redis cluster)

### DIFF
--- a/test/api/core/hotelClerk/addToChannels.test.js
+++ b/test/api/core/hotelClerk/addToChannels.test.js
@@ -1,4 +1,5 @@
-var
+require('reify');
+const
   should = require('should'),
   Request = require('kuzzle-common-objects').Request,
   Dsl = require('../../../../lib/api/dsl'),
@@ -7,7 +8,7 @@ var
   KuzzleMock = require('../../../mocks/kuzzle.mock');
 
 describe('Test: hotelClerk.addToChannels', () => {
-  var
+  let
     kuzzle,
     context = {
       connection: {id: 'connectionid'},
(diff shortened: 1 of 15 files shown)


test/api/core/hotelClerk/addToChannels.test.js
```diff
@@ -1,4 +1,5 @@
-var
+require('reify');
```
ballinette: what is the goal of `refiy` ?

scottinet: It allows NodeJS to correctly interpret ES6 class import commands (`import ... from ...`).
Normally there is only 1 instance of reify required at the start of Kuzzle, but I found while working on this PR that some unit tests files also need it to allow running them independantly.

Issue #731, opened 2017-03-21
Do not generate core dump when elasticsearch raise unmanaged errors


stafyniaksacha:
Today we [generate a standard `Error` object](https://github.com/kuzzleio/kuzzle/blob/master/lib/services/elasticsearch.js#L944) when we get unmanaged error from elasticsearch, which cause Kuzzle to crash.

We have to implement a new error type in [kuzzle-common-objects](https://github.com/kuzzleio/kuzzle-common-objects) to handle this kind of errors (eg: `ExternalServiceError`)

scottinet:
Requalified this issue, as generating standard Error objects does not cause a crash.
It only causes a dump generation with default diagtools settings, which might fill up disk space if these kind of errors occur too frequently.

Issue #730, opened 2017-03-21
Manage version conflict error from elasticsearch cluster 


stafyniaksacha:
We have to [include a mapping](https://github.com/kuzzleio/kuzzle/blob/master/lib/services/elasticsearch.js#L46) for version conflict error in elasticsearch to help developers to investigate when it happen.

> Raw elasticsearch error:
```js
{
    status: 409,
    displayName: 'Conflict',
    message: '[version_conflict_engine_exception] [data][AVrbg0eg90VMe4Z_dG8j]: version conflict, current version [153] is different than the one provided [152], with { index_uuid="iDrU6CfZSO6CghM1t6dl0A" & shard="2" & index="userglobaldata" }',
    path: '/userglobaldata/data/AVrbg0eg90VMe4Z_dG8j/_update',
    query: {},
    body: { error: [Object], status: 409 },
    statusCode: 409,
    response: '{"error":{"root_cause":[{"type":"version_conflict_engine_exception","reason":"[data][AVrbg0eg90VMe4Z_dG8j]: version conflict, current version [153] is different than the one provided [152]","index_uuid":"iDrU6CfZSO6CghM1t6dl0A","shard":"2","index":"userglobaldata"}],"type":"version_conflict_engine_exception","reason":"[data][AVrbg0eg90VMe4Z_dG8j]: version conflict, current version [153] is different than the one provided [152]","index_uuid":"iDrU6CfZSO6CghM1t6dl0A","shard":"2","index":"userglobaldata"},"status":409}',
    toString: [Function],
    toJSON: [Function]
}

```

PR #723, opened 2017-03-16
Implement the PluginContext "storage" accessor

dbengsch:
:warning: Depends on #720

**Breaking changes:**

* Doesn't list internal indexes in `index:list` anymore (index starting with `%`).

**Feature:**

* Add a new accessor `storage` in the `PluginContext`; it contains helpers to use the `InternalEngine` that points to the plugin's internal index named `'%<pluginName>'`;
* Creation of a new plugin repository;
* Add a new constructor `Repository` in the `PluginContext`; it instantiates a `PluginRepository` and provide an adapter to avoid direct accesses to the instantiated object;
* Modify `Repository` constructor to add a new argument `index`, defaulted to `kuzzle.internalEngine.index`;
* Modify `InternalEngine` constructor to add a new argument `index`, defaulted to `'%kuzzle'`;
* Prevent the use of document API (`elasticsearch` service) for internal indexes (every indexes starting with `%`).

**Boyscout:**

* Replace var by let and const in most modified files;
* Improvement in JSDoc in most modifed files;
* Switch from function contructors to ES6 class paradigm in modified files;
* Increase test coverage of some non-tested code;
* Replace old tests that was still using a real instance of `Kuzzle` and inline mocks by `KuzzleMock`.

PR #734, opened 2017-03-22
Improve Elasticsearch errors handling

scottinet:
# Description

* Instead of returning a `KuzzleError` object on an unhandled Elasticsearch error, causing Kuzzle to dump itself when diagtools are activated, returns a `ExternalServiceError` 
* Add a new ES error mapper handling errors specific to document update conflicts
* Expose a new `onUpdateConflictRetries` option in Kuzzle configuration, taking an integer, and applying the `retryOnConflict` Elasticsearch option on update queries if no `retryOnConflict` argument is provided
* The document update route now accepts the `retryOnConflict` option, allowing clients to configure Elasticsearch's behavior on specific update queries (/cc @stafyniaksacha )

# Boy scout

* Bring the Elasticsearch service implementation in line with today's coding standards

# Solved issues

#730, #731

### DIFF
--- a/lib/services/elasticsearch.js
+++ b/lib/services/elasticsearch.js
@@ -15,6 +15,69 @@ const
   es = require('elasticsearch'),
   compareVersions = require('compare-versions');
 
+const errorMessagesMapping = [
+  {
+    // [illegal_argument_exception] object mapping [titi] can't be changed from nested to non-nested
+    regex: /^\[illegal_argument_exception] object mapping \[(.*?)] can't be changed from nested to non-nested$/,
+    replacement: 'Can not change mapping for field "$1" from nested to another type'
+  },
+  {
+    // [illegal_argument_exception] object mapping [baz] can't be changed from non-nested to nested
+    regex: /^\[illegal_argument_exception] object mapping \[(.*?)] can't be changed from non-nested to nested$/,
+    replacement: 'Can not change mapping for field "$1" from object to another type'
+  },
+  {
+    // [illegal_argument_exception] Can't merge a non object mapping [aeaze] with an object mapping [aeaze]
+    regex: /^\[illegal_argument_exception] Can't merge a non object mapping \[(.*?)] with an object mapping \[(.*?)]$/,
+    replacement: 'Can not change mapping for field "$1" from object to a scalar type'
+  },
+  {
+    // [illegal_argument_exception] [tutu.tutu] is defined as an object in mapping [aze] but this name is already used for a field in other types
+    regex: /^\[illegal_argument_exception] \[(.*?)] is defined as an object in mapping \[(.*?)] but this name is already used for a field in other types$/,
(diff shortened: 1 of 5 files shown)


scottinet:
Nevermind the coverage report: this ES service version has less lines due to the ES 2015 refactor

lib/services/elasticsearch.js
```diff
@@ -772,56 +769,50 @@ function ElasticSearch (kuzzle, options, config) {
...
     return this.client.indices.getMapping()
       .then(result => {
-        indexes = Object.keys(result);
-        indexes = indexes.filter(indexName => {
-          // @todo : manage internal index properly
-          // exclude empty results
-          return indexName !== '';
-        });
+        let indexes = Object.keys(result)
+          .filter(indexName => indexName !== ''); // @todo : manage internal index properly
```
ballinette: There will be merge conflicts here withPR #723 :
https://github.com/kuzzleio/kuzzle/pull/723/files#diff-8927ec75654a199501c9220ef47f3b5cR789

scottinet: Indeed, but from the look of it, it should take only seconds to solve these.

lib/services/elasticsearch.js
```diff
@@ -389,22 +381,27 @@ function ElasticSearch (kuzzle, options, config) {
...
       }
       if (esRequest.refresh !== 'wait_for' && esRequest.refresh !== 'false' && esRequest.refresh !== false) {
         return Promise.reject(new BadRequestError('Refresh parameter only supports the value "wait_for" or false'));
       }
     }
 
+    // retryOnConflict configuration
+    if (this.config.defaults.onUpdateConflictRetries > 0) {
+      esRequest.retryOnConflict = this.config.defaults.onUpdateConflictRetries;
+    }
```
stafyniaksacha: Can we have a way to specify this setting through an API parameter (like `refresh`)

scottinet: I know I said "no" to you yesterday, but come to think of it, this is a good idea.

I'll add this feature that right away.

scottinet:
@stafyniaksacha , @ballinette > I updated this PR by allowing `update` to accept the `retryOnConflict` option, you may want to review that new part

lib/services/elasticsearch.js
```diff
@@ -958,14 +985,14 @@ module.exports = ElasticSearch;
...
+ * @param {Array} [extraParams] [optional] An array of String values corresponding
  *                            to the extra params from the Kuzzle Request to be
  *                            included in the Elasticsearch Request.
  * @return {object} data the data with cleaned attributes
  */
 function getElasticsearchRequest(request, kuzzle, extraParams) {
-  var data = {};
-  extraParams = extraParams || new Array();
+  const data = {};
+  extraParams = extraParams || [];
```
dbengsch: We could default `extraParams` in the function signature.

PR #748, opened 2017-03-28
Enable to setup kuzzle with a Elasticsearch cluster

### LINKED ISSUE #744: Enable Elasticsearch cluster in service configuration
If we use a cluster of elasticsearch, we should be able to init ES client with an array of hosts instead of a single one.

ballinette:
We can now configure kuzzle to use 
* either a single Elasticsearch server:
```javascript
db: {
// (...)
  host: 'elasticsearch',
  port: 9200,
// (...)
}
```

* or a Elasticsearch cluster, if we give an array of nodes, using [ES Client Hosts parameters](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/host-reference.html):
```javascript
db: {
// (...)
  hosts: ['elastic1:9200', 'elastic2:9200']
// (...)
}

db: {
// (...)
  hosts: [
    {"host": "elastic1"},
    {"host": "elastic2", port: 9300, protocol: 'https'}
  ]
// (...)
}

```

Fix #744


AnthonySendra:
Hello @ballinette. What if an Elasticsearch node crash? All requests are forwarded only on the other one?

ballinette:
Hi @AnthonySendra 
Yes, when the ES client has a connection error to a node, it automatically switches to another one

Note: we could also add the support of options like `sniffInterval` and `sniffOnConnectinoFault` to maintain an up-to-date cluster state (see https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/configuration.html )

ballinette:
@benoitvidis totally agree with that ;)
We just need this quick enhancement now without introducing breaking changes, but I agree that it would be far better to suit the ES client configuration format

PR #749, opened 2017-03-29
Wait plugin initialization with promises

dbengsch:
**Enhancements:**

* Wait plugin initialization with promises
* Allow all http valid methods in plugin routes

### DIFF
--- a/lib/api/core/plugins/pluginsManager.js
+++ b/lib/api/core/plugins/pluginsManager.js
@@ -117,7 +117,7 @@ function PluginsManager (kuzzle) {
         initTimeout = this.config.common.initTimeout;
 
       if (plugin.config.threads) {
-        debug('[%s] starting worker with %d treads', pluginName, plugin.config.threads);
+        debug('[%s] starting worker with %d threads', pluginName, plugin.config.threads);
 
         return initWorkers(plugin, pluginName, this.kuzzle.config)
           .timeout(initTimeout)
--- a/lib/api/core/plugins/pluginsManager.js
+++ b/lib/api/core/plugins/pluginsManager.js
@@ -402,6 +402,7 @@ function initControllers (controllers, routes, plugin, pluginName) {
       return true;
     });
   });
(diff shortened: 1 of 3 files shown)


lib/api/core/plugins/pluginsManager.js
```diff
@@ -98,70 +98,75 @@ function PluginsManager (kuzzle) {
...
-          debug('[%s] starting plugin in %s mode', pluginName, plugin.config.privileged ? 'privileged' : 'standard');
+    promises = Object.keys(this.plugins).map((pluginName) => {
+      const
+        plugin = this.plugins[pluginName],
+        pipeWarnTime = this.config.common.pipeWarnTime,
+        pipeTimeout = this.config.common.pipeTimeout,
+        initTimeout = this.config.common.initTimeout;
+
+      if (plugin.config.threads) {
+        debug('[%s] starting worker with %d treads', pluginName, plugin.config.threads);
```
jenow: typo on "treads"

lib/api/core/plugins/pluginsManager.js
```diff
@@ -431,9 +439,9 @@ function initControllers (controllers, routes, plugin, pluginName) {
           return false;
         }
 
-        if (['post', 'get'].indexOf(route.verb.toLowerCase()) === -1) {
+        if (['get', 'head', 'post', 'put', 'delete', 'connect', 'options', 'trace', 'patch'].indexOf(route.verb.toLowerCase()) === -1) {
```
benoitvidis: is the `httpRouter` already able to handle all these methods? As far as I can see, only `get`, `post`, `put` and `delete` are accepted.

Further, if I am not mistaken, the `options` verb has a specific behaviour within kuzzle &/|| the proxy.

scottinet: The httpRouter now handles `head`, but you're right for the other methods.

dbengsch: @benoitvidis Done

Issue #742, opened 2017-03-24
Impossible to fetch more than 10K KuzzleUser


BaptisteKaliop:
Hello,

Since `%kuzzle` is an internal index, it's impossible to use the search/scroll action from document controller on this index.
So, the only option here is to retrieve users with securityController/searchUsers. But seeing https://github.com/kuzzleio/kuzzle/blob/master/lib/services/internalEngine/index.js#L69, this action does not use ES's scroll API so it will be impossible to retrieve more than 10K users.

PR #750, opened 2017-03-29
Add methods scrollUsers, scrollProfiles and scrollSpecifications

scottinet:
# Description

* The following routes now accept a `scroll` argument, making these searches scrollable: `security:searchUsers`, `security:searchProfiles`, `collection:searchSpecifications`
* Add corresponding `scroll...` methods: `security:scrollUsers`, `security:scrollProfiles`, `collection:scrollSpecifications`
* Scroll IDs generated by various search methods can only be used by the corresponding scroll method, meaning that the scroll IDs obtained by, for instance, `document:search` cannot be used by `security:scrollProfiles`.

# Other changes

* Search functions in repositories classes now accept a `options` optional argument (type: object), instead of `from` and `size`. This optional object may have the following properties: `from`, `size`, `scroll`
* Bring impacted source files to today's coding standards (ES6, classes, avoiding uses of .call and .bind)
* Add missing `collection:searchSpecifications` functional tests
* Update the `ms` module to v1, the main reason being that this version is now accurately documented using jsdoc

# Notes

The `security:searchRoles` has not been impacted by these changes because the `%kuzzle/roles` collection is not searchable.

# Solved issue

#742 

# Documentation

https://github.com/kuzzleio/documentation/pull/177


scottinet:
@ballinette > Good catch. I performed a global search to check if I haven't missed other calls, it seems that you got the only one left. :+1:

Issue #740, opened 2017-03-23
Error when removing room for customer leaves the room marked as destroyed


benoitvidis:
Following a client disconnection:

```
kuzzle_1         | 0|KuzzleSe | Unhandled rejection TypeError: Cannot read property 'collection' of undefined
kuzzle_1         | 0|KuzzleSe |     at removeFromTestTables (/var/app/lib/api/dsl/storage/index.js:427:29)
kuzzle_1         | 0|KuzzleSe |     at Storage.remove (/var/app/lib/api/dsl/storage/index.js:210:3)
kuzzle_1         | 0|KuzzleSe |     at RealtimeEngine.remove (/var/app/lib/api/dsl/index.js:96:23)
kuzzle_1         | 0|KuzzleSe |     at HotelClerk.cleanUpRooms (/var/app/lib/api/core/hotelClerk.js:468:28)
kuzzle_1         | 0|KuzzleSe |     at HotelClerk.removeRoomForCustomer (/var/app/lib/api/core/hotelClerk.js:534:23)
kuzzle_1         | 0|KuzzleSe |     at async.each (/var/app/lib/api/core/hotelClerk.js:156:31)
kuzzle_1         | 0|KuzzleSe |     at /var/app/node_modules/async/dist/async.js:3025:16
kuzzle_1         | 0|KuzzleSe |     at eachOfArrayLike (/var/app/node_modules/async/dist/async.js:941:9)
kuzzle_1         | 0|KuzzleSe |     at eachOf (/var/app/node_modules/async/dist/async.js:991:5)
kuzzle_1         | 0|KuzzleSe |     at Object.eachLimit (/var/app/node_modules/async/dist/async.js:3089:3)
kuzzle_1         | 0|KuzzleSe |     at Promise (/var/app/lib/api/core/hotelClerk.js:155:13)
kuzzle_1         | 0|KuzzleSe |     at HotelClerk.hotelClerkRemoveCustormerFromAllRooms [as removeCustomerFromAllRooms] (/var/app/lib/api/core/hotelClerk.js:154:12)
```

At a first glance, besides the error handling problem, it looks like  [a wrong argument is passed from the HotelClerk to the DSL](https://github.com/kuzzleio/kuzzle/blob/master/lib/api/core/hotelClerk.js#L465), which expects a filter id.

scottinet:
Nice catch!

Found a way to reproduce it 100% of the time. Here are the steps:

[...]

PR #751, opened 2017-03-29
TypeError exception when removing a subscription

benoitvidis:
# Description

Fix the case where an unhandled Exception was raised when trying to deregister a subscription which would contain only existing test conditions at the time it was created.

# Related issue

* #740

### DIFF
--- a/lib/api/core/hotelClerk.js
+++ b/lib/api/core/hotelClerk.js
@@ -366,7 +366,7 @@ function createRoom (index, collection, filters) {
         return {diff, roomId};
       }
 
-      return Promise.fromNode(asyncCB => async.retry({interval: 200}, callback => {
+      return Promise.fromNode(asyncCB => async.retry(callback => {
         // if the room is about to be destroyed, we have to delay its re-registration until its destruction has completed
         if (this.rooms[roomId] && this.rooms[roomId].destroyed) {
           return callback(new InternalError(`Cannot create the room ${roomId} because it has been marked for destruction`));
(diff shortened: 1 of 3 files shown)


lib/api/dsl/storage/index.js
```diff
@@ -14,237 +14,244 @@ const
...
-  let result = addFilter(this.hashKey, this.filters, index, collection, filters);
+          this.storeOperand[cond.keyword](this.foPairs[index][collection][cond.keyword], subfilter, cond);
+        });
+      }
+    });
 
-  if (!result.created) {
-    return {diff: false, id: result.id};
-  }
+    // ref https://github.com/kuzzleio/kuzzle/issues/740
```
benoitvidis: The actual fix is these 4 lines.

lib/api/core/hotelClerk.js
```diff
@@ -604,7 +604,7 @@ function subscribeToRoom (roomId, request) {
           m: request.input.metadata
         }};
 
-      if (!this.customers[request.context.connectionId] || !this.customers[request.context.connectionId][roomId]) {
+      if (!this.customers[request.context.connectionId] || this.customers[request.context.connectionId][roomId] === undefined) {
```
ballinette: can `this.customers[request.context.connectionId]` be falsy and not `undefined` ?
I mean: can we write this here: 
```javascript
if (this.customers[request.context.connectionId] === undefined || this.customers[request.context.connectionId][roomId] === undefined) { 
```
?

PR #774, opened 2017-04-04
Emit UnauthorizedError when a valid token link to a removed user

### LINKED ISSUE #773: Error on isActionAllowed when user is deleted and token still available
When you delete an user, but the token is still available (so there is no more userId linked to this token) and you try to do actions with this token, the funnel controller triggers an error : 

```
error: 2017-04-04T09:32:58+02:00 [LOG:ERROR] Cannot read property 'isActionAllowed' of null
1|kuzzle   | TypeError: Cannot read property 'isActionAllowed' of null
1|kuzzle   |     at kuzzle.repositories.token.verifyToken.then.then.user (/home/elastic/kuzzle/kuzzle/lib/api/controllers/funnelController.js:237:20)
```

Since : ```return kuzzle.repositories.user.load(request.context.token.userId);``` can't load the user

BaptisteKaliop:
This fix issue #773 when user is deleted but the token linked to this user is still available

### DIFF
--- a/lib/api/controllers/funnelController.js
+++ b/lib/api/controllers/funnelController.js
@@ -234,6 +234,12 @@ function FunnelController (kuzzle) {
       .then(user => {
         request.context.user = user;
 
+          if (!user) {
+              return Promise.reject(new UnauthorizedError(
+                  `Bad request [${request.input.resource.index}/${request.input.resource.collection}/${request.input.controller}/${request.input.action}] for user ${request.context.token.userId}: user not found`)
+              );
+          }
+
         return user.isActionAllowed(request, kuzzle);
       })
       .then(isAllowed => {


scottinet:
The problem comes from the user `load` method: when loading from the database, if elasticsearch answers with a 404, we explicitly return `null`.

Instead, we should return an anonymous user object.

I'm closing this pull request in favor of a new fix I'll submit shortly.

BaptisteKaliop:
This means that with an available JWT token that was linked to a specific user, you'll receive an anonymous user?

That looks strange to me

scottinet:
It is. Since my last comment I continued to investigate, and to me there are two problems:

1. the token should have been revoked by Kuzzle in the first place, so the request should never pass through the `verifyToken` part. An explicit "invalid token" error should be returned to the client

2. kuzzle should not try to use a potentially `null` object. I'll change my previous conclusion here: in normal conditions (i.e. considering the current issue has been resolved), if the token is valid but the retrieved user is `null`, then it is not a 4xx error we should throw, but a 5xx one. At this point, the user object should NEVER be null, if that's the case, then we most certainly have a more serious problem at hand (for instance, ES instances went down). I'm hesitating between throwing a 500 or a 503 on this one.

scottinet:
New fix: https://github.com/kuzzleio/kuzzle/pull/777

PR #777, opened 2017-04-05
Revoke associated JWTs on user deletion

scottinet:
# Description

When deleting a user using `security:deleteUser`, the associated JWT were kept for their entire duration.
This lead to bugs when a deleted user tried to use one of these tokens, as Kuzzle tried to load a non-existing user, trusting the token to know if such a user exist.

This pull request maintain a list of JWT per user, and deletes all associated JWT when a user gets deleted.

Functional tests have been updated to reproduce the problem, ensuring it has been resolved.

# Other changes

* solve sonarqube issues on modified files
* rewrite most of the `authController` unit tests file, as it wasn't using mocks and because many tests  succeeded but not on the expected behavior
* fix unit tests on the `TokenRepository` class

# Solved issue

#773

### DIFF
--- a/lib/api/controllers/securityController.js
+++ b/lib/api/controllers/securityController.js
@@ -1,23 +1,48 @@
+/*
+ * Kuzzle, a backend software, self-hostable and ready to use
+ * to power modern apps
+ *
+ * Copyright 2017 Kuzzle
+ * mailto: support AT kuzzle.io
+ * website: http://kuzzle.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 'use strict';
 
 const _ = require('lodash'),
   uuid = require('node-uuid'),
-  Promise = require('bluebird'),
+  Bluebird = require('bluebird'),
   formatProcessing = require('../core/auth/formatProcessing'),
-  NotFoundError = require('kuzzle-common-objects').errors.NotFoundError,
-  BadRequestError = require('kuzzle-common-objects').errors.BadRequestError,
-  PartialError = require('kuzzle-common-objects').errors.PartialError,
-  SizeLimitError = require('kuzzle-common-objects').errors.SizeLimitError,
   Request = require('kuzzle-common-objects').Request,
   User = require('../core/models/security/user'),
   Role = require('../core/models/security/role'),
   Profile = require('../core/models/security/profile'),
-  assertHasBody = require('./util/requestAssertions').assertHasBody,
-  assertBodyHasAttribute = require('./util/requestAssertions').assertBodyHasAttribute,
-  assertBodyHasNotAttribute = require('./util/requestAssertions').assertBodyHasNotAttribute,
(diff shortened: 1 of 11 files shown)


lib/api/controllers/securityController.js
```diff
@@ -389,6 +412,7 @@ class SecurityController {
     assertHasId(request);
 
     return this.kuzzle.repositories.user.delete(request.input.resource._id)
+      .then(() => this.kuzzle.repositories.token.deleteByUserId(request.input.resource._id))
```
scottinet: fix: now clean up JWTs after the user has been successfully deleted

lib/api/core/models/repositories/tokenRepository.js
```diff
@@ -1,47 +1,68 @@
...
   }
 
   /**
    * @param {Token} requestToken
    * @returns {Promise<Object>}
    */
-  expire (requestToken) {
+  expire(requestToken) {
     return super.expireFromCache(requestToken)
+      .then(() => super.delete(requestToken.userId + '#' + requestToken._id))
```
scottinet: Clean up userid<=>token link on a forced token expiration

lib/api/core/models/repositories/tokenRepository.js
```diff
@@ -100,21 +121,22 @@ class TokenRepository extends Repository {
       .then(() => {
         this.kuzzle.tokenManager.add(token, request.context);
 
-        return token;
+        return this.persistToCache({_id: user._id + '#' + token._id}, expiresIn / 1000);
```
scottinet: Add a userid<=>token link on a token creation

lib/api/core/models/repositories/tokenRepository.js
```diff
@@ -189,25 +211,62 @@ class TokenRepository extends Repository {
...
     return this.serializeToCache (token);
   }
+
+  /**
+   * Deletes tokens affiliated to the provided user identifier
+   *
+   * @param {string} userId
+   * @return {Promise}
+   */
+  deleteByUserId(userId) {
```
scottinet: New method: delete all existing tokens created by a user ID

lib/api/controllers/securityController.js
```diff
@@ -1,23 +1,48 @@
+/*
```
ballinette: I don't know if the licence preamble is mandatory in the beginning of all the source files, or if the `LICENCE.md` file at the project's root is enough.
But if this preamble is mandatory here, it should be in ALL the files (should be done in a dedicated PR, thus)

lib/api/core/auth/tokenManager.js
```diff
@@ -22,38 +43,43 @@ const TIMEOUT_MAX = Math.pow(2, 31) - 1;
...
     if (this.tokens.array.length > 0) {
-      let delay = Math.min(this.tokens.array[0].expiresAt - Date.now(), TIMEOUT_MAX);
+      const delay = Math.min(this.tokens.array[0].expiresAt - Date.now(), TIMEOUT_MAX);
 
       if (this.timer) {
         clearTimeout(this.timer);
       }
 
-      this.timer = setTimeout(this.checkTokensValidity, delay);
+      this.timer = setTimeout(this.checkTokensValidity.bind(this), delay);
```
ballinette: Why do we need `bind(this)` ?

scottinet: because otherwise the method is run with an empty context

lib/api/controllers/securityController.js
scottinet: It is required by the Apache 2.0 license. And a PR is already underway, I'm updating files in this PR to ease up future merges.

lib/api/core/auth/tokenManager.js
ballinette: Oh yes, I did not notice the setTimeout !! (yes, my glasses may be dirty... :/ )

ballinette: Oh yes, I did not notice the `setTimeout` !! (probable my glasses are dirty... :/ )

Issue #762, opened 2017-03-30
 Detect strategies duplicate: in case of 2 plugins implementing the same strategy


dbengsch:

samniisan:
Fixed in #763

PR #780, opened 2017-04-05
Added a new route to list registrated plugin passport strategies

### LINKED ISSUE #763: Add a new route to list all available strategies

samniisan:
fixes #763

### DIFF
--- a/features/HTTP.feature
+++ b/features/HTTP.feature
@@ -1564,3 +1564,7 @@ Feature: Test HTTP API
     And There is no specifications for index "kuzzle-test-index" and collection "kuzzle-collection-test"
     When I delete the specifications again for index "kuzzle-test-index" and collection "kuzzle-collection-test"
     Then There is no error message
+
+  @usingHttp
+  Scenario: Get authentication strategies
+    Then I get the registrated authentication strategies
--- a/features/support/apiHttp.js
+++ b/features/support/apiHttp.js
@@ -729,6 +729,15 @@ ApiHttp.prototype.deleteProfile = function (id) {
   return this.callApi(options);
 };
 
+ApiHttp.prototype.getAuthenticationStrategies = function () {
+  const options = {
+    url: this.apiPath('_authenticationStrategies'),
+    method: 'GET'
+  };
+
+  return this.callApi(options);
+};
+
 ApiHttp.prototype.getUser = function (id) {
   const options = {
     url: this.apiPath('users/' + id),
--- a/features/support/apiHttp.js
+++ b/features/support/apiHttp.js
@@ -731,7 +731,7 @@ ApiHttp.prototype.deleteProfile = function (id) {
 
 ApiHttp.prototype.getAuthenticationStrategies = function () {
   const options = {
-    url: this.apiPath('_authenticationStrategies'),
+    url: this.apiPath('strategies'),
     method: 'GET'
   };
 
--- a/lib/api/core/plugins/pluginContext.js
+++ b/lib/api/core/plugins/pluginContext.js
@@ -277,6 +277,7 @@ function registerStrategy(Strategy, name, context, verify, options = {}) {
 
   try {
     this.passport.use(name, new Strategy(opts, verify.bind(context)));
+    this.pluginsManager.registeredStrategies.push(name);
   }
   catch (e) {
     // There might not be any logger active when an authentication plugin registers its strategy
--- a/lib/api/core/plugins/pluginContext.js
+++ b/lib/api/core/plugins/pluginContext.js
@@ -275,6 +275,10 @@ function instantiateRequest(request, data, options) {
 function registerStrategy(Strategy, name, context, verify, options = {}) {
   const opts = Object.assign(options, {passReqToCallback: true});
 
+  if (this.pluginsManager.registeredStrategies.indexOf(name) !== -1) {
+    throw new PluginImplementationError('An identical authentication strategy has already been registered');
+  }
+
   try {
     this.passport.use(name, new Strategy(opts, verify.bind(context)));
     this.pluginsManager.registeredStrategies.push(name);
--- a/lib/api/core/plugins/pluginsManager.js
+++ b/lib/api/core/plugins/pluginsManager.js
@@ -38,6 +38,7 @@ function PluginsManager (kuzzle) {
   this.workers = {};
   this.config = kuzzle.config.plugins;
   this.silent = false;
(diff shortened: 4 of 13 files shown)


lib/api/core/plugins/pluginsManager.js
```diff
@@ -200,6 +201,14 @@ function PluginsManager (kuzzle) {
       kuzzle.funnel.controllers[name] = controller;
     });
   };
+
+  /**
+   * Returns the list of registered passport strategies
+   * @returns {Promise}
+   */
+  this.listStrategies = function pluginListStrategies () {
+    return Promise.resolve({'authenticationStrategies': _.uniq(this.registeredStrategies)});
```
ballinette: Note: `uniq` could be unnecessary here if we forbid to register 2 strategies with the same name (and indeed, we should).

lib/api/core/plugins/pluginContext.js
```diff
@@ -277,6 +277,7 @@ function registerStrategy(Strategy, name, context, verify, options = {}) {
 
   try {
     this.passport.use(name, new Strategy(opts, verify.bind(context)));
+    this.pluginsManager.registeredStrategies.push(name);
```
ballinette: Note: we can here fix also issue #762 : add the strategy to passport only if there is no strategy with the same name yet, and throw an error otherwise.

lib/api/core/plugins/pluginsManager.js
samniisan: I will update this PR with fix for #762

ballinette:
Note for unit tests : you should split into 3 parts :
* verify that `authController.getStrategies` calls `pluginsManager.listStrategies` (using Kuzzle's mock for this one)
* verify that `registerStrategy.registerStrategy` registers correctly the strategy into the pluginManager (this test should be added in test/api/core/pluginContext/pluginContext.test.js)
* verify that `pluginsManager.listStrategies` returns a correct array with the registered strategies (this test should be added in `test/api/core/pluginsManager/`)

features/HTTP.feature
```diff
@@ -1575,3 +1575,7 @@ Feature: Test HTTP API
     And There is no specifications for index "kuzzle-test-index" and collection "kuzzle-collection-test"
     When I delete the specifications again for index "kuzzle-test-index" and collection "kuzzle-collection-test"
     Then There is no error message
+
+  @usingHttp
+  Scenario: Get authentication strategies
+    Then I get the registrated authentication strategies
```
scottinet: Additionally to the retrieval of authentication strategies, I wonder if we shouldn't verify if they are well-formed/if we retrieved what is expected by default.

lib/config/httpRoutes.js
```diff
@@ -4,6 +4,7 @@ module.exports = [
   {verb: 'get', url: '/users/_me/_rights', controller: 'auth', action: 'getMyRights'},
   {verb: 'get', url: '/_login/:strategy', controller: 'auth', action: 'login'},
   {verb: 'get', url: '/_logout', controller: 'auth', action: 'logout'},
+  {verb: 'get', url: '/_authenticationStrategies', controller: 'auth', action: 'getStrategies'},
```
scottinet: This HTTP route should be changed as we have more routes to be implemented in the near future.

Given our current paradigm, I believe this particular route should be defined like that:

`{verb: 'get', url: '/strategies', controller: 'auth', action: 'getStrategies'},`

This opens the possibility to add more route while being consistent with our current API.
For instance:

`{verb: 'get', url:'/strategies/_info/:strategyName', controller: 'auth', action: 'getStrategyInfo'}`

features/support/apiHttp.js
```diff
@@ -729,6 +729,15 @@ ApiHttp.prototype.deleteProfile = function (id) {
   return this.callApi(options);
 };
 
+ApiHttp.prototype.getAuthenticationStrategies = function () {
+  const options = {
+    url: this.apiPath('_authenticationStrategies'),
```
scottinet: See my comment below on the `httpRoutes.js` file

lib/api/core/plugins/pluginsManager.js
scottinet: Why do we return an object instead of the array of known strategies directly?

ballinette: other suggestion: return an object with infos related for each strategies:
```javascript
{
  local: {
    description: 'Local authenticate with user/password credentials',
    credential_fields: ['username', 'password']
  },
  github: {
    description: 'Github authenticated with oauth2 protocol'
  },
  // etc.
}
```

scottinet: wouldn't that make the `getInfo` route redundant?

lib/api/core/plugins/pluginContext.js
```diff
@@ -277,6 +277,10 @@ function registerStrategy(Strategy, name, context, verify, options = {}) {
 
   try {
     this.passport.use(name, new Strategy(opts, verify.bind(context)));
+    
+    if (this.pluginsManager.registeredStrategies.indexOf(name) === -1) {
```
ballinette: if a strategy with the same name already exists, we should throw with an error and abort the plugin registration.
(and it should be done before call of `this.passport.use`)

samniisan: Indeed! Will do

test/api/controllers/authController.test.js
```diff
@@ -266,4 +268,20 @@ describe('Test the auth controller', () => {
         });
     });
   });
+
+  describe('#getAuthenticationStrategies', () => {
+    it('should return a valid response', () => {
+      kuzzle.pluginsManager = new PluginManager(kuzzle);
```
ballinette: no need to do that. Kuzzle Mock already has a mocked plujginManager.
We should just add `listStrategies` method here:
https://github.com/kuzzleio/kuzzle/blob/413ff1a8e004e710df617005384a1ce945dcd1e4/test/mocks/kuzzle.mock.js#L156

test/api/core/pluginsManager/listStrategies.test.js
```diff
@@ -0,0 +1,184 @@
...
+
+describe('Test plugins manager listStrategies', () => {
+  let
+    sandbox,
+    plugin,
+    kuzzle,
+    pm2Mock,
+    context;
+
+  before(() => {
```
ballinette: Not sure that you need to mock pm2 just to test `listStrategies()` method.

Issue #770, opened 2017-03-31
DSL FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory


benoitvidis:
Possible candidate (not sure):
```
data:{
			 "controller": "realtime",
				 "action": "subscribe",
				 "index": "ireland",
				 "collection": "cork",
				 "body": {
					 "bool": {
						 "must": [
						 {
							 "or": [
							 {
								 "range": {
									 "nested.int": {
										 "gte": 118
									 }
								 }
							 },
							 {
								 "exists": {
									 "field": "boolean"
								 }
							 }
							 ]
						 },
						 {
							 "and": [
							 {
								 "range": {
									 "int": {
										 "lt": 109
									 }
								 }
							 },
							 {
								 "ids": {
									 "values": [
										 "fb712cbf-3981-4896-9d7d-5bdf768a3c56",
									 "f22bb388-c478-4679-a209-17c553e150b4",
									 "d97d0848-23ab-41f1-8065-cb25a7aec02c",
									 "d77dbf45-df5a-4016-baf1-a2cb003b29f1",
									 "33c65737-5c26-4bc7-9f29-4e45320c2d49",
									 "6308dcd5-b742-4556-a27a-dd1acffd5f3b",
									 "0068b892-173c-4476-b5c4-064b2cdc5b5b",
									 "e0cc7c8b-0de2-4a73-88a4-038a5d6df38c",
									 "607bf79c-ad45-47ae-b035-8159d104630d",
									 "df41947d-6ce0-4ff8-bf86-55c7f470ba8e"
									 ]
								 }
							 },
							 {
								 "regexp": {
									 "string": {
										 "value": "Mai.*"
									 }
								 }
							 }
							 ]
						 }
						 ],
						 "should": [
						 {
							 "bool": {
								 "must_not": [
								 {
									 "ids": {
										 "values": [
											 "6752bd0c-a36a-4165-8d8a-49474d386e02",
										 "b3ca6321-7ffa-42ea-b81d-a62058860c52",
										 "addb04aa-ecba-4a35-8183-5a44a8c97108",
										 "c24766ab-4c9a-4634-8249-00578520c7c5"
										 ]
									 }
								 }
								 ],
								 "should": [
								 {
									 "ids": {
										 "values": [
											 "0285d3d9-0450-4cf8-8a01-fe13fd646912",
[...]

PR #783, opened 2017-04-06
Fix dsl & canonization heap space exhaustion

benoitvidis:
# Description

This PR fixes the case where the canonization would extract too many conditions from the subscription filters and exhaust the Javascript memory heap.
The issue is coming from the canonization process, which computes a truth table with a size of 2^n elements, n being the number of conditions.

# Changes

* reduced the number of conditions by considering or and and filters that do not contain any `and` or `or` child clause as a condition (+ several small tricks in the standardizer)
* Introduced a configurable limit of possible conditions to submit to the canonization.

# Related issue:

* #770

### DIFF
--- a/lib/api/dsl/transform/canonical.js
+++ b/lib/api/dsl/transform/canonical.js
@@ -1,9 +1,10 @@
 'use strict';
 
-var
-  _ = require('lodash'),
-  combinatorics = require('js-combinatorics'),
-  espresso = require('espresso-logic-minimizer'),
+const
+  BadRequestError = require('kuzzle-common-objects').errors.BadRequestError,
+  config = require('../../../config'),
+  Combinatorics = require('js-combinatorics'),
+  Espresso = require('espresso-logic-minimizer').Espresso,
   strcmp = require('../util/stringCompare');
 
 /**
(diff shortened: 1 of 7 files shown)


lib/api/dsl/transform/standardize.js
```diff
@@ -513,52 +566,99 @@ class Standardizer {
...
       });
   }
 
   /**
    * Checks that a filters array is well-formed and standardized it
    *
    * @private
-   * @param {Array} filters array
-   * @param {string} keyword used to combine filters
+   * @param {{ string:  }} filter array
```
dbengsch: parameter type seams strange

lib/api/dsl/transform/canonical.js
```diff
@@ -32,60 +33,192 @@ function Canonical () {
...
+      result = [];
 
     if (Object.keys(filters).length === 0) {
       return [[{'everything': true}]];
     }
 
-    conditions = extractConditions(clone);
-    pla = buildPLA(clone, conditions.length);
-    normalized = espresso(pla);
+    const conditions = extractConditions(this._cloneFilters(filters));
```
AnthonySendra: `this._cloneFilters` can't count the conditions length?

benoitvidis: This would make the function quite hairy. 

The clone is just there to prevent messing up with the internal reformatting that is done later on by the actual canonization.

AnthonySendra: Okay but what about performance? We parse this object twice, this is not problem for performance?

PR #786, opened 2017-04-07
Add crash reports history limits

scottinet:
# Description

Crash reports can be generated by Kuzzle when a problem is detected (without service downtime), or in case of a crash.

These reports generate a memory dump of Kuzzle for debugging purposes by the Kuzzle support team. These memory dumps weight usually about 1~2GB depending on Kuzzle activity at the time of the dump process.

These space-consuming reports are a problem in case of repeated errors: they can (and will) quickly saturate the filesystem.

The aim of this PR is to prevent this FS saturation by making the diagtools module handle its reports history, using configurable limits.

By defaults, these limits have been set to a maximum of 3 coredumps and 10 reports at any given time.
(coredumps will be erased from crash reports if the core limit is inferior to the reports one)

# Other changes

* remove the `mkdirp` dependency, redundant with `fs-extra.mkdirp` method
* solve sonarqube issues on modified files

### DIFF
--- a/test/api/controllers/cli/dump.test.js
+++ b/test/api/controllers/cli/dump.test.js
@@ -36,7 +36,7 @@ describe('Test: dump', () => {
         dump: {
           history: {
             coredump: 3,
-            traces: 5
+            reports: 5
           },
           path: '/tmp',
           dateFormat: 'YYYY'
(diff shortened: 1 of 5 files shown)


test/api/controllers/cli/dump.test.js
```diff
@@ -42,89 +49,152 @@ describe('Test: dump', () => {
         }
       },
       statistics: {
-        getAllStats: getAllStatsSpy
+        getAllStats: getAllStatsStub
       }
     };
 
-    dump = rewire('../../../../lib/api/controllers/cli/dump');
+    mockrequire('fs-extra', fsStub);
```
dbengsch: :+1:

Issue #785, opened 2017-04-06
404 errors on multiple concurrent delete requests


benoitvidis:
Following code produces random 404 errors (you'll probably need to tweak the number of documents.
At a first glance, it seems to be related to the overload mechanism which triggers some requests twice.

```js
const
  Promise = require('bluebird'),
  Kuzzle = require('kuzzle-sdk'),
  kuzzle = new Kuzzle('localhost');

const collection = kuzzle.collection('c', 'test');

return collection.createDocumentPromise({fake: 1})
  .then(() => Promise.promisify(kuzzle.query.bind(kuzzle))({index: 'test'}, {controller: 'index', action: 'delete'}))
  .then(() => {
    const bulk = [];
    for (let i = 0; i < 300; i++) {
      bulk.push({create: {_index: 'test', _type: 'c', _id: `${i}`}});
      bulk.push({foo: i});
    }
    return Promise.promisify(kuzzle.query.bind(kuzzle))({}, {controller: 'bulk', action: 'import', body: {bulkData: bulk}}, {refresh: 'wait_for'});
  })
  .then(() => console.log('created'))
  .then(() => collection.fetchAllDocumentsPromise())
  .then(docs => {
    console.log(docs.length);
    return docs;
  })
  .then(docs => Promise.all(docs.map(doc => collection.deleteDocumentPromise(doc.id))))
  .then(() => process.exit(0))
  .catch(e => {
    console.error(e);
    process.exit(1);
  });
```

PR #792, opened 2017-04-10
Fix Multiple concurrent requests being executed multiple times

benoitvidis:
:warning: depends on #788 

# Description

This PR fixes the case when while running too many concurrent simultaneous requests, Kuzzle would try to execute them multiple times, leading for instance to some 404 errors in case of massive deletions.

# Related issue

* #785 

## Internal changes

3 issues fixed:

1. `funnel:execute` would still process the request if it had been put in cache
2. `funnel:getRequestSlot` would accept duplicates in its request cache
3. `funnel:_playCachedRequests` would call execute with the wrong callback, leading to case 1 again.


benoitvidis:
Putting the pr in wip to wait for the backport on master and resolve possible conflicts

lib/api/controllers/funnelController.js
```diff
@@ -127,7 +131,10 @@ class FunnelController {
 
     // resolves the callback immediately if a slot is available
     if (this.concurrentRequests < this.kuzzle.config.limits.concurrentRequests) {
-      return callback(null);
+      if (this.cachedIds[request.id]) {
+        delete this.cachedIds[request.id];
```
scottinet: Won't this leave the request in the cache queue, making the funnel processing it twice?
Once here and now if the request slot is immediately available, and once later by the `playCachedRequests` method?

benoitvidis: as the following `return` statement is made out of the `if`, I cannot figure any situation where this would occur.
The `return true` will make the request being executed right away and it wont be added to the cache queue either (made line 159 below)

benoitvidis: As seen together, this particular situation cannot occur be we still may have had a problem where a request could potentially be dropped from the cache queue before it is actually processed.

Should be fix by last commit

PR #793, opened 2017-04-11
Hotfix - concurrent requests can be executed multiple times

benoitvidis:
# Description

This PR fixes the case when while running too many concurrent simultaneous requests, Kuzzle would try to execute them multiple times, leading for instance to some 404 errors in case of massive deletions.

# Related issue

* #785 

## Internal changes

3 issues fixed:

1. `funnel:execute` would still process the request if it had been put in cache
2. `funnel:getRequestSlot` would accept duplicates in its request cache
3. `funnel:_playCachedRequests` would call execute with the wrong callback, leading to case 1 again.

### DIFF
--- a/lib/api/controllers/funnelController.js
+++ b/lib/api/controllers/funnelController.js
@@ -74,7 +74,7 @@ function FunnelController (kuzzle) {
    * Rejects it with a ServiceUnavailable error if Kuzzle is overloaded
    * and if the buffer limit has been reached
    *
-   * @param {Request} request
+   * @param {Request} request - can be mutated in case of overload error
    * @param {Function} callback
    */
   this.getRequestSlot = function funnelGetRequestSlot (request, callback) {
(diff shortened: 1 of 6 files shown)


lib/api/controllers/funnelController.js
```diff
@@ -147,32 +148,74 @@ function FunnelController (kuzzle) {
    * @param {Function} callback
    */
   this.execute = function funnelExecute (request, callback) {
-    this.getRequestSlot(request, overloadError => {
-      if (overloadError) {
-        // "handleErrorDump" shouldn't need to be called for 503 errors
-        return callback(overloadError, request);
-      }
+    const process = this.getRequestSlot(request, callback);
```
AnthonySendra: Could we name it differently? `process` variable already exists ;)

scottinet:
See my comment on the non-hotfix version of this PR before merging this one: https://github.com/kuzzleio/kuzzle/pull/792#discussion_r112156300

Issue #482, opened 2016-10-20
Document Metadata - Step 2 - Delete notifications


xbill82:
cf https://confluence.kaliop.net/display/LABS/2016-02-26+-+Delete+notifications
Implement some pseudo-deletion mechanism.
Instead of actually deleting a document from ES, mark it as deleted in the metadata.
- All read actions should check the metadata field and return a 404 if the document is marked as deleted.
- All search queries should be injected an additional filter to exclude the documents that are marked for deletion.
- The deletion notifications should be triggered each time a document is marked for deletion. Now that the document can be fetched, notifications can then be sent even if the document was created before the user subscribed to its scope.
- Some cleanup mechanism should be put in place to actually trash the data after a configurable period of time.

Keep in mind that this feature can be later on be used to handle a trash in the BO. 
The code should be ready to bypass the filter injections in order to list the objects in the trash.

PR #806, opened 2017-04-20
Delete now deactivate a document and notify by filters

### LINKED ISSUE #699: Finish the trash bin feature and improve the delete notifications
The recycle bin filtering has been developed for now.
We still lake the actual mechanism to put a document in the trash bin. This functionality also allow to fix the deletion notifications.

jenow:
# Description

Delete now set the active attribute of _meta to false and then call the notifier.
We can now search and get documents which are also in the trash by setting a `includeTrash` attribute to `true` in the input.args attribute of the request.

# Related issue

* fix #699 #482

### DIFF
--- a/lib/api/core/notifier.js
+++ b/lib/api/core/notifier.js
@@ -258,7 +258,7 @@ class NotifierController {
    * @param {Array} ids - list of deleted document IDs
    */
   notifyDocumentDelete(request, ids) {
-    return Bluebird.each(ids, (id, callback) => {
+    return Bluebird.all(ids.map(id => {
       const
         getRequest = new Request({
           action: 'delete',
@@ -267,10 +267,9 @@ class NotifierController {
           index: request.input.resource.index,
           _id: id,
           requestId: request.id,
-          volatile: request.input.volatile
+          volatile: request.input.volatile,
+          includeTrash: true
         });
-
-      getRequest.input.args.includeTrash = true;
 
       return this.kuzzle.services.list.storageEngine.get(getRequest)
         .then(result => {
@@ -293,8 +292,7 @@ class NotifierController {
             state: 'done'
           });
 
-          return this.kuzzle.services.list.internalCache.search(this.cacheKeyPrefix + deletedDocument._id)
-            .asCallback(callback);
+          return this.kuzzle.services.list.internalCache.search(this.cacheKeyPrefix + deletedDocument._id);
         })
         .then(cachedRooms => {
           const
@@ -307,14 +305,8 @@ class NotifierController {
 
           this.kuzzle.notifier.notify(cachedRooms, request, notification);
           return this.kuzzle.services.list.internalCache.remove(this.cacheKeyPrefix + id);
-        })
-        .asCallback(callback);
-    }, error => {
-      if (error) {
-        this.kuzzle.pluginsManager.trigger('log:error', error);
-        return false;
-      }
-    });
+        });
+    }));
   }
 }
 
--- a/lib/api/core/notifier.js
+++ b/lib/api/core/notifier.js
@@ -273,7 +273,6 @@ function NotifierController (kuzzle) {
           this.notify(matchedRooms, request, {
             action: 'delete',
             _id: deletedDocument._id,
-            _source: deletedDocument._source,
             _meta,
             scope: 'out',
             state: 'done'
--- a/lib/api/core/notifier.js
+++ b/lib/api/core/notifier.js
@@ -259,51 +259,36 @@ class NotifierController {
    */
   notifyDocumentDelete(request, ids) {
     return Bluebird.all(ids.map(id => {
-      return this.kuzzle.services.list.internalCache.search(this.cacheKeyPrefix + id)
-        .then(cachedRooms => {
-          const
-            notification = {
-              action: 'delete',
-              scope: 'out',
-              state: 'done',
-              _id: id
-            };
+      const
+        getRequest = new Request({
+          collection: request.input.resource.collection,
+          index: request.input.resource.index,
+          _id: id,
+          includeTrash: true
+        });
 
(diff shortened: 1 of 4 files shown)


jenow:
You're right @benoitvidis, I missed mget. Did you see any other?

benoitvidis:
@jenow nope but I did not search for it. May be worth checking all possible candidates.

scottinet:
@jenow > can you notify us once you've done a swipe on all ES exposed methods so that we can review this PR?

jenow:
@benoitvidis @scottinet I reviews the es service and it's good.

lib/services/elasticsearch.js
```diff
@@ -260,9 +261,8 @@ class ElasticSearch extends Service {
     return this.client.get(esRequest)
       .then(result => {
         if (result._source) {
-          if (result._source._kuzzle_info && !result._source._kuzzle_info.active) {
-            // todo Feedback how to get it from the 'trash' once it is implemented
-            return Bluebird.reject(new NotFoundError(`Document ${result._id} was already deleted`));
+          if (result._source._kuzzle_info && !result._source._kuzzle_info.active && !request.input.args.includeTrash) {
+            return Bluebird.reject(new NotFoundError(`Document ${result._id} was already deleted. If you want to also get trashed documents set includeTrash argument to true`));
```
scottinet: I'm not at ease with this error message on a `get` route. I see no reason to mention that the document has been "already deleted", or that it can be retrieved in the trash. At least not with the current error message.
To me, there is a conflict with returning a 404 while indicating that the document somehow exists.

I see several possibilities we need to discuss:
* throw a simple 404 "document not found" error
* throw a new 423 error (resource locked), with a message like: "Cannot get a deleted document without the "includeTrash" flag'. This new error could be, for instance, a ResourceLockedError

ballinette: I personally vote for a simple `404 "document not found"` error

lib/api/core/notifier.js
```diff
@@ -258,8 +258,42 @@ class NotifierController {
    * @param {Array} ids - list of deleted document IDs
    */
   notifyDocumentDelete(request, ids) {
-    return Bluebird.each(ids, (id, callback) => {
-      return this.kuzzle.services.list.internalCache.search(this.cacheKeyPrefix + id)
+    return Bluebird.all(ids.map(id => {
+      const
+        getRequest = new Request({
```
jenow: The whole point of this work was to send back the body of the document we deleted so we can be notified so I don't really understand why you want me not to fetch the document to get the body.. 
Also at the time we want to notify rooms for the delete notification, no rooms are in the internalCache.

benoitvidis: @ballinette @jenow , to me, the original aim of this issue was not about sending the body or not but to be able to notify clients on deletions at all.

Problem is: we are relying on the cache to detect the notifications to be sent. If a document has no entry in the cache, then no notification is sent.

Let's consider the following situation:

1. I am the only user to subscribe to one collection (no filter).
2. Another user deletes a document from this collection

In this situation, I won't get any notification at all.

@jenow : 
* We should, in case the document in cache, call, the DSL and the matching rooms in cache
* I don't see any reason why we would need to remove any entry from the cache

@ballinette , @jenow : I believe the question about if we should notify the whole document should be discussed with the team.
My personnal point of view if that, as it requires some extra work for Kuzzle, it should not be done by default, in which case, users should eventually be able to search the recycle bin in case they need to get the full document.

ballinette: @benoitvidis good point about your situation

But with the actual PR's implementation, users may be notified twice for a document's deletion, if a matched room is already in cachedRoom, or did I miss something ?

I agree that a conception workshop is needed for this point.

scottinet: I don't see why the content of a _deleted_ document should be sent: what's the point, since it has been deleted?
The only users interested by receiving deleted documents notifications should already have their ID stored somewhere in order to sync their app and remove the deleted document.
So to me, we should only send the document id.

Other than that, I agree with the way @benoitvidis describes how the deletion notification function should behave.

ballinette: OK now about the notification message, only the document's id is given.

But you still have the issue of potential duplicates, as you don't exclude from dsl query results the rooms that were in cache.

Maybe following implementation could fix that ?
```javascript
notifyDocumentDelete(request, ids) {
  let cachedRooms
  return this.kuzzle.services.list.internalCache.search(this.cacheKeyPrefix + id)
    .then(result => {
      cachedRooms = result;
      const
        notification = { ... };
      this.kuzzle.notifier.notify(cachedRooms, request, notification);
      return this.kuzzle.services.list.internalCache.remove(this.cacheKeyPrefix + id);
    )}
    .then(() => {
      // (...)
      matchedRooms = this.kuzzle.dsl.test(...);
      matchedRooms = _.difference(matchedRooms, cachedRooms);
      // (...)
    });
```

ballinette: or ignore search from cachedRooms and only get the matchedRooms from DSL (as the `dsl.test` method should be called anyway, fetching some rooms from internalCache before will not improve performances)

jenow:
I add the wip label as I have to remove the cached rooms logic

lib/api/core/notifier.js
```diff
@@ -272,13 +272,39 @@ class NotifierController {
...
           return this.kuzzle.services.list.internalCache.remove(this.cacheKeyPrefix + id);
         })
-        .asCallback(callback);
-    }, error => {
-      if (error) {
-        this.kuzzle.pluginsManager.trigger('log:error', error);
-        return false;
-      }
-    });
+        .then(()=> {
```
xbill82: These 2 `then` blocks overlap in functionality. Refer to the conversation about dropping the internal cache.

PR #819, opened 2017-04-27
Fix dump exceptions

benoitvidis:
# Description 

This PR fixes several issues with the dump

1. Repair suffix handling + internal errors minimum interval
2. Dumps are not resolved only once they are actually processed (was not the case, notably regarding the core compression)
3. Dumps are locked. A requested dump will be discarded if one is already running
4. Tries to remove several cases where the dump itself would trigger an unhandled exception, leading to a dead loop for Kuzzle (cf for instance the check on the dump directory creation)

# Related issue 

* https://github.com/kuzzleio/kuzzle-private-backlog/issues/60 (private)

### DIFF
--- a/default.config.js
+++ b/default.config.js
@@ -9,6 +9,29 @@
  * @class KuzzleConfiguration
  */
 module.exports = {
+
+  dump: {
+    enabled: true,
+    history: {
+      coredump: 3,
+      reports: 5
+    },
+    path: './dump/',
+    gcore: 'gcore',
+    dateFormat: 'YYYYMMDD-HHmmss',
+    handledErrors: {
+      enabled: true,
+      whitelist: [
+        // 'Error',
+        'RangeError',
+        'TypeError',
+        'KuzzleError',
+        'InternalError'
+      ],
+      minInterval: 10 * 60 * 1000
+    }
+  },
+
   /*
    routes: list of Kuzzle API exposed HTTP routes
    accessControlAllowOrigin: sets the Access-Control-Allow-Origin header used to
@@ -188,26 +211,6 @@ module.exports = {
 
   /** @type {DocumentSpecification} */
   validation: {
-  },
-
-  dump: {
-    enabled: true,
-    history: {
-      coredump: 3,
-      reports: 5
-    },
-    path: './dump/',
-    gcore: 'gcore',
-    dateFormat: 'YYYYMMDD-HHmm',
-    handledErrors: {
-      enabled: true,
-      whitelist: [
-        // 'Error',
-        'RangeError',
-        'TypeError',
-        'KuzzleError',
-        'InternalError'
-      ]
-    }
   }
+
 };
--- a/lib/api/controllers/cli/dump.js
+++ b/lib/api/controllers/cli/dump.js
@@ -34,7 +34,9 @@ const
   zlib = require('zlib'),
   glob = require('glob');
 
-let _kuzzle;
+let
+  _lock = false,
+  _kuzzle;
 
 /**
  * Create a dump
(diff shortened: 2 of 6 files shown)


default.config.js
```diff
@@ -9,6 +9,29 @@
...
+
+  dump: {
+    enabled: true,
+    history: {
+      coredump: 3,
+      reports: 5
+    },
+    path: './dump/',
+    gcore: 'gcore',
+    dateFormat: 'YYYYMMDD-HHmmss',
```
benoitvidis: added: seconds

default.config.js
```diff
@@ -9,6 +9,29 @@
...
+    handledErrors: {
+      enabled: true,
+      whitelist: [
+        // 'Error',
+        'RangeError',
+        'TypeError',
+        'KuzzleError',
+        'InternalError'
+      ],
+      minInterval: 10 * 60 * 1000
```
benoitvidis: interval moved from 1m to 10

lib/api/controllers/cli/dump.js
```diff
@@ -91,68 +113,115 @@ function dump (request) {
...
+      return resolve();
     }
 
-    // core-dump
-    console.log('> generating core-dump');
-    dumpme(_kuzzle.config.dump.gcore || 'gcore', `${dumpPath}/core`);
-
-    // Gzip the core
-    glob(`${dumpPath}/core*`, (err, res) => {
+    const logDir = path.dirname(process.env.pm_err_log_path || process.env.pm_log_path);
```
benoitvidis: I am not happy with the way this is done.
Current behaviour is now to take the last created files from pm2 log directory (stderr > stdout) and to compress them.

We may have several cases where this is not suitable imo

* too much dependency on pm2
* no control on what the log directory actually contains
* depends on the logger configuration. For instance, if error logs are sent to syslog, we have no way to retrieve them

I suppose this should be discussed first and would deserve a dedicated implementation out of this PR.

lib/api/controllers/funnelController.js
```diff
@@ -273,11 +271,12 @@ class FunnelController {
           errorMessage = errorMessage.toLowerCase().replace(/[^a-zA-Z0-9-_]/g, '-').replace(/[-]+/g, '-').split('-');
           errorMessage = errorMessage.filter(value => value !== '').join('-');
 
-          request.input.body.suffix = 'handled-'.concat(errorType.toLowerCase(), '-', errorMessage);
+          const suffix = `handled-${errorType.toLocaleLowerCase()}-${errorMessage}`;
+          request.input.args.suffix = suffix;
 
-          if (!this.lastDumpedErrors[request.input.body.suffix] || this.lastDumpedErrors[request.input.body.suffix] < lastSeen - MIN_TIME_BEFORE_DUMP_PREV_ERROR) {
+          if (!this.lastDumpedErrors[errorType] || this.lastDumpedErrors[errorType] < lastSeen - this.kuzzle.config.dump.handledErrors.minInterval) {
```
benoitvidis: the suffix can include too many variations

benoitvidis:
Some debug traces remaining.. putting the pr in wip

lib/api/controllers/cli/dump.js
```diff
@@ -91,68 +113,112 @@ function dump (request) {
...
+    fs.readdir(logDir, (err, files) => {
       if (err) {
-        console.error(err);
-        return;
+        console.log('... error reading log directory');
+        resolve();
```
scottinet: Missing return statement interrupting the execution flow if an error occurs

lib/api/controllers/funnelController.js
```diff
@@ -273,11 +271,12 @@ class FunnelController {
           errorMessage = errorMessage.toLowerCase().replace(/[^a-zA-Z0-9-_]/g, '-').replace(/[-]+/g, '-').split('-');
           errorMessage = errorMessage.filter(value => value !== '').join('-');
 
-          request.input.body.suffix = 'handled-'.concat(errorType.toLowerCase(), '-', errorMessage);
+          const suffix = `handled-${errorType.toLocaleLowerCase()}-${errorMessage}`;
+          request.input.args.suffix = suffix;
```
ballinette: why adding a `suffix` variable ?
can't we just have one line ?
```javascript
request.input.args.suffix = `handled-${errorType.toLocaleLowerCase()}-${errorMessage}`;
```

lib/api/controllers/cli/dump.js
stafyniaksacha: @benoitvidis I totally agree with this, we need to find something better

lib/api/controllers/cli/dump.js
benoitvidis: fixed

lib/api/controllers/funnelController.js
benoitvidis: yes we can. done

PR #821, opened 2017-04-28
Add Document:exists route

samniisan:
Usage:
```
GET /:index/:collection/:id/_exists
```

Linked SDK global ticket: https://github.com/kuzzleio/kuzzle-sdk/issues/25


lib/api/controllers/documentController.js
```diff
@@ -89,6 +89,18 @@ class DocumentController {
 
   /**
    * @param {Request} request
+   * @returns {Promise<Boolean>}
+   */
+  exists(request) {
+    assertHasId(request);
+
+    return this.engine.get(request)
+      .then(() => Promise.resolve(true))
```
scottinet: Use `Bluebird` instead of the native promises. Those are far slower than bluebird (will be fixed in future versions of V8)

samniisan: Done :)

Issue #822, opened 2017-04-28
Notification not sent on valid room


benoitvidis:
# Situation

In some circumstances (to investigate), some notifications are sent to only a partial subset of the rooms they should match.

Trying to reproduce the issue with 2 clients and some manual registrations has been unsuccessful til now (Kuzzle properly broadcast the notifications to all rooms in these attempts), which makes me think of a more complex intrication of causes.

To reproduce:

Using https://github.com/kuzzleio/kuzzle-client-tests/commit/ebc3ccb63f91b20bf70ce1858a32ae5129fcec0e:

```
SEED=317882535 node index.js
```

PR #823, opened 2017-04-28
Fix missing notifications

benoitvidis:
# Description

Fixes some cases where Kuzzle DSL would send notifications to a subset of matching subscription rooms only.
In depth, the internal test table used by the DSL could assign the same index to different filters, making the notifier trigger only one notification for these.

The fix simplifies the way the test tables indices are computed upon a reinxing (following an unregistration).

# Related issue

* #822

### DIFF
--- a/lib/api/dsl/storage/index.js
+++ b/lib/api/dsl/storage/index.js
@@ -228,7 +228,7 @@ class Storage {
 
     const {index, collection} = this.filters[filterId];
 
-    removeFromTestTables(this.filters, this.subfilters, this.testTables, index, collection, this.filters[filterId]);
+    this._removeFromTestTables(index, collection, this.filters[filterId]);
 
     this.filters[filterId].subfilters.forEach(subfilter => {
       if (subfilter.filters.length === 1) {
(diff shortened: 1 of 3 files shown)


lib/api/dsl/storage/index.js
```diff
@@ -266,6 +266,124 @@ class Storage {
...
+    if (tt.removedConditions.array.length < tt.clength / 10) {
+      return;
+    }
+
+    // rebuild conditions index
+    const
+      conditionKeys = Object.keys(this.subfilters[index][collection]),
+      conditions = new Uint8Array(tt.conditions.length);
+
+    for (let i = 0; i < conditionKeys.length; i++) {
```
scottinet: `i` should be declared outside the for loop for better performances

lib/api/dsl/storage/index.js
```diff
@@ -266,6 +266,124 @@ class Storage {
...
+
+      conditions[i] = sf.conditions.length;
+      sf.cidx = i;
+    }
+    tt.conditions = conditions;
+    tt.clength = conditionKeys.length;
+
+    // rebuild filter index pointers
+    if (tt.removedFiltersCount) {
+      for (let i = 0; i < Object.keys(this.filtersIndex[index][collection]).length; i++) {
```
scottinet: Same here

benoitvidis: @scottinet do we really care during unregistration?

scottinet: Unfortunately yes: unregistration and reindexation are both very time-consuming tasks, and these therefore should be considered critical code.

benoitvidis: ok. fixed.

Issue #824, opened 2017-05-02
Cannot read property \'low\' of undefined


benoitvidis:
Occurs during unsubscriptions

To reproduce:

* Kuzzle: https://github.com/kuzzleio/kuzzle/commit/cbdd5f76b5117829c28b8fbb6392564101458cde
* https://github.com/kuzzleio/kuzzle-monkey-tests/commit/33d3d72b2c9c7d21f010d3721ba0d1d3af5114bf
* seed: 498147141

```
SEED=498147141 node index.js
```

Without private repo:

```
> { "controller": "realtime", "action": "subscribe", "index": "i", "collection": "c", "body": { "and": [ { "not": { "range": { "foo": { "lt": 41 } } } }, { "not": { "range": { "foo": { "lt": 20 } } } }, { "not": { "range": { "foo": { "lt": 16 } } } } ] } }
< {"requestId":"3178121f-d87d-4e2a-bde9-3d9bb5a1c437","status":200,"error":null,"controller":"realtime","action":"subscribe","collection":"c","index":"i","volatile":null,"result":{"roomId":"98ff637ed96bc51e","channel":"98ff637ed96bc51e-01e0d28a3161b5f8"},"room":"3178121f-d87d-4e2a-bde9-3d9bb5a1c437"}
> {"controller": "realtime", "action": "unsubscribe", "body": {"roomId": "98ff637ed96bc51e"}}
[...]

PR #825, opened 2017-05-02
Fix unhandled exception on unsubscribing multiple range operators

benoitvidis:
# Description

This PR fixes the case where after subscribing to a filter containing several `range` or `not range` conditions inside a `and` conditions, would crash Kuzzle on deregistering the filter.

# Related issue

* #824

### DIFF
--- a/lib/api/dsl/storage/removeOperands.js
+++ b/lib/api/dsl/storage/removeOperands.js
@@ -167,11 +167,15 @@ class OperandsRemoval {
       field = Object.keys(condition.value)[0];
 
     if (operand.fields[field].count > 1) {
-      const info = operand.fields[field].subfilters[subfilter.id];
+      const info = operand.fields[field].subfilters[subfilter.id][condition.id];
 
       operand.fields[field].tree.remove(info.low, info.high, info.subfilter);
       operand.fields[field].count--;
-      delete operand.fields[field].subfilters[subfilter.id];
+      delete operand.fields[field].subfilters[subfilter.id][condition.id];
+
+      if (Object.keys(operand.fields[field].subfilters[subfilter.id]).length === 0) {
+        delete operand.fields[field].subfilters[subfilter.id];
+      }
     }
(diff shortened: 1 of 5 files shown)


lib/api/dsl/storage/removeOperands.js
```diff
@@ -167,11 +167,15 @@ class OperandsRemoval {
...
+      const info = operand.fields[field].subfilters[subfilter.id][condition.id];
 
       operand.fields[field].tree.remove(info.low, info.high, info.subfilter);
       operand.fields[field].count--;
-      delete operand.fields[field].subfilters[subfilter.id];
+      delete operand.fields[field].subfilters[subfilter.id][condition.id];
+
+      if (Object.keys(operand.fields[field].subfilters[subfilter.id]).length === 0) {
+        delete operand.fields[field].subfilters[subfilter.id];
+      }
```
scottinet: There is an unnecessary deletion occuring if there is exactly 1 condition left in the structure: the condition deletion, and the subfilter one.

This can be prevented with the following code, replacing the lines 174-178:

```js
    if (Object.keys(operand.fields[field].subfilters[subfilter.id]).length === 1) {
      delete operand.fields[field].subfilters[subfilter.id];
    }
    else {
      delete operand.fields[field].subfilters[subfilter.id][condition.id];
    }
```

lib/api/dsl/storage/removeOperands.js
```diff
@@ -209,7 +213,11 @@ class OperandsRemoval {
       }
 
       operand.fields[field].count--;
-      delete operand.fields[field].subfilters[subfilter.id];
+      delete operand.fields[field].subfilters[subfilter.id][condition.id];
```
scottinet: See my previous comment about double deletion.

lib/api/dsl/storage/removeOperands.js
benoitvidis: fixed.

PR #826, opened 2017-05-03
improve debug function to allow toggle one/multiple lines

stafyniaksacha:
# Description

Debug are complex to read due to a lot information given by messages.
This PR improves options to allow developers to chose what and how they want theirs debug messages

## Added
DEBUG_EXPAND=off
> This variable will act on how objects are printed:
> - on/enabled : objects are printed in a pretty json format on a new line after debug messages
> - off/disabled : objects are printed without linebreak in a oneline format
> default: on

## Always here
DEBUG=kuzzle:*
DEBUG_DEPTH=2
DEBUG_MAX_ARRAY_LENGTH=100
DEBUG_SHOW_HIDDEN=off
DEBUG_COLORS=on
> These environment variables was already available, values are theses by defaults


lib/debug.js
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const debug = require('debug');
+const util = require('util');
+
+debug.formatters.a = value => {
+  const inspectOpts = debug.inspectOpts;
```
benoitvidis: can you confirm formatters are called only when debug is enable?

stafyniaksacha: Nothing is done when `DEBUG` env variable is not set
see: https://github.com/visionmedia/debug/blob/master/src/debug.js#L67

lib/api/core/plugins/pluginsManager.js
```diff
@@ -22,7 +22,7 @@
 'use strict';
 
 const
-  debug = require('debug')('kuzzle:plugins'),
+  debug = require('../../../debug')('kuzzle:plugins'),
```
scottinet: I find it far too easy to forget that there is a `debug` source file in Kuzzle's directories, and another one as a node module.
I bet that before long we'll have `require('debug')` lines in our code, making our debug capabilities inconsistent.

To make things easier I have these proposals:

* rename `../../../debug` into `../../../kuzzleDebug`

or:

* add `debug` as a `kuzzle` property, making it accessible almost everywhere in the project without needing any require beforehand. That would force several changes throughout Kuzzle code, though.

or:

* I hesitate to propose this one, but we also could create a global `debug` constant

PR #828, opened 2017-05-03
Fixes broker consistency 

stafyniaksacha:
:warning: Need to be merged after #826

# Description
- Ensure that event registration work no matter connection status (`listen` and `unsubscribe` methods)
- Ensure that event registrations status are well propagated when connection with server are established (connection / reconnection / already connected) 
- Ensure that ping can timeout only when socket is really connected (and not in a connecting state)
- Ensure custom handler are called at the right time with good arguments (`onConnectHandler`, `onErrorHandler` and `onCloseHandler`)
- Changed arguments of events emitted by `pluginManager` (see kuzzleio/documentation#212)

# Boyscout
- Update `ws` depedency to 2.3.1 (aligned with proxy)
- Removed usage of custom connection state: use native one from `ws` library
- Remove deprecated `preversion` and ` postversion` from `npm run` commands
- Added debug on proxy entry point and global services


features/HTTP.feature
```diff
@@ -1573,6 +1573,7 @@ Feature: Test HTTP API
     Then There is no error message
     When I delete the specifications for index "kuzzle-test-index" and collection "kuzzle-collection-test"
     Then There is no error message
+    Then I wait 1s
```
stafyniaksacha: This is a hack to allow specification to be refreshed in cluster nodes

Issue #816, opened 2017-04-27
Add  a graceful shutdown command line


scottinet:
We need a command asking Kuzzle to:

1. send an instruction to the proxy, asking it to stop sending new requests
2. process the remaining pending requests
3. shut itself down when the request queue is empty

This command would allow us to patch & restart Kuzzle nodes one by one, cleanly, without any downtime client-wise.

PR #833, opened 2017-05-09
Graceful shutdown implementation

scottinet:
# Description

:warning: this PR depends on https://github.com/kuzzleio/kuzzle-proxy/pull/72 :warning: 

Make Kuzzle gracefully exit, *after* the remaining requests have been processed, on a SIGINT signal (`kill -s SIGINT` or `Ctrl+C`).

Also add a new `bin/kuzzle shutdown` command-line interface, asking Kuzzle to gracefully exit.

Here is what happens when a graceful shutdown is triggered:

* Kuzzle sends to the proxy a command instructing it to stop sending new requests to the node being shut down
* Once done, and after some delay to allow requests on the way to this node to be registered, Kuzzle waits for the remaining requests to be finished
* Then:
  * if Kuzzle is NOT launched by PM2, it exits immediately
  * if Kuzzle is launched by PM2: if a PM2 watcher is active (typically on a development environment), Kuzzle allows PM2 to restart the process. If there is no PM2 watcher (production environment), Kuzzle deletes itself from the PM2 processes list and exits

# Other changes

Residual worker plugins were cleaned up on a Kuzzle start process indiscriminately, based on their prefixed process names. This might cause problems if multiple Kuzzle instances are running on the same PM2 daemon. And it causes PM2 processes interaction problems when Kuzzle deletes itself from the process list.

To prevent this, Kuzzle now stops and cleans up worker plugins when it exits, either because of a graceful shutdown, or because of a process crash (unhandled exception, SIGTERM signal, and so on).
This clean up is performed using the registered worker process identifiers.

The only case where worker plugins are not stopped is if Kuzzle is killed with an uncatchable signal (SIGKILL, SIGBUS, SIGILL, SIGSEGV, ...).
In this case, launching `pm2 delete` with the appropriate workers process identifiers is necessary to prevent zombie processes.

# Solved issue

#816

### DIFF
--- a/lib/api/core/plugins/pluginsManager.js
+++ b/lib/api/core/plugins/pluginsManager.js
@@ -243,6 +243,28 @@ class PluginsManager {
   listStrategies () {
     return Bluebird.resolve(this.registeredStrategies);
   }
+
+  /**
+   * Stops and unregisters plugins workers
+   * @returns {Promise}
+   */
+  shutdownWorkers() {
+    if (Object.keys(this.workers).length === 0) {
+      return Bluebird.resolve();
+    }
+
+    return new Bluebird(resolve => {
+      Object.keys(this.workers).forEach(worker => {
+        this.workers[worker].pmIds.getArray().forEach(id => pm2.delete(id));
+      });
+
+      this.workers = {};
+
+      // gives time for pm2 to delete plugins before resolving
+      // the promise
+      setTimeout(() => resolve(), 100);
+    });
+  }
 }
 
 /**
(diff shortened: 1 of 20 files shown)


lib/api/core/plugins/pluginsManager.js
```diff
@@ -243,6 +243,30 @@ class PluginsManager {
...
+    return new Bluebird(resolve => {
+      names.forEach(name => {
+        this.workers[name].pmIds.getArray().forEach(id => pm2.delete(id));
+      });
+
+      this.workers = {};
+
+      // gives time for pm2 to delete plugins before resolving
+      // the promise
+      setTimeout(() => resolve(), 100);
```
benoitvidis: nitpicking: 100ms sounds very short to me depending on the plugins complexity. I would rather make it configurable, or __better__ monitor the pids and retry till they are up.

Further, using `Bluebird.delay` instead of a Promise constructor + setTimeout here would be a little clearer imo.

scottinet: I didn't know about Bluebird.delay, I like that :-)

About plugins, I didn't thought about that. Normally, once PM2 has received the `delete` command, it waits a small delay (1.6s by default) for the process to stop, and it sends a SIGKILL after that delay.
Buuuut... we may have to change that default delay. And waiting for plugins to stop and free allocated resource before creating new instances seems a cleaner to deal with plugins to me.

I'll fix that, and add the appropriate tests.

scottinet: @benoitvidis > code updated

PR #838, opened 2017-05-11
Add healthCheck action

### LINKED ISSUE #580: Create a ok route
We need a basic healthcheck route to monitor a kuzzle node.

ballinette:
Fix #580

### DIFF
--- a/lib/api/controllers/serverController.js
+++ b/lib/api/controllers/serverController.js
@@ -110,7 +110,7 @@ class ServerController {
       })
       .then(response => {
         if (['yellow', 'green'].indexOf(response.status) < 0) {
-          throw new ExternalServiceError('es is down: ' + JSON.stringify(response));
+          throw new ExternalServiceError('Elasticsearch is down: ' + JSON.stringify(response));
         }
         return Bluebird.resolve({status: 'ok'});
       });
--- a/lib/api/controllers/serverController.js
+++ b/lib/api/controllers/serverController.js
@@ -23,6 +23,7 @@
 
 const
   Bluebird = require('bluebird'),
+  ExternalServiceError = require('kuzzle-common-objects').errors.ExternalServiceError,
   os = require('os');
 
 /**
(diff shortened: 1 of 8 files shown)


AnthonySendra:
@scottinet the part about cluster is also in mind (and planned I guess?), but in a short time, a simple route for checking if at least one Kuzzle is up is useful.
On AWS we have the Load-Balancer with health check, and currently we call `-XGET localhost:7512/`. The response and the calculation on this route is too big for this simple health check. We just want to know if someone can handle the request.

But I agree in a second time we absolutely need a route for checking all node of the cluster.

lib/api/controllers/serverController.js
```diff
@@ -100,6 +101,24 @@ class ServerController {
...
+  healthCheck() {
+    return this.kuzzle.services.list.memoryStorage.getInfos()
+      .then(() => this.kuzzle.services.list.internalCache.getInfos())
+      .then(() => this.kuzzle.services.list.storageEngine.getInfos())
+      .catch(err => {
+        throw new ExternalServiceError(err);
+      })
+      .then(response => {
+        if (['yellow', 'green'].indexOf(response.status) < 0) {
+          throw new ExternalServiceError('es is down: ' + JSON.stringify(response));
```
AnthonySendra: Nitpick: Can we return "Elasticsearch is down" instead of "es"? If a user is not comfortable with Kuzzle stack, he probably doesn't know what "es" means.

scottinet:
Ok, why not. But then we need to stop and think for a moment about the `status` response, in order to prevent introducing a breaking change when we extend this route to a whole cluster.

ballinette:
we can make as ElasticSearch Cluster, responding with a "color" status ? 
* `red` => The stack is down
* `yellow` => The stack is working correctly, but some nodes are done
* `green` => The stack is up and fully working

And for this 1st implemenation, just respond by `red` or `green`

Open question: if service as ES or Redis is not working, what do we respond ?
* a `200` result with message status `red` ?
* a `503`result with an `ExternalServiceError` (as actually) ?

scottinet:
Good question. 

I think we should act exactly as other "check" routes, e.g. `checkToken`: the route itself works well, Kuzzle is able to respond, but some services are inactive => it should respond with a status `200` and a `red` status

ballinette:
@scottinet @AnthonySendra Changed made according to our discussion. 

NB: we could improve the result with the detail of which service fail: instead of a single `{status: "green|red"}` result, we could have something like: 
```javascript
{
  status: "green|red",
  services: {
    internalCache: "green|red",
    memoryStorage: "green|red",
    storageEngine: "green|red",
  }
}
```

What do you think ?

lib/api/controllers/serverController.js
```diff
@@ -100,6 +104,25 @@ class ServerController {
...
+      .then(() => this.kuzzle.services.list.internalCache.getInfos())
+      .then(() => this.kuzzle.services.list.storageEngine.getInfos())
+      .then(response => {
+        if (['yellow', 'green'].indexOf(response.status) < 0) {
+          throw new ExternalServiceError('Elasticsearch is down: ' + JSON.stringify(response));
+        }
+        return Bluebird.resolve({status: 'green'});
+      })
+      .catch(error => {
+        request.setError(new ServiceUnavailableError(error));
```
ballinette: set here a `ServiceUnavailableError` to force a `503` error code (`ExternalServiceError` code is `500`)

AnthonySendra:
@ballinette great idea :)

PR #840, opened 2017-05-15
Login API reshape

### LINKED ISSUE #754: Login defaults strategy to "local" instead of throwing an error
Because authentication strategies are plugins, and because there is no default authentication strategy implemented in Kuzzle, I think we should not default the login strategy to `local`. Instead, we should make this argument required, and throw a `BadRequestError` if no strategy is provided (see `AuthController:login`)

xbill82:
:warning: This PR breaks the current behavior of the `login` method in the SDK. The following PRs need to be reviewed on SDK

* kuzzleio/sdk-php#66
* kuzzleio/sdk-javascript#218
* kuzzleio/sdk-android#126

# Description

This PR aims to reshape the API and the behavior of the `login` and `logout` actions in order to achieve better API consistency and more reliable behavior.

* `_login` and `_logout` HTTP routes are now POST verbs;
* `login` action now requires a mandatory `strategy` field, otherwise throwing a `BadRequestError` (fixes #754 )


xbill82:
WIP-ing the PR to figure out what broke the tests.

Issue #831, opened 2017-05-04
Login and Logout should not be GET methods


dbengsch:
Currently one of login's route and the logout route are GET methods in HTTP:

```
  {verb: 'get', url: '/_login/:strategy', controller: 'auth', action: 'login'},
  {verb: 'get', url: '/_logout', controller: 'auth', action: 'logout'},
```

Both should (only) be POST methods.

By the way the login routes where strategy is provided in the url don't work at the moment.
After discussion, only the POST route without strategy in the url should be kept, but the body structure should be adapted to something like:
```json
{
  "local": {
    "username": "MyUser",
    "password": "MyPassword"
  },
  "anotherStrategy": {...}
}

ballinette:
:warning:  We need the get method for `login` action, as oauth plugin though passport.js use it: when the user is authenticated in the remote identity provider, he is redirected to Kuzzle's login action through a HTTP GET request containing the oauth token in the query-string.

PR #845, opened 2017-05-18
Re-add HTTP GET route for user login

ballinette:
\+ inject both request body and query string into passport.authenticate arguments.

(So that strategies like OAuth or CAS can handle authentication queries with a GET request and arguments coming from the query string)

Refs #831 

**Breaking change**
The `expiresIn` attribute is now moved from `input.body` to `input.args`
That means:

* HTTP requests will now look like:
```
POST http://kuzzle:7512/_login/local?expiresIn=1h
{
  "username": "jdoe",
  "password": "foobar"
}
```

* Other protocols request will look like:
```javascript
{
  controller: 'auth',
  action: 'login',
  strategy: 'local',
  expiresIn: '1h',
  body: {
    username:  'jdoe',
    password: 'foobar'
  }
}
```

### DIFF
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -26,16 +26,14 @@ const
   _ = require('lodash'),
   Token = require('../core/models/security/token'),
   formatProcessing = require('../core/auth/formatProcessing'),
-  {
-    InternalError: KuzzleInternalError,
-    BadRequestError
-  } = require('kuzzle-common-objects').errors,
+  KuzzleInternalError = require('kuzzle-common-objects').errors.InternalError,
   {
     assertIsStrategyRegistered,
     assertIsConnected,
     assertHasBody,
     assertHasStrategy,
     assertBodyHasNotAttribute,
+    assertArgsHasAttribute,
     assertBodyHasAttribute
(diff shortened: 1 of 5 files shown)


lib/api/controllers/authController.js
```diff
@@ -76,20 +74,16 @@ class AuthController {
...
-
-    if (!request.input.args.strategy) {
-      throw new BadRequestError('No strategy specified');
-    }
+    assertArgsHasAttribute(request, 'strategy');
 
     const strategy = request.input.args.strategy;
 
-    return this.kuzzle.passport.authenticate({query: request.input.body, original: request}, strategy)
+    return this.kuzzle.passport.authenticate({body: request.input.body, query: request.input.args}, strategy)
```
ballinette: `passport.authenticate` want as first argument a JSON object representing an HTTP request, as described in Express: http://expressjs.com/fr/api.html#req
Then, it's the responsibility of each strategy to use `query` or `body` attributes, depending of its needs.

dbengsch: The original attribute is really important as it provides a way for plugins to use the `execute` accessor.

Issue #861, opened 2017-06-08
Credentials deleted when unable to create a user because it already has credentials


scottinet:
Fun bug: the `security:createUser` method tries to create all of a user's credentials, sequentially.

If a credential creation fails, then a rollback is performed and the user is deleted.

Problem: the current code has plugins return a failure because they cannot create an already existing user.
This leads to users been deleted because they already exist, when trying to create them.

PR #867, opened 2017-06-12
Fix incoherent behaviors regarding user credentials management

scottinet:
# Description

* The API now forbids to create or update credentials on non-existing users
* Several fixes on credentials plugin calls in the security controllers (wrong strategy argument passed)
* The `reset` CLI command now thoroughly deletes users and their credentials, preventing inconsistent data when a kuzzle reset is performed
* The `resetStorage` function performing a Kuzzle reset has been moved from the main `kuzzle.js` file to the `cleanDb` CLI action implementation
* Kuzzle now throws internal errors when credentials are found on a non-existing user when attempting to create it

# Other changes

CLI commands now provide feedbacks to their caller (kuzzle or a CLI user) instead of writing them exclusively in Kuzzle logs.
This has been done primarily to make CLI users wait when performing a `reset`, as deleting users may take minutes depending on the amount of users to delete, and feedback is necessary to indicate that the CLI is performing actions (prior to this, the command would provide no feedback and users may think that the CLI has frozen)
CLI feedbacks have been extended to the `dump` and `shutdown` CLI commands.

# Fixed issue

#861

### DIFF
--- a/lib/api/controllers/cli/cleanDb.js
+++ b/lib/api/controllers/cli/cleanDb.js
@@ -40,15 +40,11 @@ function resetStorage (kuzzle, statusUpdate) {
   return deleteUsers(kuzzle, statusUpdate)
     .then(() => kuzzle.internalEngine.deleteIndex())
     .then(() => statusUpdate('Kuzzle internal database deleted'))
-    .then(() => {
-      const promises = ['internalCache', 'memoryStorage']
-        .map(id => kuzzle.services.list[id].flushdb());
-
-      return Bluebird.all(promises);
-    })
-    .then(() => statusUpdate('Kuzzle memory cache deleted'))
+    .then(() => kuzzle.services.list.internalCache.flushdb())
     .then(() => {
       kuzzle.indexCache.remove(kuzzle.internalEngine.index);
+      statusUpdate('Kuzzle internal cache flushed');
+
       return kuzzle.internalEngine.bootstrap.all();
     });
 }
--- a/lib/api/controllers/cli/cleanDb.js
+++ b/lib/api/controllers/cli/cleanDb.js
@@ -19,6 +19,94 @@
  * limitations under the License.
  */
 
+const 
+  Bluebird = require('bluebird'),
+  Request = require('kuzzle-common-objects').Request;
+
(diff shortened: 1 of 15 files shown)


lib/api/controllers/cli/cleanDb.js
```diff
@@ -19,6 +19,94 @@
...
+ * @returns Promise
+ */
+function resetStorage (kuzzle, statusUpdate) {
+  statusUpdate('Kuzzle reset initiated: this may take a while...');
+  
+  return deleteUsers(kuzzle, statusUpdate)
+    .then(() => kuzzle.internalEngine.deleteIndex())
+    .then(() => statusUpdate('Kuzzle internal database deleted'))
+    .then(() => {
+      const promises = ['internalCache', 'memoryStorage']
```
dbengsch: I would be more confortable if we only flushed the `internalCache` as we do for ES data, `memoryStorage` should not be flushed.

lib/api/controllers/securityController.js
```diff
@@ -711,12 +690,25 @@ class SecurityController {
...
-      updateMethod = this.kuzzle.pluginsManager.getStrategyMethod(request.input.args.strategy, 'update'),
-      validateMethod = this.kuzzle.pluginsManager.getStrategyMethod(request.input.args.strategy, 'validate');
+    const 
+      id = request.input.resource._id,
+      strategy = request.input.args.strategy;
+
+    return this.kuzzle.repositories.user.load(id)
+      .then(user => {
+        if (user === null) {
+          throw new BadRequestError(`Cannot create credentials: unknown kuid "${id}"`);
```
dbengsch: `update` instead of `create`

lib/api/controllers/cli/cleanDb.js
scottinet: This is legacy code, I only moved this function without looking what it does. But I agree, clients storage shouldn't be resetted, at least until we add more options to the `reset` CLI action.

I'll remove the memory storage reset ASAP.

lib/api/controllers/securityController.js
scottinet: bart

PR #876, opened 2017-06-14
Changed deleteSpecifications API return type

samniisan:
Small PR to change the `deleteSpecifications` API return type (when resolved) to a more logical one (empty `array` to `boolean`).
Doc change underway.


samniisan:
@ballinette Thanks for the advise, changed to `{acknowledged: true}`

Issue #790, opened 2017-04-09
I don't find pipe to work on data before notify


neolas:
Hello,

I need to work on data sent by a subscription. I try to use proxy:broadcast but it's in read-only, and proxy:notify is not triggered.

It's a blocker for my personal project, so for the moment I hack the notify function in notifier.js like this : 
```
      kuzzle.pluginsManager.trigger('proxy:beforeNotify', data.payload)
            .then( payload => { data.payload = payload; kuzzle.entryPoints.proxy.dispatch(eventName, data)})
            .catch(err => { console.log("Error : ",err); }); 
```
With this, I can manipulate content before notify with a classic pipe.

Cheers,

ballinette:
Hi @neolas. need to be discussed with the other team members, but your solution looks good to me.
Feel free to suggest a Pull Request with your patch ;)

scottinet:
I disagree. 

I don't see why the already existing `proxy:notify` event cannot be used for this purpose. To me, the best solution is to allow pipe plugins to use it and interact with the notification data before the message is broadcasted.

I'll propose a pull request with this solution.

scottinet:
After reviewing the current (and old) core notifier code, I propose a refactor. This should be fairly quick to implement.

I propose the following:

- `proxy:notify` and `proxy:broadcast` events are removed
- new `notify` specialized events are added:
  - `notify:message`: notifications about realtime messages
  - `notify:document`: notifications about document changes
  - `notify:channel`: notifications about users joining or leaving channels
  - `notify:token`: notifications about token expirations
- pipe plugins can intercept, modify or block notifications
- code split for better readability:
  - the core `notify` function will be split into `notifyDocument`, `notifyMessage`, `notifyChannel` and `notifyToken`
  - the ubiquitous `NotificationObject` will be split into `NotificationDocument`, `NotificationMessage`, `NotificationChannel` and `NotificationToken` classes

The expected gain of these actions are:

- better performances: replacing the polymorphic `NotificationObject` class will allow V8 to treat notification contents as C++ classes instead of dictionaries. Given the possible amount of notifications to process per second, this change might be noticable.
- better separation of concerns and smaller functions/classes => better maintenability
- if necessary, it should be fairly easier to add more events compared to the current code
- this refactor will bring this module in line with our current way of implementing stuff

scottinet:
For the moment I'll delete these two events, so the only breaking change of this refactor will be included in the RC10 version.

[...]

PR #882, opened 2017-06-16
Modernize notifier core module

scottinet:
# Description

Following discussions on #790, several changes have been implemented:

* split the ubiquitous `notify` method and the `NotificationObject` into more specialized versions for documents, users and server notifications
* notification methods now trigger specialized events: `notify:document`, `notify:user` and `notify:server`. Pipe plugin can update the notification content or even block it
* notification methods also trigger a global event at the very end of the process, just before sending the notification: `notify:dispatch`. Pipe plugins can update or block notifications
* notifications have been rationalized to make them more consistent (they were derived from old `RequestObject` objects). The `jwtTokenExpired` notification in particular has been changed to a `TokenExpired` specialized notification (BREAKING CHANGE)

# Boyscout

* modernized old notifier unit tests
* resolved promise rejection warnings in some unit tests
* `funnelController.processRequest`: there was an unit test on this method that was failing "silently" because of a lack of `return` statement at the end of it. This is fixed, and the bug causing this test to fail has also been fixed

# Other tasks to perform

* SDKs must be adapted to the `jwtTokenExpired` notification change
* documentation will also be updated

# Solved issue

#790


lib/api/controllers/documentController.js
```diff
@@ -162,7 +162,7 @@ class DocumentController {
         modifiedRequest = newRequest;
 
         request.input.args.state = 'pending';
```
ballinette: is `request.input.args.state` still necessary ?

scottinet: It appears to be useless, and for a long time to boot.

Removed this assignation from all of this controller's methods.

lib/api/core/models/notifications/NotificationDocument.js
```diff
@@ -22,78 +22,46 @@
...
+    this.scope = scope;
+    this.state = state;
+    this.result = content;
+    this.type = 'document';
   }
 
-  toJson() {
-    const object = {
-      error: null,
+  toJSON() {
```
scottinet: No we don't

scottinet:
@benoitvidis > done

Also, while documenting the newly added events, I found out that the payload was inconsistent. This is now fixed.

PR #885, opened 2017-06-19
update packages dependencies

stafyniaksacha:


package.json
```diff
@@ -19,70 +19,70 @@
...
+    "readline-sync": "^1.4.7",
+    "reify": "^0.11.24",
+    "rimraf": "^2.6.1",
     "sorted-array": "^2.0.1",
-    "utf-8-validate": "^3.0.1",
-    "validator": "^6.2.1",
-    "ws": "2.3.1"
+    "utf-8-validate": "^3.0.2",
+    "validator": "^7.0.0",
+    "ws": "3.0.0"
```
benoitvidis: ws@3 introduces a [breaking change](https://github.com/websockets/ws/releases/tag/3.0.0) in not storing the `upgradeReq` object in the socket anymore.
Not that critical here as we are only using it in debug & log message but still...

stafyniaksacha: Done !

benoitvidis: :+1:

Issue #581, opened 2016-12-15
max subscription # setting


benoitvidis:
We need a way to bind Kuzzle memory usage.

The main memory consumer in Kuzzle is the subscriptions. A first simple approach is to limit their number.

NB: This setting should be stored in the db.

PR #899, opened 2017-06-28
Add a configuration limiting the number of real-time rooms

scottinet:
# Description

Add a new configuration limit: `subscriptionRooms`
This configuration limits the number of unique rooms (not the number of users nor the number of channels), allowing to control the maximum amount of memory used by Kuzzle.
This configuration has been set to 1 000 000 by default.

This limit can be set to 0 for an unlimited number of rooms.

To help with this task, two new methods have been introduced in the DSL API: `normalize` and `store`.
Used together, these methods are strictly equivalent to the already existing `register` one.
These two methods allow computing the Room unique identifier without storing it, allowing the hotelClerk module to check the rooms count against the configured limit, withholding DSL storage until all checks are passed.

# Other changes

* Upgrade sinonjs from 1.17 to 2.3.5. Updated unit tests impacted by breaking changes on the sinonjs API
* Stub `console` commands on the `PluginsManager` unit tests to remove useless messages
* `subscribe:unsubscribe` now returns a `NotFoundError` if a user tries to unsubscribe from an unknown room or from a room they did not subscribe to
* modernize unit tests on the hotelClerk core module
* if an invalid server limit is configured, instead of issuing a warning and autocorrecting the value, Kuzzle now throws and exits with an explicit error message, to prevent running with a different configuration than the one intended by administrators
* add missing unit tests on the configuration check method

# Solved issue

#581


scottinet:
* `dsl.register` should not be removed: it is still used by the document validator module, and it can be used by plugins. This method is easier to use and should be preferred for the general use case, the two new methods only resolve the case where we need a room ID before storing the room, so that we may cancel the registration without the extra cost of storing and then removing a room in the DSL
* I'll submit a PR right away: the `formattedFilters` were absent from this event's payload for a long time now (since the DSL engine v2 in fact)

scottinet:
@benoitvidis > good idea :+1:

scottinet:
@benoitvidis > done

Also added a new check in the configuration limits checks. Oh, and I figured that only issuing a warning and autocorrecting/defaulting invalid configuration was not a good idea, since it's like running Kuzzle with a configuration different than the one asked.

So now, Kuzzle throws and exits if an invalid limit is configured, with an explicit error message to help administrators to fix the configuration issue.

PR #908, opened 2017-07-05
Fixed missing meta in notifications

samniisan:
Fixes missing _meta in Notification responses.


lib/api/core/notifier.js
```diff
@@ -176,7 +176,8 @@ class NotifierController {
       }
 
       this.notifyDocument(rooms, request, scope, state, request.input.action, {
-        _source: request.input.body,
+        _meta: request.input.body ? (request.input.body._kuzzle_info || {}) : {},
```
scottinet: Either this is a real-time message and it has no meta, or it's a document and it's not yet created.
Unless the document is about to be replaced, and the user fetched it before changing the content and sending it back to Kuzzle. In that case, then the situation is worse: the meta data come from the soon-to-be obsolete document.

Either way, I think that the `publish` method can never emit meta data for `about to be created` notifications.

samniisan: Understood :) just removed `_meta` in this scenario.

samniisan:
@benoitvidis My bad, should have checked the tests. Fixed :+1:

PR #909, opened 2017-07-06
Expose trigger method to plugin context

xbill82:
Fix #909 

The `pluginsManager.trigger` method is wrapped and exposed in the `PluginContext` object. The wrapper performs minor checks and pre-pends the plugin name in order to avoid collisions with Kuzzle native events.

### DIFF
--- a/lib/api/core/plugins/pluginContext.js
+++ b/lib/api/core/plugins/pluginContext.js
@@ -105,6 +105,11 @@ class PluginContext {
         }
       });
 
+      Object.defineProperty(this.accessors, 'trigger', {
+        enumerable: true,
+        get: () => curryTrigger(pluginName).bind(kuzzle)
+      });
+
       /**
        * @param {string} collection
        * @param {?function} ObjectConstructor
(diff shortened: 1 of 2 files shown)


test/api/core/pluginContext/pluginContext.test.js
```diff
@@ -149,6 +155,30 @@ describe('Plugin Context', () => {
...
+      const trigger = kuzzle.pluginsManager.trigger;
+      context.accessors.trigger('event:with:colons');
+      should(trigger).be.calledWith('log:error');
+    });
+
+    it('should call trigger with the given event name and payload', () => {
+      const trigger = kuzzle.pluginsManager.trigger;
+      const eventName = 'backHome';
+      const payload = {
+        question: 'who\'s is this motorcycle?',
```
dbengsch: :joy:

lib/api/core/plugins/pluginContext.js
```diff
@@ -105,6 +105,11 @@ class PluginContext {
         }
       });
 
+      Object.defineProperty(this.accessors, 'trigger', {
+        enumerable: true,
+        get: () => curryTrigger(pluginName).bind(kuzzle)
```
stafyniaksacha: Can you avoid usage of `bind` method by setting `kuzzle` as an argument ?

lib/api/core/plugins/pluginContext.js
```diff
@@ -177,6 +182,37 @@ function execute (request, callback) {
...
+   * @param  {String} eventName The name of the custom event to trigger.
+   * @param  {Object} payload   The payload of the event.
+   */
+  return function trigger (eventName, payload) {
+    const kuzzle = this;
+
+    if (eventName.indexOf(':') !== -1) {
+      kuzzle.pluginsManager.trigger(
+        'log:error',
+        new PluginImplementationError(`Custom event invalid name (${eventName}). Colons are not allowed in custom events.`)
```
stafyniaksacha: Why did we refuse usage of colons ?

xbill82: because events can contain _one and only one colon_. And the generated event name already contains a colon (`plugin-${pluginName}:${eventName}`)

lib/api/core/plugins/pluginContext.js
xbill82: good remark, but I used the `bind` to stay consistent with the rest of the file.

stafyniaksacha: We dit it because we always dit it like this ? :joy:

PR #913, opened 2017-07-07
Fix freeze/crash on real-time unsubscriptions + rooms inconsistencies

scottinet:
# Description

* Fix freeze + GC timeouts resulting in a process crash when hundreds of real-time unsubscriptions are asked from Kuzzle (for instance while monkey testing). This was due to the HotelClerk's `removeRoomForAllCustomers` method executing a lot of needless blocking loops
* Fix zombie room creations, occurring when subscriptions are being processed after a network connection was closed
* Fix Kuzzle triggering diagnostic tools from normal 4xx errors during filter canonicalization

# Other changes

* The funnel controller now discards any request if its associated connection is closed (this check ignores HTTP connections, obviously)
* Common errors (e.g. 4xx's) are not logged anymore, preventing logs from being clogged, and also preventing Kuzzle performances to be degraded because of lots of erroneous requests (easy DDOS?)
* HotelClerk.getRealtimeCollections now returns a copy of the rooms list, instead of a copy built one by one from a copy of the rooms list
* A few local optimizations here and there (mainly replacing Array.forEach by `for(;;)` loops)


lib/api/controllers/funnelController.js
```diff
@@ -200,16 +201,19 @@ class FunnelController {
...
+    }
+
     this.checkRights(request)
       .then(modifiedRequest => this.processRequest(modifiedRequest))
       .then(processResult => callback(null, processResult))
-      .catch(err => {
-        this._executeError(err, request, true, callback);
-        result = 1;
-      });
+      .catch(err => this._executeError(err, request, true, callback));
```
scottinet: I've removed these lines because `result` is set to `1` asynchronously, but its value is returned before it can be changed by the promise result function.
Ergo, in its current state, `execute` always returns `0`.

Moreover, at this point, the result status of `execute` doesn't matter: `-1` is only used when no free request slot has been found, so that the request can be queued to be executed at a later time.

scottinet: Moreover, the return value of `execute` should only be used by `_playCachedRequests`, which only cares about a return code of `-1`, in order to rebsubmit the request when that happens.
So it doesn't really matter if execute return 0 or 1

ballinette: I agree that for now, the implemented use cases only need to distinguish `-1` from other result values...
... but the actual behaviour of this method just is not consistent with the jsdoc.

IMHO, we need
* either to update the jsdoc (and return only `0` or `-1` - that means also return `0` in line 197 above)
* or to return `1` each time we have an error, that means, for example, replace the current block code by somthing like:
```javascript
return this.checkRights(request)
  .then(modifiedRequest => this.processRequest(modifiedRequest))
  .then(processResult => {
    callback(null, processResult);
    return 0;
  })
  .catch(err => {
    this._executeError(err, request, true, callback);
    return 1;
  });
```

ballinette: OK, after reading more carefully the code, its clearer for me...
Just the jsdoc line `0: ok, 1: error, -1: delayed` is quite confusing with what kind of "errors" we are speaking at.

maybe we should just me more explicit in the jsdoc, with something like 
```
  * @returns {Number} -1: request delayed, 0: request processing, 1: error while trying to get the request slot
```
?

ballinette: Anyway, it's nitpicking and does not block the PR ;)

scottinet: Your proposal about returning a promise is a no-go.
But I agree with updating the jsdoc.

scottinet: @ballinette > done. I also updated the `mExecute` function

PR #919, opened 2017-07-11
Make package.json mandatory for plugins

### LINKED ISSUE #915: Make package.json mandatory for plugins
Plugin name will be read from the mandatory `package.json` file. A unicity constraint must be enforced on plugins (only one plugin with a given name can be installed on a Kuzzle instance).

xbill82:
fix #915 

* `package.json` is now mandatory for plugins. Plugins without it will not be loaded
* Plugin names are determined from `package.json`
* Plugin names collisions are not allowed


lib/api/core/plugins/pluginsManager.js
```diff
@@ -283,14 +283,26 @@ function loadPlugins(config, kuzzleVersion, rootPath) {
...
+
+    if (loadedPlugins[packageJson.name]) {
+      throw new PluginImplementationError(`A plugin named ${packageJson.name} already exists`);
+    }
+
     const
-      pluginPath = path.resolve(pluginsDir, pluginDirName),
       plugin = {
-        name: path.basename(pluginPath), // we use only basename of plugin path to allow multi instances of same plugin
+        name: packageJson.name, // we use only basename of plugin path to allow multi instances of same plugin
```
stafyniaksacha: Could you update or remove obsolete comment here ?

PR #927, opened 2017-07-26
Prerequisites for cluster mode

benoitvidis:
This PR adds a core event to the notifier to be consumed by the cluster.

### Boyscouting

* bash script to get chrome devtools url added (could not find a way to auto-open it as chrome forbids it).
* added a `sorted` option to `realtime:list` action to get subscription (useful to cluster debugging)


benoitvidis:
cluster needs to be able to delete rooms without throwing => wip

lib/api/core/hotelClerk.js
```diff
@@ -366,6 +395,9 @@ class HotelClerk {
 
         return this.kuzzle.pluginsManager.trigger('room:new', {index, collection, roomId: response.id})
           .then(modified => {
+            // pipes are not allowed to modified the room id
```
ballinette: nitpicking: `sed s/to modified/to modify/`

lib/api/core/hotelClerk.js
```diff
@@ -366,6 +395,9 @@ class HotelClerk {
...
+            const roomId = response.id;
```
ballinette: should also replace the return statement (line 421):
` return {diff: response.diff, roomId: modified.roomId};`
=>
` return {diff: response.diff, roomId};`

lib/api/core/hotelClerk.js
```diff
@@ -497,8 +527,18 @@ class HotelClerk {
       throw new NotFoundError(`Unsubscribe error: not subscribed to ${roomId}`);
     }
 
-    if (notify) {
-      this.kuzzle.pluginsManager.trigger('core:hotelClerk:removeRoomForCustomer', {requestContext, roomId});
+    return this._removeRoomForCustomerWithoutThrowing(requestContext, roomId, notify);
+  }
+
+  _removeRoomForCustomerWithoutThrowing (requestContext, roomId, notify = true) {
```
ballinette: what's the goal of this method ?
As far as I read, it seems to only called by `_removeRoomForCustomer`, which already throw exceptions for the cases

benoitvidis:
@ballinette : about `_removeRoomForCustomerWithouthThrowing`, beside having a terrible name, this method was originally meant to be called by the cluster to avoid cases where late syncs could try to delete a already-removed room.
The way it is done now has changed and it is not needed anymore. I am going to revert this part.

PR #928, opened 2017-07-26
Allow plugins to dynamically (un)register authentication strategies

scottinet:
# Description

Use cases were submitted and demonstrated that a way to add or remove authentication strategies dynamically was required for some businesses.

One of them involves a back-end managing several short-lived applications, each one of them needing its own unique OAUTH2 identifier. As there is roughly one new application per week, it seems very inconvenient to restart Kuzzle each time a new OAUTH2 identifier needs to be added.

This pull request adds a new `strategies` accessor in the plugin context, exposing two currified methods: `add` and `remove`

Strategy addition or removal triggers a `core:auth:strategy[Added|Removed]` event, propagating the action across a cluster

Documentation update: https://github.com/kuzzleio/documentation/pull/339
Cluster update: https://github.com/kuzzleio/kuzzle-plugin-cluster/pull/43

# Notes

Plugins are responsible of ensuring that, once a strategy is added on-the-fly, it'll also be added during startup, to ensure that there won't be inconsistencies if a cluster reboots

# Plugins Manager (slight) refactor

* To avoid going back & forth between the plugins manager class instance and external private methods, I had to move authentication initialization and registration inside the class. For the sake of consistency, I decided to move ALL of these init&register methods inside the class (e.g. initHooks, registerPipe, and so on)
* The properties `authentications` (strategies repository) and `registeredStrategies` (strategies list) have been merged into the new `strategies` property, as maintaining 2 variables seemed unnecessary

# Other changes

* Rename `assertIsConnected` to `assertIsAuthenticated`, as it makes more sense
* Fix a problem where a plugin creating an unchained, uncatchable rejected promise forces Kuzzle to trigger its diagtools and exit. This behavior is now triggered only if `NODE_ENV !== production` (as this still can be useful to quickly spot bugs during development), otherwise only an error message is logged.

### DIFF
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -29,7 +29,7 @@ const
   KuzzleInternalError = require('kuzzle-common-objects').errors.InternalError,
   {
     assertIsStrategyRegistered,
-    assertIsConnected,
+    assertIsAuthenticated,
     assertHasBody,
     assertHasStrategy,
     assertBodyHasNotAttribute,
@@ -195,7 +195,7 @@ class AuthController {
    * @returns {Promise<object>}
    */
   updateSelf(request) {
-    assertIsConnected(this.kuzzle, request);
+    assertIsAuthenticated(this.kuzzle, request);
     assertHasBody(request);
     assertBodyHasNotAttribute(request, '_id');
     assertBodyHasNotAttribute(request, 'profileIds');
@@ -218,7 +218,7 @@ class AuthController {
    * @returns {Promise.<Object>}
    */
   createMyCredentials(request) {
-    assertIsConnected(this.kuzzle, request);
+    assertIsAuthenticated(this.kuzzle, request);
     assertHasStrategy(request);
     assertIsStrategyRegistered(this.kuzzle, request);
     assertHasBody(request);
@@ -237,7 +237,7 @@ class AuthController {
    * @returns {Promise.<Object>}
    */
   updateMyCredentials(request) {
-    assertIsConnected(this.kuzzle, request);
+    assertIsAuthenticated(this.kuzzle, request);
     assertHasStrategy(request);
     assertIsStrategyRegistered(this.kuzzle, request);
     assertHasBody(request);
@@ -254,7 +254,7 @@ class AuthController {
    * @returns {Promise.<Object>}
    */
   credentialsExist(request) {
-    assertIsConnected(this.kuzzle, request);
+    assertIsAuthenticated(this.kuzzle, request);
     assertHasStrategy(request);
     assertIsStrategyRegistered(this.kuzzle, request);
 
@@ -268,7 +268,7 @@ class AuthController {
    * @returns {Promise.<Object>}
    */
   validateMyCredentials(request) {
-    assertIsConnected(this.kuzzle, request);
+    assertIsAuthenticated(this.kuzzle, request);
     assertHasStrategy(request);
     assertIsStrategyRegistered(this.kuzzle, request);
     assertHasBody(request);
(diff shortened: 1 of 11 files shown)


lib/api/controllers/authController.js
```diff
@@ -29,7 +29,7 @@ const
   KuzzleInternalError = require('kuzzle-common-objects').errors.InternalError,
   {
     assertIsStrategyRegistered,
-    assertIsConnected,
+    assertIsAuthenticated,
```
dbengsch: :+1:

lib/api/core/plugins/pluginContext.js
```diff
@@ -252,4 +262,50 @@ function instantiateRequest(request, data, options = {}) {
...
+ * @param  {string} pluginName 
+ * @return {function} function taking a strategy name and properties, 
+ *                    registering it into kuzzle, and returning
+ *                    a promise
+ */
+function curryAddStrategy(kuzzle, pluginName) {
+  return function addStrategy(name, strategy) {
+    return new Bluebird((resolve, reject) => {
+      try {
+        kuzzle.pluginsManager.registerStrategy(pluginName, name, strategy);
```
ballinette: suggestion: add a hook `strategy:add` to enable cluster plugin to propagate the action (see commente [here](https://github.com/kuzzleio/documentation/pull/339#discussion_r130080996)

lib/api/core/plugins/pluginContext.js
```diff
@@ -252,4 +262,50 @@ function instantiateRequest(request, data, options = {}) {
...
+ * @param  {string} pluginName 
+ * @return {function} function taking a strategy name and properties, 
+ *                    registering it into kuzzle, and returning
+ *                    a promise
+ */
+function curryRemoveStrategy(kuzzle, pluginName) {
+  return function removeStrategy(name) {
+    return new Bluebird((resolve, reject) => {
+      try {
+        kuzzle.pluginsManager.unregisterStrategy(pluginName, name);
```
ballinette: suggestion: add a hook `strategy:remove` to enable cluster plugin to propagate the action (see commente [here](https://github.com/kuzzleio/documentation/pull/339#discussion_r130080996)

lib/api/core/plugins/pluginContext.js
scottinet: Didn't like the name you proposed but apart from that: events added.

I'm working on the cluster code to make nodes synchronize dynamic strategies management.

lib/api/core/plugins/pluginsManager.js
```diff
@@ -238,15 +237,406 @@ class PluginsManager {
...
+              }
+              else if (result.message && typeof result.message === 'string') {
+                callback(null, false, result);
+              }
+              else {
+                callback(null, false, `Was not able to log in using the strategy "${strategyName}"`);
+              }
+            }
+            else {
+              callback(new UnauthorizedError('Login failed.'));
```
ballinette: according to passportJS specifications, the callback should be invoked with an error only if an exception occurs during the verification. (see http://passportjs.org/docs#verify-callback )
Here, it's not an exception. We should just invoke the callback with a `false` result:
`callback(null, false)`

scottinet: Yes and no.
You're right about plugins returning `false` as a result, as it should be handled as a simple login rejection, and not an exception.

But we also should handle cases where plugins return erroneous values (for instance, an array, a number, etc).

So I added a new test case: if `result` is explicitly set to `false` then we invoke the callback with that boolean value. In all other cases, we invoke the callback with a `PluginImplementationError` exception (instead of an `UnauthorizedError` one).

Issue #922, opened 2017-07-20
Subscriptions keep open when clients are gone


benoitvidis:
Running some monkey tests, once all clients have exited, open subscriptions can still be see in realtime::list.

benoitvidis:
First problem identified: Catastrophic  cartesian product cases can explode the number of subfilters in Kuzzle dsl filter tree (up to 100k for **one** subscribtion in my tests _seed 798964700_)

Leads:
* Such cases most likely occur when combining multi-value operators together (i.e. `in` or `ids`). 
In most cases, some impossible cases can be identified from the output of the canonization.

ex: 
```
         [ [ { equals: { 'nested.string': 'foo' }, not: false },
             { equals: { _id: 'acda0aba-053b-453e-98b0-a082902cbaf6' },
               not: false },
             { equals: { _id: 'f7d67838-6207-4acb-b5c3-9794b14c289d' },
               not: false },
             { regexp: { string: { value: 'foo.*' } }, not: false } ],
           [ { equals: { 'nested.string': 'Mais où retrouver à présent la trace presque effacée de l\'ancien crime ? - Sophocle, Oedipe-Roi' },
               not: false },
             { equals: { _id: 'acda0aba-053b-453e-98b0-a082902cbaf6' },
               not: false },
             { equals: { _id: 'f7d67838-6207-4acb-b5c3-9794b14c289d' },
               not: false },
             { regexp: { string: { value: 'foo.*' } }, not: false } ],
           [ { equals: { 'nested.string': 'All your base are belong to us' },
               not: false },
             { equals: { _id: 'acda0aba-053b-453e-98b0-a082902cbaf6' },
               not: false },
             { equals: { _id: 'f7d67838-6207-4acb-b5c3-9794b14c289d' },
               not: false },
             { regexp: { string: { value: 'foo.*' } }, not: false } ],
           [ { equals: { 'nested.string': 'foo' }, not: false },
             { equals: { _id: 'acda0aba-053b-453e-98b0-a082902cbaf6' },
               not: false },
             { equals: { _id: 'f7d67838-6207-4acb-b5c3-9794b14c289d' },
               not: false },
             { regexp: { string: { value: 'foo.*' } }, not: false } ],
           [ { equals: { 'nested.string': 'All your base are belong to us' },
               not: false },
             { equals: { _id: 'acda0aba-053b-453e-98b0-a082902cbaf6' },
               not: false },
[...]

Issue #768, opened 2017-03-31
DSL "equals" operator optimization 


benoitvidis:
Last update on DSL converted the "equals" operator from an plain object to a Map.
Benckmarks show that map.get is in average at least 7x slower then getting an object property.

Optim to implement:
* `equals` storage as an object: 
```js
{
  string: {},
  number: {},
  boolean: {},
  object: {}
}
```
The trick is that all scalar types we accept on the `equals` operator return a different value when calling `typeof`.
We can then safely access the matching storage _without any test_, just by hitting `storage[typeof val]`.

PR #929, opened 2017-07-26
DSL - post canonization optimizations

benoitvidis:
# Description

This PR aims to reduce the impact of possible catastrophic cardinality explosion induced by the canonization process while subscribing (cf #922 for samples).

It also refactors the implementation of the `equals` operator for better performances (#768)

### DIFF
--- a/lib/api/dsl/storage/removeOperands.js
+++ b/lib/api/dsl/storage/removeOperands.js
@@ -60,11 +60,14 @@ class OperandsRemoval {
       value = condition.value[fieldName],
       operand = foPairs[index][collection].equals;
 
-    if (operand.fields[fieldName].get(value).length > 1) {
-      operand.fields[fieldName].get(value).splice(operand.fields[fieldName].get(value).indexOf(subfilter), 1);
+    if (operand.fields[fieldName][typeof value][value].length > 1) {
+      operand.fields[fieldName][typeof value][value].splice(operand.fields[fieldName][typeof value][value].indexOf(subfilter), 1);
     }
-    else if (operand.fields[fieldName].size > 1) {
-      operand.fields[fieldName].delete(value);
+    else if (Object.keys(operand.fields[fieldName][typeof value]).length > 1) {
+      delete operand.fields[fieldName][typeof value][value];
+    }
+    else if (Object.keys(operand.fields[fieldName]).length > 1) {
+      delete operand.fields[fieldName][typeof value];
     }
     else if (operand.keys.array.length > 1) {
       operand.keys.remove(fieldName);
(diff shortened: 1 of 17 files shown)


benoitvidis:
double checking a possible issue

lib/api/dsl/match/matchNothing.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+module.exports = (storage, testTables) => {
+  testTables.addMatch(storage.fields.all);
```
scottinet: Why do we add matches on this keyword? Shouldn't we match nothing?

lib/api/dsl/storage/removeOperands.js
```diff
@@ -151,6 +154,10 @@ class OperandsRemoval {
     this.exists(foPairs, index, collection, subfilter, condition, 'notexists');
   }
 
+  nothing(foPairs, index, collection) {
```
ballinette: nitpicking: missing JSDoc.

ballinette:
looks good to me, as far as I understand the DSL implementation... but I would need a little workshop to be sure to be aware with it ;)

(NB: maybe there could be a later improvement with ranges)

lib/api/dsl/match/matchNothing.js
benoitvidis: You're right. Actually, I just realized we don't need the matcher at all.
Updating it.

benoitvidis:
@ballinette sure, I guess we can set a session :)

Regarding the anti-topology detection, there is quite some space for improvements (regex, range exclusion, nothing operator etc.), which is good news. 

For the time being, the main case I could find were the cardinality would explose is when combining equalities, i.e.
```js
const filter = {
  and: [
    or: [
       { equals: {foo: 1} },
       { equals: {foo: 2} },
       // ...
       { equals: {foo: 'n'} }
    ],
    or: [
       { equals: {bar: 1} },
       { equals: {bar: 2} },
       // ...
       { equals: {bar: 'n'} }
    ];
    or: [
       { equals: {baz: 1} },
       { equals: {baz: 2} },
       // ...
       { equals: {baz: 'n'} }
    ]
  ]
};
```

This example produces `n^3` minterms. 

The first fix was very optimistic as the worse found case would include a lot anti-topology

Just added a new parameter (disabled by default to avoid change breaking), to limit the number of minters (AND clauses) to cover a wider range of cases.

PR #934, opened 2017-08-03
Fixes an error when trying to update a non-existing SecurityDocument

samniisan:
Fixes an error when trying to update a non-existing SecurityDocument (User|Profile|Role)

### DIFF
--- a/lib/api/controllers/securityController.js
+++ b/lib/api/controllers/securityController.js
@@ -701,13 +701,7 @@ class SecurityController {
     assertHasId(request);
 
     return this.kuzzle.repositories.role.loadRole(request.input.resource._id)
-      .then(role => {
-        if (!role) {
-          return Bluebird.reject(new NotFoundError(`Cannot update role ${request.input.resource._id}: role not found`));
-        }
-
-        return updateMetadata(role, request);
-      })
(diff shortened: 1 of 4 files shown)


lib/api/controllers/securityController.js
```diff
@@ -1122,9 +1116,14 @@ function formatUserSearchResult(kuzzle, formatter, raw) {
  * Sets the metadata for an update request
  * @param {object} securityDocument
  * @param {Request} request
+ * @param {string} context
```
dbengsch: wording: `context` word should be kept for `userContext` only.

samniisan: Updated.

samniisan:
@scottinet Not sure I get your point, this PR still returns a NotFoundError, just moved things around

ballinette:
indeed, the `NotFoundError` was only triggered for `updateRole` before this PR.
Now, the code is factorized and the error is triggered for both `updateUser` and `updateProfile` as well ;)

scottinet:
Oh, right, I missed the updated code for `updateMetadata`. I thought your PR would allow updating non-existing documents by somehow creating them.

PR #964, opened 2017-09-13
Make the real-time engine an independant public library

scottinet:
# Description

Make the inner implementation of Kuzzle's real-time percolation engine an independant NPM library: [Koncorde](https://www.npmjs.com/package/koncorde)


scottinet:
It's probably the best PR description length / changes made ratio I've ever produced.

lib/api/core/hotelClerk.js
```diff
@@ -574,14 +573,14 @@ class HotelClerk {
...
 
   /**
-   * Delete room if no user has subscribed to it, and remove it also from the DSL
+   * Deletes a room if no user has subscribed to it, and removes it also from Koncorde
    *
    * @this HotelClerk
    * @param {string} roomId
    */
-  _removeRoomFromDsl (roomId) {
+  _removeRoomFromKoncorde (roomId) {
```
ballinette: nitpicking: I would use a more generic function name: `_removeRoomFromRealtime` instead of `..FromKoncorde`

scottinet: Comments and code now refer to the `real-time engine` instead of `Koncorde`

PR #966, opened 2017-09-15
Fix invalid authentication options storage

scottinet:
# Description

Fix an absent-mindedness error leading to authentication options provided by auth plugins being stored with the wrong key name, making that option useless without this fix

# Other changes

* refactor the PassportWrapper to make them 1/ up-to-date with current practices and 2/ test for correct options handling
* document a part of the passport wrapper that leads to think that there is a promise resolution problem without a thorough investigation on the matter


scottinet:
@ballinette > `authenticate`'s prototype modified

PR #975, opened 2017-09-29
Fix the documents garbage collector service

scottinet:
# Description

I'm working on the document metadata documentations, so I had to read the documents garbage collector service, and found a few problems:

* it's started 2 times: once in the Kuzzle startup sequence, and the other instance comes from the service initializer, automatically launching the service `init` method (I don't think this code should be a "service", but requalifying it as a core module would introduce a breaking change in Kuzzle configuration files)
* only 1 timer instance was kept, but this module should really be a singleton, removing the need of keeping track of `setTimeout` calls and cleaning the timer
* it gets the list of indexes and collections from the database layer, instead of using Kuzzle's index cache 
* it tries to clean documents as soon as it is started, and I find that behavior potentially harmful in production environments
* it asks the database layer to clean all collections at once, instead of sequentially, potentially stressing the database layer especially on databases with large number of collections to clean up
* it checks if Kuzzle is overloaded only before starting the cleaning sequence, where it would be more benefical to verify Kuzzle's status before every collection clean-up, to interrupt the GC if Kuzzle becomes overloaded in the middle of the cleaning process
* the garbage collector configuration was missing from the `.kuzzlerc.sample` file

This PR fixes all these issues. Garbage collector unit tests were also converted to today's coding standards.

# Other changes

Kuzzle mock now uses `sinon` new `stub.resolves` and `stub.rejects` method, for more readability.

### DIFF
--- a/lib/services/garbageCollector.js
+++ b/lib/services/garbageCollector.js
@@ -22,120 +22,123 @@
 'use strict';
 
 const
-  OneHour = 3600000,
-  OneDay = OneHour * 24,
   Bluebird = require('bluebird'),
   Request = require('kuzzle-common-objects').Request;
 
-function GarbageCollector(kuzzle) {
-  this.timer = null;
+// Time constants
+const 
+  oneHour = 3600000,
+  oneDay = oneHour * 24;
 
-  this.init = function garbageCollectorInit() {
-    this.run();
+// There can be only one
+let _highlander = null;
+
+class GarbageCollector {
+  constructor(kuzzle) {
+    this.kuzzle = kuzzle;
+  }
+
+  init () {
+    // We do not want to trigger the garbage collector right away:
+    // if for whatever reason Kuzzle is (re)started on a production
+    // environment (after a crash, because of autoscaling, etc),
+    // the last thing a client needs is to clean up trashcans
+    setTimeout(() => this.run(), oneHour);
+
     return Bluebird.resolve(this);
-  };
+  }
 
-  this.run = function garbageCollectorRun () {
+  run () {
     let ids = [];
-    const
-      body = {
-        from: 0,
-        size: kuzzle.config.services.garbageCollector.maxDelete || 10000,
-        sort: [{ '_kuzzle_info.deletedAt': { unmapped_type: 'date'} }],
-        query: {
-          bool: {
-            should: [
-              {
-                term: {
-                  '_kuzzle_info.active': false
-                }
-              }
-            ]
-          }
-        }
-      };
 
-    if (this.timer) {
-      clearTimeout(this.timer);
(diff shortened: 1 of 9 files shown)


lib/services/garbageCollector.js
```diff
@@ -22,120 +22,128 @@
...
+class GarbageCollector {
+  constructor(kuzzle) {
+    this.kuzzle = kuzzle;
+  }
+
+  init () {
+    // We do not want to trigger the garbage collector right away:
+    // if for whatever reason Kuzzle is (re)started on a production
+    // environment (after a crash, because of autoscaling, etc),
+    // the last thing a client needs is to clean up trashcans
```
ballinette: :+1:

scottinet:
@ballinette , @dbengsch > I've spotted a few discrepancies while working on the documentation, so I made small modifications on this PR. You might want to review the last commits

lib/services/garbageCollector.js
```diff
@@ -22,120 +22,128 @@
...
+          .filter(index => index[0] !== '%'); // skipping reserved indexes
 
-            res.indexes.forEach(index => {
-              if (index === kuzzle.internalEngine.index) {
-                return;
-              }
+        const collections = [];
+
+        for (const index of indexes) {
+          collections.push(...this.kuzzle.indexCache.indexes[index].map(collection => ({index, collection})));
```
benoitvidis: nitpicking: we don't have any control on the number of collections per index. I would rather use `concat` here

lib/services/garbageCollector.js
```diff
@@ -22,120 +22,128 @@
...
+        }
+
+        return Bluebird.resolve(collections)
+          .each(icpair => {
+            return this.clearCollection(icpair.index, icpair.collection)
+              .then(deletedIds => {
+                if (deletedIds.length) {
+                  // using concat instead of push + spread operator because the list
+                  // of deleted IDs may be large
+                  ids = ids.concat(deletedIds);
```
benoitvidis: nitpicking: why keeping an array of possibly long ids as we only use their count if I am not mistaken?

scottinet: I agree. Moreover, I don't see what anyone can do with a bunch of permanently deleted document identifiers. 

But that would introduce a breaking change since this is the current payload for the `gc:end` event.
So for now I chose to not change that behavior, but I'm not at ease with it either. 

If we do change that, then I would rather provide a list of index-collection pairs that have been cleaned up instead. This is the most useful piece of information I can think of, even if it's by a small margin over not providing any payload atr all.

lib/services/garbageCollector.js
scottinet: fixed

PR #994, opened 2017-11-15
Bug fix - deleteByQuery on invalid index/collections result in a dump

### LINKED ISSUE #992: deleteByQuery on invalid index/collections result in a dump...
Trying to delete a index / collection that doesn't existe resulted in a core dump

dump_no-such-index.zip available on my hard drive if needed

benoitvidis:
* fixes #992 

Discussions during the implementation of this fix led to rethink a little `deleteByQuery`, introducing some small breaking changes (which is acceptable IMO considering the previous behaviour never was documented):

* deleteByQuery response switched from the list of ids to `{deleted: <num of docs deleted>}`
* as a consequence, deleteByQuery does not send any notification anymore

### features changes

* param `includeTrash` added. If set to `true`, also deletes from the trash
* param `bypassTrash` added. If set to `true`, actually deletes documents from es instead of updating them.


benoitvidis:
need to rethink optional query parameters => wip

lib/services/elasticsearch.js
```diff
@@ -488,70 +509,44 @@ class ElasticSearch extends Service {
...
-          return this.client.bulk(esRequestBulk)
-            .then(() => this.refreshIndexIfNeeded(esRequestBulk, {ids}))
-            .catch(error => reject(this.esWrapper.formatESError(error)));
-        });
-      });
+    // refresh is needed here to prevent versions conflicts and is acceptable in a delete by query context
+    return this.client.indices.refresh({index: esRequest.index})
+      .then(() => esMethod(esRequest))
+      .then(response => {
+        this.refreshIndexIfNeeded(esRequest, response);
```
ballinette: This will return the result before the resolution of `refreshIndexIfNeeded()` promise.

shoudln't it be rather:
```
then(response => this.refreshIndexIfNeeded(esRequest, {deleted: response[responseMember]}))
```
?

benoitvidis: oops, you're right. Fixed.

benoitvidis:
If OK with you, I propose to close this PR and create a new one using the old `scroll` + `bulk` mechanism.

Using ES `(delete|update)ByQuery` seems to always lead to version conflicts errors and it is a pitty IMO to deal with it manually on the client side.

scottinet:
@benoitvidis > Before doing so, wouldn't it be a good idea to check Elasticsearch's changelogs just to make sure the version we use doesn't have a problem that has since been fixed? If not, then I agree with emulating these couple commands manually on our side

PR #1006, opened 2017-11-28
kuzzleContext.constructors.Request: Allow to override request input args with falsy value

### LINKED ISSUE #1004: Bad request construction when using kuzzleContext.constructors.Request in plugin context
When building a new Request using the plugin context (`kuzzleContext.constructors.Request`) the args property is not build properly when the data part is a number value of 0 or a boolean value of false (Must come from lib/api/core/plugins/pluginContext.js line around 290 : `if (!target.input.args[arg]) {`).

Example :

With a request.input.args.from = 4 the following code :
```
var r = new context.constructors.Request(request, {
  from: 0
})
console.log(r.input.args.from)
```
Will output `4` instead of the expected `0`

ballinette:
fix #1004

### DIFF
--- a/lib/api/core/plugins/pluginContext.js
+++ b/lib/api/core/plugins/pluginContext.js
@@ -295,12 +295,12 @@ function instantiateRequest(request, data, options = {}) {
     }
 
     for (const arg of Object.keys(_request.input.args)) {
-      if (!target.input.args[arg]) {
+      if (target.input.args[arg] === undefined) {
         target.input.args[arg] = _request.input.args[arg];
       }
     }
 
-    if (!target.input.jwt) {
+    if (_data.jwt === undefined) {
       target.input.jwt = _request.input.jwt;
     }
   }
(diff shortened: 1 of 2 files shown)


lib/api/core/plugins/pluginContext.js
```diff
@@ -295,12 +295,12 @@ function instantiateRequest(request, data, options = {}) {
     }
 
     for (const arg of Object.keys(_request.input.args)) {
-      if (!target.input.args[arg]) {
+      if (target.input.args[arg] === undefined) {
```
scottinet: I think you meant `!== undefined`

But then, undefined values cannot be iterated over using `for..of`, so this test can be removed entirely.

lib/api/core/plugins/pluginContext.js
```diff
@@ -295,12 +295,12 @@ function instantiateRequest(request, data, options = {}) {
...
         target.input.args[arg] = _request.input.args[arg];
       }
     }
 
-    if (!target.input.jwt) {
+    if (_data.jwt === undefined) {
```
scottinet: I think you meant `!== undefined`

PR #1007, opened 2017-11-29
Allow strict mapping definition

### LINKED ISSUE #1002: The elasticsearch mapping option dynamic=strict is not usable
When a mapping is inserted in Kuzzle with the dynamic option set to strict, any action to create a document fails due to the automatically added _kuzzle_info field.

benoitvidis:
Fixes #1002 by auto-injecting `_kuzzle_info` internal propery mapping.

## Boyscouting 

(big one, sorry -_-;...)
Modified all tests that would use patterns like `mock.returns(Bluebird.rejects())` to avoid premature escalation of UnhandledRejection before the test promise is checked.

### DIFF
--- a/lib/services/internalEngine/bootstrap.js
+++ b/lib/services/internalEngine/bootstrap.js
@@ -82,32 +82,6 @@ class InternalEngineBootstrap {
       properties: {
         controllers: {
           enabled: false
-        },
-        _kuzzle_info: {
-          properties: {
-            author: {
-              type: 'text',
-              fields: {
-                keyword: {
-                  type: 'keyword'
-                }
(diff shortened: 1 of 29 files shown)


lib/services/internalEngine/bootstrap.js
```diff
@@ -82,32 +82,6 @@ class InternalEngineBootstrap {
...
-            },
-            updater: {
-              type: 'text',
-              fields: {
-                keyword: {
-                  type: 'keyword'
-                }
-              }
-            }
-          }
```
benoitvidis: internalEngine does not use meta data.

PR #1008, opened 2017-11-30
Hotfix plugin execute

benoitvidis:
**:warning: depends on #1007**

Kuzzle plugin::execute method would never clean up the funnel pendingRequests when the overload protection is active.

### DIFF
--- a/lib/api/controllers/funnelController.js
+++ b/lib/api/controllers/funnelController.js
@@ -443,6 +443,9 @@ class FunnelController {
         this.handleErrorDump(e);
         callback(e);
         return null;
+      })
+      .finally(() => {
+        delete this.pendingRequests[request.id];
       });
 
     return 0;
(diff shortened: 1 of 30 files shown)


benoitvidis:
@ballinette because unit-tests now fail on my system (and probably travis) without the related fix about premature unhandled rejections.

lib/api/controllers/funnelController.js
```diff
@@ -443,6 +443,9 @@ class FunnelController {
         this.handleErrorDump(e);
         callback(e);
         return null;
+      })
+      .finally(() => {
+        delete this.pendingRequests[request.id];
```
scottinet: I would rather delete this request just after `getRequestSlot` answers us that we can process it, since it means that it isn't pending anymore. 

And since the overload protection mechanism can optionally be bypassed by plugins, this means I would add this [line 429](https://github.com/kuzzleio/kuzzle/blob/2381e29b0c68fa398321021287c8f0e68c3bc928/lib/api/controllers/funnelController.js#L429)

benoitvidis: Following our discussion, I just removed `pendingRequests`.
@ballinette: You may want to review again. We could not find any use for this queue, while it obviously induces some risks.

PR #1016, opened 2017-12-04
 [cli] Log fixtures import errors

### LINKED ISSUE #1001: The fixtures feature at start does not report errors
When fixtures are imported, no error are reported and it is stated that all documents were imported well.

Way to reproduce: declare a collection a dynamic = strict and try to import document with new fields

benoitvidis:
Fixes #1001

### DIFF
--- a/lib/services/elasticsearch.js
+++ b/lib/services/elasticsearch.js
@@ -679,17 +679,24 @@ class ElasticSearch extends Service {
         // If some errors occured during the Bulk, we send a "Partial Error" response :
         if (result.errors) {
           result.partialErrors = [];
+          const items = [];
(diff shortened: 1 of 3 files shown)


lib/services/elasticsearch.js
```diff
@@ -679,17 +679,24 @@ class ElasticSearch extends Service {
...
+            if (item.error) {
+              item.action = action;
+              result.partialErrors.push(item);
+            }
+            else {
+              items.push(item);
             }
           }
+
+          result.items = items;
```
benoitvidis: boyscouting: split valid from errored items in result.items vs result.partialErrors

PR #1024, opened 2018-01-09
raise an error if the document is not active

jenow:
We now raise an error if we try to delete a document which is already deleted and has the _kuzzle_info.active attribute to false

### DIFF
--- a/lib/services/elasticsearch.js
+++ b/lib/services/elasticsearch.js
@@ -474,28 +474,28 @@ class ElasticSearch extends Service {
 
     return this.client.get({index: esRequest.index, type: esRequest.type, id: esRequest.id})
       .then(result => {
-        if (result._source._kuzzle_info && result._source._kuzzle_info.active) {
-          // injecting retryOnConflict default configuration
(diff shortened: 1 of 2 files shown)


lib/services/elasticsearch.js
```diff
@@ -472,24 +472,31 @@ class ElasticSearch extends Service {
 
     assertWellFormedRefresh(esRequest);
 
-    // injecting retryOnConflict default configuration
-    if (!esRequest.hasOwnProperty('retryOnConflict') && this.config.defaults.onUpdateConflictRetries > 0) {
-      esRequest.retryOnConflict = this.config.defaults.onUpdateConflictRetries;
-    }
+    return this.client.get({index: esRequest.index, type: esRequest.type, id: esRequest.id})
+      .then(result => {
+        if (result._source._kuzzle_info && result._source._kuzzle_info.active) {
```
ballinette: if the document has no `_kuzzle_info` attribute, we consider that it is active.

we should rather 
* first reject if `_kuzzle_info` is set AND is `false`
* then update the document with the metadata otherwise.

like in `get` method.

PR #1040, opened 2018-02-05
fixes #1036 `callback is not a function`

### LINKED ISSUE #1036: PluginExecute `callback is not a function`
It appears this error is coming when Kuzzle is under pressure, exceeding its max concurrent requests.
As a work around, increasing it to 500 makes it work perfectly well.

benoitvidis:
cf #1036 for further explanation on the issue.

This PR takes a quite radical approach by excluding any plugin:execute action from the overload protection.
The reasons behind this decision are:

* the overload is a complex part of Kuzzle code and any additional complexity makes it more difficult to maintain
* plugin controllers already go through the "regular" overload mechanism. 

### Boyscouting

* increased default overload trigger (a simple kuzzle installation on my laptop can handle more than 200 req/s)
* increased default pipe timers, which tend to pollute logs on production systems.

### DIFF
--- a/lib/api/core/plugins/pluginContext.js
+++ b/lib/api/core/plugins/pluginContext.js
@@ -164,6 +164,12 @@ function execute (kuzzle, request, callback) {
     reject,
     deferred;
 
+  if (callback && typeof callback !== 'function') {
+    error = new PluginImplementationError(`Invalid argument: Expected callback to be a function, received "${typeof callback}"`);
+    kuzzle.pluginsManager.trigger('log:error', error);
+    return Bluebird.reject(error);
+  }
+
   if (!callback) {
     deferred = new Bluebird((res, rej) => {
       resolve = res;
--- a/lib/api/core/plugins/pluginContext.js
+++ b/lib/api/core/plugins/pluginContext.js
@@ -157,48 +157,37 @@ class PluginContext {
  * @param {boolean} [overloadProtect]
  * @param {Function} [callback]
  */
-function execute (kuzzle, request, overloadProtect = true, callback = null) {
-  let resolve = null,
-    reject = null, 
-    deferred = null,
-    _callback = callback,
-    _overloadProtect = overloadProtect;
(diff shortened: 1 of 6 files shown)


lib/api/core/plugins/pluginContext.js
```diff
@@ -157,48 +157,37 @@ class PluginContext {
...
-  if (!_callback) {
+  if (!callback) {
     deferred = new Bluebird((res, rej) => {
       resolve = res;
       reject = rej;
     });
   }
 
-  if (!error && (!request || !(request instanceof Request))) {
+  if (!request || !(request instanceof Request)) {
```
ballinette: Nitpicking: it seems `error` can now only set by next line...
So lines 174->185 can be simplified like that:
```javascript
if (!request || !(request instanceof Request)) {
  const error = new PluginImplementationError('Invalid argument: a Request object must be supplied');
  if (callback) {
    return callback(error);
  }
  reject(error);
  return deferred;
}
```

lib/api/core/plugins/pluginContext.js
```diff
@@ -157,48 +157,37 @@ class PluginContext {
...
-  }
+function execute (kuzzle, request, callback) {
+  let
+    error,
+    resolve,
+    reject,
+    deferred;
 
-  if (!_callback) {
+  if (!callback) {
```
scottinet: As this PR is dealing with the `callback is not a function` errror, I would also add a test before that one to verify that the `callback` argument is, indeed, a callable function, before we start using it in the remaining of this method.

That way we could intercept and raise a proper `PluginImplementationError` instead of letting Kuzzle deal with an uncatched error.

benoitvidis: done

lib/api/core/plugins/pluginContext.js
benoitvidis: done

PR #1041, opened 2018-02-05
Add validation for profile edition

### LINKED ISSUE #1038: [security] Validation for profiles
Currently, when updating a profile, any json entry is accepted as-is.

For instance, this is accepted:
```
  "policies": [
    {
      "roleId": "admin",
      "restrictedTo": [
        {
          "index": "qwant",
          "collection": "a"
        }
      ]
    }
  ]
}
```

While `restrictedTo` only accepts `collections` as an array.
I suppose Kuzzle should throw a BadRequestError when the submitted json is not valid.

benoitvidis:
fixes #1038

### DIFF
--- a/lib/api/core/models/security/profile.js
+++ b/lib/api/core/models/security/profile.js
@@ -77,12 +77,51 @@ class Profile {
    * @return {Promise<boolean>}
    */
   validateDefinition() {
-    if (!_.isArray(this.policies)) {
-      return Bluebird.reject(new BadRequestError('The roles member must be an array'));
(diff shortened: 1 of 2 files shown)


ballinette:
OK for me, but it seems some behaviour tests should be fixed now...

benoitvidis:
Hum.. after second look, the current validation does not seem correct to me or at least does not match the documentation:
http://docs.kuzzle.io/guide/essentials/security/

`restrictedTo` should accept a array of collections in a `collections` member.
=> double check => wip

PR #1044, opened 2018-02-08
Identify requests using their immutable identifier

scottinet:
(1.x equivalent of the following hotfix: https://github.com/kuzzleio/kuzzle/pull/1043)

The overload-protection mechanism relied on the request's `requestId` property, which is a mutable, client-defined identifier, intended to help clients linking Kuzzle responses to the corresponding requests they emitted.

This means that if a malicious party floods a Kuzzle server with requests using the same `requestId` value,  this forces the overload-protection mechanism to intervene, and it tries to buffer these new requests. Since a request with the same identifier is already buffered, then the new ones are discarded.
This leads to an unrecoverable memory leak in Kuzzle's entry point system, as long as in protocol plugins, eventually saturating a Kuzzle processus memory, forcing it to shut down.

This pull request fixes that issue by relying on the new `kuzzle-common-object` module version, which adds an immutable, internal identifier when a request object is created server-side. This identifier is now used by the overload-protection mechanism in place of the mutable `requestId` one.

### DIFF
--- a/lib/api/controllers/funnelController.js
+++ b/lib/api/controllers/funnelController.js
@@ -138,8 +138,8 @@ class FunnelController {
 
     // resolves the callback immediately if a slot is available
     if (this.concurrentRequests < this.kuzzle.config.limits.concurrentRequests) {
-      if (this.requestsCacheById[request.id]) {
-        delete this.requestsCacheById[request.id];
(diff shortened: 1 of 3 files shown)


scottinet:
merging as its hotfix counterpart has already been validated

PR #1049, opened 2018-02-20
Improve docker build

benoitvidis:
This PR brings the following new abilities:

* do not use pm2 anymore
* allow to pass options to kuzzle binary directly from command
* boyscouting: replace deprecated MAINTAINER directive with LABELs

### DIFF
--- a/docker-compose/scripts/run.sh
+++ b/docker-compose/scripts/run.sh
@@ -2,26 +2,29 @@
 
 set -e
 
+log () {
+  echo "[$(date --rfc-3339 seconds)] - $1"
+}
+
 elastic_host=${kuzzle_services__db__client__host:-http://elasticsearch:9200}
 
-echo "[$(date --rfc-3339 seconds)] - Waiting for elasticsearch to be available"
+log "Waiting for elasticsearch"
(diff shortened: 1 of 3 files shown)


docker-compose/scripts/run.sh
```diff
@@ -2,26 +2,29 @@
...
+    log "Could not connect to elasticsearch in time. Aborting..."
     exit 1
 fi
 
-echo "[$(date --rfc-3339 seconds)] - Starting Kuzzle..."
+log "Starting Kuzzle..."
 
-pm2 start --silent /config/pm2.json
-pm2 logs --lines 0 --raw
+exec ./bin/kuzzle "$@"
```
ballinette: As current script is now copied into /usr/local/bin/kuzzle (according to your Dockerfile's changes), and outside application directory, shouldn't we call `/var/app/bin/kuzzle` (with absolute path) instead ?

benoitvidis: The path the script is located does not matter, only the one where it is run, being the container WORKDIR (/var/app), so everything should be OK (and is, tested locally)

PR #1064, opened 2018-03-08
Dependencies update

alexandrebouthinon:


features/run.sh
```diff
@@ -18,6 +18,6 @@ for endpoint in Embedded Proxy; do
       host="$embedded_host"
       port="$embedded_port"
     fi
-    ./node_modules/.bin/cucumberjs --format progress-bar -p "${protocol}${endpoint}" --world-parameters "{\"host\": \"${host}\", \"port\": \"${port}\"}"
+    ./node_modules/.bin/cucumber-js --exit --format progress-bar -p "${protocol}${endpoint}" --world-parameters "{\"host\": \"${host}\", \"port\": \"${port}\"}"
```
scottinet: I'm against the use of the `--exit` option, as this forces cucumber to exit even if tests are still in the event loop, running.
If functional tests never exit, then it may be either one of two things:
- there is a bug in functional tests, meaning than there may be a test not filling its purpose
- there is a bug in kuzzle, and an API route never answers, forcing test scenarios to wait indefinitely

This is also what the authors of cucumber say in their "exiting" documentation: https://github.com/cucumber/cucumber-js/blob/master/docs/cli.md#exiting

This argument should be removed and the neverending test issue should be investigated

PR #1034, opened 2018-01-24
Do not inject http headers into request's input args

ballinette:
# Bugfix
#1030: catch any exception thrown inside an HTTP request scenario, and send back a response with errors

# Other improvements
 #1031 : do not inject HTTP headers to request's input args. (they are already accessible via `request.input.headers`)

PR #1066, opened 2018-03-14
Hotfix: head and options http headers

scottinet:
https://github.com/kuzzleio/kuzzle/pull/1034 fixes HTTP headers being incorrectly assigned to the `args` section of the `RequestInput` object, instead of the `headers` one.

But the default HEAD and OPTIONS HTTP requests were left out of these changes, so this PR fixes that.

### DIFF
--- a/lib/api/core/httpRouter/index.js
+++ b/lib/api/core/httpRouter/index.js
@@ -68,7 +68,6 @@ class Router {
 
     // Add an automatic HEAD route on the '/' url, answering with default headers
     attach('/', (request, cb) => {
-      Object.assign(request.input.args, this.defaultHeaders);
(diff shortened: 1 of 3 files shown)


lib/api/core/httpRouter/index.js
```diff
@@ -68,7 +68,6 @@ class Router {
 
     // Add an automatic HEAD route on the '/' url, answering with default headers
     attach('/', (request, cb) => {
-      Object.assign(request.input.args, this.defaultHeaders);
```
scottinet: Default headers are automatically assigned to the `RequestResponse` object, and they serve no purpose in the input part of the request, so this line is actually useless.

PR #1073, opened 2018-03-28
Use bulk insertions for multi-documents API routes

scottinet:
# Description

Multi-documents routes (mcreate, mupdate, and so on) featured subpar performances (approx. 1000 documents/min on my laptop), due to how they invoked repetitively their mono-document routes counterparts. For instance, a `mCreate` with 1.000 documents needed to emulate 1.000 API calls to the `create` API route.
The other downside of that method is that it could very quickly fill up the overload-buffer, hence the default limit of 200 documents per request.

This PR completely rewrites the m* routes using Elasticsearch's bulk API. Performances are now measured around 500.000 documents/min on my laptop (w/ real-time subscriptions).
The default limit of 200 documents per request has been kept, but the check ensuring that this limit was not too high has been removed.

# Other changes

* Remove the funnel controller's request history, as it takes too much memory space for little to no gain
* Fix functional tests: they were running 6x on the default "embedded-websocket" target, because of changes on how parameters are passed to the world constructor since cucumberjs@4. 
* Fix how the `bulk:import` API call detects partial errors: it now relies on the `status` response property instead of the non-existent `error` one
* Fix a bug in the statistics component, where frames may not be retrieved when using a redis cluster
* Change the way the notifier component interacts with Redis, using string values instead of sets, to allow mass-getting keys from cache
  * Remove custom redis methods, as they were almost exclusively used by the now-obsolete code in the notifier component
* Rename `notifyDocumentDelete` to `notifyDocumentMDelete` to fit the new notifier methods naming convention
  * Rewrite that method to make it work on a set of values instead of _n_ calls to a function handling 1 deletion at a time
* Minor dependencies update
* Revert `rewire` to version `2.5.2` due to a bug with nyc (see https://github.com/istanbuljs/nyc/issues/747)
* GLaDOS is now GLaD again


lib/services/elasticsearch.js
```diff
@@ -907,6 +909,302 @@ class ElasticSearch extends Service {
...
+  /**
+   * Delete multiple documents with one request
+   * 
+   * @param  {Request} request - Kuzzle API request
+   * @return {Promise}
+   */
+  mdelete(request) {
+    const 
+      esRequest = initESRequest(request, ['consistency', 'refresh', 'timeout', 'fields']),
+      index = request.input.resource.index,
```
kuzzle: MINOR Use destructuring syntax for these assignments from "request.input.resource". [rule](http://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS3514)

kuzzle:
* ![MINOR][MINOR] 1 minor

features/run.sh
```diff
@@ -18,6 +18,8 @@ for endpoint in Embedded Proxy; do
       host="$embedded_host"
       port="$embedded_port"
     fi
-    ./node_modules/.bin/cucumber-js --format progress-bar -p "${protocol}${endpoint}" --world-parameters "{\"host\": \"${host}\", \"port\": \"${port}\"}"
+
+    # profiles are defined in the cucumber.js file at the root of this project
```
ballinette: so the env variables $CUCUMBER_PROXY_HOST, $CUCUMBER_EBEDDED_HOST, etc, are not used anymore ?

scottinet: Good catch, I removed those because I thought they were redundant with the cucumber.js configuration file.
Change reverted.

Aschen:
I encounter a memory allocation problem when I try to insert 1 000 000 documents by batch of 50 000.  
The problem occurs around the 650 000 document inserted.  
I am using the NYC Taxi Database.

scottinet:
will investigate ASAP

scottinet:
@Aschen > the memleak was caused by the funnel controller's request history feature. It keeps the last few requests (default: 50), to provide context to coredumps analysis.
This history is near-useless and it makes memory management impossible, so @benoitvidis and I decided to remove it altogether.

PR #1075, opened 2018-03-28
Add templates for issues and pull request

Aschen:
Add a bug issue template to help user to write good bug reports

Here is the template : 

## Steps to Reproduce
1.
2.
3.
4.


appacea:
add version and/or build and os

PR #1090, opened 2018-04-13
Hotfix: JWTs TTL reset to default cache TTL (24min)

scottinet:
# Description

The global repository system has an auto-refresh feature. The idea behind that feature is that, if a cache item is frequently accessed, then it should be kept in the cache indefinitely, while entries that are accessed only once expire after some time (default: 1440 seconds), freeing up cache space.

That feature had 2 issues that this PR fixes: 

* it ignored the configured entry TTL and always used the configure default TTL. If entries are created with a different TTL to reflect different needs, the auto-refresh will override that value whenever the entry is accessed
* JWTs are handled by a specialized repository (the `tokenRepository` module), which creates cache entries with the same TTL than the token's one. Unfortunately, the auto-refresh feature was not disabled for this repository, leading to entries TTL being reset to the default 1440s each time a token was being accessed

Stricter unit tests have been added to make sure that this problem never arises again

# Boyscout 

The `repository` unit tests now use the mocked Kuzzle stubs instead of creating new ones.

### DIFF
--- a/lib/api/core/models/repositories/tokenRepository.js
+++ b/lib/api/core/models/repositories/tokenRepository.js
@@ -248,6 +248,18 @@ class TokenRepository extends Repository {
         }));
       });
   }
+
+  /**
+   * The repository main class refreshes automatically the TTL
+   * of accessed entries, letting only unaccessed entries expire
+   *
+   * But tokens' TTL must remain the same than their expiration time,
+   * refreshing a token entry has no meaning.
+   *
+   * So we need to override the TTL auto-refresh function to disable it
+   */
+  refreshCacheTTL() {
+  }
 }
 
 module.exports = TokenRepository;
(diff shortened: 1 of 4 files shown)


lib/api/core/models/repositories/tokenRepository.js
```diff
@@ -248,6 +248,18 @@ class TokenRepository extends Repository {
...
+  /**
+   * The repository main class refreshes automatically the TTL
+   * of accessed entries, letting only unaccessed entries expire
+   *
+   * But tokens' TTL must remain the same than their expiration time,
+   * refreshing a token entry has no meaning.
+   *
+   * So we need to override the TTL auto-refresh function to disable it
+   */
+  refreshCacheTTL() {
```
kuzzle: CRITICAL Add a nested comment explaining why this function is empty or complete the implementation. [rule](http://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS1186)

kuzzle:
* ![CRITICAL][CRITICAL] 1 critical

Issue #1091, opened 2018-04-14
Login failed:  Cannot get roles for uninitialized profile anonymous


Tyki:
Just installed Kuzzle 1.2.11 and on the first access to the backoffice, i'm creating an administrator and resetting anonymous rights

Note that this issue is not showing up in Kuzzle 1.2.6 (pre 1.2.7 with refactor of security controller)
## Expected Behavior

Be able to connect to backoffice

## Current Behavior

Login failed
## Possible Solution

Not resetting the anonymous rights

## Steps to Reproduce

1. Setup a new Kuzzle instance
2. Go to the Backoffice Kuzzle
3. Create an admin and check "reset anonymous rights"
4. Constat that it's impossible to connect with an admin account due to anonymous rights 

## Context (Environment)
Kuzzle 1.2.11, ES 5.4

ballinette:
Confirmed.
Bug present since release 1.2.7

PR #1094, opened 2018-04-18
add feature tests for createFirstAdmin method

ballinette:
### What does this PR do ?

Issue #1091 revealed that we did not test `createFirstAdmin` action.
This PR add features test for that, as well as hooks to correclty clean-up kuzzle instance before and after the tests.

### Other changes

* ~~Because of the implementation of HTTP router, the route `/:_id/_createFirstAdmin` did not work (`_id` parameter was not injected correctly).
That is because it has conflicts with other routes starting with `/:index/` (we do not allow different parametric names at uri root level).
Workaround: replace by '/_createFirstAdmin/:_id'~~

* Fix HTTP router, which did not work well when we have 2 parametric routes with 2 different parameter's name at the same level 
As a consequence, because many routes match `/:index/*`, the route `:_id/_createFirstAdmin` did not work (the http router injected the first parameter to `index` variable instead of `_id`)

* In feature tests, `BeforeAll` and `AfterAll` hooks did not use `world-parameters` variables coming from command line (because cucumber does not inject the world context to these hooks), and so we were not able to run tests with other host than `localhost`.
It's fixed by manually getting and parsing the command line parameter, using `minimist`

### Boyscout

Fix natural languages in a test description.


features/support/hooks.js
```diff
@@ -124,6 +132,33 @@ function cleanSecurity () {
     });
 }
 
+function grantDefaultRoles () {
```
ballinette: this hook is called after `createFirstAdmin`: it's goal is to re-grant all rights to anonymous user, so that we can then call the other hooks and clean-up Kuzzle database after the tests.

scottinet:
```
Because of the implementation of HTTP router, the route /:_id/_createFirstAdmin did not work (_id parameter was not injected correctly).
That is because it has conflicts with other routes starting with /:index/ (we do not allow different parametric names at uri root level).
```

I didn't know such a case existed when I created our http router. But in hindsight I should not be surprised, and this is a limitation we need to drop, by replacing the current definition of the `parametric` property in the `RoutePart` class:

```
    this.parametric = {
      name: '',
      subparts: null
    };
```

With something along these line:

```
   this.parametric = {
     <parameter name>: <subpart object>,
     <parameter name2>: subpart object>,
     ...
   };
```

Obviously we need a few adaptations in our tree parsing when routing requests to the right controller, but I don't think it would be less efficient.

At the very least, we should right now make the HTTP router complain when a parametric value is set with a different name than the one(s) already registered, and it should prevent Kuzzle from starting until the problem is fixed.

scottinet:
Thanks. I apologize for that limitation, I simply didn't think that through when I designed this router. :-(

features/step_definitions/profiles.js
```diff
@@ -92,6 +95,26 @@ Then(/^I'm ?(not)* able to find the profile with id "([^"]*)"(?: with profile "(
             return callbackAsync(new Error(`Profile with id ${id} exists`));
           }
 
+          if (profile) {
+            const
+              compare = (a, b) => {
+                return a.roleId < b.roleId;
```
scottinet: Comparison functions must return consistent results, or else the sort result may be undefined.
Here, if `a == b`, then that function:
* returns `true` if called with `(a, b)`
* also returns `true` if called with `(b, a)`

features/step_definitions/profiles.js
```diff
@@ -92,6 +95,26 @@ Then(/^I'm ?(not)* able to find the profile with id "([^"]*)"(?: with profile "(
...
+              },
+              policies = stringify(body.result._source.policies.sort(compare)),
```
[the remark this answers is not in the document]
scottinet: I have good news and bad news: 
* Javascript's `sort` algorithm does not guarantee stability :-(
* json-stable-stringify claims to be stable, and it does sort JSON arrays too

So I suggest you remove the call to `sort`

scottinet:
It's fine as long as you don't have different objects with the same role ID, as js's sort doesn't not guarantee stability. :)

PR #1097, opened 2018-04-23
Add cache expiration for document change subscription

Aschen:
## What does this PR do ?

Redis is used by Kuzzle to cache documents that enter or left a scope.  
Whenever a document is updated, replaced or deleted, these cached item are used to tell if the document enter or left scopes.   
When a document is created, catched by a real-time filter and never updated again, it will stay cached in Redis forever.  

This PR add a TTL to this cache. By default the TTL is set to 72 hours.  
It can be configured in the `.kuzzlerc` under the value `limits.subscriptionDocumentTTL`.  
If a value of 0 is provided, cache will live in Redis forever.  

User should set this value with cautious because it can lead to memory leaks.  

See [backlog#245](https://github.com/kuzzleio/kuzzle-backlog/issues/245)

  - Step 1 : Add this line to the `docker-compose/dev.yml` file before launching the Kuzzle stack `      - kuzzle_limits__subscriptionDocumentTTL=20`
  - Step 2 : Run `docker-compose -f docker-compose/dev.yml up`
  - Step 3 : Start a subscription with no filter with the admin console
  - Step 4 : Create a document with the admin console
  - Step 5 : List all keys in Redis : `docker-compose -f docker-compose/dev.yml exec redis redis-cli KEYS \*`. You should see a key starting with `{notif/<index name>/<collection name>/<document id>`
  - Step 6 : Wait 20sec and run the same command again. The key should have disappear.

### DIFF
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -50,10 +50,16 @@
     //      transformed in their Canonical Disjunctive Normal Form (CDNF).
     //      Set to 0 for no limit.
     // * subscriptionRooms
-    //      Maximum number of different subscription rooms 
+    //      Maximum number of different subscription rooms
     //      (i.e. different index+collection+filters subscription configurations)
     //      Depends primarily on available memory.
     //      If set to 0, an unlimited number of rooms can be created
+    // * subscriptionDocumentTTL
+    //      Maximum time (in seconds) a document will be kept in cache for real-time subscriptions.
+    //      This cache is used to notify subscriber when a document enter or leave a scope after modifications.
+    //      By default, subscription will be kept 72 hour.
+    //      /!\ Remember that keeping subscription over a long period of time can lead to memory leak. /!\
+    //      Set to 0 to keep subscriptions forever.
     "concurrentRequests": 50,
     "documentsFetchCount": 10000,
     "documentsWriteCount": 200,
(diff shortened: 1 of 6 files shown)


.kuzzlerc.sample
```diff
@@ -50,18 +50,25 @@
     //      transformed in their Canonical Disjunctive Normal Form (CDNF).
     //      Set to 0 for no limit.
     // * subscriptionRooms
-    //      Maximum number of different subscription rooms 
+    //      Maximum number of different subscription rooms
     //      (i.e. different index+collection+filters subscription configurations)
     //      Depends primarily on available memory.
     //      If set to 0, an unlimited number of rooms can be created
+    // * subscriptionDocumentTTL
```
Aschen: @appacea Can you check my english please ?

appacea: //      Maximum time (in seconds) a document will be kept in cache for real-time subscriptions.
//      This cache is used to notify subscribers when a document enters or leaves a scope after an update.
//      By default, subscriptions will be kept 72 hours.
//      /!\ Please note that keeping subscriptions over a long period of time can lead to a memory leak. /!\
//      If set to 0, the subscription will be kept in cache forever.

scottinet: TTLs set to extended periods of time won't lead to a memory leak fortunately, but most certainly to more memory consumed.
However, setting that property to 0 **will** (and not *can*) lead to a memleak **if and only if** documents enter a real-time subscription's scope, and never leave it again.

I think that it is very important to be precise here, because memory leaks are serious issues, and almost always the result of bugs, so we need to explain why a memory leak may occur.

ballinette:
why did you set `master` as target branch ?
Is it a hotfix ?

Issue #1099, opened 2018-04-24
Add http:request event hook/pipes


ballinette:
In some use case, we can have the JWT included within cookies instead of Bearer HTTP header.

So we should be able to get token from the cookies and inject it into the request payload before the request processing.

To do that, we need a `http.routed` event in which we can setup pipes.

Also, adding `http.get`, `http.post`, `http.put` and `http.delete` events, in the same way as `http.options`, can be useful

Issue #1020, opened 2017-12-13
An unhandled promise rejection in a plugin controller produces a dump and restart kuzzle


dbengsch:
When promise is rejected without being returned or handled within the plugin, an `unhandledRejection` is produced and provokes a core dump and a kuzzle restart.

Log available if asked

Aschen:
An `unhandledRejection` won't make Kuzzle to dump and restart in a production environment (see [kuzzle.js](https://github.com/kuzzleio/kuzzle/blob/master/lib/api/kuzzle.js#L205).  
I don't know if we can really do something about `unhandledRejection` in plugin..

Aschen:
Otherwise there is a mistake in the `unhandledRejection` event handler parameters, the handler take two parameters `reason` and `promise` (see [event unhandledRejection](https://nodejs.org/api/process.html#process_event_unhandledrejection))
The `reason` parameter can also be `undefined` if the promise is rejected with no reason. Maybe we can update the code : 
```
  process.on('unhandledRejection', (reason, promise) => {

    if (reason !== undefined) {
      console.error(`ERROR: unhandledRejection: ${reason.message}`, reason.stack, promise); // eslint-disable-line no-console
    } else {
      console.error('ERROR: unhandledRejection:', promise); // eslint-disable-line no-console
    }
```

PR #1108, opened 2018-05-04
Add events on http routing

Aschen:
## What does this PR do ?

This PR add 6 events related to the HTTP method used. `http:get`, `http:post`, `http:put`, `http:patch`, `http:delete`, `http:head` (`http:options` was already present).  
See [Jira#KZL-42](https://jira.kaliop.net/browse/KZL-42) or https://github.com/kuzzleio/kuzzle/issues/1099 for more context.  

The events are fired in `Router.route()` and now this method always return a promise.  

Changes are documented here : https://github.com/kuzzleio/documentation/pull/478

  - Step 1 : Run a Kuzzle stack with plugins debug : `DEBUG=kuzzle:plugins docker-compose -f docker-compose/dev.yml up`
  - Step 2 : Try running a curl request and verify if the correct event is fired : `curl -X PUT localhost:7512`
image

  - Step 3 : Repeat with all HTTP methods: `methods=(GET POST PUT PATCH DELETE HEAD); for method in $methods; do curl -X $method localhost:7512/_getInfo; done`

### Boyscout

 - Fix listener arguments on `unhandledRejection` (Related to #1020 )
 - Fix `should().be.called()` syntax in tests
 - Allow user to remove `node_modules` created by docker without `sudo`
 - Replace references to back-office by admin console

### DIFF
--- a/docker-compose/scripts/install-plugins.sh
+++ b/docker-compose/scripts/install-plugins.sh
@@ -13,7 +13,7 @@ for target in ${plugins_dir}/* ${protocols_dir}/* ; do
     cd "$target"
     npm install --unsafe
     find node_modules/ -type f -exec chmod 666 {} \;
-    find node_modules/ -type d -exec chmod 777 {} \;
+    find node_modules/ -type d -exec chmod 755 {} \;
     cd "$working_dir"
   fi
 done
--- a/docker-compose/scripts/install-plugins.sh
+++ b/docker-compose/scripts/install-plugins.sh
@@ -12,7 +12,7 @@ for target in ${plugins_dir}/* ${protocols_dir}/* ; do
     echo 'Installing dependencies for ' $(basename "$target")
     cd "$target"
     npm install --unsafe
+    chmod 777 -R node_modules
     cd "$working_dir"
   fi
 done
-
--- a/docker-compose/scripts/install-plugins.sh
+++ b/docker-compose/scripts/install-plugins.sh
@@ -12,7 +12,8 @@ for target in ${plugins_dir}/* ${protocols_dir}/* ; do
     echo 'Installing dependencies for ' $(basename "$target")
     cd "$target"
     npm install --unsafe
-    chown -R ${DEV_UID}:${DEV_GID} node_modules
+    find -L node_modules/.bin -type f -exec chmod 776 {} \;
+    find node_modules/ -type d -exec chmod 755 {} \;
     cd "$working_dir"
   fi
 done
--- a/docker-compose/scripts/install-plugins.sh
+++ b/docker-compose/scripts/install-plugins.sh
@@ -12,7 +12,7 @@ for target in ${plugins_dir}/* ${protocols_dir}/* ; do
     echo 'Installing dependencies for ' $(basename "$target")
     cd "$target"
     npm install --unsafe
-    find node_modules/ -type f -exec chmod 666 {} \;
+    find -L node_modules/.bin -type f -exec chmod 776 {} \;
     find node_modules/ -type d -exec chmod 755 {} \;
     cd "$working_dir"
   fi
--- a/docker-compose/scripts/install-plugins.sh
+++ b/docker-compose/scripts/install-plugins.sh
@@ -12,8 +12,7 @@ for target in ${plugins_dir}/* ${protocols_dir}/* ; do
     echo 'Installing dependencies for ' $(basename "$target")
     cd "$target"
     npm install --unsafe
-    find -L node_modules/.bin -type f -exec chmod 776 {} \;
-    find node_modules/ -type d -exec chmod 755 {} \;
+    chown -R ${DEV_UID}:${DEV_GID} node_modules
     cd "$working_dir"
   fi
 done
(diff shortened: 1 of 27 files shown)


Aschen:
I noticed that there were no functional test for the plugin system.  
Is it on purpose or just due to lack of time ?

lib/api/core/httpRouter/index.js
```diff
@@ -158,48 +158,56 @@ class Router {
           .catch(error => replyWithError(cb, request, error));
       }
 
-      return replyWithError(cb, request, new BadRequestError(`Unrecognized HTTP method ${httpRequest.method}`));
+      return Promise.reject(replyWithError(cb, request, new BadRequestError(`Unrecognized HTTP method ${httpRequest.method}`)));
```
ballinette: what is the reason to reject the promise ?

lib/api/core/httpRouter/index.js
```diff
@@ -158,48 +158,56 @@ class Router {
...
-        return replyWithError(cb, request, new NotFoundError(`API URL not found: ${routeHandler.url}`));
-      }
+      return this.kuzzle.pluginsManager.trigger(httpEvent, request)
+        .then(mutatedRequest => {
+          routeHandler.request = mutatedRequest;
 
-      if (httpRequest.content.length <= 0) {
-        return routeHandler.invokeHandler(cb);
-      }
+          if (routeHandler.handler === null) {
```
ballinette: should be moved before the event trigger, to avoid calling plugin pipes if the URL does not match any route.

ballinette:
@Aschen what do you mean by "functional test for the plugin system" ?

lib/api/core/httpRouter/index.js
Aschen: To ensure consistency on return value. This function always return a promise, either fulfilled or rejected but maybe it's not mandatory ?

lib/api/core/httpRouter/index.js
Aschen: Are you sure ? Even if the request does not match any route, it's still a HTTP request received by Kuzzle, shouldn't we trigger the event anyway ?

Aschen:
@ballinette I mean tests with cucumber to ensure the events can be hooked or piped to a plugin

docker-compose/scripts/install-plugins.sh
```diff
@@ -12,7 +12,7 @@ for target in ${plugins_dir}/* ${protocols_dir}/* ; do
     echo 'Installing dependencies for ' $(basename "$target")
     cd "$target"
     npm install --unsafe
+    chmod 777 -R node_modules
```
alexandrebouthinon: Why 777 ?

docker-compose/scripts/run-dev.sh
```diff
@@ -6,6 +6,7 @@ elastic_host=${kuzzle_services__db__client__host:-http://elasticsearch:9200}
 
 npm install --unsafe
 npm rebuild all --unsafe
+chmod 777 -R node_modules
```
alexandrebouthinon: Why 777 ?

docker-compose/scripts/run-test.sh
```diff
@@ -6,6 +6,7 @@ elastic_host=${kuzzle_services__db__client__host:-http://elasticsearch:9200}
 
 npm install --unsafe
 npm install --unsafe --only=dev
+chmod 777 -R node_modules
```
alexandrebouthinon: Why 777 ?

alexandrebouthinon:
Use chown instead of chmod for permissions :wink:

docker-compose/scripts/install-plugins.sh
Aschen: Because I'm lazy.. I will fix that :)

Aschen:
Seen with @alexandrebouthinon , we can't use `chown` because we are inside the docker container and we don't know the host user uid. We can assume it's 1000 but since we are not sure I prefer to use a permissive `chmod`

lib/api/core/httpRouter/index.js
```diff
@@ -158,48 +158,58 @@ class Router {
...
-
-      if (httpRequest.content.length <= 0) {
-        return routeHandler.invokeHandler(cb);
-      }
-
-      if (httpRequest.headers['content-type']
-        && !httpRequest.headers['content-type'].startsWith('application/json')) {
-        return replyWithError(cb, request, new BadRequestError(`Invalid request content-type. Expected "application/json", got: "${httpRequest.headers['content-type']}"`));
-      }
+      return this.kuzzle.pluginsManager.trigger(httpEvent, request)
```
scottinet: These new events must be triggered **after** the request has been validated. It makes no sense sending a request to plugins if an URL is invalid or if a request is malformed.

lib/api/core/httpRouter/index.js
ballinette: No, it doesn't. This method works with callbacks and not promises (see also comment from scottinet)

Aschen:
Very clever solution @ballinette, I like it :)

benoitvidis:
@ballinette I do not agree with your solution to an almost non-existing problem.
Injecting some host-specific data just for this looks quite ugly to me. 

In any case, this is for dev only and not blocking imo.

Aschen:
I revert the change suggested by @ballinette because travis wasn't happy

ballinette:
hmmm not blocking, I agree, except that I'm not fan of having a script that silently change the filemod on my host... :/

but travis should work if we fix the command in  kuzzle/test/travis-bin/run-tests.sh like this:
```
DEV_UID=`id -u` DEV_GID=`id -g` docker-compose up -f docker-compose/test.yml run kuzzle 
```

Issue #1061, opened 2018-03-07
The `collection/create` route should accept a mapping object


scottinet:
The current `collection/create` API route allows creating a new collection in Kuzzle, but without any mapping associated to it.

This seems wrong, as it forces users to then call `collection/updateMapping` to update a default mapping (or to create the collection directly with `updateMapping`, an undocumented feature taking advantage of how Elasticsearch works)

The `create` route should accept a mapping object.

PR #1114, opened 2018-05-11
Allow mapping in collection creation route

Aschen:
## What does this PR do ?

This PR allow to specify the mapping when you create a collection.  
Subsequent call on the same route with different mapping will only update the mapping.  

Fix https://github.com/kuzzleio/kuzzle/issues/1061  
Documentation https://github.com/kuzzleio/documentation/pull/480

  - Step 1 : Run Docker Compose stack: `docker-compose -f docker-compose/dev.yml up`
  - Step 2 : Create an index : `curl -X POST localhost:7512/my-index/_create`
  - Step 3 : Create a collection with a mapping: `curl -H "Content-Type: application/json" -X PUT -d '{"properties": {"gordon": {"type": "text"}}}' http://localhost:7512/my-index/my-collection`
  - Step 4 : Check the mapping: `curl -X GET localhost:7512/my-index/my-collection/_mapping?pretty`

### DIFF
--- a/lib/services/elasticsearch.js
+++ b/lib/services/elasticsearch.js
@@ -564,7 +564,7 @@ class ElasticSearch extends Service {
   }
 
   /**
-   * Create an empty collection. Mapping will be apply if supplied.
+   * Create an empty collection. Mapping will be applied if supplied.
    *
    * @param {Request} request
    * @returns {Promise}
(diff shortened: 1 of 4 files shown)


lib/services/elasticsearch.js
```diff
@@ -564,17 +564,20 @@ class ElasticSearch extends Service {
   }
 
   /**
-   * Create an empty collection with no mapping
+   * Create an empty collection. Mapping will be apply if supplied.
```
scottinet: applied

PR #1117, opened 2018-05-16
Document validators bugfixes

scottinet:
## What does this PR do ?

Fix several bugs in the document validators, a few spotted by @dariuspranskus (see https://gitter.im/kuzzleio/kuzzle?at=5ae9c536f2d2d5370459353b), and others found while I investigated Darius' issues:

* The `validateFieldSpecification` method in the `BaseType` class (at the root of all document field datatype validators) has an inconsistent prototype, returning either a `false` boolean value if the specs are erroneous, or a plain `object` (updated options object) otherwise. This leads to confusion and some inheriting classes return `true` when the options are valid, instead of an object (_e.g._ the `string` datatype validator), leading to options being ignored. This is fixed by adopting a consistent prototype: these functions now always return the provided options (updated, defaulted or left as-is), and throw in case of an error
  * this new prototype allows custom rejection messages, allowing users to know exactly why their validator specifications are rejected
  * test cases have been added accross the board to make these validations much stricter
  * since plugins can add new datatypes, non-`KuzzleError` errors thrown are reinstantiated as `PluginImplementationError` errors
* Document `validators` (using Koncorde filters) registration was done in a never returned promise, making these validators ignored
* In case of a document update, we need to fetch the document from the database first, before trying to validate it. The controller/action values used to detect that case were referring to an old version of Kuzzle's API (`write:update` instead of `document:update`)
  * moreover, when fetching a document fails, then the promise chain never gets resolved (nor rejected), remaining in the event loop forever (promise leak). And Kuzzle never responds to the user neither. This is fixed by forcing the chain to be rejected when this happens
* Erroneous validator specifications now throw `PreconditionError` instead of `InternalError` errors, the latter triggering a diagnostic dump
* Fix how Epoch-like values are handled by the `date` field validator:
  * `epoch_millis` and `epoch_second` formats now only accept values provided as `Integer` instead of `String`. This format change is necessary because many unrelated date formats are valid when an Epoch-like date value is provided
  * The `range.min` and `range.max` options now properly handle Epoch-like dates provided as numbers
* The `object` validator now properly rejects `null` values, as well as arrays

### Other changes

* Upgrade old unit tests from ES5 to ES6
* MUCH stricter unit tests
* Simplify some complex parts of the code
* Remove unnecessary promise creations
* Add a few debug messages
* Remove unnecessary method overloads in child classes

### Boyscout

* Fix a few typos here and there
* Fix erroneous or outdated JSDoc comments
* Update dependencies


kuzzle:

Note: The following issues were found on lines that were not modified in the pull request. Because these issues can't be reported as line comments, they are summarized here:

1. ![CRITICAL][CRITICAL] [geoShape.js#L68](https://github.com/kuzzleio/kuzzle/blob/edc415c5471a8bcf7f258667a3d6747ac755a30f/lib/api/core/validation/types/geoShape.js#L68): Add a "default" clause to this "switch" statement. [rule](http://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3ASwitchWithoutDefault)

lib/api/core/validation/types/type.js.template
```diff
@@ -15,6 +15,7 @@ class MyNewType extends BaseType {
    * @param {TypeOptions} typeOptions
    * @param {*} fieldValue
    * @param {string[]} errorMessages
+   * @return {boolean}
```
Aschen: Maybe you can also add the `require` for `PreconditionError` because this will be used in any new type defined

PR #1123, opened 2018-05-30
remove SIGPIPE from the list of abnormal termination signals

scottinet:
## What does this PR do?

Upon receiving a SIGPIPE signal, Kuzzle creates a coredump and shuts itself down. Simply put: SIGPIPE is currently considered by Kuzzle as an abnormal termination signal.

This should not be the case, as a SIGPIPE signal may be emitted for various reasons, for instance when a network socket we're currently reading is abruptly closed.

This means that there is a way to force-close Kuzzle from its API: simply bombard it with requests, and then close the connection while doing it. Repeating the process enough time (or doing it in parallel by opening multiple sockets) will sooner or later trigger a SIGPIPE.

This PR removes SIGPIPE from the list of abnormal termination signals. I simplified the signal registering code a bit in the process.

Start a [Kuzzle Benchmark](https://github.com/kuzzleio/kuzzle-benchmark) with at least 1000 clients, and then interrupt the benchmark in the middle of it.

### DIFF
--- a/lib/api/kuzzle.js
+++ b/lib/api/kuzzle.js
@@ -181,16 +181,6 @@ class Kuzzle extends EventEmitter {
  * @param {Kuzzle} kuzzle
  */
 function registerErrorHandlers(kuzzle) {
-  const coreDumpSignals = {
-    SIGQUIT: 3,
-    // SIGILL: 4, can not be handled
-    SIGABRT: 6,
-    // SIGFPE: 8, can not be handled
-    // SIGKILL: 9, can not be handled
-    // SIGSEGV: 11, can not be handled
-    SIGPIPE: 13
-    // SIGBUS: 10, can not be handled
-  };
   const request = new Request({
     controller: 'actions',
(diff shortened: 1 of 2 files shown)


lib/api/kuzzle.js
```diff
@@ -226,15 +216,15 @@ function registerErrorHandlers(kuzzle) {
   });
 
   // abnormal termination signals => generate a core dump
-  Object.keys(coreDumpSignals).forEach(signal => {
+  for (const signal of ['SIGQUIT', 'SIGABRT']) {
```
Aschen: nitpicking (optional) : 
I'd rather have the list of signals in a variable, it's clearer I thought

scottinet: I tend to remove variables whenever they are used in only one place, because I find it simpler that way.
I guess this is a matter of personnal preferences.

Aschen: I guess :)

PR #1126, opened 2018-05-31
Node.js v10 compatibility

scottinet:
## What does this PR do ?

Modules written in C and C++ need to be updated to make Kuzzle compatible with node10, due to breaking changes in Joyent's `nan` V8 wrapper (e.g. Koncorde, dumpme, bufferutils, ...)

### Other changes

* Update all other dependencies
* Adapt unit tests to make them compatible with the new major version of `sinon`
* Apparently, it's not possible to override `console` in a module anymore, using rewire. This trick now overrides `console` globally, making tests silent. So all `console` overrides have been removed
* Replaced uses of the `Buffer` constructor with `Buffer.from`: the `Buffer` constructor has been deprecated in node.js v6, and using it in Node.js 10 makes node complain with warning messages

### Boyscout

* Remove unnecessary uses of Bluebird in updated unit tests
* Remove unnecessary uses of sinon's sandboxes


test/services/garbageCollector.test.js
```diff
@@ -64,6 +63,7 @@ describe('Test: GarbageCollector service', () => {
           clock.tick(oneDay);
 
           should(kuzzle.pluginsManager.trigger).called();
+          return null;
```
ballinette: why this ?

scottinet: To prevent bluebird from complaining about the promise chain not returning anything

ballinette: ok

PR #1130, opened 2018-06-07
Improve access logs performances

scottinet:
## What does this PR do?

Recent benchmark tests showed a performance bottleneck, mainly due to:
* winston v2 (logging lilbrary): winston v3 shows non-negligible performances improvements, but introduces a lot of breaking changes due to an overhaul of their API
* our `logAccess` method, called for all requests submitted to Kuzzle, and creating the corresponding log data for winston: a few optimizations have been introduced to it, improving its performances. This function must now be viewed as a critical section of code, and any modification must be done with care

You may use the [kuzzle-benchmark project](https://github.com/kuzzleio/kuzzle-benchmark)

### Other changes

* Force the `net.core.somaxconn` system parameter in the dev docker-compose file


lib/api/core/entrypoints/embedded/index.js
```diff
@@ -281,7 +285,7 @@ class EmbeddedEntryPoint extends EntryPoint {
     let
```
Aschen: nitpicking : usage of string interpolation rather than concatenation
(not here but line 356 and 315)

scottinet: fixed

PR #1133, opened 2018-06-13
Fix document:mGet "includeTrash" option

jenow:
## What does this PR do ?

Fix a bug in mGet where we give the raw request to the elasticSearch client without removing the attribute includeTrash.


ballinette:
I come too late, but we shouldn't have the `includeTrash` option within the body (the option is positioned in `request.input.args` and not `request.input.body`)
So I don't see which bug this PR fixes ?

PR #1148, opened 2018-07-10
set ulimits for elasticsearch

scottinet:
## What does this PR do ?

Sets the maximum number of open files to 65k+ as required by ES (see https://www.elastic.co/guide/en/elasticsearch/reference/current/setting-system-settings.html#limits.conf)


Aschen:
Maybe we can put the `vm.max_map_count=262144` setting in the docker-compose as well ?  
(Currently in `test/travis/travis-bin/run-tests.sh` )

PR #1152, opened 2018-07-11
Add script to build and push docker images

Aschen:
## What does this PR do ?

:warning: This won't work until https://github.com/kuzzleio/kuzzle-containers/pull/39 is merged :warning: 

This PR add a script to build the docker images `plugin-dev` and `kuzzle`.  
Actually `kuzzle` is named `kuzzle-test` to ensure everything is ok before replacing the old `kuzzle` image hosted on Dockerhub by the new one.

The script will be triggered in 2 cases:
 - When a PR is merged on `1.x` : the script will build the images using the `1.x` branch of Kuzzle, then tag them with `develop` and push them to Dockerhub
 - When a release is merged on `master` : the script will build the images using the `master` branch, then tag them with the release tag (inside `TRAVIS_TAG` env variable) and the `latest` tag and push everything to Dockerhub

  - Step 1 : Merge of release on `master` simulation: `TRAVIS_BRANCH=master TRAVIS_TAG=1.4.0 bash deploy.sh`
  - Step 2 : Merge of development branch on `1.x`: `TRAVIS_BRANCH=1.x bash deploy.sh`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,16 @@ group: edge
 
 jdk: oraclejdk8
 
+env:
+  global:
+    - secure: U0+F9Dk03aCycppFX83SZnSvo7syf9NJ2Hns/87wpafjKRXlyUKt3XEsqVUKW+hpI3Me0Av8bjJEBGxchv9xEk+8HcNfzIV8qdfOClym2GF/fY2M2Jjdy4XW7eNsZIYXKo4GHw/+flVRSVR+IG0XP74CuVcAUR/bjMiZomRPJJmfeZ8vaUXbM5P2rTKj4pvtxhR9dFDIzz4XgMBaGvuOVQmmYcms6zlanYkuk2ZDRgZRrnL7BGPkxYjhYvcNLjG4EA9VLqMs/CSdUQMI4BcW3+mI0bGmB25FhjUnMFHLuZctyw4zc42Lzi+Mjept9ynUnKZeTa2W3p5qEYsCr1grT2kScr3NpzacRkRtwEa943zrLFThMkXITVIFACAfelDcDxzAO16Gmqr1Khai6pSAyiQFDJsap/WlZ9JK+mFjsW1hI+jO3HZ30F+TVeaiLp9UFZEg8+DpHueHlwEz3NhPzdg1tWdH8e+itX33Ro0LBjEtX/9MH/ztDycVIs4R+ouSRvrSWH3/vj/653b3xHpJ7/n9OtM8OIpq1xDMpZxBLBtleCD3dUkbDxy8G5qSxPAOk7GZ1GDmiufCYTQXu1rgY7dXj+vCaXyH9f7zwu8WielEoU/0GPUb36usnUGyJkXXgdgGpVdSe73NmGyZgUwp7IQEkOb5507KVy6nEPpIWLI=
+
 services:
-- docker
+ - docker
 
 script:
-- bash test/travis-bin/run-tests.sh
-- bash test/travis-bin/sonar.sh
+ - bash test/travis-bin/run-tests.sh
+ - bash test/travis-bin/sonar.sh
 
 cache:
   directories:
@@ -19,7 +23,7 @@ cache:
 notifications:
   webhooks:
     urls:
-    - https://webhooks.gitter.im/e/8e145155fbaaf37cffea
+     - https://webhooks.gitter.im/e/8e145155fbaaf37cffea
     on_success: change
     on_failure: always
     on_start: never
@@ -41,3 +45,7 @@ deploy:
 
 addons:
   sonarqube: true
+
+after_success:
+  - docker login -u aschen -p $DOCKER_PASSWORD
+  - bash build-docker-images.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ jdk: oraclejdk8
 
 env:
   global:
+    # DockerHub password for aschen (amaret@kuzzle.io) account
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,12 +4,16 @@ group: edge
 
 jdk: oraclejdk8
 
+env:
+  global:
+    - secure: U0+F9Dk03aCycppFX83SZnSvo7syf9NJ2Hns/87wpafjKRXlyUKt3XEsqVUKW+hpI3Me0Av8bjJEBGxchv9xEk+8HcNfzIV8qdfOClym2GF/fY2M2Jjdy4XW7eNsZIYXKo4GHw/+flVRSVR+IG0XP74CuVcAUR/bjMiZomRPJJmfeZ8vaUXbM5P2rTKj4pvtxhR9dFDIzz4XgMBaGvuOVQmmYcms6zlanYkuk2ZDRgZRrnL7BGPkxYjhYvcNLjG4EA9VLqMs/CSdUQMI4BcW3+mI0bGmB25FhjUnMFHLuZctyw4zc42Lzi+Mjept9ynUnKZeTa2W3p5qEYsCr1grT2kScr3NpzacRkRtwEa943zrLFThMkXITVIFACAfelDcDxzAO16Gmqr1Khai6pSAyiQFDJsap/WlZ9JK+mFjsW1hI+jO3HZ30F+TVeaiLp9UFZEg8+DpHueHlwEz3NhPzdg1tWdH8e+itX33Ro0LBjEtX/9MH/ztDycVIs4R+ouSRvrSWH3/vj/653b3xHpJ7/n9OtM8OIpq1xDMpZxBLBtleCD3dUkbDxy8G5qSxPAOk7GZ1GDmiufCYTQXu1rgY7dXj+vCaXyH9f7zwu8WielEoU/0GPUb36usnUGyJkXXgdgGpVdSe73NmGyZgUwp7IQEkOb5507KVy6nEPpIWLI=
```
benoitvidis: nitpicking: could we add a comment to tell what the secret is for (I know this is quite obvious in this case but may not be in the future if we add another one)?

.travis.yml
```diff
@@ -4,12 +4,17 @@ group: edge
 
 jdk: oraclejdk8
 
+env:
+  global:
+    # DockerHub password for aschen (amaret@kuzzle.io) account
```
scottinet: not sure you want to make that kind of information public

.travis.yml
```diff
@@ -41,3 +46,7 @@ deploy:
 
 addons:
   sonarqube: true
+
+after_success:
+  - docker login -u aschen -p $DOCKER_PASSWORD
```
scottinet: I'd prefer a generic account dedicated to the whole team

build-docker-images.sh
```diff
@@ -0,0 +1,83 @@
...
+  mv kuzzle-containers/kuzzle kuzzle-containers/kuzzle-test
+}
+
+if [ -z "$TRAVIS_BRANCH" ]; then
+  echo "TRAVIS_BRANCH not found"
+  exit 1
+fi
+
+
+if [ "$TRAVIS_BRANCH" == "1.x" ]; then
```
scottinet: Does this mean we'll have only 1 `develop` tag? What happens when we start working on v2?

.travis.yml
Aschen: The password is encrypted

build-docker-images.sh
Aschen: When we start working on `2.x`, we will have to rethink our `develop` tag I think.  
Maybe we can simply use `kuzzleio/kuzzle:1.x` and `kuzzleio/kuzzle:2.x` tags.

.travis.yml
Aschen: So I have to create an account. Which email should I use ?

build-docker-images.sh
Aschen: I think we need a workshop on this subject.  
Between Kuzzle, the SDK, the Documentation, we are going to handle a lot of different versions at the same time

build-docker-images.sh
```diff
@@ -0,0 +1,83 @@
...
+  docker_build 'kuzzle-test' $tag current_tag=$tag
+
+  docker_push 'plugin-dev' $tag
+  docker_push 'kuzzle-test' $tag
+
+  rm -rf kuzzle-containers
+elif [ "$TRAVIS_BRANCH" == "master" ]; then
+  git_clone
+
+  tag=$TRAVIS_TAG
```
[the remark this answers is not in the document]
Aschen: Travis will provide a `TRAVIS_TAG` variable if the merge corresponds to a release.  
Since we always merge features on master with a release I assume this variable will be present.

build-docker-images.sh
scottinet: @ballinette > I don't see why we should have a generic tag name for development, for the same reason I'm resolutely against having one on github: because of version collisions, especially during transition phases.
Since we're going to maintain multiple major versions, I think it'd be best to tag our docker images accordingly, for instance `<major version number>-develop` (this comes off the top of my head, I didn't think that through yet). 
Moreover, these tags could be dynamically generated, and having explicit tag names would also lower the probability of human errors.

Aschen:
I fixed the submodule @ballinette

build-docker-images.sh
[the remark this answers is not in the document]
ballinette: ok

build-docker-images.sh
ballinette: Many projects use a `develop` name for current development branch and `x.y` for old LTS versions, and they don't seem to have issue with that.
Many other use only numeric version branches also.

I really don't care on which option we choose, we just need a quick workshop to make a decision and be consistent... but it's not an issue for this PR anyway.

ballinette: But I also like your proposal with `<major version number>-develop` ;)

build-docker-images.sh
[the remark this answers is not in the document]
ballinette: hmmm... we also may have (rare) hotfix PRs that are directly merged to master :thinking:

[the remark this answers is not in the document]
Aschen: There is two build triggered with our hotfix process : 
 - the first one occur when we merge the hotfix branch on master
 - the second one when we tag the commit by releasing on Github

When the first build occur, I didn't build and push the images because we didn't have the tag name by now. I only build and push when the second build occur.

build-docker-images.sh
Aschen: I think we need to formalize all that with the team before applying it in the naming of the tags of our images.  
Otherwise I also like @scottinet proposal `<major version number>-develop `

build-docker-images.sh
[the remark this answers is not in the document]
ballinette: ok fine ;)

PR #1153, opened 2018-07-12
fixes #1146, passport login request missing headers

### LINKED ISSUE #1146: Authentication plugins: missing headers in the request passed to passport
In `login` action, while authenticating through passport, we do not have a `headers` property in the created request object:
https://github.com/kuzzleio/kuzzle/blob/master/lib/api/controllers/authController.js#L85

but some authentication strategies (especially oauth2) need to get data from HTTP headers.

This causes the  following error:
```
"TypeError: Cannot read property 'x-forwarded-proto' of undefined
    at Object.exports.originalURL (..../plugins/available/kuzzle-plugin-auth-passport-oauth/node_modules/passport-oauth2/lib/utils.js:26:27)
    at Strategy.OAuth2Strategy.authenticate (.../plugins/available/kuzzle-plugin-auth-passport-oauth/node_modules/passport-oauth2/lib/strategy.js:141:39)
(...)
```

benoitvidis:
## What does this PR do ?

Fixes #1146 .
After digging in passport code, I finally found the expected type for the Request is an `http.IncomingMessage`.

1. install [kuzzle-plugin-auth-passport-oauth](https://github.com/kuzzleio/kuzzle-plugin-auth-passport-oauth)
2. configure it in kuzzlerc, i.e:
```json
{
  "plugins": {
    "kuzzle-plugin-auth-passport-oauth": {
      "strategies": {
        "github": {
          "credentials": {
            "clientID": "<my client id>",
              "clientSecret": "<my secret>",
              "callbackURL": "http://localhost:8080"
          },
            "persist": [
              "login",
            "avatar_url",
            "name",
            "email"
            ],
            "scope": [
              "email",
            "public_profile"
            ],
            "kuzzleAttributesMapping": {
              "userMail": "email"
            },
            "identifierAttribute": "id"
        }
      },
        "defaultProfiles": [
          "default"
        ]
    }
  }
}
```
3. Test it in your browser:
index.html
```html
<!DOCTYPE html>
<html>
    <head>
      <script src="https://cdn.rawgit.com/kuzzleio/sdk-javascript/master/dist/kuzzle.js"></script>
      <script src="https://cdn.rawgit.com/kuzzleio/kuzzle-sdk-login-oauth-popup/master/index.js"></script>
    </head>
    <body>
        <script>

        </script>
    </body>
</html>
```
in the console:
```js
const k = new Kuzzle('localhost');
k.loginOauthPopup('github')
```

`console.log` anything you want.

NB: no tests, sorry (could not find anything relevant, any input is welcome).

### DIFF
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -27,6 +27,9 @@ const
   User = require('../core/models/security/user'),
   formatProcessing = require('../core/auth/formatProcessing'),
   {
+    IncomingMessage
+  } = require('http'),
+  {
     InternalError: KuzzleInternalError,
     PluginImplementationError,
     KuzzleError
@@ -81,8 +84,23 @@ class AuthController {
     assertArgsHasAttribute(request, 'strategy');
 
     const strategy = request.input.args.strategy;
+    const passportRequest = new IncomingMessage();
 
-    return this.kuzzle.passport.authenticate({body: request.input.body, query: request.input.args, original: request}, strategy)
+    passportRequest.url = `/login?strategy=${strategy}`;
+    passportRequest.method = 'POST';
+    passportRequest.httpVersion = '1.1';
+    passportRequest.httpVersionMajor = 1;
+    passportRequest.httpVersionMinor = 1;
+    passportRequest.body = request.input.body;
+    passportRequest.query = request.input.args;
+    passportRequest.headers = Object.assign({}, request.input.headers);
+    for (const h of Object.keys(passportRequest.headers)) {
+      passportRequest.rawHeaders.push(h);
+      passportRequest.rawHeaders.push(passportRequest.headers[h]);
+    }
+    passportRequest.original = request;
+
+    return this.kuzzle.passport.authenticate(passportRequest, strategy)
       .then(content => {
         // do not trigger the "auth:strategyAutenticated" pipe if the result is not a User object,
         // ie. if we are a intermediate step of a multi-step authentication strategy (example: first redirection call for oAuth strategies)
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -86,6 +86,8 @@ class AuthController {
     const strategy = request.input.args.strategy;
     const passportRequest = new IncomingMessage();
 
+    // even in http, the url and the method are not pushed back to the request object
+    // set some arbitrary values to get a pseudo-valid object.
     passportRequest.url = `/login?strategy=${strategy}`;
     passportRequest.method = 'POST';
     passportRequest.httpVersion = '1.1';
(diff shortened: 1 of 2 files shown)


scottinet:
Does it have to be merged in `master`?

ballinette:
agree with @scottinet : it's a bugfix, but can wait for a future relase (there is a workaround for now)
so it should rather be merged to `1.x`

lib/api/controllers/authController.js
```diff
@@ -26,6 +26,9 @@ const
   _ = require('lodash'),
   User = require('../core/models/security/user'),
   formatProcessing = require('../core/auth/formatProcessing'),
+  {
+    IncomingMessage
+  } = require('http'),
```
Aschen: Nitpicking: 
Why don't use `IncomingMessage = require('http').IncomingMessage` ? I'm surprised that our linter allow that syntax for single import.

alexandrebouthinon: It's the ES6 import (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)

scottinet: Nope. It's a [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)
Anyway, @Aschen is right, this is not useful to use that syntax when there is only 1 item to extract (and it's a nitpicking issue)

benoitvidis: Nodejs tends to promote destructured imports even for single inclusions.
cf https://nodejs.org/dist/latest-v8.x/docs/api/url.html#url_constructor_new_url_input_base for instance of node.js unit tests on github.

lib/api/controllers/authController.js
```diff
@@ -81,8 +84,23 @@ class AuthController {
     assertArgsHasAttribute(request, 'strategy');
 
     const strategy = request.input.args.strategy;
+    const passportRequest = new IncomingMessage();
+
+    passportRequest.url = `/login?strategy=${strategy}`;
+    passportRequest.method = 'POST';
```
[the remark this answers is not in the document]
benoitvidis: I'll check if it can be retrieved from the request for http and keep it as POST (and add some comment about it).
The same applies for the url. The idea was just to create an as-valid-as-possible object here.

benoitvidis:
@scottinet you're right. There is a workaround for the issue and can wait for the next release.

lib/api/controllers/authController.js
Aschen: I prefer this syntax too, even for a single import.  
I thought our linter won't allow this because I remember that I had to fix an import like that recently. Maybe it was before Node8 support

PR #1155, opened 2018-07-17
Inject protocol informations into the RequestContext object

scottinet:
## What does this PR do ?

Fill the updated version of the `RequestContext` object (see [kuzzle-common-objects@3.0.8](https://github.com/kuzzleio/kuzzle-common-objects/releases/tag/3.0.8)) with additional information, either from a direct connection to Kuzzle, or from one established to a Kuzzle Proxy.

Internal backlog task: KZL-49
Fixes: https://github.com/kuzzleio/kuzzle/issues/1106

### Other changes

* Update all references to `RequestContext`, preventing the use of deprecated properties
* Fix direct websocket connections IP address retrieval (`uws` has a different internal structure than `ws`)
* There is no need to manually copy protocol headers into `Request.input.headers` anymore, the 3.0.9 version of the kuzzle-common-objects module takes care of that now. It's that module responsability to handle its own properties
* Remove a bluebird warning (occuring only in dev. mode) generated by the router and the funnel controllers

PR #1158, opened 2018-07-18
Replace ParseError errors with BadRequestError objects

scottinet:
:warning: depends on #1155 :warning: 

## What does this PR do ?

The `ParseError` object is a duplicate of `BadRequestError` and should be considered as deprecated.


benoitvidis:
OK, fine by me.

PR #1159, opened 2018-07-19
Fix: authenticators are not registered when there is no strategies 

scottinet:
## What does this PR do ?

* Kuzzle fails to register `authenticators` when no `strategies` object is defined by a plugin (for instance when that plugin intends to dynamically register strategies at a later time)
* Whenever a plugin tries to use an authenticator without ever declaring an `authenticators` object, Kuzzle fails at startup with a generic (and thus hard to debug) Javascript error. It now properly throws a `PluginImplementationError` giving some insight to plugin developers

This is a hotfix because the 1st fix is expected by one of our integrators, and no workaround is readily available (Kuzzle also forbids empty `strategies` objects)

### DIFF
--- a/lib/api/core/plugins/pluginsManager.js
+++ b/lib/api/core/plugins/pluginsManager.js
@@ -187,8 +187,12 @@ class PluginsManager {
             this._initControllers(plugin);
           }
 
-          if (plugin.object.strategies || plugin.object.authenticators) {
-            this._initAuthentication(plugin);
+          if (plugin.object.authenticators) {
+            this._initAuthenticators(plugin);
+          }
+
+          if (plugin.object.strategies) {
+            this._initStrategies(plugin);
           }
 
           if (plugin.object.hooks) {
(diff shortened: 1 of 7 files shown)


lib/api/core/plugins/pluginsManager.js
```diff
@@ -678,13 +682,25 @@ class PluginsManager {
...
    */
-  _initAuthentication (plugin) {
+  _initStrategies (plugin) {
     const errorPrefix = `[${plugin.name}]:`;
 
     if (!_.isPlainObject(plugin.object.strategies) || _.isEmpty(plugin.object.strategies)) {
       throw new PluginImplementationError(`${errorPrefix} the exposed "strategies" plugin property must be a non-empty object`);
     }
 
+    for (const name of Object.keys(plugin.object.strategies)) {
```
Aschen: Nitpicking:
Why not make use of `Object.entries` ? 
```
for (const [name, strategy] of Object.entries(plugin.object.strategies) {
  this.registerStrategy(plugin.name, name, strategy);
}
```

Issue #1165, opened 2018-07-26
createOrReplaceRole and createOrReplaceProfile throwing TypeError


lethak:
## Expected Behavior
[createOrReplaceRole](https://docs.kuzzle.io/api-documentation/controller-security/create-or-replace-role) and [createOrReplaceProfile](https://docs.kuzzle.io/api-documentation/controller-security/create-or-replace-profile) should be working without throwing a TypeError

## Current Behavior
A TypeError is thrown upon plugin initialization when kuzzle is starting

when using createOrReplaceRole:
```
ERROR  TypeError: Cannot read property 'createOrReplace' of null
    at RoleRepository.persistToDatabase (/var/app/lib/api/core/models/repositories/repository.js:231:
31)
    at role.validateDefinition.then (/var/app/lib/api/core/models/repositories/roleRepository.js:176:
21)
    at tryCatcher (/var/app/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/var/app/node_modules/bluebird/js/release/promise.js:512:31
)
    at Promise._settlePromise (/var/app/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromise0 (/var/app/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (/var/app/node_modules/bluebird/js/release/promise.js:693:18)
    at Async._drainQueue (/var/app/node_modules/bluebird/js/release/async.js:133:16)
    at Async._drainQueues (/var/app/node_modules/bluebird/js/release/async.js:143:10)
    at Immediate.Async.drainQueues (/var/app/node_modules/bluebird/js/release/async.js:17:14)
    at runCallback (timers.js:672:20)
    at tryOnImmediate (timers.js:645:5)
    at processImmediate [as _immediateCallback] (timers.js:617:5)

```

when using createOrReplaceProfile:
```
ERROR  TypeError: Cannot read property 'get' of null
    at RoleRepository.loadOneFromDatabase (/var/app/lib/api/core/models/repositories/repository.js:83
:31)
    at RoleRepository.loadRoles (/var/app/lib/api/core/models/repositories/roleRepository.js:66:31)
    at fromDTO.then.profile (/var/app/lib/api/core/models/repositories/profileRepository.js:268:46)
    at tryCatcher (/var/app/node_modules/bluebird/js/release/util.js:16:23)
[...]

PR #1167, opened 2018-07-27
Initialize repositories before plugins

Aschen:
## What does this PR do ?

This PR change the initialization sequence of Kuzzle.  
Now the repositories are initialized before the plugin so the plugin can use the repositories at initialization.  

Fix: https://github.com/kuzzleio/kuzzle/issues/1165

  - Step 1 : Execute this request in a plugin `init()` function
```
    return this.context.accessors.execute(new this.context.constructors.Request({
        controller: 'security',
        action: 'createOrReplaceRole',
        refresh: 'wait_for',
        _id: 'admin',
        body: {
          "controllers": {
            "*": {
              "actions": {
                "*": true
              }
            }
          }
        }
    }))
    .catch(err => console.warn('ERROR ', err))
```
 
### Other changes

 - Propagate `NODE_LTS` env variable to the proxy

### DIFF
--- a/lib/api/kuzzle.js
+++ b/lib/api/kuzzle.js
@@ -129,6 +129,7 @@ class Kuzzle extends EventEmitter {
       .then(() => this.validation.init())
       .then(() => this.indexCache.init())
       .then(() => this.funnel.init())
+      .then(() => this.repositories.init())
       .then(() => this.pluginsManager.init())
       .then(() => this.pluginsManager.run())
       .then(() => this.pluginsManager.trigger('log:info', 'Services initiated'))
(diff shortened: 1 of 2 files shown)


lib/api/kuzzle.js
```diff
@@ -143,7 +144,7 @@ class Kuzzle extends EventEmitter {
           maxMinTerms: this.config.limits.subscriptionMinterms
         });
 
-        return this.repositories.init();
+        return true;
```
scottinet: Why return a boolean?
You should instead return `this.validation.curateSpecification()`, removing one `.then`  level

PR #1169, opened 2018-07-31
Fix: authentication rejection not forwarded

scottinet:
## What does this PR do?

When an authentication plugin rejects a login attempt while providing a reason, that reason is not forwarded by Kuzzle in the resulting error response.

This prevents plugin developers to forward the rejection reason, and can lead to confusion on the user side.

Attempt to log in using our default `local` strategy plugin, with an invalid login name or password. The result is a generic `UnauthorizedError` error without a message, while a `wrong login name or password` message (provided by the auth. plugin) is expected.

### Other changes

If a plugin's `verify` function throws a non-KuzzleError error, that error is forwarded as is, instead of being cast as a `PluginImplementationError` error.

### DIFF
--- a/lib/api/core/auth/passportWrapper.js
+++ b/lib/api/core/auth/passportWrapper.js
@@ -58,26 +58,34 @@ class PassportWrapper {
       // (Proof: HTTP redirection unit test)
       response.addEndListener(() => resolve(response));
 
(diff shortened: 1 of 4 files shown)


scottinet:
right problem, wrong fix

PR #1178, opened 2018-08-06
Use local multi-stage build for kuzzleio/plugin-dev and kuzzleio/kuzzle

Aschen:
## What does this PR do ?

Before this PR, we were using the Dockerfile in the `kuzzle-containers` repository to build `kuzzleio/plugin-dev` and `kuzzleio/kuzzle` images.  
However, keep the Dockerfile in the same repository as the product is more common.  

This PR update the Dockerfile with a multi-stage build : 
 - `plugin-dev` is used to build an image for plugin developer with Kuzzle source code and the complete toolchain
 - `kuzzle` is used to build a minimal production image for Kuzzle

Theses build are executed by the `build-docker-images.sh` script.  
This script can be triggered in two ways: 
 - when a pull request is merged on `1.x` (or on the futur `2.x` branch), in this case the Docker build happen in the usual Travis build. This will build and push the following images: `kuzzleio/plugin-dev:1.x`, `kuzzleio/kuzzle:1.x` (aliases with `develop` tag will also be pushed for backward compat)
 - when a new release is published (eg: `1.4.2`), in this case a [Github Webhook](https://github.com/kuzzleio/kuzzle/settings/hooks/42671706) will trigger a Travis build via an [AWS Lambda](https://us-west-2.console.aws.amazon.com/lambda/home?region=us-west-2#/functions/github-travis-gateway?tab=graph). This will build and push the following images: `kuzzleio/plugin-dev:1.4.2`, `kuzzleio/kuzzle:1.4.2` (aliases with tag `latest` will also be pushed)

:warning: For now the production image is pushed as `kuzzleio/kuzzle8`, when this PR get merged I will wait at least a release of Kuzzle to be sure that everything is ok with the images and after that I will do another PR to push the production image as `kuzzleio/kuzzle`.

Related PR: [AWS Lambda code](https://bitbucket.kaliop.net/projects/LABS/repos/kuzzle.io-aws/pull-requests/5/overview)

#### Emulate a merge on branch `1.x`
  - Step 1 : Run the script like Travis: `DOCKER_PASSWORD=xxx TRAVIS_BRANCH=1.x bash build-docker-images.sh`
  - Step 2 : Check the last commit: `docker run --rm -it kuzzleio/plugin-dev:1.x git log --pretty=oneline`
  - Step 3 : Check if it match the last commit on `1.x`: `git fetch origin 1.x && git log 1.x --pretty=oneline`

#### Emulate the release `1.4.1`
  - Step 1 : Run the script like Travis: `DOCKER_PASSWORD=xxx RELEASE_TAG=1.4.1 bash build-docker-images.sh`
  - Step 2 : Check the kuzzle version in the images, it should be `1.4.1`: `docker run --rm -it kuzzleio/plugin-dev:latest head package.json`

### Other changes

  - Update readme part about Kuzzle enterprise


Dockerfile
```diff
@@ -1,37 +1,77 @@
...
-  && chmod a+x /usr/local/bin/kuzzle \
-  && chmod a+x /var/app/docker-compose/scripts/docker-entrypoint.sh \
-  \
-  && apt-get clean \
-  && apt-get remove -y \
-    build-essential \
-    g++ \
-    python \
-  && apt-get autoremove -y \
+       curl \
```
Aschen: I'm not sure that we need `curl` in our production image but some of our script (the one that is waiting Elasticsearch to be up for example) need `curl`.

Dockerfile
```diff
@@ -1,37 +1,77 @@
...
+
+RUN  git clone --depth 1 --branch $kuzzle_tag https://github.com/kuzzleio/kuzzle /var/app \
+  && git clone --depth 1 https://github.com/kuzzleio/kuzzle-plugin-auth-passport-local /var/app/plugins/available/kuzzle-plugin-auth-passport-local \
+  && git clone --depth 1 https://github.com/kuzzleio/kuzzle-plugin-logger /var/app/plugins/available/kuzzle-plugin-logger
 
 WORKDIR /var/app
 
+RUN  npm install --unsafe-perm \
+  && npm rebuild all --unsafe-perm \
+  && for plugin in plugins/enabled/*; do cd "$plugin"; npm install --unsafe; cd /var/app; done
```
scottinet: `--unsafe` => `--unsafe-perm`

Dockerfile
scottinet: Since we already install `wget`, maybe we could use it to remove curl... in a later PR.

README.md
```diff
@@ -94,12 +94,12 @@ Check our [contributing documentation](./CONTRIBUTING.md) to know about our codi
...
-## Kuzzle Enterprise
+## Kuzzle in production
 
-Kuzzle Enterprise is production-proof, and provides all the business-critical features your need for your business, as
-the scalability, the high-availability (multi-nodes), probes for BI, diagnostic tools & professional services,
+Kuzzle is production-proof, and provides all the business-critical features your need for your business, as
+the scalability, the high-availability (multi-nodes), probes for BI & diagnostic tools.  
 
-[Compare editions to learn more](https://kuzzle.io/pricing/)
+See our [Professional services](https://kuzzle.io/pricing/).
```
scottinet: Proposal: `Check out our [support plans](<link>)`

Dockerfile
Aschen: `wget` is not present on the production image. We don't install any packet except curl

Dockerfile
```diff
@@ -1,37 +1,80 @@
...
+  && npm install -g \
+    pm2 \
+  && echo "" > /opt/node-v$NODE_VERSION-linux-x64/lib/node_modules/pm2/lib/keymetrics \
+  && rm -rf /var/lib/apt/lists/* \
+  && echo "alias ll=\"ls -lahF --color\"" >> ~/.bashrc
+
+ARG kuzzle_tag
+
+RUN  git clone --depth 1 --branch $kuzzle_tag https://github.com/kuzzleio/kuzzle /var/app \
+  && git clone --depth 1 https://github.com/kuzzleio/kuzzle-plugin-auth-passport-local /var/app/plugins/available/kuzzle-plugin-auth-passport-local \
```
ballinette: what if master branch of kuzzle-plugin-auth-passport-local is not compatible with `$kuzzle_tag` version of kuzzle ?

Aschen: Well, we have a problem then I think.  
I didn't figure out a simple solution.. I just assume these plugin will be compatible with Kuzzle when the image will be built.  
When the `2.x` come out we will have to think to a compatibility matrix or something

scottinet: @ballinette > the way I see it, official plugins must be compatible with all available versions of Kuzzle at the moment of the plugin's release.
This means that we'll have to have fallbacks depending on the target Kuzzle version.

I think this is way easier that way, instead of maintaining multiple plugin major versions + compatibility matrixes everywhere.

ballinette: But there is no reason to clone these plugins separately from Kuzzle core as they are already included as submodule (so `git clone --recursive --depth 1 --branch $kuzzle_tag https://github.com/kuzzleio/kuzzle /var/app` should be sufficient)

... and for the next major release, my advice is that:
* plugin passport local should not be forced anymore (we can want to not authenticate with login/password, but with other strategies)
* plugin logger should not be a plugin anymore and integrated within the core (winston is configurable enough to set up the logger as we want, without need to create another plugin)

Aschen: We talked about the submodules and we weren't sure we'd keep them, I think.
At least with the git clone we decide exactly what we want in the image

PR #1179, opened 2018-08-10
KZL-174 Expose to plugins a SDK instance directly mapped to the funnel

ballinette:
## What does this PR do ?

Apply https://jira.kaliop.net/browse/KZL-176 and https://jira.kaliop.net/browse/KZL-177

This creates a custom light network wrapper for SDK that is directly mapped to the funnel.
And exposes a SDK instance with this wrapper to the plugins.

Create a plugin and add hooks or custom routes which calls `context.accessors.sdk.<somectrl>.<someaction>`

### Other changes

* `funnel.executePluginRequest` now returns a promise (no need to use callbacks anymore since we disabled the overload protection for this method)
* deprecate `context.accessors.execute` method (will be removed in V2)

### Boyscout

fix some jsdocs

### DIFF
--- a/test/api/controllers/funnelController/executePluginRequest.test.js
+++ b/test/api/controllers/funnelController/executePluginRequest.test.js
@@ -25,74 +25,32 @@ describe('funnelController.executePluginRequest', () => {
     funnel.handleErrorDump = sinon.stub();
   });
 
-  it('should fail if an unknown controller is invoked', done => {
+  it('should fail if an unknown controller is invoked', () => {
     const rq = new Request({controller: 'foo', action: 'bar'});
 
-    const ret = funnel.executePluginRequest(rq, (err, res) => {
-      try {
-        should(res).be.undefined();
+    return funnel.executePluginRequest(rq)
+      .then(() => Promise.reject(new Error('Should not resolves')))
+      .catch(err => {
         should(err).be.instanceOf(BadRequestError);
         should(err.message).be.eql('Unknown controller foo');
         should(kuzzle.pluginsManager.trigger).not.be.called();
-        done();
-      }
-      catch (e) {
-        done(e);
-      }
-    });
-
-    should(ret).be.eql(1);
+      });
   });
 
-  it('should execute the request', done => {
-    const rq = new Request({controller: 'testme', action: 'action'});
-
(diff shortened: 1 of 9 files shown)


Aschen:
What about the realtime capabilities of the SDK ?

ballinette:
@Aschen I don't see how using realtime  is relevant inside a plugin (which actions are scoped inside a client's request)
If you need to subscribe to filters and keep connected to kuzzle, you do not need a plugin, you need an external client.

ballinette:
put in WIP:
need to merge https://github.com/kuzzleio/sdk-javascript/pull/305, and then resolve the whole response instead of only `response.result`

scottinet:
I'm not at ease with that one. I mean, I obviously agree with the idea itself, but I believe it's too soon: we're adding an experimental and as-of-now undocumented SDK to plugin interfaces, so this is pretty much unusable. :confused:

test/api/controllers/funnelController/executePluginRequest.test.js
```diff
@@ -25,71 +25,29 @@ describe('funnelController.executePluginRequest', () => {
...
 
-  it('should fail if an unknown controller is invoked', done => {
+  it('should fail if an unknown controller is invoked', () => {
     const rq = new Request({controller: 'foo', action: 'bar'});
 
-    const ret = funnel.executePluginRequest(rq, (err, res) => {
-      try {
-        should(res).be.undefined();
+    return funnel.executePluginRequest(rq)
+      .then(() => Promise.reject(new Error('Should not resolves')))
```
scottinet: s/resolves/resolve/

ballinette:
@scottinet I agree that it is quite soon for this feature, and of course it can wait for the SDK 6.0 release

However: 
* the sdk refacto is almost done and is just waiting for the documentation to be released
* it could be fine to have this feature before the V2 to let early developers to test it.

A suggestion could be: release the feature, but tag it as "experimental" for now. (and of course remove back the "deprecated" tag on `context.accessors.execute`)

This way we will be able to use it and test it for demos, POCs, etc, and some "pilote" projects could also try it and give their feedback.

ballinette:
put in "WIP" as we will wait for a beta release of SDK 6 to merge this PR

PR #1180, opened 2018-08-10
KZL 293 Allow plugin developer to return a Request object from controller route

Aschen:
## What does this PR do ?

With the `execute` method, plugin developer often write the same code when returning in controller actions : `return execute(some_request).then(result => result.result)`

But if the status if different from 200 (206 for example), they have to open the request returned by `execute` and extract the result, the status, the errors...

This PR add the ability for plugin developer to return directly a Request object in a controller action.  

This comes with few restrictions:
 - the returned Request object must be the one received in parameter
 - the request status must have been set

  - Step 1 : Create a plugin with the following code: https://gist.github.com/Aschen/fd2baf0faa7c6e65dfab394751cb296f
  - Step 2 : Curl the route: `curl localhost:7512/_plugin/ctoi-le-plugin/test?pretty`


Aschen:
With the `execute` method, plugin developer often write the same code when returning in controller actions : `return execute(some_request).then(result => result.result)`

But if the status if different from 200 (206 for example), they have to open the request returned by `execute` and extract the result, the status, the errors...

This PR allow them to return directly the request returned by execute

Aschen:
Partial errors will be refactored in Kuzzle v2.
`execute` will be removed in Kuzzle v2.

PR #1181, opened 2018-08-10
Plugins/Protocols manifest.json overhaul

scottinet:
## What does this PR do?

Re-architecture the way plugins `manifest.json` files are handled:

* The `package.json` file was required for plugins (but not for protocols), but Kuzzle actually doesn't care if a plugin is a valid NPM module or not
* Kuzzle now displays (lengthy) warning messages during startup to encourage plugin developers to expose a manifest file: manifests will become **required** in the next major release of Kuzzle
* Plugins `package.json` files and Protocols `protocol` property are now **deprecated fallbacks** to manifests files, and only used by Kuzzle if necessary
* Protocols now also need to expose a manifest file
* Manifests files are now thoroughly checked to make sure that their format is correct

List of currently supported manifest files properties:

* `name` (**required**): Plugin/Protocol identifier. Names can only contain lowercase letters, numbers, hyphens and underscores. If the name is empty or absent, Kuzzle falls back to the `package.json` file (plugins) or the `protocol` property (protocols)
* `kuzzleVersion`: must be a non-empty string in semver format, limiting the range of Kuzzle versions supported by this Plugin/Protocol. If not set, a warning is displayed on the console, and Kuzzle assumes that the Plugin/Protocol is compatible only with Kuzzle v1.x
* `privileged`: must be a boolean (no change made)

(internal backlog task: KZL-430)

### Other changes

* Unit tests improvements:
  * Mocha's reporter has been set to `dot` instead of `spec`. This makes easier to spot uncaught exceptions or rejected promises, and thus makes testing safer
  * All console prints during unit-tests have been mocked to keep the tests results clean and easy to follow

### Boyscout

* Remove a call to `fs.existsSync`, which is deprecated
* Kuzzle's version in the `kuzzle.config` object is now handled by the configuration modules, instead of being injected by Kuzzle: better separation of concerns + makes some unit tests easier by allowing the `KuzzleMock` object to expose the current Kuzzle version at no cost
* Fix a promise that was rejected with a non-error object in the auth controller unit tests


lib/api/core/plugins/pluginsManager.js
```diff
@@ -714,130 +708,87 @@ class PluginsManager {
...
+    debug('loading plugins: %a', plugins);
 
-    try {
-      plugin.manifest = Object.assign(plugin.manifest, require(path.resolve(plugin.path, 'manifest.json')));
-    }
-    catch(e) {
-      // do nothing, simply use default manifest
-    }
+    for (let pluginPath of plugins) {
+      pluginPath = path.resolve(this.pluginsDir, pluginPath);
```
kuzzle: CRITICAL Remove this assignment of "pluginPath". [rule](http://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS2310)
MINOR Introduce a new variable instead of reusing the foreach variable "pluginPath". [rule](http://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS1226)

lib/api/core/plugins/pluginsManager.js
```diff
@@ -699,7 +693,7 @@ class PluginsManager {
    * @throws {PluginImplementationError} If strategies registration fails
    */
   _initAuthenticators (plugin) {
-    const errorPrefix = `[${plugin.name}]:`;
+    const errorPrefix = `[${plugin.manifest.name}]:`;
 
     // @todo the _.isNil test need to be removed as soon as the
```
kuzzle: INFO Complete the task associated to this TODO comment. [rule](http://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS1135)

kuzzle:
SonarQube analysis reported 3 issues
* ![CRITICAL][CRITICAL] 1 critical
* ![MINOR][MINOR] 1 minor
* ![INFO][INFO] 1 info

[INFO]: https://sonarsource.github.io/sonar-github/severity-info.png 'Severity: INFO'

lib/api/core/plugins/pluginsManager.js
```diff
@@ -714,130 +708,87 @@ class PluginsManager {
...
-    debug('[%s] loading plugin from directory "%s"', plugin.name, plugin.path);
+    debug('loading plugins: %a', plugins);
 
-    try {
-      plugin.manifest = Object.assign(plugin.manifest, require(path.resolve(plugin.path, 'manifest.json')));
-    }
-    catch(e) {
-      // do nothing, simply use default manifest
-    }
+    for (let relativePluginPath of plugins) {
```
kuzzle: CRITICAL Make "relativePluginPath" "const". [rule](http://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS3353)

kuzzle:
* ![CRITICAL][CRITICAL] 1 critical
* ![INFO][INFO] 1 info

[INFO]: https://sonarsource.github.io/sonar-github/severity-info.png 'Severity: INFO'

lib/api/core/plugins/pluginsManager.js
```diff
@@ -714,130 +708,87 @@ class PluginsManager {
...
+    debug('loading plugins: %a', plugins);
 
-    try {
-      plugin.manifest = Object.assign(plugin.manifest, require(path.resolve(plugin.path, 'manifest.json')));
-    }
-    catch(e) {
-      // do nothing, simply use default manifest
-    }
+    for (const relativePluginPath of plugins) {
+      const pluginPath = path.resolve(this.pluginsDir, relativePluginPath);
```
kuzzle: CRITICAL Remove this assignment of "pluginPath". [rule](http://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS2310)
MINOR Introduce a new variable instead of reusing the foreach variable "pluginPath". [rule](http://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS1226)

kuzzle:
SonarQube analysis reported 3 issues
* ![CRITICAL][CRITICAL] 1 critical
* ![MINOR][MINOR] 1 minor
* ![INFO][INFO] 1 info

[INFO]: https://sonarsource.github.io/sonar-github/severity-info.png 'Severity: INFO'

kuzzle:
* ![INFO][INFO] 1 info

[INFO]: https://sonarsource.github.io/sonar-github/severity-info.png 'Severity: INFO'

lib/api/core/plugins/manifest.js
```diff
@@ -0,0 +1,108 @@
...
+      if (fallback !== null) {
+        console.warn(`[${this.path}] Plugins without a manifest.json file are deprecated.`);
+        fallback(this);
+        return this;
+      }
+
+      throw new PluginImplementationError(e);
+    }
+
+    if (!_.isNil(raw.privileged)) {
```
benoitvidis: Protocols are per definition privileged (and are not plugins).
If we really want to go with a manifest class, I'd use a separate implementation, possibly implementing a common abstract class for plugins vs protocols.

lib/api/core/entrypoints/embedded/index.js
```diff
@@ -158,22 +164,30 @@ class EmbeddedEntryPoint extends EntryPoint {
...
-        .filter(d => fs.statSync(d).isDirectory());
+    dirs = dirs.map(d => path.join(dir, d))
+      .filter(d => fs.statSync(d).isDirectory());
 
-    for (const protoDir of dirs) {
+    return Bluebird.map(dirs, protoDir => {
       const
         protocol = new (require(protoDir))(),
-        name = protocol.protocol || path.basename(protoDir);
+        manifest = new Manifest(this.kuzzle, protoDir, m => {
```
benoitvidis: Why implement a relatively complex  fallback function mechanism when this is the only place it will ever be used?

lib/api/core/plugins/manifest.js
```diff
@@ -0,0 +1,108 @@
...
+      if (typeof raw.name !== 'string' || raw.name.length === 0) {
+        throw new PluginImplementationError(`[${this.path}/manifest.json] Invalid "name" property: expected a non-empty string`);
+      }
+
+      // Plugin names are used to create a database index, and those cannot
+      // contain upercased caracters. We should also limit the
+      // characters set to the standard \w regexp, minus uppercased
+      // characters, and with the widely used '-' char
+      if (!/^[a-z0-9_-]+$/.test(raw.name)) {
+        throw new PluginImplementationError(`[${this.path}/manifest.json] Invalid plugin name. The name must be comprised only of lowercased letters, numbers, hyphens and underscores`);
```
benoitvidis: This is quite true for plugins, but not for protocols.

lib/api/core/plugins/manifest.js
```diff
@@ -0,0 +1,108 @@
...
+      this.privileged = raw.privileged;
+    }
+
+    if (!_.isNil(raw.kuzzleVersion)) {
+      if (typeof raw.kuzzleVersion !== 'string' || raw.kuzzleVersion.length === 0) {
+        throw new PluginImplementationError(`[${this.path}/manifest.json] Invalid "kuzzleVersion" property: expected a non-empty string`);
+      }
+      this.kuzzleVersion = raw.kuzzleVersion;
+    } else {
+      console.warn(`[${pluginPath}/manifest.json] No "kuzzleVersion" property found: assuming the target is Kuzzle v1`);
```
benoitvidis: Do we really need so many type checking (and thus lodash) to validate a string that will eventually be sent to compareVersion?

lib/api/core/plugins/manifest.js
scottinet: * Disagree: why are protocols "per definition" privileged? It seems to me that no one should need to access the `kuzzle` internal object. In fact, I believe in that so strongly that I want to propose to remove "privileged mode" altogether from Kuzzle from the next major release. If more accessors are needed because we didn't anticipate some use case, then we shall add these specifically
* Agree: protocols are not plugins. I will add an abstract class for manifests. This will resolve at least one of your comments below.

lib/api/core/entrypoints/embedded/index.js
scottinet: I had a much simpler way to do it, directly in the manifest class, but protocols need a fallback too, but a different one.
I will simplify that with a different implementation of an abstract Manifest class in the coming batch of fixes.

lib/api/core/plugins/manifest.js
scottinet: There are 2 type checks:

* We need to check against null/undefined in order to display a warning and use a default semver version range, preventing the introduction of a breaking change. I indeed use lodash for that because I find `_.isNul(raw.kuzzleVersion)` easier to read than `raw.kuzzleVersion !== undefined && raw.kuzzleVersion !== null`, at no extra cost
* I then check against other types because semver is inconsistent when a non-string value is provided: it may return `false`, thus triggering a (rather imprecise) `version mismatch` error from our side, or it may throw an `invalid version` error. Handling these two cases seemed simpler to me with a simple test before using semver

scottinet: semver inconsistencies showcases:

```
false
> semver.satisfies(false, '1.2.3')
false
> semver.satisfies(true, '1.2.3')
TypeError: Invalid Version: true
    at new SemVer (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:291:11)
    at compare (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:578:10)
    at eq (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:617:10)
    at cmp (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:649:41)
    at Comparator.test (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:716:10)
    at testSet (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:1123:17)
    at Range.test (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:1115:9)
    at Function.satisfies (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:1161:16)
> semver.satisfies(12, '1.2.3')
TypeError: Invalid Version: 12
    at new SemVer (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:291:11)
    at compare (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:578:10)
    at eq (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:617:10)
    at cmp (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:649:41)
    at Comparator.test (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:716:10)
    at testSet (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:1123:17)
    at Range.test (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:1115:9)
    at Function.satisfies (/home/scottinet/git/kuzzle/node_modules/semver/semver.js:1161:16)
```

scottinet:
About the quantity of code: this is mainly due because we already handled manifest files, but:
* this was hard-coded in the (very) lengthy `loadPlugins` function
* it was only made for plugins, not for protocols

Also, I found that a lot of manifest values weren't properly tested. All in all, I agree: a lot of code for a simple feature, but that's mainly _old_ code that was moved around :-)

lib/api/core/plugins/manifest.js
scottinet: Hm. Dang.

```
semver.satisfies('foosdflkj fsjfd jg1.2', '1.2.3')
TypeError: Invalid Version: foosdflkj fsjfd jg1.2
```

I will have to rethink that one a bit.

lib/api/core/plugins/manifest.js
benoitvidis: I don't know any library that tries to hide its internal to developers as strong as we do.
My position is:
1. Why would we block developers as long as it is clear undocumented features are not supported?
2. Javascript is a nice language that in most cases allows developers to achieve what they want eventually
3. Why spend some time & effort in "protecting" our objects?

About the "why": protocols are given an entrypoint instance. Entrypoints rely on kuzzle instance. So protocols can access it (but this is not supported).
The concept of "privileged" is actually meaningless for protocols.

scottinet: @benoitvidis Any library in OOP languages will have protected/private properties, allowing to expose a clean and simple API to its users, while keeping implementation details "hidden" to those who explicitly inherit those objects. Usually, unless their purpose is to be extended, these implementation details are poorly documented, oftentimes on purpose: you do not want to provide support to anyone who want to extend an ever-evolving object.

And we come to the crux of the problem you state: there is no such mechanism in Javascript, everything is public. This is neither good or bad, it's simply different. But there is the problem of how to define a clear line, that is easy to understand for our users, between what is stable for an entire major version, and what one should avoid to use, unless at his own risks.

This is why we provide dedicated, stable and well-documented interfaces for plugins and protocols developers. 
About those, our (my :disappointed: ) first approach was to protect everything, and we clearly overdid it. We stepped back a bit from that position, the only well-protected objects remaining are the Request ones, because securing them allow us to spare a lot of redundant tests to make sure those are correct. And we did that only because you recently found a way to secure them without diminshing their usability: as discussed together, I was ready to un-protect them, since as you put it, in Javascript it's the users responsibility to not break what a library provides.

Do we have to over-protect everything? It's been a while that I think that we went over the top, so my position is a lot more lax now. But do we need to expose internal objects without purpose? I think not.

So, where does that leave us with privileged mode?
As I said, I want to remove this entirely, I see no purpose in int.
And if someone finds a `kuzzle` object by inspecting the innerworks of the interfaces we provide, then so be it. It's their resposibility, and we should not spend time going againt Javascript itself.

scottinet:
@benoitvidis > I implemented all your requested changes, except for the "privileged mode as a default for protocols" remark. After thinking about it, I think we are (somewhat) agreeing that this mode has no sense in itself for protocols, since they have access to the kuzzle object anyway. So the question remains: should I expose a kuzzle object on purpose or not (I think we shouldn't, and that this "mode" has to go)

scottinet:
After sleeping on it, it occured to me that the privileged mode can only be activated by plugins anyway, the property does nothing to protocols. So I moved that part to the plugin-specific manifest class, the property is now completely ignored by protocols.
This should put our argument to rest.

PR #1182, opened 2018-08-10
KZL 290 - Add fixtures for securities

### LINKED ISSUE #1184: Missing additionnal informations in partial error with mDeleteUsers
## Expected Behavior

When I delete several users with `mDeleteUser` and one of the user does not exists, I should have a partial error with the missing user id

## Current Behavior

I have a partial error without the missing user id

## Possible Solution

When a user is loaded, if he does not exists in the database the method resolves `null` instead of reject with a NotFoundError.

## Steps to Reproduce

```js
const Kuzzle = require('kuzzle-sdk');

const kuzzle = new Kuzzle('localhost', (err, resp) => {

  const args = {
    controller: 'security',
    action: 'mDeleteUsers'
  };

  const query = {
    body: {
      ids: ['hey', 'hi']
    }
  };

  kuzzle
    .queryPromise(args, query)
    .then(response => console.log(response))
    .catch(error => console.log(error))
    .finally(() => kuzzle.disconnect());
});
```

## Context (Environment)

Kuzzle 1.4.3

Aschen:
## What does this PR do ?

This PR allow to load fixtures for users, profiles and roles at Kuzzle startup.  
The roles and profiles are loaded with `createOrReplace` and the users are first deleted with `mDeleteUsers` and then created with `createUser`.  

Documentation PR: https://github.com/kuzzleio/documentation/pull/521

The securities fixtures must have the following format: 

```json
{
  "roles": {
    "driver": {
      "controllers": {
        "auth": {
          "actions": {
            "checkToken": true,
            "getCurrentUser": true,
            "getMyRights": true,
            "login": true
          }
        },
        "document": {
          "actions": {
            "*": true
          }
        }
      }
    }
  },
  "profiles": {
    "driver": {
      "policies": [
        {
          "roleId": "driver",
          "restrictedTo": [
            { "index": "nyc-open-data", "collections": ["yellow-taxi"] }
          ]
        }
      ]
    }
  },
  "users": {
    "bcalhoun": {
      "content": {
        "profileIds": ["driver"]
      },
      "credentials": {
        "local": {
          "username": "bcalhoun",
          "password": "blueshift"
        }
      }
    }
  }
}
```

  - Step 1 : Start only `elasticsearch` and `redis` with docker-compose (don't forget to expose the ports 9200 and 6379 respectively)
  - Step 2 : Run Kuzzle from the commande line: `./bin/kuzzle start --fixtures test/mocks/securities.json`
  - Step 3 : Check the roles, profiles and users presence in the admin console

### Other changes

  - Move mappings and fixtures loading to Janitor
  - Fix node inspector in docker-compose
  - Fix #1184 : Refactor a little the method `load` in the repositories to reject with an explicit `NotFoundError`. Some methods (`persistUser` generator from SecurityController and `load` from PluginRepository) needed to return `null` instead of rejecting with a 404 so I added an optional `returnNull` options to the `loadOneFromDatabase` method.

### DIFF
--- a/lib/api/core/janitor.js
+++ b/lib/api/core/janitor.js
@@ -58,10 +58,10 @@ class Janitor {
   loadSecurities (filePath) {
     return this._loadJson(filePath)
       .then(securities => {
-        return this._createRoles(securities.roles)
-          .then(() => this._createProfiles(securities.profiles))
+        return this._createSecurity('createOrReplaceRole', securities.roles)
+          .then(() => this._createSecurity('createOrReplaceProfile', securities.profiles))
           .then(() => this._deleteUsers(securities.users))
-          .then(() => this._createUsers(securities.users));
+          .then(() => this._createSecurity('createUser', securities.users));
       });
   }
 
@@ -490,54 +490,18 @@ class Janitor {
     });
   }
 
-  _createRoles (roles) {
-    if (! roles) {
+  _createSecurity (action, objects) {
+    if (! objects) {
       return Bluebird.resolve();
     }
 
-    return Bluebird.map(Object.keys(roles), roleId => {
+    return Bluebird.map(Object.keys(objects), id => {
       const request = new Request({
         controller: 'security',
-        action: 'createOrReplaceRole',
+        action,
         refresh: 'wait_for',
-        _id: roleId,
-        body: roles[roleId]
-      });
-
-      return this.kuzzle.funnel.processRequest(request);
-    });
-  }
-
-  _createProfiles (profiles) {
-    if (! profiles) {
-      return Bluebird.resolve();
-    }
-
-    return Bluebird.map(Object.keys(profiles), profileId => {
-      const request = new Request({
-        controller: 'security',
-        action: 'createOrReplaceProfile',
-        refresh: 'wait_for',
-        _id: profileId,
-        body: profiles[profileId]
-      });
-
-      return this.kuzzle.funnel.processRequest(request);
-    });
-  }
-
-  _createUsers (users) {
-    if (! users) {
-      return Bluebird.resolve();
-    }
-
-    return Bluebird.map(Object.keys(users), userId => {
-      const request = new Request({
-        controller: 'security',
-        action: 'createUser',
-        refresh: 'wait_for',
-        _id: userId,
-        body: users[userId]
+        _id: id,
+        body: objects[id]
       });
 
       return this.kuzzle.funnel.processRequest(request);
--- a/lib/api/core/janitor.js
+++ b/lib/api/core/janitor.js
@@ -495,13 +495,13 @@ class Janitor {
       return Bluebird.resolve();
     }
 
-    return Bluebird.map(Object.keys(objects), id => {
+    return Bluebird.map(Object.keys(objects), _id => {
       const request = new Request({
-        controller: 'security',
         action,
+        _id,
+        controller: 'security',
         refresh: 'wait_for',
-        _id: id,
-        body: objects[id]
+        body: objects[_id]
       });
 
       return this.kuzzle.funnel.processRequest(request);
--- a/lib/api/core/janitor.js
+++ b/lib/api/core/janitor.js
@@ -49,6 +49,122 @@ class Janitor {
   }
 
   /**
+   * Load database fixtures into Kuzzle
+   *
+   * @param {String} filePath
+   * @returns {Promise}
+   */
+  loadFixtures (filePath) {
+    return new Bluebird((resolve, reject) => {
+      fs.readFile(filePath, 'utf8', (error, data) => {
+        if (error) {
+          return reject(error);
+        }
+
+        try {
+          const fixtures = JSON.parse(data);
+
+          resolve(fixtures);
+        }
+        catch (e) {
+          reject(e);
+        }
+      });
+    })
+    .then(fixtures => {
+      const requests = [];
+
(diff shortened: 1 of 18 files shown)


Aschen:
I'm not quite sure for the label, `enhancement` or `new-feature` ? @ballinette

ballinette:
We can consider that fixtures loading is a Kuzzle's feature.
So as we add a fixture, we can consider it as a `new-feature`

lib/api/core/janitor.js
```diff
@@ -362,6 +465,109 @@ class Janitor {
     process.exit(0);
   }
 
+  _loadJson (filePath) {
```
scottinet: I'm puzzled: since we're loading JSON files, why not just requiring them?

lib/api/core/janitor.js
```diff
@@ -362,6 +465,109 @@ class Janitor {
...
+      });
+    });
+  }
+
+  _createRoles (roles) {
+    if (! roles) {
+      return Bluebird.resolve();
+    }
+
+    const promises = Object.keys(roles).map(roleId => {
```
scottinet: (nitpicking) This can be simplified using `Bluebird.map`:

```js
return Bluebird.map(Object.keys(roles), roleId => {
  const request = new Request({
    controller: 'security',
    action: 'createOrReplaceRole',
    refresh: 'wait_for',
    _id: roleId,
    body: roles[roleId]
  });
  return this.kuzzle.funnel.processRequest(request);
});
```

If you approve of this proposal, then it should also be applied to other _createXxx functions

lib/api/core/janitor.js
```diff
@@ -362,6 +465,109 @@ class Janitor {
...
+    return Bluebird.all(promises);
+  }
+
+  _deleteUsers (users) {
+    if (! users) {
+      return Bluebird.resolve();
+    }
+
+    const
+      ids = Object.keys(users).map(userId => userId),
```
scottinet: Tautological use of Array.map, the following is sufficient: `ids = Object.keys(users)`

lib/api/core/janitor.js
```diff
@@ -362,6 +465,109 @@ class Janitor {
...
+        controller: 'security',
+        action: 'mDeleteUsers',
+        refresh: 'wait_for',
+        body: { ids }
+      });
+
+    return this.kuzzle.funnel.processRequest(request)
+      .catch(error => {
+        if (error.status === 206) {
+          return Bluebird.resolves();
```
scottinet: `Bluebird.resolves` is not a function. This still works because you end up returning `undefined` instead of a new empty promise resolution, which leads to the same result. 
Replace with `return null;`

lib/api/core/janitor.js
```diff
@@ -362,6 +465,109 @@ class Janitor {
...
+        }
+
+        return Bluebird.reject(error);
```
scottinet: (nitpicking) Proposal: I find the following easier to read `throw error;`

lib/api/core/models/repositories/repository.js
```diff
@@ -97,15 +103,19 @@ class Repository {
...
+        if (error.status !== 404) {
+          return Bluebird.reject(error);
+        }
+
+        if (! options.returnNull) {
+          return Bluebird.reject(new NotFoundError(`Unable to find ${this._objectName()} with id '${id}'`));
         }
 
-        return Bluebird.reject(error);
+        return Bluebird.resolve(null);
```
scottinet: I strongly disagree with this, as this introduces an unneeded inconsistency in the repositories API.
I would prefer that you remove that `returnNull` option and, in the persistXxx functions, that you try...catch the result of the yielded call and act appropriately if an exception status equal to 404 is received.

lib/api/core/models/repositories/repository.js
```diff
@@ -97,15 +103,19 @@ class Repository {
...
         }
+
         return null;
       })
       .catch(error => {
-        if (error.status === 404) {
-          // no content found, we return null without failing
-          return Bluebird.resolve(null);
+        if (error.status !== 404) {
+          return Bluebird.reject(error);
```
scottinet: (nitpicking) You can use `throw` instead of `return Bluebird.reject`
That remark can also be applied to the other rejection just below

lib/api/core/models/repositories/repository.js
Aschen: Much clever solution. I wasn't comfortable with the `persistUser` function, I never use this kind of javascript black magic before. I will find out :)

lib/api/core/janitor.js
scottinet: Why not. This may warrant an explanatory comment, though

lib/api/core/janitor.js
```diff
@@ -362,6 +465,73 @@ class Janitor {
...
+    });
+  }
+
+  _createSecurity (action, objects) {
+    if (! objects) {
+      return Bluebird.resolve();
+    }
+
+    return Bluebird.map(Object.keys(objects), id => {
+      const request = new Request({
```
kuzzle: MINOR Group all shorthand properties at either the beginning or end of this object declaration. [rule](http://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS3499)

kuzzle:
* ![MINOR][MINOR] 1 minor

ballinette:
@scottinet are the latest changes OK for you ?

PR #1183, opened 2018-08-13
Kzl 443 arm

benoitvidis:
**:warning: Depends on https://github.com/kuzzleio/kuzzle-containers/pull/42 + docker hub automation**

## What does this PR do ?

This PR attemps to make Kuzzle arm-friendly.
For the time being, the only uncompatiblity identified is with highwayhash, which targets x85 and amd64 only.

This PR replaces this module in Kuzzle and its dependency with murmurhash3, which seems to be an acceptable candidate for replacement and compiles on arm (no collision seen on several 16M batches and performance is slightly above highwayhash).

You can use regular kuzzle tests.


alexandrebouthinon:
Is this really need to be merge into master ?

benoitvidis:
Travis test added.

benoitvidis:
For some reason, once pushed to docker hub, the arm images do not work on travis.
Needs some further investigation.

alexandrebouthinon:
Need a Travis job splitting to reduce build duration. I'm on it :smile:

PR #1201, opened 2018-09-26
fixes #1200 - no access log with http:options

### LINKED ISSUE #1200: No connection retrieved for connection id: null for http OPTION queries
When querying Kuzzle in http, using the `OPTIONS` verb, we get the response but we see in the access logs:

`2018-09-26T07:14:52+00:00 [LOG:WARN] [access log] No connection retrieved for connection id: null\nMost likely, the connection was closed before the response was received.`

benoitvidis:
## What does this PR do ?

This PR fixes #1200 - no access log with http options verb

```
curl -XOPTIONS kuzzle:7512/_now
```
and look at the access log


jenow:
@benoitvidis I am not merging this PR because a test is failing, can you take a look?

PR #1211, opened 2018-10-20
Add missing coma in .kuzzlerc.sample

stafyniaksacha:
only boyscouting (copy/pasta issue)


Aschen:
Sonarqube is not happy because he didn't have access to the credentials in your forked repo.
We will fix that for future PR

PR #1230, opened 2019-01-08
switched from uws to ws

etrousset:
## What does this PR do ?
Remove dependence to uws and switched to ws. 

This should allow running Kuzzle on NodeJS 10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -121,6 +121,26 @@ jobs:
         - docker-compose -f .ci/test.yml run kuzzle
 
 # ---------------------------------------
+# Node 10 - x64
+# ---------------------------------------
+    - stage: Functional tests
+      name: "Functionnal Tests - amd64 - Node v10"
+
+      env:
+        - DOCKER_COMPOSE_VERSION=1.9.0
+        - NODE_LTS=10
+        - DOCKER_PROXY_TAG=$(if [ "${TRAVIS_BRANCH}" != "master" ]; then echo ":develop"; fi)
+
+      services:
+        - docker
+
+      before_install:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -120,6 +120,26 @@ jobs:
       script:
         - docker-compose -f .ci/test.yml run kuzzle
 
+# ---------------------------------------
+# Node 10 - x64
+# ---------------------------------------
+    - stage: Functional tests
+      name: "Functionnal Tests - amd64 - Node v10"
```
xbill82:
```suggestion
      name: "Functional Tests - amd64 - Node v10"
```

etrousset:
@benoitvidis : As I expected the two packages had the same interface, I just replaced package `uws` with `ws` and launched tests. Has nothing seemed to be broken, I thought it was ok.

But actually, you are right the `require('uws') ` should have failed.... so it looks like the tests doesn't really test the websocket protocol?

scottinet:
@Aschen > Did you try it? From what I know, the node bindings are unmaintained and do not work with node.js 10

Aschen:
Hmm I just tested it and it seems to be a different library.
```
console.log('uWebSockets.js');
{ App: [Function],
  SSLApp: [Function],
  nextTick: [Function],
  processNextTickQueue: [Function],
  print: [Function],
  us_listen_socket_close: [Function] }
```
Nevertheless, I still think that we can use Discord fork or use our own fork based on their till the API didn't change.

scottinet:
obsoleted by https://github.com/kuzzleio/kuzzle/pull/1301

PR #1235, opened 2019-01-21
Allow to set request error for request:onError pipes

Aschen:
## What does this PR do ?

This PR allows to manually set another error than the original one when using a pipe on `request:onError` event.  
Before this PR, the only way to change the error returned to the client was to instantiate a new request: 
```
      const
        {
          data,
          options
        } = request.serialize(),
        customRequest = this.context.constructors.Request(data, options);

      customRequest.setError(new this.context.errors.UnauthorizedError('WRONG_CREDENTIALS'));

      return customRequest;
```

  - Step 1 : In a plugin, add the following hook on `request:onError`
```
  async postLoginError (request) {
    if (request.input.controller === 'auth' && request.input.action === 'login' && request.status === 401) {
      const
        {
          data,
          options
        } = request.serialize(),
        customRequest = this.context.constructors.Request(data, options);

      customRequest.setError(new this.context.errors.UnauthorizedError('WRONG_CREDENTIALS'));

      return customRequest;
    }

    return request;
  }
```
  - Step 2 : Then try to login into Kuzzle with wrong credentials, you should see a 401 error with `WRONG_CREDENTIALS`

### DIFF
--- a/lib/api/controllers/funnelController.js
+++ b/lib/api/controllers/funnelController.js
@@ -443,6 +443,13 @@ class FunnelController {
             }
 
             return Bluebird.reject(modifiedRequest.error);
+          })
+          .catch(err => {
+            if (err instanceof KuzzleError) {
+              throw err;
+            }
+
+            throw new PluginImplementationError(err);
           });
       });
   }
--- a/lib/api/controllers/funnelController.js
+++ b/lib/api/controllers/funnelController.js
@@ -443,8 +443,7 @@ class FunnelController {
             }
 
             return Bluebird.reject(modifiedRequest.error);
-          })
-          .catch(() => Bluebird.reject(_error));
+          });
       });
   }
 
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Kuzzle enables you to build modern web applications and complex IoT networks in 
 
 ### Quick install
 
-The easyest way to setup a kuzzle server for Linux-like systems without prerequisites is to download and run our installation script:
+The easiest way to setup a kuzzle server for Linux-like systems without prerequisites is to download and run our installation script:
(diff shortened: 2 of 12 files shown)


lib/api/controllers/funnelController.js
```diff
@@ -443,8 +443,7 @@ class FunnelController {
             }
 
             return Bluebird.reject(modifiedRequest.error);
-          })
-          .catch(() => Bluebird.reject(_error));
+          });
```
scottinet: I disagree: we _have_ to enforce an error inheriting the `KuzzleError` abstract class, if only to make sure that the expected fields are there, and preventing reading possible undefined values.

To fix the problem mentionned in the PR, I would instead do this:

```js
          .catch(error => {
            if (!(error instanceof Error) || !(error instanceof KuzzleError)) {
              throw new PluginImplentationError(error);
            } else {
              throw error;
            }
          }
```

Also, PR fixing bugs should as often as possible include new unit tests reproducing the problem, with the PR fixing those. And in that specific case, I believe that this bug does warrant new unit tests, since we're fixing a behavior directly visible to plugin developers.

Aschen: They can't set another error than a KuzzleError on a request because the `setError` method wrap any other error into an `InternalError` and you can't manually set the error: https://github.com/kuzzleio/kuzzle-common-objects/blob/master/lib/request.js#L232
I still kept your proposal and I added tests.

scottinet: That's true for `setError`, but since we're in a catched promise, we have to assume that the resulting error comes from a `throw` (or a `reject`), and that `setError` hasn't necessarily been called

PR #1239, opened 2019-02-06
Fix createFirstAdmin HTTP route

ballinette:
## What does this PR do ?

Old HTTP route for createFirstAdmin (`:_id/_createFirstAdmin`) had a conflict with other routes.
For example, we could not create an admin user with `admin` uid, because `/admin/_createFirstAdmin` failed.

This PR fixes that isse by renaming the route in  `/_createFirstAdmin/:_id`... and keep the old one as deprecated

In an empty Kuzzle instance, call the following CURL request and verify that the admin user is created:

```
curl --request GET \
  --url http://<kuzzlehost>:<kuzzleport>/_createFirstAdmin/admin \
  --header 'content-type: application/json' \
  --data '{
  "content": {
    "name": "John Doe"
  },
  "credentials": {
    "local": {
      "username": "userAdmin", 
      "password": "myPassword"
    }
  }
}'
```

### DIFF
--- a/features/support/api/http.js
+++ b/features/support/api/http.js
@@ -305,7 +305,7 @@ class HttpApi {
     };
 
     if (id !== undefined) {
-      options.url = this.apiPath('_createFirstAdmin/' + id);
+      options.url = this.apiPath(`_createFirstAdmin/${id}`);
     }
 
     if (reset) {
--- a/features/support/api/http.js
+++ b/features/support/api/http.js
@@ -299,13 +299,13 @@ class HttpApi {
 
   createFirstAdmin (body, id, reset) {
     const options = {
-      url: this.apiPath('/_createFirstAdmin'),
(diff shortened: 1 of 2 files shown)


features/support/api/http.js
```diff
@@ -299,13 +299,13 @@ class HttpApi {
...
     const options = {
-      url: this.apiPath('/_createFirstAdmin'),
+      url: this.apiPath('_createFirstAdmin'),
       method: 'POST',
       body
     };
 
     if (id !== undefined) {
-      options.url = this.apiPath('/' + id + '/_createFirstAdmin');
+      options.url = this.apiPath('_createFirstAdmin/' + id);
```
Aschen:
```suggestion
      options.url = this.apiPath(`_createFirstAdmin/${id}`);
```

PR #1245, opened 2019-02-18
KZL-1005 -  Add method to use request context in SDK queries

Aschen:
EDIT: following the discussion we had here about this functionality, I have to edit this PR but I keep the original one bellow.

## What does this PR do ?

Adds a method to use the original request user inside the requests sent by the embedded SDK.  
The main purpose of this feature is to properly set Kuzzle metadata (`author` and `updater`) on documents.

Call to `sdk.as(user)` method will instantiate a new SDK with a `FunnelProcotol` that takes the user in the constructor and inject it in the request context before passing the request to the`Funnel`.  

```
// This will create a document with `author` metadata to `null`
this.context.accessors.sdk.document.create(...); 

// This will create a document with `author` metadata to the original request user
this.context.accessors.sdk.as(request.context.user).document.create(...);

// This will throw a PluginImplementationError, no user impersonation for now
this.context.accessors.sdk.as({ user: { _id: 'gordon' } }).document.create(...);
```

Documentation: https://github.com/kuzzleio/documentation/pull/253

Consider the following plugin called `ctoi-le-plugin`
```
class Plugin {
  constructor () {
    this.context = null;
  }

  init (config, context) {
    this.context = context;

    this.controllers = {
      test: {
        createDocs: 'createDocs'
      }
    };

    this.routes = [
      { verb: 'get', url: '/test/createDocs', controller: 'test', action: 'createDocs' }
    ];

    this.pipes = { }

    this.hooks = { }
  }

  async createDocs(request) {
    this.context.accessors.sdk.as(request.context.user).document.create('nyc-open-data', 'yellow-taxi', { toto: 'lol'});

    this.context.accessors.sdk.document.create('nyc-open-data', 'yellow-taxi', { toto: 'lol'});
    return true;
  }
}
```
  - Step 1 : Call the plugin route as anonymous, 2 document should be created, one will have `_kuzzle_info.author === '-1'` and the second `_kuzzle_info.author === null`

  - Step 1 : Call the plugin route [http://localhost:7512/_plugin/ctoi-le-plugin/test/createDocs](http://localhost:7512/_plugin/ctoi-le-plugin/test/createDocs)
  - Step 3 :  Call the plugin route, 2 document should be created, one will have `_kuzzle_info.author === 'admin'` and the second `_kuzzle_info.author === null`

## Other changes

 - throw an error when accessing the `sdk.realtime` controller

# ORIGINAL PR
## What does this PR do ?

When making a request with the embedded SDK, the executed request does not have the same context as the original request.  
As a result, the documents created do not have the `author` metadata for example.

This PR adds a `sdk.context()` method that take the original request JWT and send it along with the following SDK request. Queries that does not have a `sdk.context()` call before their execution will have the `author` metadata` set to `null`.  

Consider this controller action:
```
  async createDocs (request) {
    this.context.accessors.sdk.context(request).document.create('nyc-open-data', 'yellow-taxi', { foo: 'bar'})

    this.context.accessors.sdk.document.create('nyc-open-data', 'yellow-taxi', { foo: 'bar'})
  }
```
If I am logged as the admin user as I execute this controller action, the first call to `document.create` will result to a document with `author` metadata set to `'admin'`.  
The second call is not contextualized, so the document will be created with the `author` metadata set to `null`

Consider the following plugin
```
class Plugin {
  constructor () {
    this.context = null;
  }

  init (config, context) {
    this.context = context;

    this.controllers = {
      test: {
        createDocs: 'createDocs'
      }
    };

    this.routes = [
      { verb: 'get', url: '/test/createDocs', controller: 'test', action: 'createDocs' }
    ];

    this.pipes = { }

    this.hooks = { }
  }

  async createDocs(request) {
    this.context.accessors.sdk.context(request).document.create('nyc-open-data', 'yellow-taxi', { toto: 'lol'});

    this.context.accessors.sdk.document.create('nyc-open-data', 'yellow-taxi', { toto: 'lol'});
    return true;
  }
}
```
  - Step 1 : Call the plugin route as anonymous, 2 document should be created, one will have `_kuzzle_info.author === '-1'` and the second `_kuzzle_info.author === null`

  - Step 1 : Log as admin
  - Step 3 :  Call the plugin route, 2 document should be created, one will have `_kuzzle_info.author === 'admin'` and the second `_kuzzle_info.author === null`


Aschen:
@scottinet I want your review on this one before finishing writing unit test.

Concerning rights checking, I intentionally skip this part because I found it irrelevant, I assume that is plugin developer responsibility.

Aschen:
You're right, your solution with instantiating a new SDK is much simpler and clever :)

Concerning the race condition, I was aware of it and this is why I created a new variable for each `query` call to indicate whether the query must have a context or not.

lib/api/core/sdk/funnelProtocol.js
```diff
@@ -22,23 +22,34 @@
...
 
 const
-  KuzzleEventEmitter = require('kuzzle-sdk').KuzzleEventEmitter,
-  Request = require('kuzzle-common-objects').Request;
+  Request = require('kuzzle-common-objects').Request,
+  KuzzleEventEmitter = require('kuzzle-sdk').KuzzleEventEmitter;
 
 class FunnelProtocol extends KuzzleEventEmitter {
-  constructor(funnel) {
+  constructor(funnel, originalRequest = null) {
```
ballinette: Instead of `originalRequest` I would use a `RequestContext` object as second parameter:

If we want to inject a custom context inside a Hook or a Pipe, we don't want to instantiate a full dummy Request as all we want is the context.

So the SDK instantiation would be:
* without context:
```js
this.context.accessors.sdk.document.create(...)
```

* inside a plugin controller, if we want the context of the original request:
```js
this.context.accessors.sdk.context(request.context).document.create(...)
```

* inside a Hook or a Pipe, if we want to inject a custom context (for example: user = my-plugin-identifier):
```js
const customContext = new RequestContext({my-custom-context-info});
this.context.accessors.sdk.context(customContext).document.create(...);
```

ballinette: That means we should also expose a RequestContext contructor to PluginContext...

Aschen: For me this syntax is a little heavy: `this.context.accessors.sdk.context(request.context).document.create`  

I'm thinking about two solutions:
  
  - `context()` can take either a Request or a RequestContext
  - `context()` always take a Request, even if we didn't need a full dummy Request, it's much simpler to understand since developers are more used to work with Request objects than RequestContext objects directly.

ballinette: > For me this syntax is a little heavy: `this.context.accessors.sdk.context(request.context).document.create`

Really more heavy than `this.context.accessors.sdk.context(request).document.create` ?

>     * `context()` can take either a Request or a RequestContext

=> NO WAY It's always a bad solution to allow different types for a single parameter.
 
>     * `context()` always take a Request, even if we didn't need a full dummy Request, it's much simpler to understand since developers are more used to work with Request objects than RequestContext objects directly.

But:
* the Request object contains lots of attributes that are totally unused here 
* developers are not much used to **create** Request objects (and which `controller`, `action`, etc, should we give to the constructor here ???)
* it's not less easy to understand how to create a RequestContext object with a `user` attribute

ballinette:
other suggestion: do not expose the SDK as an accessor, with this odd `context` method, but rather expose a SDK constructor with the context as optional parameter, as we already have with Koncorde, Repository or so on:

* instantiate a SDK instance:

```js
// without context:
const sdk = new this.context.constructors.sdk();

// with the context of the original request:
const sdk = new this.context.constructors.sdk(request.context);

// with a custom context:
const
  customContext = new RequestContext({my-custom-context-info}),
  sdk = new this.context.constructors.sdk(customContext);
```

* use the SDK:
```js
sdk.document.create(...);
```

We can then use the same SDK instance for multiple API calls that need the same context..

lib/api/core/sdk/funnelProtocol.js
Aschen: A little bit, but I admit it doesn't really change much...

After some thought I agree that passing a RequestContext object is more logic.  

The point is that for most of the request we might want to use the original user id and for some we want to use a custom user id.
Maybe we can pass directly an object representing the user ?

```
this.context.accessors.sdk.context(request.context.user).document.create(...)

const user = { _id: 'admin' }
this.context.accessors.sdk.context(user).document.create(...)
```

In this case maybe we can think about another method name than `context`

ballinette: about method name, see also my other suggestion below, about exposing a sdk constructor instead of an accessor.

about the type of the "contextual" parameter, what do we really need (what from the request's context is relevant ?)
Actually, the RequestContext contains following attributes:
* `token`: the current's user JWT => not usable in a plugin context
* `connection`: object representing the original network connection (as well as the protocol name) => no need to transmit the original request's `connection` object, but we could inject automatically in each request a custom object with "internal" as protocol name, for sample ?
* `user`: object representing the current logged in user... do we need the whole object, or only his ID ? What if we inject a custom user ID that do not exist inside our database ? Do we accept it anyway or do we need to validate his existence ? (I really do not know what is the best approach, but we should ask ourselves this question).

ballinette: ... To resume, I suggest, as you, that we only need the `user` from the context (the whole object or his ID... to be decided), and we could also add a custom `connection.protocol` object to the "internal" RequestContext, like this:
```js
class FunnelProtocol extends KuzzleEventEmitter {
  constructor(funnel, user = null) {
    super();

    this.funnel = funnel;
    this.id = 'funnel';
    this.user = user;
  }

  isReady () {
    return true;
  }

  query (request) {
    const kuzzleRequest = new Request(request, {protocol: "funnel"});

    if (this.user) {
      kuzzleRequest.context.user = this.user;
    }

    return this.funnel.executePluginRequest(kuzzleRequest)
      .then(result => ({ result }));
  }
}
```

scottinet:
@ballinette > this is subjective but I really don't like your proposal about asking plugin developers to instantiate new SDK objects just to have a request context.
I find it hard to explain and thus hard to document why a new SDK must be instantiated for each request.
Even if `.context(request)` does that under the hood, from a purely syntaxical perspective I find it far more understandable "to execute an API action using a request context" rather than "to instantiate an entirely new SDK to execute an API action" 

Also, I feel that we should ask plugin developers to pass the entire Request object, and neither the RequestContext nor the User object embedded in it:

* it costs nothing to us to fetch the information we need from a Request object
* in the long term, it's probably less prone to breaking changes than asking developers to provide only the exact properties we need right now
* it does not introduce unnecessary notions to the plugin developer: they already are accustomated to the Request object, we do not need to burden them with additional complexity

lib/api/core/sdk/funnelProtocol.js
benoitvidis: I agree with this last proposal, passing the complete user. Only, I would then rename `context` to `as`:

`this.context.accessors.sdk.as(request.context.user).document.create(...)`

ballinette:
I don't see why it would be hard to understand for a plugin developer that, to make a request with the SDK, he has to write
```js
const sdk = new this.context.constructors.sdk();
```
while it is exactly what he does, for example, with the Koncorde engine...

(NB: he does need to instantiate a SDK for each request, but for each context... if he wants to call multiple requests within the same context, he just needs to instantiate a single SDK instance)

but anyway, if you still want to expose the SDK as an accessor, I prefer the suggestion from @benoitvidis about the "context" method name:
```js
this.context.accessors.sdk.as(request.context.user).document.create(...)
```

scottinet:
I like Benoit's suggestion a lot as well, I think it's both readable, short and almost natural to explain and document.

Aschen:
I like @benoitvidis solution as well :+1:

lib/api/core/plugins/pluginContext.js
```diff
@@ -135,19 +135,36 @@ class PluginContext {
...
+        collection: sdk.collection,
+        document: sdk.document,
+        index: sdk.index,
+        ms: sdk.ms,
+        realtime: sdk.realtime,
+        security: sdk.security,
+        server: sdk.server
+      });
+
+      // TODO: Wait for the end of Node.js 6 support to use the spread operator
```
kuzzle: INFO Complete the task associated to this TODO comment. [rule](https://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS1135)

kuzzle:
* ![INFO][INFO] 1 info

[INFO]: https://sonarsource.github.io/sonar-github/severity-info.png 'Severity: INFO'

lib/api/core/sdk/funnelProtocol.js
```diff
@@ -22,23 +22,44 @@
 'use strict';
 
 const
-  KuzzleEventEmitter = require('kuzzle-sdk').KuzzleEventEmitter,
-  Request = require('kuzzle-common-objects').Request;
+  {
+    Request,
+    errors: {
+      InternalError
```
kuzzle: MAJOR Remove this override of "InternalError". [rule](https://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS2424)

kuzzle:
* ![MAJOR][MAJOR] 1 major
* ![INFO][INFO] 1 info

[INFO]: https://sonarsource.github.io/sonar-github/severity-info.png 'Severity: INFO'

Aschen:
@ballinette @scottinet @benoitvidis you might review this PR again

kuzzle:
* ![INFO][INFO] 1 info

[INFO]: https://sonarsource.github.io/sonar-github/severity-info.png 'Severity: INFO'

lib/api/core/sdk/funnelProtocol.js
```diff
@@ -22,23 +22,44 @@
...
     return true;
   }
 
+  /**
+   *  Hydrate the user and execute SDK query
+   */
   query (request) {
-    return this.funnel.executePluginRequest(new Request(request))
-      .then(result => ({result}));
+    const kuzzleRequest = new Request(request);
```
ballinette: Suggestion: maybe we could also inject a hard-coded `connection` objet to the context (for example with attribute `{protocol: 'funnel'}`)?
I mean construct the `Request` like this:
```js
    const kuzzleRequest = new Request(request, {
      connection: {protocol: "funnel"}
    });
```

Aschen: Good idea!

lib/api/core/plugins/pluginContext.js
```diff
@@ -135,19 +136,46 @@ class PluginContext {
...
-          auth: sdk.auth,
-          bulk: sdk.bulk,
-          collection: sdk.collection,
-          document: sdk.document,
-          index: sdk.index,
-          ms: sdk.ms,
-          realtime: sdk.realtime,
-          security: sdk.security,
-          server: sdk.server
+          query: kuzzleSdk.query.bind(kuzzleSdk),
```
scottinet: Why don't you use your new `getPluginSdk` method to prevent duplicated code?

Something like:

```js
get: () => {
  const wrapped = getPluginSdk(kuzzleSdk);
  wrapped.as = /* your "as" method is fine */

  return wrapped;
}
```

lib/api/core/sdk/funnelProtocol.js
```diff
@@ -22,23 +22,46 @@
...
-  constructor(funnel) {
+  constructor(funnel, user = null) {
     super();
+
     this.funnel = funnel;
     this.id = 'funnel';
+    this.user = user;
+
+    if (this.user && !(this.user instanceof User)) {
+      throw new KuzzleInternalError('FunnelProtocol.constructor: Invalid User object');
```
scottinet: (nitpicking) I agree with that check, but if we do get that error in production, then it's probably a good idea to add the received `user` argument in the error message, to make debugging easier.

lib/api/core/sdk/funnelProtocol.js
```diff
@@ -22,23 +22,46 @@
...
+   *  Hydrate the user and execute SDK query
+   */
   query (request) {
-    return this.funnel.executePluginRequest(new Request(request))
-      .then(result => ({result}));
+    const kuzzleRequest = new Request(request, {
+      connection: { protocol: 'funnel' }
+    });
+
+    if (this.user) {
```
scottinet: This condition should be replaced with a proper call to the `Request` object constructor:

```js
const kuzzleRequest = new Request(request, {
  connection: { protocol: 'funnel' },
  user: this.user
});
```

PR #1263, opened 2019-03-15
Change start sequence for mappings, fixtures and securities

Aschen:
## What does this PR do ?

The actual start sequence is the following:
 - init internal engine, services, cache, funnel, ...
 - init plugins
 - load fixtures, mappings and securities

The problem is that if a plugin need to do some request on an index when he start (on the `core:kuzzleStart` event), the mappings and fixtures won't be loaded for now.  

Moreover, when the event `core:kuzzleStart` is triggered plugin assume that Kuzzle is ready to receive request but in fact Kuzzle is still loading mappings, fixtures and securities.  

This PR change the start sequence to load fixtures and mappings before plugins initialization and securities after plugins initialization (we need the plugin for strategies before loading users with credentials) and after that the event `core:kuzzleStart` is fired.

  - Step 1 : Start docker-compose without Kuzzle (and nginx since it depends on kuzzle service)
  - Step 2 : Start kuzzle with the command line with mappings, fixtures and securities (if you have the ffly4u you repo, lucky you! `./bin/kuzzle start --mappings ../ffly4u/docker/default-mappings.json --fixtures ../ffly4u/docker/default-fixtures.json --securities ../ffly4u/docker/default-securities.json`)


lib/api/kuzzle.js
```diff
@@ -128,9 +128,31 @@ class Kuzzle extends EventEmitter {
       .then(() => this.indexCache.init())
       .then(() => this.repositories.init())
       .then(() => this.funnel.init())
+      .then(() => {
+        if (! params.mappings) {
+          return null;
+        }
+
+        return this.janitor.loadMappings(params.mappings);
```
scottinet: (nitpicking) Wouldn't be cleaner if the `janitor.loadXxx` methods had a defined behavior if no mappings are provided to them?

This way you could simply write: `then(() => this.janitor.loadMappings(params.mappings))`

Aschen: I leave the `loadSecurities` to keep the message about the loading time

PR #1265, opened 2019-03-19
KZL-508 - auth:refreshToken

benoitvidis:
:warning: depends on #1267 

## What does this PR do ?

This PR adds a new `refreshToken` action to the `auth` controller.

Given a valid JWT, the action invalidates the given JWT and provides a new fresh token.
NB: The given jwt is expired only after a grace period to let concurrent client requests end up successfully during the refresh.


scottinet:
I've put the "do not merge" tag temporarily: we need to fix 2 bugs before this PR can be merged and make its way into production (I'm on it, PRs should be submitted real soon)

lib/api/controllers/authController.js
```diff
@@ -340,6 +369,34 @@ class AuthController {
...
+
+    const options = {};
+    if (request.input.args.expiresIn) {
+      options.expiresIn = request.input.args.expiresIn;
+    }
+
+    return this.kuzzle.repositories.token.generateToken(request.context.user, request, options)
+      .then(token => {
+        // we allow a grace period before expiring the current token
+        setTimeout(
```
scottinet: After a long and hard think about this "grace period" feature: I thought it was a good idea, but now I'm not so sure.
As you implemented it, the same token can be refreshed multiple times, generating many tokens from a single one. And we can also keep a token alive by constantly refreshing it.
I think that this introduces a security risk.

Still, I find this option interesting.
I think that this can work _only_ if this token cannot be refreshed anymore during the grace period, maybe by adding a flag in the information held by Redis?
This would allow the use of the token during the grace period, while preventing the token leak I described above.

Also, instead of using `expire` with a TTL of `0` (at least I assume that's what the redis client does), you can directly provide the grace period in milliseconds to it (`expire` does not expire a key, but it changes its TTL)

scottinet: Since time is short, I'm continuing this work. As seen together, I'll solve the token bomb problem (even if we both don't know if this is a real risk or not), and also the realtime subscriptions expiring with the 1st token instead of being refreshed too.

lib/api/controllers/authController.js
```diff
@@ -369,6 +372,44 @@ class AuthController extends BaseController {
...
     return Bluebird.resolve({});
   }
+
+  /**
+   * @param {Request} request
+   */
+  refreshToken(request) {
+    assertIsAuthenticated(this.anonymousId, request);
+
+    // do not refresh an already refreshed token
```
ballinette: why ?

scottinet: Because otherwise, you can generate many tokens during a token's grace period. And then, from each one of these new tokens, you can once again re-generate other tokens.

Neither Benoit nor I were able to correctly evaluate the risk, but we thought that preventing this potential token bomb was not a bad thing.

PR #1266, opened 2019-03-19
Fix controllers action leak

scottinet:
# Description

Controllers actions were implicitly exposed: whenever an API request is submitted, the funnel checks if the requested action is exposed as a function by the requested controller, and if so, the action was allowed to be executed.
This means that internal methods, which are obviously not API actions, are executable too.

Hopefully, I've extensively tested this, and it appears that *for now* this has little consequence. But this is a major bug that needs to be fixed ASAP, just in case.

This PR does that, by making controllers action exposure explicit. 

# Other changes

* Entrypoints unit test: solve a randomly failing test because of a mock-require leakage

# How to test it

We can execute the `_mChanges` internal method from the document controller, just by sending the following request:

```json
{"controller":"document","action":"_mChanges","body":{"documents":[] },"index":"foo","collection":"bar"}
```

With this PR active, we get a `BadRequestError` instead, with the following message: `No corresponding action _mChanges in controller document`

### DIFF
--- a/lib/api/controllers/bulkController.js
+++ b/lib/api/controllers/bulkController.js
@@ -32,20 +32,8 @@ const
  */
 class BulkController extends BaseController {
   constructor (kuzzle) {
-    super(kuzzle);
+    super(kuzzle, ['import']);
     this.engine = kuzzle.services.list.storageEngine;
-  }
-
-  /**
-   * Check if the provided action name exists within that controller.
-   * This check's purpose is to prevent actions leak by making actions exposure
-   * explicit.
-   *
-   * @param  {string} name
-   * @return {boolean}
-   */
-  isAction(name) {
-    return name === 'import';
   }
 
   /**
--- a/lib/api/controllers/controller.js
+++ b/lib/api/controllers/controller.js
@@ -19,21 +19,17 @@
  * limitations under the License.
  */
 
-'use strict';
-
 const
-  BaseController = require('./controller'),
-  {errors: {PartialError}} = require('kuzzle-common-objects'),
-  {assertHasBody, assertBodyHasAttribute} = require('../../util/requestAssertions');
+  {
+    errors: {
+      InternalError: KuzzleInternalError
+    }
+  } = require('kuzzle-common-objects');
 
-/**
- * @class BulkController
- * @param {Kuzzle} kuzzle
- */
-class BulkController extends BaseController {
-  constructor (kuzzle) {
-    super(kuzzle);
-    this.engine = kuzzle.services.list.storageEngine;
+// Base class for all controllers (except the funnel one)
+class BaseController {
+  constructor(kuzzle) {
+    this.kuzzle = kuzzle;
   }
 
   /**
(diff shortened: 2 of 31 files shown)


test/api/core/entrypoints/embedded/protocols/mqtt.test.js
```diff
@@ -10,33 +10,32 @@ const
...
+    mockrequire('mosca', {
+      Server: moscaMock
+    });
+
+    MqttProtocol = mockrequire.reRequire('../../../../../../lib/api/core/entrypoints/embedded/protocols/mqtt');
   });
 
   after(() => {
-    clock.restore();
+    mockrequire.stopAll();
```
scottinet: fix a mock-require leakage (failing that, the "mqtt" protocol is still mocked)

test/api/core/entrypoints/embedded/index.test.js
```diff
@@ -120,6 +121,10 @@ describe('lib/core/api/core/entrypoints/embedded/index', () => {
     }
   });
 
+  after(() => {
+    mockrequire.stopAll();
```
scottinet: fix a mock-require leakage (failing that, the "mqtt" protocol is still mocked)

That fix prevents the MQTT test below from randomly failing

lib/api/controllers/bulkController.js
```diff
@@ -22,18 +22,32 @@
...
   {assertHasBody, assertBodyHasAttribute} = require('../../util/requestAssertions');
 
 /**
  * @class BulkController
  * @param {Kuzzle} kuzzle
  */
-class BulkController {
+class BulkController extends BaseController {
   constructor (kuzzle) {
+    super(kuzzle);
```
Aschen: Why aren't you declaring the action in the `actions` property for this one?

lib/api/controllers/controller.js
```diff
@@ -0,0 +1,49 @@
...
+   * Check if the provided action name exists within that controller.
+   * This check's purpose is to prevent actions leak by making actions exposure
+   * explicit.
+   *
+   * @param  {string} name
+   * @return {boolean}
+   */
+  isAction(name) {
+    throw new KuzzleInternalError(
+      `Call to incomplete controller when invoking the action ${name}`);
```
Aschen: Why this method has to be re-implemented for each controller ? It can just check the `actions` property content so we can avoid code duplication in every controllers

lib/api/controllers/controller.js
```diff
@@ -0,0 +1,49 @@
...
+const
+  {
+    errors: {
+      InternalError: KuzzleInternalError
+    }
+  } = require('kuzzle-common-objects');
+
+// Base class for all controllers (except the funnel one)
+class BaseController {
+  constructor(kuzzle) {
```
Aschen: Proposal: pass the `actions` list as constructor argument and do the `this.actions = new Set(actions)` in the BaseController constructor to avoid unnecessary code duplication

lib/api/controllers/bulkController.js
scottinet: Because I prefer my code duplicated. DRY is for the weak.

2wf3wi


(fix incoming :expressionless: )

scottinet: Fixed.

lib/api/controllers/controller.js
scottinet: Resolving: same as the previous comment.

lib/api/controllers/controller.js
scottinet: Resolving: same as the previous comment.

PR #1273, opened 2019-03-26
Fix incomplete access logs

scottinet:
# Description

Context: access logs are invoked every time an API route is invoked, to keep a trace about Kuzzle's activity.

There are two problems with how access logs are currently handled:
1. If a connection drops before the end of an API request, then the warning `No connection retrieved for connection id: <connection id>` is logged. This is a problem, because, even if the connection had since dropped, the query was executed. This means that our access logs are incomplete
2. Whenever we log messages such as the warning previously described, or the ones printed when an authentication token cannot be decoded, this invokes `pluginsManager.trigger` which creates a promise in the event loop. With enough activity, this can quickly clogs up the event loop and endanger Kuzzle's stability

This PR solves the 1st problem by simply using the provided `Request` object, which holds basic connection and token informations, when the raw network connection object is not available (meaning that the connection no longer exists).
The second problem is solved by simply removing the usages of the `log:xxx` events, as those didn't provide valuable information at all.

:warning: access logging is a sensitive feature of kuzzle, please be extra careful when reviewing this PR

# Other changes

When printing an access log, we now print `(anonymous)` instead of `-1` when a query is executed by an unauthenticated user.

# How to test

1. Activate access logs by setting `server.logs.transports.silent = false` in the kuzzlerc configuration file
2. Run a script that stresses Kuzzle to the point that it starts buffering requests to prevent overload
3. Shut down the script suddenly

Before this PR: many `No connection retrieved for connection id...` warnings are printed on the console. With this PR: access logs are properly printed even on dead connections.

Moreover, Kuzzle takes far less time to recover with this PR than without, because less promises are put on the event loop.

Script that you can use to test this PR:

```js
const WebSocket = require('uws');

const
  nbjobs = 10000,
  ws = new WebSocket('ws://172.18.0.5:7512', {perMessageDeflate: false}),
  payload = JSON.stringify({controller:'server', action:'info'});

ws.on('open', () => {
  let done = 0;
  const start = Date.now();

  ws.on('message', data => {
    done++;

    if (done === nbjobs) {
      clearInterval(interval);
      const elapsed = (Date.now() - start) / 1000;
      console.log(`Elapsed: ${elapsed}s, Throughput: ${Math.round(nbjobs/elapsed)} requests/s`);
      ws.close();
    }
  });

  for (let i = 0; i < nbjobs; i++) {
    ws.send(payload);
  }
});
```


test/api/core/entrypoints/embedded/index.test.js
```diff
@@ -104,7 +104,7 @@ describe('lib/core/api/core/entrypoints/embedded/index', () => {
 
     entrypoint = new EntryPoint(kuzzle);
 
-    Object.defineProperty(entrypoint, 'log', {
+    Object.defineProperty(entrypoint, 'logger', {
```
scottinet: Unit tests were broken because they checked that the logger was never called... but sinon's spies were put on the wrong entrypoint property name, so the checks always passed :expressionless:

lib/api/core/entrypoints/embedded/index.js
```diff
@@ -358,20 +381,28 @@ class EmbeddedEntryPoint extends EntryPoint {
...
 
-    this.logger.info(connection.ips[connection.ips.length - 1 - this.config.logs.accessLogIpOffset] +
+    const ip = connection.ips[connection.ips.length - 1 - this.config.logs.accessLogIpOffset]
+      || '-';
+
+    if (user === null) {
+      user = '-';
+    }
+
+    this.logger.info(ip +
```
alexandrebouthinon: Why don't you use string literals?

scottinet: Because that would be impractical and hard to read in this very specific case, given the amount of variables used to create an access log message.

lib/api/core/entrypoints/embedded/index.js
```diff
@@ -358,20 +381,28 @@ class EmbeddedEntryPoint extends EntryPoint {
...
         try {
           const
             b64Payload = request.input.jwt.split('.')[1],
             payload = Buffer.from(b64Payload, 'base64').toString('utf8');
           user = JSON.parse(payload)._id;
         }
         catch (err) {
-          this.kuzzle.pluginsManager.trigger('log:warn', `Unable to extract user from jwt token: ${request.input.jwt}`);
+          // do nothing: we don't know anything about the token, it may be
+          // invalid on-purpose for all we know
```
Yoann-Abbes: I don't understand this comment : how and in wich case the token would be invalid on-purpose ?

scottinet: Forgery.

We're dealing with data coming from the outside, so nothing can be trusted.

lib/api/core/entrypoints/embedded/index.js
```diff
@@ -276,12 +282,15 @@ class EmbeddedEntryPoint extends EntryPoint {
    * @param {object} extra
    */
   logAccess (request, extra = null) {
-    const connection = this.clients[request.context.connection.id];
+    let connection = this.clients[request.context.connection.id];
```
ballinette: I wonder if we really need to get the whole `ClientConnection` object, and then use a fallback if the connecion is lost.

Can't we get all the information we need from `request.context.connection` object ?

ballinette: I mean we only need that (or I miss something ?):
```
const connection = {
  id: request.context.connection.id,
  protocol: request.context.connection.protocol || '-',
  ips: request.context.connection.ips || [},
  headers: request.context.connection.misc.headers || []
};
```

lib/api/core/entrypoints/embedded/index.js
```diff
@@ -296,20 +305,33 @@ class EmbeddedEntryPoint extends EntryPoint {
...
+      verb = 'DO';
 
-    if (connection && connection.protocol.startsWith('HTTP/')) {
+    if (connection.protocol.startsWith('HTTP/')) {
       verb = extra.method;
       url = extra.url;
 
-      // try to get plain user name, 1st form jwt token if present, then from basic auth
-      if (connection.headers.authorization) {
+      // try to get plain user name, 1st form jwt token if present, then from
```
ballinette: very-nitpicking: `s/1st form jwt token/1st from jwt token/` ;)

lib/api/core/entrypoints/embedded/index.js
```diff
@@ -296,20 +305,33 @@ class EmbeddedEntryPoint extends EntryPoint {
...
+      // basic auth
+      if (user === null && connection.headers.authorization) {
```
ballinette: Is there any case where `logAccess` could be called before the jwt is set ?

scottinet: I guess on a HTTP `OPTIONS` request?

lib/api/core/entrypoints/embedded/index.js
scottinet: I didn't consider that option :thinking: 
You're right, that shouldn't make a difference, neither data nor performance wise (at this point `ClientConnection` should be promoted to a pure C++ class by turbofan, and allocated on the stack, so the overhead should be comparable to getting a ref from the `clients` dictionnary).

I'll make the necessary changes. :+1:

scottinet: @ballinette > after extensive testing, there are cases when we have a connection but no `Request` associated to it. For instance, when we log an error about the HTTP query itself, before we can even parse it. So we do need a fallback.

ballinette: OK, let's keep this code as is ;)

PR #1277, opened 2019-03-27
Fix documentation links

Aschen:
## What does this PR do ?

Fix documentation links


Aschen:
> why is the auth-local plugin updated?

@scottinet I have no idea >< It seems that nothing change :o

PR #1282, opened 2019-04-02
[bugfix] scan keys on redis cluster

benoitvidis:
## What does this PR do ?

Fixes Kuzzle way of fetching keys when redis is running in cluster mode. 
cf: https://github.com/luin/ioredis/issues/175 

### How should this be manually tested?

An easy way to test this method is to call `server:getStats`.
You can get a ready-to-use redis cluster from the cluster development stack.

PR #1284, opened 2019-04-02
Cluster/functional tests

benoitvidis:
:warning: depends on https://github.com/kuzzleio/kuzzle/pull/1282

## What does this PR do ?

This PR prepares kuzzle to run functional tests under cluster mode.

- adds a new cucumber profile that excludes redis tests, which are incompatible with redis cluster mode
- fixes a little bug with user unsubscription notifications not being send/received in cluster mode.


features/kuzzle.feature
```diff
@@ -351,6 +353,7 @@ Feature: Kuzzle functional tests
   @realtime
   Scenario: get list of subscriptions
     Given A room subscription listening to "lastName" having value "Hopper"
+    And I wait 0.1s
```
scottinet: (nitpicking) wouldn't this latency be better located in the subscription task above?

lib/api/core/hotelClerk.js
```diff
@@ -555,17 +555,18 @@ class HotelClerk {
...
+    }
 
-      const request = new Request({
-        controller: 'realtime',
-        action: 'unsubscribe',
-        index: room.index,
-        collection: room.collection,
-        volatile
-      }, requestContext);
+    // even if the room is deleted for this node, another one may need the notification
```
scottinet: Ok I had a _very_ hard time following what's going on here while in cluster mode, so please correct me if I'm wrong.

I have an issue about that room removal function.
This is not exactly about that line in particular, but that entire function bothers me, and here is why: whenever a user executes the `realtime:subscribe` API route, the cluster plugin makes other nodes artificially create an empty Koncorde room.
If I had to guess, I would say that it's to allow the `realtime:join` route to work.

But what this bugfix shows is that a room can be destroyed on one node, while still existing on other ones. Which, in other terms, means that we have a desync.

I'm not quite sure that this should be resolved in this one PR. But if I had to work on it, I would remove the artifical room creation made by the cluster plugin, and rely on the redis status to check whether a room can be joined or not.

What do you think?

benoitvidis: I need to check again, but I think we still need a room in Koncorde on every node to check if a document write matches a possible subscription.

Regarding the room deletion, we do not sync the hotelclerk and on that level, it is fine. We still keep the room in Koncorde if needed by using an ugly hack in the cluster: https://github.com/kuzzleio/kuzzle-plugin-cluster/blob/master/lib/index.js#L270 though.

Still, I agree, I am not very happy with the way the cluster behaves for the time being. I would tend to think that the next move would be to include it in Kuzzle core. This way, we could get rid of this kind of ugly overrides and make things a little clearer.

features/kuzzle.feature
benoitvidis: I am not sure I get it. 
The cluster latency is between the time the subscription is sent and the time it is replicated to be able to compute the list.

lib/api/core/hotelClerk.js
scottinet: Ohh I missed that hack. And it's ugly all right. :expressionless: 

I agree, integrating the cluster plugin directly into the core is the right move, probably to be done for the v2 though :+1:

features/kuzzle.feature
scottinet: What I'm saying is that you're waiting 0.1s every time you call `Given A room subscription listening...`, so I thought it might be best to wait that 0.1s in the code of that "given" task, instead of in the features files.

benoitvidis: ah ok. I am not sure about that. I would prefer to keep it explicit in the scenario that we are adding a small replication delay rather than hiding it in the subscription step.

test/api/core/hotelClerk/removeCustomerFromAllRooms.test.js
```diff
@@ -64,7 +64,7 @@ describe('Test: hotelClerk.removeCustomerFromAllRooms', () => {
     hotelClerk.removeCustomerFromAllRooms(context);
 
     should(kuzzle.realtime.remove).be.calledOnce();
-    should(kuzzle.notifier.notifyUser).be.calledOnce();
+    should(kuzzle.notifier.notifyUser).be.calledTwice();
```
ballinette: niptickings:
* Add a comment to explain why the `notifyUser` method shoud be called twice
* The test title is not about notifications, only about cleanup: to be more "unit", the call to "notifyUser" should be tested in another test.

benoitvidis: - about the comment: I disagree. From the test perspective, the customer is registered to 2 rooms. IMO, it is more natural that we send 2 notifications than one. There should have been a comment on the original version though.
- OK, I split the tests

PR #1286, opened 2019-04-10
[bugfix] allows bulk inserts on aliases

benoitvidis:
## What does this PR do ?

This PR fixes the case where a bulk operation would fail on an elasticsearch index alias.

```bash
#!/bin/bash
  
set -x
  
es_container=kuz_elasticsearch_1
  
es=$(docker inspect --format='{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' "$es_container"):9200
kuzzle="localhost:7512"
  
curl -XPOST $kuzzle/index/_create
curl -XPUT $kuzzle/index/collection
  
curl -d '{"actions": {"add": {"index": "index", "alias": "alias"}}}' $es/_aliases
  
curl -H "Content-type: application/json" -d '{
    "bulkData": [
      {"index": {"_type": "collection", "_index": "alias"}},
      {"foo": "bar"}
    ]
}' $kuzzle/_bulk
```

### DIFF
--- a/lib/services/elasticsearch.js
+++ b/lib/services/elasticsearch.js
@@ -665,11 +665,22 @@ class ElasticSearch extends Service {
       updatedAt: dateNow
     };
 
+    const esCache = {};
+
     return this.client.indices.getMapping()
-      .then(mappings => {
-        // Keep a local cache of existents index/collections on elasticsearch
-        const reducer = (indexes, index) => Object.assign(indexes, { [index]: Object.keys(mappings[index].mappings) });
-        const esCache = Object.keys(mappings).reduce(reducer, {});
+      .then(raw => {
+        for (const index of Object.keys(raw)) {
+          esCache[index] = Object.keys(raw[index].mappings);
+        }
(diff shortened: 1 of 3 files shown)


stafyniaksacha:
There is no way to create es indexes through kuzzle yet, no ?

scottinet:
Of course there is: https://docs.kuzzle.io/api/1/controller-index/create/

stafyniaksacha:
> Of course there is: https://docs.kuzzle.io/api/1/controller-index/create/

I mean, indexes aliases

scottinet:
Ah, no. I don't think there will ever be a standard API way to create aliases, because this is specific to Elasticsearch.

But we do have some leads about how to standardize database services, with some kind of abstraction layer above it (e.g. using Koncorde's DSL as a primary way to access data), and maybe let the ES service add some custom controllers on top of the standard API. 
This would allow changing the storage solution, while still keeping what makes ES specific and powerful.

PR #1289, opened 2019-04-18
[WebSocket] Handle ping/pong packets

scottinet:
# Description

PING/PONG packets are described in the WebSocket specifications, and are used to:
* efficiently detect and clean up dead sockets (a socket can stay opened indefinitly if the client couldn't send a socket closed packet)
* keep idle sockets alive

This PR:
* adds a new `heartbeat` configuration under the `protocols.websocket` part of the .kuzzlerc file
* adds a heartbeat to the WebSocket protocol, sending PING packets to client sockets, and mark them as "dead" until a PONG response is received
* destroys and cleans up sockets that stayed dead after the heartbeat delay

# Other change

Deduplicated the test made on the `enabled` part of a protocol configuration: the test was made in each protocol implementation, AND in the global entrypoint object during protocols initialization.

# How to test

An easy way to test this is to use `wscat`:
* (if not already done) install it with `npm -g wscat`
* connect it to your local kuzzle stack: `wscat --connect http://localhost:7512`
* suspend it with a SIGSTOP signal (e.g. Ctrl+Z on the terminal)

Without this PR: you can go have a lunch, return to your terminal and wake wscat up (with `fg`), and you can use it again as if nothing happend, even if the socket should have been marked as dead by Kuzzle since it can never respond to any request.

With this PR: if you wait at least 2 minutes by default, and wake wscat up with `fg`, it will exit immediately with the following message: `disconnected`
(you need to wait for 2 heartbeats before the socket is disconnected: the 1st one to mark the socket temporarily as "dead", and the second one to clean up the dead socket)

```
$ wscat --connect http://localhost:7512
connected (press CTRL+C to quit)
> 
[1]  + 24777 suspended  wscat --connect http://localhost:7512
$ fg
[1]  + 24777 continued  wscat --connect http://localhost:7512
disconnected
$
```

You can follow what happens with this test using debug messages (set the `DEBUG` environment variable to `kuzzle:entry-point:protocols:websocket` before starting Kuzzle):

```
  kuzzle:entry-point:protocols:websocket [090cacd2-8ca5-49b1-bf8a-8b9b37b5882d] creating Websocket connection +4s
  kuzzle:entry-point:protocols:websocket [WebSocket] Heartbeat +56s
  kuzzle:entry-point:protocols:websocket [WebSocket] Heartbeat +60s
  kuzzle:entry-point:protocols:websocket [090cacd2-8ca5-49b1-bf8a-8b9b37b5882d] received a `close` event +2ms
  kuzzle:entry-point:protocols:websocket [090cacd2-8ca5-49b1-bf8a-8b9b37b5882d] onClientDisconnection +1ms
```

These logs show that the 2nd heartbeat after the wscat connection correctly force closes the socket, since it couldn't respond to the PING packet in time.

### DIFF
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -376,8 +376,14 @@
         // value of Access-Control-Allow-Origin header to answer the upgrade request
         "origins": "*:*"
       },
+      // * enabled:
+      //    Set to true to enable WebSocket support
+      // * heartbeat:
+      //    The time, in milliseconds, between the server's PING requests.
+      //    Setting this value to 0 disables PING/PONG requests entirely.
       "websocket": {
-        "enabled": true
+        "enabled": true,
+        "heartbeat": 60000
       }
     }
   },
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -373,15 +373,16 @@
       },
       "socketio": {
         "enabled": true,
-        // value of Access-Control-Allow-Origin header to answer the upgrade request
+        // value of Access-Control-Allow-Origin header to answer the upgrade
+        // request
         "origins": "*:*"
       },
-      // * enabled:
-      //    Set to true to enable WebSocket support
-      // * heartbeat:
-      //    The time, in milliseconds, between the server's PING requests.
-      //    Setting this value to 0 disables PING/PONG requests entirely.
       "websocket": {
+        // * enabled:
+        //    Set to true to enable WebSocket support
+        // * heartbeat:
+        //    The time, in milliseconds, between the server's PING requests.
+        //    Setting this value to 0 disables PING/PONG requests entirely.
         "enabled": true,
         "heartbeat": 60000
       }
--- a/lib/api/core/entrypoints/embedded/index.js
+++ b/lib/api/core/entrypoints/embedded/index.js
@@ -97,24 +97,15 @@ class EmbeddedEntryPoint extends EntryPoint {
     this.httpServer = http.createServer();
     this.httpServer.listen(this.config.port, this.config.host);
 
-    if (this.config.protocols.http.enabled) {
-      this.protocols.http = new HttpProtocol();
-      this.protocols.http.init(this);
-    }
+    this.protocols = {
+      http: new HttpProtocol(),
+      mqtt: new MqttProtocol(),
+      websocket: new WebSocketProtocol(),
+      socketio: new SocketIoProtocol()
+    };
 
-    if (this.config.protocols.mqtt.enabled) {
-      this.protocols.mqtt = new MqttProtocol();
-      this.protocols.mqtt.init(this);
-    }
-
-    if (this.config.protocols.websocket.enabled) {
-      this.protocols.websocket = new WebSocketProtocol();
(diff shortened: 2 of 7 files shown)


lib/api/core/entrypoints/embedded/index.js
```diff
@@ -97,24 +97,15 @@ class EmbeddedEntryPoint extends EntryPoint {
...
-
-    if (this.config.protocols.websocket.enabled) {
-      this.protocols.websocket = new WebSocketProtocol();
-      this.protocols.websocket.init(this);
-    }
-
-    if (this.config.protocols.socketio.enabled) {
-      this.protocols.socketio = new SocketIoProtocol();
-      this.protocols.socketio.init(this);
+    this.protocols = {
```
scottinet: Unrelated change: the test on the `enabled` boolean property made in the previous version is duplicated in all protocols implementations, so I decided that protocols were responsible of their own configuration and deduplicated the test by removing the ones made by the global entrypoint object.

.kuzzlerc.sample
```diff
@@ -376,8 +376,14 @@
         // value of Access-Control-Allow-Origin header to answer the upgrade request
         "origins": "*:*"
       },
+      // * enabled:
```
ballinette: nitpicking: to stay coherent with other settings above in the file, comments should be placed "inside" the object:
```
       "websocket": { 
           // <!-- Comments here -->
           enabled: true,
           heartbeat: 60000
         }
```

lib/api/core/entrypoints/embedded/protocols/websocket.js
```diff
@@ -52,12 +53,27 @@ class WebsocketProtocol extends Protocol {
    */
   init (entryPoint) {
     super.init(entryPoint);
```
ballinette: Nitpicking: no need to init the superclass if the protocol is disabled. So we could move this line after the next conditional statement ?

ballinette:
@benoitvidis 
> I suppose that as the PING/PONG implementation is part of the Websocket RFC, the answer to the following question is yes, but are we sure the client implementation of Websockets used by the SDK will reply to the ping requests issued by Kuzzle?

The SDK use either the native WebSocket object when used inside a browser, or the "ws" package in node, which implements the ping/pong mechanism according to the specs:
https://www.npmjs.com/package/ws#how-to-detect-and-close-broken-connections
`Pong messages are automatically sent in response to ping messages as required by the spec.`

scottinet:
@benoitvidis > Yes: any websocket library worth its salt will answer to a PING request with a PONG one, as this is part of the RFC.
In fact, I checked it, and this is true for Kuzzle itself: sending it a PING request will make it respond with a PONG one.

lib/api/core/entrypoints/embedded/protocols/websocket.js
scottinet: Good point. I made the change for other protocols as well, as this can be confusing to have a debug message saying that we're initializing an actually disabled protocol.

scottinet:
@ballinette > I found yourremarks to be more than nitpickings. I made the necessary changes :+1:

PR #1296, opened 2019-05-03
KZL-997/Restrict index and collection list

Aschen:
## What does this PR do ?

Take the profiles restrictions into account for `index:list` and `collection:list`.  

  - Step 1 : Create a user with a profile restricted to one particular index and one collection
  - Step 2 : Log as this user and then call `index:list` and `collection:list`.  
  - Step 3 : You should see only the index and collection allowed for this user

### DIFF
--- a/features/kuzzle.feature
+++ b/features/kuzzle.feature
@@ -1,4 +1,22 @@
 Feature: Kuzzle functional tests
+
+  @security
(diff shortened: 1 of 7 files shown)


scottinet:
Closed, at least for now. This subject may rise again in a more global approach taking multi-tenants into account (maybe...)

PR #1297, opened 2019-05-09
Fix timeouts on plugin action returing the request

benoitvidis:
## What does this PR do ?

This PR fixes the case where Kuzzle would timeout without error when/if a plugin controller action returns a non-serializable response.
This is notably the case if a controller action returns the origin request.

### DIFF
--- a/lib/api/controllers/funnelController.js
+++ b/lib/api/controllers/funnelController.js
@@ -374,6 +374,17 @@ class FunnelController {
       .then(responseData => {
         modifiedRequest.setResult(responseData, {status: request.status === 102 ? 200 : request.status});
 
+        if (!this.isNativeController(modifiedRequest) && !modifiedRequest.response.raw) {
+          // check if the plugin response can be serialized
+          try {
+            JSON.stringify(modifiedRequest.response.toJSON());
+          }
+          catch (e) {
+            modifiedRequest.setResult(null);
+            throw new Error('Unable to serialize response. Are you trying to return the request?');
+          }
+        }
+
         return this.kuzzle.pluginsManager.trigger(this.getEventName(request, 'after'), modifiedRequest);
       })
       .then(newRequest => this.kuzzle.pluginsManager.trigger('request:onSuccess', newRequest))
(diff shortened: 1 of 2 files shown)


lib/api/controllers/funnelController.js
```diff
@@ -374,6 +374,17 @@ class FunnelController {
...
         modifiedRequest.setResult(responseData, {status: request.status === 102 ? 200 : request.status});
 
+        if (!this.isNativeController(modifiedRequest) && !modifiedRequest.response.raw) {
+          // check if the plugin response can be serialized
+          try {
+            JSON.stringify(modifiedRequest.response.toJSON());
+          }
+          catch (e) {
+            modifiedRequest.setResult(null);
+            throw new Error('Unable to serialize response. Are you trying to return the request?');
```
Aschen: I think it's better to throw a `PluginImplementationError` since the error came from a plugin.  
It's only when returning the original request or any Request object ?

EDIT: in fact the `Error` will be wrapped in a `PluginImplementationError` but still I think it's better to be explicit

lib/api/controllers/funnelController.js
```diff
@@ -374,6 +374,17 @@ class FunnelController {
       .then(responseData => {
         modifiedRequest.setResult(responseData, {status: request.status === 102 ? 200 : request.status});
 
+        if (!this.isNativeController(modifiedRequest) && !modifiedRequest.response.raw) {
+          // check if the plugin response can be serialized
+          try {
+            JSON.stringify(modifiedRequest.response.toJSON());
```
scottinet: This check should be performed directly on `responseData` to prevent the unnecessary overhead of `RequestResponse.toJSON`.

lib/api/controllers/funnelController.js
benoitvidis: :+1:

lib/api/controllers/funnelController.js
benoitvidis: :+1:

Yoann-Abbes:
The fact a routs plugin cannot send non-serialisable response anymore shoudnt be documented ?

scottinet:
@Yoann-Abbes > it is documented, and plugins can send non-serializable responses by setting the `response.raw` property to `true`. If you check the code again, you'll see that the check is not performed if that property is set.

PR #1298, opened 2019-05-09
Fix disabled protocol initialization

Aschen:
## What does this PR do ?

Even with the `disabled` property set to `false`, the protocols are all loaded into memory.  
The footprint is small but I ran into a bug with realtime notification where Kuzzle try to send notification with the `dispatch` method even if the protocol was disabled.  

Now the protocols are initialized and if they are `disabled` Kuzzle does not keep them in memory.  

  - Step 1 : Subscribe to realtime notification with MQTT disabled
  - Step 2 : When a notification arrive, you should'nt see an error message in the console

### Other changes
 
 - Protocol `init` method now return a Promise
 - Rename `Protocol.protocol` property to `Protocol.name`

### DIFF
--- a/lib/api/core/entrypoints/embedded/index.js
+++ b/lib/api/core/entrypoints/embedded/index.js
@@ -35,7 +35,8 @@ const
     },
     errors: {
       InternalError: KuzzleInternalError,
-      ServiceUnavailableError
+      ServiceUnavailableError,
+      PluginImplementationError
     }
   } = require('kuzzle-common-objects'),
   MqttProtocol = require('./protocols/mqtt'),
@@ -97,18 +98,27 @@ class EmbeddedEntryPoint extends EntryPoint {
     this.httpServer = http.createServer();
     this.httpServer.listen(this.config.port, this.config.host);
 
-    this.protocols = {
-      http: new HttpProtocol(),
-      mqtt: new MqttProtocol(),
-      websocket: new WebSocketProtocol(),
-      socketio: new SocketIoProtocol()
-    };
+    const initPromises = [];
 
-    for (const protocol of Object.keys(this.protocols)) {
-      this.protocols[protocol].init(this);
+    for (const ProtocolClass of [
+      HttpProtocol,
+      MqttProtocol,
+      WebSocketProtocol,
+      SocketIoProtocol
+    ]) {
+      const protocol = new ProtocolClass();
+
+      initPromises.push(
+        protocol.init(this).then(enabled => {
+          if (enabled) {
+            this.protocols[protocol.name] = protocol;
+          }
+        })
+      );
     }
 
-    return this.kuzzle.repositories.user.anonymous()
+    return Bluebird.all(initPromises)
+      .then(() => this.kuzzle.repositories.user.anonymous())
       .then(anonymous => {
         this.anonymousUserId = anonymous._id;
         return this.loadMoreProtocols();
@@ -161,7 +171,7 @@ class EmbeddedEntryPoint extends EntryPoint {
   }
 
   /**
-   * Loads installed plugins in memory
+   * Loads installed protocols in memory
    */
   loadMoreProtocols () {
     const dir = path.join(__dirname, '../../../../../protocols/enabled');
@@ -169,7 +179,7 @@ class EmbeddedEntryPoint extends EntryPoint {
     let dirs;
     try {
       dirs = fs.readdirSync(dir);
-    } catch(e) {
+    } catch (e) {
       return Bluebird.resolve();
     }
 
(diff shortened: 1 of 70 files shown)


lib/api/core/entrypoints/embedded/index.js
```diff
@@ -191,6 +200,9 @@ class EmbeddedEntryPoint extends EntryPoint {
         })
         .timeout(this.kuzzle.config.services.common.defaultInitTimeout)
         .then(() => {
+          if (this.protocols[manifest.name]) {
+            throw new KuzzleInternalError(`Conflicting protocol name "${manifest.name}".`);
```
scottinet: Should be a `PluginImplementationError` error

lib/api/core/entrypoints/embedded/protocols/http.js
```diff
@@ -374,7 +382,7 @@ class HttpProtocol extends Protocol {
       };
 
     if (typeof allowCompression !== 'boolean') {
-      throw new Error('Invalid HTTP "allowCompression" parameter value: expected a boolean value');
+      throw new KuzzleInternalError('Invalid HTTP "allowCompression" parameter value: expected a boolean value');
```
scottinet: No, this is not a Kuzzle internal error, but a configuration error (which is probably a lacking Error object).

This doesn't really matter because, in the end, the error is caught during initialization, and that aborts Kuzzle's start sequence. The error type is only really useful when a response is sent back to a client, or when it's a plugin error.

test/api/core/entrypoints/embedded/index.test.js
```diff
@@ -112,6 +136,11 @@ describe('lib/core/api/core/entrypoints/embedded/index', () => {
         error: sinon.spy()
       }
     });
+
+    HttpMock.prototype.init = sinon.stub().resolves(true);
```
scottinet: DRY: putting `this.init = sinon.stub().resolves(true);` in the FakeProtocol constructor allows you to get rid of these lines

test/api/core/entrypoints/embedded/protocols/mqtt.test.js
```diff
@@ -58,48 +58,59 @@ describe('/lib/api/core/entrypoints/embedded/protocols/mqtt', () => {
...
     protocol = new MqttProtocol();
-    return protocol.init(entrypoint);
   });
 
-  describe('#init', () => {
+  describe('#init', () => {    
+    it('should return false if the protocol is disabled', () => {
+      entrypoint.config.protocols.mqtt.enabled = false;
+
+      return protocol.init(entrypoint)
```
scottinet: (nitpicking) can be simplified with [fulfilledWith](https://shouldjs.github.io/#assertion-fulfilledwith):

```js
return should(protocol.init(entrypoint)).fulfilledWith(false);
```

lib/api/core/entrypoints/embedded/protocols/http.js
scottinet: Actually I opted for a different approach myself, while working on you-know-what: I'm using the `assert` native module for checks made during startup.
It's cleaner as in most cases we expect failed assertions to stop the program execution at some point.

Aschen: I agree with the usage of `assert`

test/api/core/entrypoints/embedded/index.test.js
Aschen: No because I want to know separately when I call one of the `init` function

scottinet: I don't understand: with my proposal, won't each class instance have its own init function?

Aschen: Since I am instantiating theses classes in the EmbeddedEntrypoint, I have to set the stub directly on the prototype if I want to be able to change the returned value (eg: `MqttMock.prototype.init = sinon.stub().resolves(false);`)

lib/api/core/entrypoints/embedded/index.js
```diff
@@ -97,18 +98,27 @@ class EmbeddedEntryPoint extends EntryPoint {
...
-      http: new HttpProtocol(),
-      mqtt: new MqttProtocol(),
-      websocket: new WebSocketProtocol(),
-      socketio: new SocketIoProtocol()
-    };
+    const initPromises = [];
 
-    for (const protocol of Object.keys(this.protocols)) {
-      this.protocols[protocol].init(this);
+    for (const ProtocolClass of [
```
alexandrebouthinon: Are you sure removing `this.protocols` property does not break the custom protocols implementation capability?

scottinet: `this.protocols` is not removed
What this PR does is to go back to a previous state where only active embedded protocols are stored in that object (I changed that not that long ago, when even inactive protocols where stored)

alexandrebouthinon: Ok tanks for those explanations :slightly_smiling_face:

PR #1299, opened 2019-05-14
[KZL-1046] Add Kuzzle Nightly & Redis 3 and 4 test

alexandrebouthinon:
## What does this PR do ?
Update our Travis configuration to launch heavy and time consuming tests by night using Travis CRON.
  
Look at Travis

### DIFF
--- a/.ci/test-aarch64.yml
+++ b/.ci/test-aarch64.yml
@@ -42,7 +42,7 @@ services:
       - TRAVIS_REPO_SLUG
 
   redis:
-    image: redis:3.2
+    image: redis:${REDIS_VERSION:-5}
 
   elasticsearch:
     image: kuzzleio/elasticsearch:5.6.10
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ jobs:
 # Unit Testing
 # ---------------------------------------
     - stage: "Tests & Linters"
-      if: type = pull_request OR type = cron OR type = push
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -170,13 +170,69 @@ jobs:
...
+      services: docker
+
+      cache:
+        directories:
+          - $HOME/.pm2 
+
+      env:
+        - DOCKER_COMPOSE_VERSION=1.9.0
+        - NODE_LTS=8
+        - REDIS_VERSION=latest
```
benoitvidis: Shouldn't we rather stick to a tag (can be set to `5` imo to follow last versions)?

alexandrebouthinon: Yes you're right :slightly_smiling_face:

alexandrebouthinon:
@Aschen: I update the Redis version in all repo Docker Compose files. I'll update other docker-compose soon

.ci/test-aarch64.yml
```diff
@@ -42,7 +42,7 @@ services:
       - TRAVIS_REPO_SLUG
 
   redis:
-    image: redis:3.2
+    image: redis:${REDIS_VERSION:-5}
```
scottinet: Why do we default to redis 5 if we're only testing v3 and v4?

alexandrebouthinon: Redis 5 is tested in `Integration tests` > `Node v8`. Because this is the new default version.

PR #1300, opened 2019-05-15
[KZL-522] Collection specifications methods cloisoned to a collection

thomasarbona:
## What does this PR do ?

Validate specifications:
- add new endpoint `http://kuzzle:7512/<index>/<collection>/_validateSpecifications`
- deprecate the previous endpoint `http://kuzzle:7512/_validateSpecifications`

Update specifications:
- add new endpoint `http://kuzzle:7512/<index>/<collection>/_specifications`
- deprecate the previous endpoint `http://kuzzle:7512/_specifications`

- Step 1 : call the validate specifications with the new endpoint
- Step 2 : same with update specifications

### DIFF
--- a/lib/config/httpRoutes.js
+++ b/lib/config/httpRoutes.js
@@ -153,7 +153,10 @@ module.exports = [
   {verb: 'post', url: '/credentials/:strategy/_me/_create', controller: 'auth', action: 'createMyCredentials'},
   {verb: 'post', url: '/credentials/:strategy/_me/_validate', controller: 'auth', action: 'validateMyCredentials'},
 
+  /* DEPRECATED - will be removed in v2 */  
   {verb: 'post', url: '/_validateSpecifications', controller: 'collection', action: 'validateSpecifications'},
+
+  {verb: 'post', url: '/:index/:collection/_validateSpecifications', controller: 'collection', action: 'validateSpecifications'},
   {verb: 'post', url: '/validations/_search', controller: 'collection', action: 'searchSpecifications'},
 
   {verb: 'post', url: '/_bulk', controller: 'bulk', action: 'import'},
@@ -290,7 +293,10 @@ module.exports = [
 
   {verb: 'put', url: '/:index/:collection', controller: 'collection', action: 'create'},
   {verb: 'put', url: '/:index/:collection/_mapping', controller: 'collection', action: 'updateMapping'},
+
+  /* DEPRECATED - will be removed in v2 */
   {verb: 'put', url: '/_specifications', controller: 'collection', action: 'updateSpecifications'},
+  {verb: 'put', url: '/:index/:collection/_specifications', controller: 'collection', action: 'updateSpecifications'},
 
   {verb: 'put', url: '/:index/:collection/:_id', controller: 'document', action: 'createOrReplace'},
   {verb: 'put', url: '/:index/:collection/_mCreateOrReplace', controller: 'document', action: 'mCreateOrReplace'},
--- a/lib/config/httpRoutes.js
+++ b/lib/config/httpRoutes.js
@@ -296,6 +296,7 @@ module.exports = [
 
   /* DEPRECATED - will be removed in v2 */
   {verb: 'put', url: '/_specifications', controller: 'collection', action: 'updateSpecifications'},
+  
   {verb: 'put', url: '/:index/:collection/_specifications', controller: 'collection', action: 'updateSpecifications'},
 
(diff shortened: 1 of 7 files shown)


test/api/controllers/collectionController.test.js
```diff
@@ -256,6 +258,80 @@ describe('Test: collection controller', () => {
...
+      };
+
+      kuzzle.validation.isValidSpecification.resolves({
+        isValid: false,
+        errors: ['bad bad is a bad type !']
+      });
+
+      return collectionController.updateSpecifications(request)
+        .catch(error => {
+          should(kuzzle.pluginsManager.trigger).be.calledOnce();
```
Aschen: You can chain assertions on the same stub:
```
          should(kuzzle.pluginsManager.trigger)
            .be.calledOnce()
            .be.calledWith('validation:error', 'Some errors with provided specifications.');
```

test/api/controllers/collectionController.test.js
```diff
@@ -256,6 +258,80 @@ describe('Test: collection controller', () => {
...
+      };
+
+      kuzzle.validation.isValidSpecification.resolves({isValid: true});
+      kuzzle.validation.curateSpecification.resolves();
+
+      return collectionController.updateSpecifications(request)
+        .then(response => {
+          should(kuzzle.internalEngine.refresh).be.calledOnce();
+          should(kuzzle.validation.curateSpecification).be.called();
+          should(kuzzle.internalEngine.createOrReplace).be.calledOnce();
```
Aschen: You can chain assertions on stubs

test/api/controllers/collectionController.test.js
```diff
@@ -295,6 +371,78 @@ describe('Test: collection controller', () => {
...
+      CollectionController.__set__({
+        createSpecificationList: sinon.stub().resolves({
+          _id: 'indexcollection',
+          _source: {
+            validation: 'validation',
+            index: 'index',
+            collection: 'collection'
+          }
+        }),
+        validateSpecificationList: sinon.stub().resolves({valid: true})
```
Aschen: You should test that this stub is called with the good arguments. (Which is not the case here)

test/api/controllers/collectionController.test.js
```diff
@@ -295,6 +371,78 @@ describe('Test: collection controller', () => {
...
+      CollectionController.__set__({
+        createSpecificationList: sinon.stub().resolves({
+          _id: 'indexcollection',
+          _source: {
+            validation: 'validation',
+            index: 'index',
+            collection: 'collection'
+          }
+        }),
+        validateSpecificationList: sinon.stub().resolves(errorResponse)
```
Aschen: Same comment about testing the stub call

test/api/controllers/collectionController.test.js
```diff
@@ -256,6 +258,80 @@ describe('Test: collection controller', () => {
...
+          should(kuzzle.internalEngine.createOrReplace).not.be.called();
+
+          should(error).be.an.instanceOf(BadRequestError);
+          should(error.message).be.exactly('Some errors with provided specifications.');
+          should(error.details).match([ 'bad bad is a bad type !' ]);
+        });
+    });
+
+    it('should create or replace specifications when requested with deprecated way', () => {
+      request.input.resource.index = undefined;
```
Aschen: You could DRY this by putting it in a `beforeEach()` hook

lib/config/httpRoutes.js
```diff
@@ -153,7 +153,10 @@ module.exports = [
   {verb: 'post', url: '/credentials/:strategy/_me/_create', controller: 'auth', action: 'createMyCredentials'},
   {verb: 'post', url: '/credentials/:strategy/_me/_validate', controller: 'auth', action: 'validateMyCredentials'},
 
+  /* DEPRECATED - will be removed in v2 */
```
scottinet: (open to debate)

since this deprecation notice is introduced so late in the v1, I wonder if this shouldn't stay for the next major version
Usually, you deprecated something during a major, it stays in the next one, and (sometimes) removed after that

Aschen: Since there is only very few people that use the validation module I think we can remove this route.
If we setup a `compat` configuration option (cf [trello](https://trello.com/c/WAcB9MHT/74-adm-ajout-dune-partie-compat-dans-la-config-de-kuzzle-pour-pouvoir-d%C3%A9sactiver-des-fonctionnalit%C3%A9s-deprecated)) users will be able to keep deprecated route if they want but IMHO it's better to push users to embrace change

PR #1302, opened 2019-05-20
Add write and mWrite

Aschen:
## What does this PR do ?

(Open discussion)

This PR add two method for the bulk controller that behave like `createOrReplace` and `mCreateOrReplace`:
  - `write`: write a document without adding metadata or performing data validation
  - `mWrite`: write several documents without adding metadata or performing data validation

The name is open to debate, I have thought of `put` and `mPut` also

The goal is to have a low level method that allow to arbitrary write document into the storage layer to bypass automatic metadata or the data validation layer.  

It's better to have these two endpoints than an option like `bypassMeta` on existing methods because developer can choose to not expose these methods to users.

  - Step 1 :
  - Step 2 :
  - Step 3 :  
  ...

### Other changes

  - Clone kuzzle configuration before passing it in the plugin context

### DIFF
--- a/lib/api/controllers/bulkController.js
+++ b/lib/api/controllers/bulkController.js
@@ -83,7 +83,6 @@ class BulkController extends BaseController {
       .then(response => {
         this.kuzzle.indexCache.add(request.input.resource.index, request.input.resource.collection);
 
-
         if (notify && response.created) {
           this.kuzzle.notifier.notifyDocumentCreate(request, response);
         }
--- a/lib/api/controllers/bulkController.js
+++ b/lib/api/controllers/bulkController.js
@@ -78,7 +78,7 @@ class BulkController extends BaseController {
     assertIdStartsNotUnderscore(request);
 
     const notify = this.tryGetBoolean(request, 'args.notify');
-    console.log(request)
+
     return this.engine.createOrReplace(request, false)
       .then(response => {
         this.kuzzle.indexCache.add(request.input.resource.index, request.input.resource.collection);
--- a/lib/api/controllers/bulkController.js
+++ b/lib/api/controllers/bulkController.js
@@ -77,7 +77,7 @@ class BulkController extends BaseController {
     assertHasIndexAndCollection(request);
     assertIdStartsNotUnderscore(request);
 
-    const notify = this.ensureBooleanFlag(request, 'input.args.notify');
+    const notify = this.tryGetBoolean(request, 'args.notify');
 
     return this.engine.createOrReplace(request, false)
       .then(response => {
@@ -106,7 +106,7 @@ class BulkController extends BaseController {
     assertBodyAttributeType(request, 'documents', 'array');
     assertHasIndexAndCollection(request);
 
-    const notify = this.ensureBooleanFlag(request, 'input.args.notify');
+    const notify = this.tryGetBoolean(request, 'args.notify');
 
     return this.engine.mcreateOrReplace(request, false)
       .then(response => {
--- a/lib/api/controllers/bulkController.js
+++ b/lib/api/controllers/bulkController.js
@@ -118,7 +118,7 @@ class BulkController extends BaseController {
           this.kuzzle.notifier.notifyDocumentMChanges(request, response.result, true);
         }
 
-        return {hits: response.result, total: response.result.length};
+        return { hits: response.result, total: response.result.length };
       });
   }
 
--- a/lib/api/controllers/bulkController.js
+++ b/lib/api/controllers/bulkController.js
@@ -78,7 +78,7 @@ class BulkController extends BaseController {
     assertIdStartsNotUnderscore(request);
 
     const notify = this.tryGetBoolean(request, 'args.notify');
-
+    console.log(request)
     return this.engine.createOrReplace(request, false)
       .then(response => {
         this.kuzzle.indexCache.add(request.input.resource.index, request.input.resource.collection);
--- a/lib/api/controllers/bulkController.js
+++ b/lib/api/controllers/bulkController.js
@@ -23,7 +23,7 @@
 
 const
   BaseController = require('./controller'),
-  { PartialError } = require('kuzzle-common-objects').errors,
+  { errors: { PartialError } } = require('kuzzle-common-objects'),
   {
     assertHasBody,
     assertBodyHasAttribute,
--- a/lib/api/controllers/bulkController.js
+++ b/lib/api/controllers/bulkController.js
@@ -79,7 +79,7 @@ class BulkController extends BaseController {
 
     const notify = this.ensureBooleanFlag(request, 'input.args.notify');
 
-    return this.engine.writeDocument(request)
+    return this.engine.createOrReplace(request, false)
       .then(response => {
         this.kuzzle.indexCache.add(request.input.resource.index, request.input.resource.collection);
 
@@ -109,7 +109,7 @@ class BulkController extends BaseController {
 
     const notify = this.ensureBooleanFlag(request, 'input.args.notify');
 
-    return this.engine.mwriteDocuments(request)
+    return this.engine.mcreateOrReplace(request, false)
       .then(response => {
         if (response.error.length > 0) {
           request.setError(new PartialError('Some document creations failed', response.error));
--- a/lib/api/controllers/bulkController.js
+++ b/lib/api/controllers/bulkController.js
@@ -23,8 +23,14 @@
 
 const
   BaseController = require('./controller'),
-  {errors: {PartialError}} = require('kuzzle-common-objects'),
-  {assertHasBody, assertBodyHasAttribute} = require('../../util/requestAssertions');
+  { PartialError } = require('kuzzle-common-objects').errors,
+  {
+    assertHasBody,
+    assertBodyHasAttribute,
+    assertHasIndexAndCollection,
+    assertBodyAttributeType,
+    assertIdStartsNotUnderscore
+  } = require('../../util/requestAssertions');
 
 /**
  * @class BulkController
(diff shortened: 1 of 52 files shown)


scottinet:
@Aschen > wouldn't the `bulk` controller be more suited for such tasks, and easier to configure regarding rights management?
And in that regard, I would make real-time notifications optional (and off by default).

What do you think?

Aschen:
@scottinet Yes I agree, it's more logical to put this kind of low-level operation together!  

I also agree for the notifications, I needed them for my usecase but again since these are low-level method It's better to let the developer choose what he want to do.

lib/services/elasticsearch.js
```diff
@@ -1109,6 +1137,35 @@ class ElasticSearch extends Service {
...
+   * @return {Promise}
+   */
+  mwriteDocuments(request) {
+    const
+      esRequest = initESRequest(request, ['consistency', 'refresh', 'timeout', 'fields']),
+      extracted = extractMDocuments(request, {});
+
+    esRequest.body = [];
+
+    let i; // NOSONAR
```
thomasarbona: Why don't you declare `i` in the `for` loop ?

lib/services/elasticsearch.js
```diff
@@ -1109,6 +1137,35 @@ class ElasticSearch extends Service {
...
+    for (i = 0; i < extracted.documents.length; i++) {
+      esRequest.body.push({
+        index: {
+          _index: request.input.resource.index,
+          _type: request.input.resource.collection,
+          _id: extracted.documents[i]._id
+        }
+      });
+      esRequest.body.push(extracted.documents[i]._source);
```
thomasarbona: Why not call `push` once ?

lib/api/controllers/bulkController.js
```diff
@@ -55,6 +66,63 @@ class BulkController extends BaseController {
...
+
+    return this.engine.writeDocument(request)
+      .then(response => {
+        this.kuzzle.indexCache.add(request.input.resource.index, request.input.resource.collection);
+
+
+        if (notify && response.created) {
+          this.kuzzle.notifier.notifyDocumentCreate(request, response);
+        }
+        else if (notify) {
```
thomasarbona: Isn't this condition redundant ?

Aschen: No because I want to notify only if the user specified that he want realtime notifications

lib/services/elasticsearch.js
Aschen: It's a performance issue on Node.js 6 if you didn't declare the let outside the loop (see [this](https://stackoverflow.com/questions/37792934/why-is-let-slower-than-var-in-a-for-loop-in-nodejs) )
Since this method will potentially handle thousands of documents it's better to use this tricks

lib/services/elasticsearch.js
Aschen: Elasticsearch bulk syntax need the element N to be some kind of header and the element N+1 is actually the document body. 
https://doc.kuzzle.io/api/1/controller-bulk/import/

lib/services/elasticsearch.js
```diff
@@ -1109,6 +1137,35 @@ class ElasticSearch extends Service {
...
+   */
+  mwriteDocuments(request) {
+    const
+      esRequest = initESRequest(request, ['consistency', 'refresh', 'timeout', 'fields']),
+      extracted = extractMDocuments(request, {});
+
+    esRequest.body = [];
+
+    let i; // NOSONAR
+    for (i = 0; i < extracted.documents.length; i++) {
```
Yoann-Abbes: Just because I can't find another thing to say :
Feel weird to read `i++`, maybe you should always use the pre-increment syntax `++i` even if in javascript there is apparently no difference compare to C/C++/Java, I think this is good habit.

scottinet: Now I'm curious: care to explain why tihs should be a good practice?
I have so many examples right now where, unless you're code golfing, the only good habit with that operator is to make sure that its use is isolated...

thomasarbona: For me the good practice is to always use `+=` instead of `++` because it standardizes the code (all the incrementation are the same).
In the `++` case, it can make potential errors when you don't know the difference (but I think everyone knows the difference here) between pre and post incrementation and you could make code like this:
```javascript
const y = x++;
```

Aschen: There is almost no difference in Node.js and when I'm using iterator in C++ I didn't forget ;)
<details><summary>Benchmarks</summary>
Node.js 6
```
i++ x 4,154 ops/sec ±0.39% (89 runs sampled)
++i x 4,934 ops/sec ±0.29% (96 runs sampled)
Fastest is ++i
```
Node.js 8
```
i++ x 11,567 ops/sec ±0.14% (97 runs sampled)
++i x 11,493 ops/sec ±0.33% (96 runs sampled)
Fastest is i++
```
</details>

Yoann-Abbes: @scottinet I think its a good practice because if you write it always the same way you won't forget it when you are with low languages where it has its importance. That is just an opinion :)

lib/api/controllers/bulkController.js
Aschen: see on slack, it's redundancy vs imbrication

lib/services/elasticsearch.js
Aschen: Actually, it's faster to call 2 time `Array.push` with one element than 1 time with 2 elements:

<details>
<pre>
.add('2 push', () => {
  const tab = []
  for (let i = 0; i < 10000; ++i) {
    tab.push(elem);
    tab.push(elem2);
  }
})
.add('1 push', () => {
  const tab = [];
  for (let i = 0; i < 10000; ++i) {
    tab.push(elem, elem2)
  }
})
</pre>
<pre>
1 push x 5,475 ops/sec ±2.01% (91 runs sampled)
2 push x 10,486 ops/sec ±0.49% (96 runs sampled)
Fastest is 2 push
</pre>
</details>

lib/services/elasticsearch.js
```diff
@@ -339,13 +339,27 @@ class ElasticSearch extends Service {
...
+    if (injectKuzzleMeta) {
+      kuzzleMeta = {
+        author: userId,
+        createdAt: Date.now(),
+        updatedAt: Date.now(),
+        updater: userId,
+        active: true,
+        deletedAt: null
+      };
+    };
```
kuzzle: MINOR Remove this empty statement. [rule](https://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS1116)

lib/services/elasticsearch.js
```diff
@@ -339,13 +339,27 @@ class ElasticSearch extends Service {
    * Create a new document to ElasticSearch, or replace it if it already exist
    *
    * @param {Request} request
+   * @param {boolean} injectKuzzleMeta
    * @returns {Promise} resolve an object that contains _id
    */
-  createOrReplace(request) {
+  createOrReplace(request, injectKuzzleMeta = true) {
+    let kuzzleMeta = undefined;
```
kuzzle: MINOR Use null instead. [rule](https://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS2138)

kuzzle:
* ![MINOR][MINOR] 2 minor

lib/services/elasticsearch.js
scottinet: What is this, I don't even

First, it's very unlikely that there is a difference in performances between the pre- and post-increment operators, whatever the language. And even if there is a slight difference (again, very unlikely), it would be premature optimization to bother about it.

Second, there is nothing to remember but this one rule: do not ever mix pre-/post- increment operators with anything else in the same line to prevent unpredictability.
For instance: `foo(i, ++i)` => what are the values passed to "foo"? 

Now that's out of the way: writing the same thing always the same way doesn't seem a bad idea to me, so I can hear the consistency argument, but it's a pretty minor one to me regarding that operator, since, well, it should always be alone.

Yoann-Abbes: I was talking about loops only for the "pre-increment reflex", maybe I should have precised to avoid you writing an entire paragraph :p

Aschen: Guys please leave my PR alone :see_no_evil:

Aschen:
@scottinet @thomasarbona @Yoann-Abbes You might want to review this PR, I added the tests

lib/api/controllers/bulkController.js
```diff
@@ -23,16 +23,27 @@
 
 const
   BaseController = require('./controller'),
-  {errors: {PartialError}} = require('kuzzle-common-objects'),
-  {assertHasBody, assertBodyHasAttribute} = require('../../util/requestAssertions');
+  { PartialError } = require('kuzzle-common-objects').errors,
```
scottinet: (nitpicking)

```suggestion
  { errors: { PartialError } } = require('kuzzle-common-objects'),
```

lib/api/controllers/controller.js
```diff
@@ -82,8 +82,11 @@ class BaseController {
       const booleanValue = flagValue !== undefined ? true : false;
 
       _.set(request, flagPath, booleanValue);
+
+      return booleanValue;
     }
 
+    return Boolean(flagValue);
```
scottinet: this is not `ensureBooleanFlag` anymore, since you're now retrieving the value

another name must now be found, for instance `tryGetBoolean`?

also, the jsdoc is now outdated

lib/api/controllers/bulkController.js
```diff
@@ -55,6 +66,62 @@ class BulkController extends BaseController {
...
+   * @param {Request} request
+   * @returns {Promise<Object>}
+   */
+  write(request) {
+    assertHasBody(request);
+    assertHasIndexAndCollection(request);
+    assertIdStartsNotUnderscore(request);
+
+    const notify = this.tryGetBoolean(request, 'args.notify');
+    console.log(request)
```
kuzzle: MINOR Add a semicolon at the end of this statement. [rule](https://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3ASemicolon)

kuzzle:
* ![MINOR][MINOR] 1 minor

PR #1307, opened 2019-05-23
Hotfix: correctly compute mUpdate notifications against the updated document

scottinet:
# Description

When doing a `document:mUpdate`, real-time notifications are erroneously computed on the update query content, instead of the full updated document.

This is because for mUpdate we use Elasticsearch's `source` option, asking ES to return the updated document source, saving us from performing a `get` query for each updated document. But the returned source doesn't replace the `_source` property, which still contains only the update content. Instead, the full document source is put in `get._source`.

This PR correctly replaces _source with get._source before returning the result to the document controller.

# How to test

* Subscribe with the following filters: `{exists: 'foo'}`
* Create a document with the following body: `{foo: true, bar: 'baz'}`
* Use mUpdate to update it, **without** changing `foo`. Body example: `documents: [{_id: <documentId>, body: {bar: 'qux'}}]`

Without this fix, an `out` notification is sent, instead of an `in` one, because the notifier module thinks that the `foo` field has been removed and does not exist anymore.

### DIFF
--- a/lib/services/elasticsearch.js
+++ b/lib/services/elasticsearch.js
@@ -1153,7 +1153,21 @@ class ElasticSearch extends Service {
       }
     }
 
-    return this._mexecute(esRequest, toImport, extracted.rejected);
+    return this._mexecute(esRequest, toImport, extracted.rejected)
+      .then(response => {
+        // with _source: true, ES returns the updated document in
+        // response.result.get._source
+        // => we replace response.result._source with it so that the notifier
+        // module can seamlessly process all kind of m* response*
+        let j; // NOSONAR
+        for (j = 0; j < response.result.length; j++) {
+          if (response.result[j].get && response.result[j].get._source) {
+            response.result[j]._source = response.result[j].get._source;
+          }
+        }
+
+        return response;
+      });
   }
 
   /**
--- a/lib/services/elasticsearch.js
+++ b/lib/services/elasticsearch.js
@@ -1159,7 +1159,10 @@ class ElasticSearch extends Service {
         // response.result.get._source
         // => we replace response.result._source with it so that the notifier
         // module can seamlessly process all kind of m* response*
-        for (const result of response.result) {
+        let j; // NOSONAR
+        for (j = 0; j < response.result.length; j++) {
+          const result = response.result[j];
+
           if (result.get && result.get._source) {
             result._source = result.get._source;
           }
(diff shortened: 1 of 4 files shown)


lib/services/elasticsearch.js
```diff
@@ -1153,7 +1153,21 @@ class ElasticSearch extends Service {
...
-    return this._mexecute(esRequest, toImport, extracted.rejected);
+    return this._mexecute(esRequest, toImport, extracted.rejected)
+      .then(response => {
+        // with _source: true, ES returns the updated document in
+        // response.result.get._source
+        // => we replace response.result._source with it so that the notifier
+        // module can seamlessly process all kind of m* response*
+        let j; // NOSONAR
+        for (j = 0; j < response.result.length; j++) {
+          if (response.result[j].get && response.result[j].get._source) {
```
thomasarbona: It's a bit hard to read, why not use a variable ? `const res = response.result[j];`

scottinet: good point

scottinet: should be easier to read now

lib/services/elasticsearch.js
```diff
@@ -1153,7 +1153,20 @@ class ElasticSearch extends Service {
...
     }
 
-    return this._mexecute(esRequest, toImport, extracted.rejected);
+    return this._mexecute(esRequest, toImport, extracted.rejected)
+      .then(response => {
+        // with _source: true, ES returns the updated document in
+        // response.result.get._source
+        // => we replace response.result._source with it so that the notifier
+        // module can seamlessly process all kind of m* response*
+        for (const result of response.result) {
```
alexandrebouthinon: I think @thomasarbona was talking about doing this:
```suggestion
        let j; // NOSONAR
        const res = response.result;
        for (j = 0; j < res.length; j++) {
          if (res[j].get && res[j].get._source) {
            res[j]._source = res[j].get._source;
```
You sacrificed Node.js 6 performance for the sake of readability when you remove this:
```js
        let j; // NOSONAR
        for (j = 0; j < response.result.length; j++) {
          if (response.result[j].get && response.result[j].get._source) {
            response.result[j]._source = response.result[j].get._source;
```

scottinet: last time I checked, `for..of` runs over 2.5M ops/s which should be largely enough here

scottinet: ok I was a bit dishonest, m* routes will probably have to handle millions of documents so we should use the fastest loops available...

PR #1310, opened 2019-05-27
Fix CLI login, credentials not sent

### LINKED ISSUE #1309: CLI commands don't seem to use credentials
## Expected Behavior
Commands such as resetDatabase should work when the proper password and username are provided, eg in our case 
```bash
./bin/kuzzle -U god -P god resetDatabase --noint
```
should work, and it did in versions 1.7.0 (and I think in version 1.7.1).

## Current Behavior
Instead of working as it used to, now we get the error: 
```
{ [Error: Forbidden action [null/null/admin/resetDatabase] for user -1] status: 403 }
```
Which appears to mean that the credentials are not used and that kuzzle attempts to do it as the anonymous user, which of course we haven't allowed.
What makes me think this is the problem is that when doing it through the API (HTTP) it works fine, with the same credentials.

## Possible Solution
Correct a bug where the credentials provided in the CLI are not used I guess ?

## Steps to Reproduce
Explained above

## Context (Environment)

Kuzzle version: 1.7.4
Node.js version: 11.10.0, but same problem with other versions
SDK version: NA
I should add that this was reproduced on different workstations, but all of them running on windows 10.

Aschen:
## What does this PR do ?

Fix #1309 

The credentials wasn't send to kuzzle, so the CLI was always logged as anonymous user. 

  - Step 1 : Create an admin user and restrict anonymous rights
  - Step 2 : Run ` ./bin/kuzzle -U admin -P admin resetDatabase --noint`

## Other changes

 - update minimal node.js version in package.json (https://github.com/kuzzleio/kuzzle-containers/pull/49)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "kuzzle",
   "author": "The Kuzzle Team <support@kuzzle.io>",
-  "version": "1.7.4",
+  "version": "1.7.5",
   "description": "Kuzzle is an open-source solution that handles all the data management through a secured API, with a large choice of protocols.",
   "main": "./lib/index.js",
   "bin": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "kuzzle",
   "author": "The Kuzzle Team <support@kuzzle.io>",
-  "version": "1.7.5",
+  "version": "1.7.4",
   "description": "Kuzzle is an open-source solution that handles all the data management through a secured API, with a large choice of protocols.",
   "main": "./lib/index.js",
   "bin": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "kuzzle",
   "author": "The Kuzzle Team <support@kuzzle.io>",
-  "version": "1.7.4",
+  "version": "1.7.5",
```
scottinet: you have to update the version stored in `package-lock.json` too, otherwise it's ignored

instead of updating the version number manually, you should have used `npm version [major|minor|patch]`

alexandrebouthinon: You can also generate a new `package-lock.json` file by removing the old one and run `npm install`

PR #1311, opened 2019-05-28
Fix promise leaks

scottinet:
# Description

The functions exposed by the `notifier` core generate many promises that are not awaited, and whose rejections might not be handled. And this is bad practice.

But more than a bad practice, this is also a real issue because, well... hm... I'll let Node.js do the explaining:

```
(node:32263) UnhandledPromiseRejectionWarning: Unhandled promise rejection. 
  This error originated either by throwing inside of an async function without a 
  catch block, or by rejecting a promise which was not handled with .catch(). 
  (rejection id: 1)
(node:32263) [DEP0018] DeprecationWarning: Unhandled promise rejections are 
  deprecated. In the future, promise rejections that are not handled will 
  terminate the Node.js process with a non-zero exit code.
```

That kind of warning doesn't look nice in Kuzzle's logs, not mentioning the potential issue this might cause if Node.js starts crashing on unhandled promise rejections.

# How to test

Check the unit tests logs: no more warnings :tada:


lib/api/controllers/documentController.js
```diff
@@ -182,15 +181,16 @@ class DocumentController extends BaseController {
...
+          modifiedRequest, 'unknown', 'pending');
       })
-      .then(response => {
-        this.kuzzle.notifier.notifyDocumentCreate(modifiedRequest, response);
-
-        return response;
-      });
+      .then(() => this.engine.create(modifiedRequest))
+      .then(_response => {
+        response = _response;
```
Aschen: Why duplicate the response here ?

scottinet: because it's needed further on in the promises chain

Aschen: Oh yes sorry, I must be tired -_-

lib/api/controllers/documentController.js
```diff
@@ -224,22 +223,24 @@ class DocumentController extends BaseController {
...
+        return this.kuzzle.notifier.publish(
+          modifiedRequest, 'unknown', 'pending');
       })
-      .then(response => {
-        this.kuzzle.indexCache.add(modifiedRequest.input.resource.index, modifiedRequest.input.resource.collection);
-
+      .then(() => this.engine.createOrReplace(modifiedRequest))
+      .then(_response => {
+        response = _response;
+        return this.kuzzle.indexCache.add(
```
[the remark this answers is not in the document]
scottinet: I didn't pay attention to that line, since it was already there. But now that you mention it, I can't find why this would be useful.

That line was added by a [3-years old PR](https://github.com/kuzzleio/kuzzle/pull/137/files#diff-d1355b0c631259c190837285da4f9880R41) and my best guess is that, at that time, it was useful to put in cache the indexes/collections created directly in ES while Kuzzle was running to prevent false negatives.

This is not useful anymore since the index cache now features a `hotreload` mechanism.

So I'll remove it. Good catch :+1:

scottinet:
@benoitvidis > the first issue is not that straightforward: what kind of performances are we measuring? 
This might impact the response times of API requests where the real-time engine is involved, because now Kuzzle responds only when notifications are actually computed and submitted to the network layer.
But in the other hand, that means that we free up a request slot only after all promises are settled. And this might solve a backpressure problem, where promises generated by the real-time engine can pile up in the event loop because we don't wait before processing another request from the buffer.
So, my take on this is that: 
1. it might have a slight negative impact on API response times where real-time is involved
2. it might have a slight positive impact on notifications delay
3. it might prevent Kuzzle's event loop from being saturated

I'm using the conditional here, because in fact, I couldn't see any impact when I ran our benchmarks. So these considerations are theoretical, as far as I can see at least, and that's probably because the real-time engine is a lot faster than, say, checking user rights.

About the second issue: I'm not sure, but I think that we probably will have new errors appearing once the new Kuzzle version is released, because the warnings we got came from unhandled promise rejections. 
So, after this PR, those errors should still be unhandled I guess, but at least they should now be properly logged or returned to requesting users, so we should now be able to fix those.

scottinet:
Re-reading my answer, I agree that there are unknows with this PR. Do you know how we could be a bit more sure of the outcomes?
On my side, I'll try to see if I can think of something...

PR #1314, opened 2019-06-03
Remove _id from bulk:write 2e route

Aschen:
## What does this PR do ?

Remove the second route for `bulk:write`. The _id is now a query string parameter and not part of the url

  - Step 1 :
  - Step 2 :
  - Step 3 :  
  ...


scottinet:
@benoitvidis > this started from this comment in the corresponding documentation PR: https://github.com/kuzzleio/documentation/pull/300#discussion_r289846420

The routes you mention are old, and created prior to our own HTTP router. Passing a document ID in the querystring is now free, and having multiple URLs for a same route doesn't seem clean to me.

Thinking of that, I wonder if perhaps we should deprecated the duplicated routes for old API actions :thinking:

Aschen:
@scottinet Yes I think we should deprecate other duplicated routes

PR #1315, opened 2019-06-04
Add the new Vault module to handle encrypted application secrets

Aschen:
## What does this PR do ?

Add an integrated Vault to securely store secrets (API key, credentials, ...) in project repositories.  
This allows developers to encrypt a secrets file and then commit the encrypted version.  
Then Kuzzle is able to decrypt the secrets file and expose it to plugin developers.

Secrets file are JSON file, only the string values are encrypted:
```
{
  "aws": {
    "key": "silmaril"
  },
  "secret": "ring"
}
```

You can use the CLI to encrypt the file, the secret key can be either passed as an argument or set as the `KUZZLE_VAULT_KEY` environment variable:
```
bin/kuzzle encryptSecrets config/secrets.json --vault-key verylongandstrongkey
```

Then a file named `config/secrets.enc.json` is generated:
```
{
  "aws": {
    "key": "536553f3181ada6f700cac98100f1266.f700cac98100f1266"
  },
  "secret": "3185620ca7e490a8d9dbf79c3e78c1a3.8d9dbf79c3e78c1a3"
}
```

Kuzzle will automatically try to decrypt this file if present, or you can specify another one using `--secrets-file <file>` with the `bin/kuzzle start` command.

Then the secrets will be exposed to plugins developers under `context.secrets`.

Doc https://github.com/kuzzleio/documentation/pull/321

  - Step 1 : Encrypt a secrets file: `./bin/kuzzle encryptSecrets config/secrets.json --vault-key azerty`
  - Step 2 : Decrypt it: `./bin/kuzzle decryptSecrets config/secrets..enc.json --vault-key azerty`
  - Step 3 : Run Kuzzle with it `KUZZLE_VAULT_KEY=azerty docker-compose -f docker-compose/dev.yml up`
 - Step 4 : Access the secrets inside a plugin `init` method


kuzzle:

Note: The following issues were found on lines that were not modified in the pull request. Because these issues can't be reported as line comments, they are summarized here:

1. ![MINOR][MINOR] [vault.js#](https://github.com/kuzzleio/kuzzle/blob/0c4b04ec81726676174ed46090babdc7b628ff31/lib/api/core/vault.js#): Add a new line at the end of this file. [rule](https://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AMissingNewlineAtEndOfFile)

bin/commands/decryptSecrets.js
```diff
@@ -0,0 +1,79 @@
...
+    console.log(cout.error('[ℹ] You must provide the vault key with --vault-key <key> or in KUZZLE_VAULT_KEY environment variable'));
+    process.exit(1);
+  }
+
+  let outputFile = options.outputFile;
+  if (!options.outputFile) {
+    outputFile = `${encryptedSecretsFile.split('.enc')[0]}.json`;
+  }
+
+  console.log(cout.warn(`[ℹ] You are going to overwrite the following file: ${outputFile}`));
```
scottinet: this log line and the question following it should only be shown if the output file already exists

bin/commands/decryptSecrets.js
```diff
@@ -0,0 +1,79 @@
...
+
+  if (!options.noint) {
+    userIsSure = readlineSync.question('[❓] Are you sure? If so, please type "I am sure": ') === 'I am sure';
```
scottinet: `userIsSure` was not declared before

bin/commands/decryptSecrets.js
```diff
@@ -0,0 +1,79 @@
...
+  const cout = new ColorOutput(options);
+
+  if (!options.vaultKey && !process.env.KUZZLE_VAULT_KEY) {
+    console.log(cout.error('[ℹ] You must provide the vault key with --vault-key <key> or in KUZZLE_VAULT_KEY environment variable'));
+    process.exit(1);
+  }
+
+  let outputFile = options.outputFile;
+  if (!options.outputFile) {
+    outputFile = `${encryptedSecretsFile.split('.enc')[0]}.json`;
```
scottinet: what happens if I store my encrypted configuration in the following file: `/etc/kuzzle.encrypted/config.enc.json`?

You might want to use [path.basename](https://devdocs.io/node~10_lts/path#path_path_basename_path_ext) instead

bin/commands/encryptSecrets.js
```diff
@@ -0,0 +1,63 @@
...
+  const cout = new ColorOutput(options);
+
+  if (!options.vaultKey && !process.env.KUZZLE_VAULT_KEY) {
+    console.log(cout.error('[ℹ] You must provide the vault key with --vault-key <key> or in KUZZLE_VAULT_KEY environment variable'));
+    process.exit(1);
+  }
+
+  let outputFile = options.outputFile;
+  if (!options.outputFile) {
+    outputFile = `${secretsFile.split('.json')[0]}.enc.json`;
```
scottinet: same here: use path.basename

kuzzle:

Note: The following issues were found on lines that were not modified in the pull request. Because these issues can't be reported as line comments, they are summarized here:

1. ![MINOR][MINOR] [vault.js#](https://github.com/kuzzleio/kuzzle/blob/ece90616deb30b22313749a8767a60ad66ca9d78/lib/api/core/vault.js#): Add a new line at the end of this file. [rule](https://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AMissingNewlineAtEndOfFile)

default.config.js
```diff
@@ -155,6 +155,7 @@ module.exports = {
   },
 
   server: {
+    vaultSeed: '764c1afa2988981a',
```
scottinet: I'm not at ease with that.
Instead, I think you should do as with the JWT secret key: if it's not null in the configuration file, then use that one. Otherwise, create a fresh new key and store it in the database.

You can use the JWT key generator as an example: https://github.com/kuzzleio/kuzzle/blob/master/lib/services/internalEngine/bootstrap.js#L225

lib/api/core/vault.js
```diff
@@ -0,0 +1,218 @@
...
+    }
+    else {
+      return;
+    }
+
+    if (this.config.server.vaultSeed === '764c1afa2988981a') {
+      console.warn(
+        'You are using the default vault seed.\n',
+        'It\' STRONGLY advised to change this value.\n',
+        'Run the following commande to generate a new seed:\n',
```
scottinet:
```suggestion
        'Run the following command to generate a new seed:\n',
```

lib/api/core/vault.js
```diff
@@ -0,0 +1,218 @@
...
+      this.vaultKey = process.env.KUZZLE_VAULT_KEY;
+    }
+    else {
+      return;
+    }
+
+    if (this.config.server.vaultSeed === '764c1afa2988981a') {
+      console.warn(
+        'You are using the default vault seed.\n',
+        'It\' STRONGLY advised to change this value.\n',
```
scottinet:
```suggestion
        'It\'s STRONGLY advised to change this value.\n',
```

lib/api/core/vault.js
```diff
@@ -0,0 +1,218 @@
...
+        'Run the following commande to generate a new seed:\n',
+        '\tbin/kuzzle vaultSeed\n'
+      );
+    }
+
+    this.vaultKeyHash = crypto.createHash('md5')
```
scottinet: Never use md5, it's broken. If you need a cryptographic hash, use SHA512 (BLAKE2b is not supported in Node.js 6)

Why do we need to hash the provided key anyway? Is this to prevent storing the key in memory?

lib/api/core/vault.js
```diff
@@ -0,0 +1,218 @@
...
+      .toUpperCase();
+
+    this.cipherIV = this.config.server.vaultSeed.substr(0, 16);
+  }
+
+  /**
+   * Init the vault with a decryption key and a file containing
+   * the secrets, decrypt it and store decrypted secrets
+   *
+   * Throws an error if:
```
scottinet:
```suggestion
   * Rejects with an error if:
```

default.config.js
Aschen: That what I wanted to do in the first time but I also have to use the Vault in the CLI, without access to internal database and without generation of a key like for the JWT secret.  

Moreover the generated key changes when you restart Kuzzle (with docker-compose down) or wipe the internal database and the secrets can not be decrypted any more since the IV is not the same.   

This is why the initialization vector must be set in the configuration file and not auto generated. (I added warning messages to indicate developers to do so and I will also insist on that point in the documentation)

lib/api/core/vault.js
Aschen: In fact I does not need a cryptographic hash here, I need a key of at least 128 bits and must be a multiple of 32 bits. I only use md5 to ensure a correct size for the key but you're right, I can also use sha512

default.config.js
scottinet: About your 1st point: you can either use the CLI as an API client as with the other commands, or maybe you can bootstrap our internal engine client?

Your second point is moot, since:
1. the proposed solution still allows setting a key in the configuration files
2. it's a good thing that, after you destroy completely a kuzzle instance, a fresh one is unable to decrypt the vault without the right seed. That's what the vault is for, isn't it?

lib/api/core/vault.js
scottinet: ah ok, if that's the only purpose then we don't care

bin/commands/decryptSecrets.js
Aschen: Strange that eslint didn't notice this..

bin/commands/decryptSecrets.js
```diff
@@ -0,0 +1,79 @@
...
+  try {
+    const encryptedSecrets = JSON.parse(fs.readFileSync(encryptedSecretsFile, 'utf-8'));
+
+    const secrets = vault.decryptObject(encryptedSecrets);
+
+    fs.writeFileSync(outputFile, JSON.stringify(secrets, null, 2));
+
+    console.log(cout.ok(`[✔] Secrets successfully encrypted: ${outputFile}`));
+  } catch (error) {
+    console.log(cout.error(`[ℹ] Can not encrypt secret file: ${error.message}`));
```
Yoann-Abbes:
```suggestion
    console.error(cout.error(`[ℹ] Can not encrypt secret file: ${error.message}`));
```

bin/commands/encryptSecrets.js
```diff
@@ -0,0 +1,63 @@
...
+  try {
+    const secrets = JSON.parse(fs.readFileSync(secretsFile, 'utf-8'));
+
+    const encryptedSecrets = vault.encryptObject(secrets);
+
+    fs.writeFileSync(outputFile, JSON.stringify(encryptedSecrets, null, 2));
+
+    console.log(cout.ok(`[✔] Secrets successfully encrypted: ${outputFile}`));
+  } catch (error) {
+    console.log(cout.error(`[ℹ] Can not encrypt secret file: ${error.message}`));
```
Yoann-Abbes:
```suggestion
    console.error(cout.error(`[ℹ] Can not encrypt secret file: ${error.message}`));
```

default.config.js
Aschen: You propose to add methods to the admin controller ? I thought about it but I feel it overkill.  

About my second point, you're right on 1) and 2). But I was thinking that maybe the better solution is not to generate the key ourself and let the developers manually set it in the kuzzlerc file. 
If they does not put a value to `server.vaultKey`, the Vault is just deactivated

lib/api/core/vault.js
```diff
@@ -0,0 +1,218 @@
...
+   *  - a key is found but no file
+   *  - a file is found but no key
+   *
+   * @param {string|null} secretsFile - file containing the encrypted secrets
+   * @returns {Promise}
+   */
+  init (secretsFile = null) {
+    this.encryptedSecretsFile = secretsFile || defaultSecretsFile;
+
+    const fileExists = fs.existsSync(this.encryptedSecretsFile);
```
Yoann-Abbes: Do you check permissions on the file somewhere ?

default.config.js
scottinet: I agree about how you treated the key, and especially about deactivating the vault if no key was set. :+1: 
I was of course only talking about the seed.

Aschen: Hmm I think I misuse some words..

I was talking about the seed, deactivate the Vault if the developers does not have set a value for `server.vaultSeed`

lib/api/core/vault.js
Aschen: Node.js will throw a `[Errno 13] Permission denied` if the permissions are not sufficient

Yoann-Abbes: oh ok I didn't know about Errno in js

default.config.js
scottinet: I think that only the presence of a key should tell Kuzzle to (de)activate the vault. The seed in itself has no meaning, it can only concomitantly to a key.

But what we can do is to ignore any seed value if there isn't a key, and only auto-generate a seed if a key is provided without a seed.

lib/api/core/vault.js
Aschen: Actually it's still the system that throw this error, Node.js just forward it

default.config.js
Aschen: I can auto-generate a seed but either I have to put it in the kuzzlerc in the same time or the developer has to do it manually.  
If the seed is stored only in the database, other developers cloning the project will not have the same value for the seed

scottinet: Ah, right, I get it: auto-generating a seed binds the encrypted file to a Kuzzle instance. 
We could add a small command printing the current seed, though, but that doesn't seem practical.

And I can't shake the feeling that proposing a default seed doesn't look good at all.

New idea then: what about using the key as a seed (stretching it if needed) if none is provided?
Of course, we would have to keep that big warning you added (we would have to reword it), but in essence, that would mean the following: "adding a seed is optional, but strongly encouraging to reinforce encryption"
And we could do that without having a "default seed" (which means having no seed at all)

What do you think?

Aschen: I like this proposal ;)

.kuzzlerc.sample
```diff
@@ -257,6 +257,11 @@
       "proxy": false
     },
 
+    // [vaultSeed]
+    // Initialization vector used to encrypt secrets
+    // Random seed can be generated with: bin/kuzzle vaultSeed
+    "vaultSeed": "764c1afa2988981a",
```
scottinet: Isn't that property outdated now that we create a new seed for each encrypted value?
(same for the one in the default.config.js file)

lib/api/core/vault.js
```diff
@@ -0,0 +1,211 @@
...
+
+    if (!this.vaultKey && !fileExists) {
+      return Promise.resolve();
+    }
+
+    if (this.vaultKey && !fileExists) {
+      return Promise.reject(
+        new Error('A vault key is present and no secrets file can be found. Aborting.')
```
scottinet: (nitpicking) this test and the one below should probably be checked using the assert module

lib/api/core/vault.js
```diff
@@ -0,0 +1,211 @@
...
+
+    return this._readJsonFile(this.encryptedSecretsFile)
+      .then(encryptedSecrets => this.decryptObject(encryptedSecrets))
+      .then(secrets => {
+        this.secrets = secrets;
+
+        return null;
+      })
+      .catch(error => {
+        throw new Error(`Can not decrypt secrets: ${error.message}`);
```
scottinet: (nitpicking) can you write `Cannot` instead of `Can not`? I find it easier to the eye :-) 
(same goes for other error messages)

lib/api/core/vault.js
```diff
@@ -0,0 +1,211 @@
...
+
+  /**
+   * Encrypt data with AES CBC using the secret key and an initialization vector
+   * It's not safe to re-use an IV , so we generate a new IV each time we encrypt
+   * something and we store it next to the encrypted data.
+   * See https://www.wikiwand.com/en/Block_cipher_mode_of_operation#/Initialization_vector_(IV)
+   */
+  _encryptData (data) {
+    const
+      IV = crypto.randomBytes(8).toString('hex'),
```
scottinet: (nitpicking) names starting with an uppercase should be reserved to classes

```suggestion
      iv = crypto.randomBytes(8).toString('hex'),
```

(same thing in the _decryptData function below)

bin/kuzzle
```diff
@@ -139,6 +138,28 @@ bindOption(
     .action(require('./commands/loadSecurities'))
 );
 
+// $ kuzzle encryptSecrets
+bindOption(
+  program
+    .command('encryptSecrets <file>')
+    .option('    --vault-key <vaultKey>', 'Vault key used to encrypt secrets')
```
scottinet: I just spotted that during my review of this feature's documentation: you're mixing snake-case with camelCase.
These options should be named `vaultKey`, `secretsFile`, and so on.

Aschen: I do this because `vaultKey` will be the variable name used in the script but usually command line options are in snake-case

scottinet: I know, and commands should be in snake-case too for the sake of convention... but it's too late now: we have `--noColors` and commands are all camelCased. :-(

scottinet: I won't block that PR because of this, though

Aschen: Oh.. I think we can get ride of it in v2

PR #1319, opened 2019-06-11
Add CLI commands to dump and restore indexes

Aschen:
## What does this PR do ?

It's not possible to change a collection mapping with Kuzzle due to Elasticsearch limitations. The only way to change a mapping is to delete the entire index, recreate it with different mappings and then import the data.

This PR add 2 actions to the CLI to dump and restore indexes:
  - `bin/kuzzle dumpIndex <index> <path>`: dump an entire index in specified directory
  - `bin/kuzzle restoreIndex <path>`: restore the content of a previously dumped index

Each collection of the index is dumped in a separate JSON file. Theses files are loaded with `bulk:import` method.  

The index and the collections must exists before running the `restoreIndex` command.

  - Step 1 : Create an index and collections with some data
  - Step 2 : Dump the index with the CLI
  - Step 3 : Remove the index and collections
  - Step 4: Recreate them and load the dump with the CLI


bin/commands/createFirstAdmin.js
```diff
@@ -74,21 +74,24 @@ function confirm (username, resetRoles) {
...
-    controller: 'server',
-    action: 'adminExists'
-  };
 
-  return sendAction(action, options)
+  return getSdk(options)
+    .then(response => {
+      sdk = response;
+
+      return null;
```
scottinet: (nitpicking) can be simplified with `return sdk.server.adminExists();` (1 promise layer less)
Same goes for similar lines in other commands.

bin/commands/indexDump.js
```diff
@@ -0,0 +1,121 @@
...
+
+/* eslint-disable no-console */
+
+const
+  fs = require('fs'),
+  ColorOutput = require('./colorOutput'),
+  path = require('path'),
+  getSdk = require('./getSdk');
+
+function mkdirp (fullPath) {
```
scottinet: (nitpicking) can you add a TODO comment asking to remove that function for when our min. Node.js version will be at least Node.js 10?
(starting from node 10, we can use the option `recursive: true` with [fs.mkdir](https://devdocs.io/node~10_lts/fs#fs_fs_mkdir_path_options_callback))

bin/commands/indexDump.js
```diff
@@ -0,0 +1,121 @@
...
+  if (parts.length > 1) {
+    mkdirp(parts.slice(0, parts.length - 1).join('/'));
+  }
+
+  if (! fs.existsSync(fullPath)) {
+    fs.mkdirSync(fullPath);
+  }
+}
+
+function dumpCollectionPart (results, bulkDocuments) {
```
scottinet: This won't do: you're storing the entire collection in memory, but that means that we won't be able to dump even moderately large collections.

You need to use a [writable stream](https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_createwritestream_path_options) and dump the result page per page, to keep memory consumption under control.

bin/commands/indexRestore.js
```diff
@@ -0,0 +1,84 @@
...
+  fs = require('fs'),
+  ColorOutput = require('./colorOutput'),
+  getSdk = require('./getSdk');
+
+function importCollection (sdk, cout, dumpFiles) {
+  if (dumpFiles.length === 0) {
+    return null;
+  }
+
+  const bulkData = JSON.parse(fs.readFileSync(dumpFiles[0]));
```
scottinet: You can't put the entire data file in memory, and even less parse it into a JSON object: this method won't allow this CLI tool to handle even moderately large dumped collections.

You need to open a [readable stream](https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_createreadstream_path_options), and use it with some kind of buffer to send bulk imports to Kuzzle sequentially.

bin/commands/indexRestore.js
```diff
@@ -0,0 +1,84 @@
...
+
+  if (bulkData.length === 0) {
+    console.log(cout.notice(`[ℹ] File ${dumpFiles[0]} is empty. Skip import.`));
```
scottinet:
```suggestion
    console.log(cout.notice(`[ℹ] File ${dumpFiles[0]} is empty. Skipping import.`));
```

bin/commands/indexRestore.js
```diff
@@ -0,0 +1,84 @@
...
+    .then(() => {
+      console.log(cout.ok(`[✔] Dump file ${dumpFiles[0]} imported`));
+
+      return null;
+    })
+    .catch(error => {
+      console.log(cout.warn(`[ℹ] Error importing ${dumpFiles[0]}`));
+      if (error.status === 206) {
+        for (const err of error.errors) {
+          console.log(`  ${err.error.reason}`);
```
scottinet: You should probably put partial errors in a log file: this'll flood the terminal even if <1% of an import failed on a collection containing millions of documents.

bin/commands/indexRestore.js
scottinet: Also, we may want to make the import buffer size configurable, to circumvent potentially strict limits put in the .kuzzlerc file by Kuzzle admins

bin/kuzzle
```diff
@@ -160,6 +160,22 @@ bindOption(
     .action(require('./commands/decryptSecrets'))
 );
 
+// $ kuzzle indexDump
+bindOption(
+  program
+    .command('indexDump <index> <path>')
+    .description('dump an entire index in specified directory')
```
scottinet:
```suggestion
    .description('dump an entire index in the specified directory')
```

bin/commands/createFirstAdmin.js
Aschen: I need to reuse the `sdk` few line below ([#L130](https://github.com/kuzzleio/kuzzle/pull/1319/files/aa23c1b1c4093be4f1f532ab645238fa7d3208b7#diff-30b2cf2a4cfe977c24b2f003ceb1c96cR130) to execute `security.createFirstAdmin`

bin/commands/indexDump.js
```diff
@@ -0,0 +1,117 @@
...
+    mkdirp(parts.slice(0, parts.length - 1).join('/'));
+  }
+
+  if (! fs.existsSync(fullPath)) {
+    fs.mkdirSync(fullPath);
+  }
+}
+
+function writeData(stream, data) {
+  return new Promise(resolve => stream.write(data, () => resolve()));
```
scottinet: To be tested, but that might be very slow, as this is like waiting synchronously for each `data` line to be handled by the system.
You may want to discard that callback argument when invoking `write` and instead use the returned value: you can resolve the promise immediately if it returns `true`, otherwise this means that backpressure is building up and you need to wait for the `drain` event before resolving the promise.

This might not be important in this case, and I didn't test the performances difference, so this is your call.

bin/commands/indexDump.js
```diff
@@ -0,0 +1,117 @@
...
+  let
+    promises = [],
+    results = await sdk.document.search(index, collection, {}, options);
+
+  do {
+    process.stdout.write(`  ${results.fetched}/${results.total} documents dumped`);
+    process.stdout.write('\r');
+
+    for (const hit of results.hits) {
+      promises.push(writeData(ndjsonStream, {
```
scottinet: Ah, I didn't saw that part when I wrote [this comment](https://github.com/kuzzleio/kuzzle/pull/1319/files#r299345682).

Seeing this, I now think that you should definitely completely remove the `writeData` function and directly invoke `write` here, yielding only if it returns `false`, until the drain event is triggered. That would make writing a dump much, much faster (and simpler)

bin/commands/indexDump.js
```diff
@@ -0,0 +1,117 @@
...
+  } while (results = await results.next());
+
+  ndjsonStream.end();
+  writeStream.end();
+
+  return waitWrite;
+}
+
+async function indexDump (sdk, cout, index, batchSize, directoryPath) {
+  console.log(cout.notice(`Dumping index ${index} in ${directoryPath} ...`));
```
scottinet: To prevent a silent regression, you might want to make that PR depend on https://github.com/kuzzleio/kuzzle/pull/1342 and remove all usages of `console` (after #1342, `cout` directly prints the arguments on the console, you don't need to do that yourself anymore)

bin/commands/indexRestore.js
```diff
@@ -0,0 +1,137 @@
...
+  getSdk = require('./getSdk');
+
+function handleError(cout, dumpFile, error) {
+  if (error.status === 206) {
+    const
+      errorFile = `${dumpFile.split('.').slice(0, -1).join('.')}-errors.jsonl`,
+      writeStream = fs.createWriteStream(errorFile, { flags: 'a' }),
+      serialize = ndjson.serialize();
+
+    serialize.on('data', line => {
```
scottinet: same here: just pipe the writable stream to `serialize`

bin/commands/indexRestore.js
```diff
@@ -0,0 +1,137 @@
...
+          documents.push(obj);
+
+          if (documents.length === batchSize) {
+            mWriteRequest.body.documents = documents;
+            documents = [];
+
+            readStream.pause();
+
+            sdk.query(mWriteRequest)
+              .then(() => {
```
scottinet: Skipped if there is an error, even if it's a mere 206 one, meaning that the stream is never unpaused and that the output is never updated.

bin/commands/indexRestore.js
```diff
@@ -0,0 +1,137 @@
...
+          mWriteRequest.collection = obj.collection;
+        }
+      })
+      .on('end', () => {
+        if (documents.length > 0) {
+          mWriteRequest.body.documents = documents;
+
+          sdk.query(mWriteRequest)
+            .catch(error => handleError(cout, dumpFile, error))
+            .then(() => resolve());
```
scottinet: you should also update the stdout log line to prevent the last log to report a number of imported documents less than the expected one (confusing for users)

bin/commands/indexDump.js
```diff
@@ -0,0 +1,117 @@
...
+}
+
+function writeData(stream, data) {
+  return new Promise(resolve => stream.write(data, () => resolve()));
+}
+
+async function dumpCollection (sdk, index, collection, batchSize, directoryPath) {
+  const
+    filename = `${directoryPath}/${index}--${collection}--data.jsonl`,
+    writeStream = fs.createWriteStream(filename),
```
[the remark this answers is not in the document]
Aschen: I already tried that but I got this error `TypeError: Invalid non-string/buffer chunk`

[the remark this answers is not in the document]
scottinet: :-(

bin/commands/indexDump.js
Aschen: 104 sec to 20 sec for 1 000 000 documents, indeed :+1:

bin/commands/indexDump.js
```diff
@@ -0,0 +1,129 @@
...
+    waitWrite = new Promise(resolve => writeStream.on('finish', resolve)),
+    ndjsonStream = ndjson.serialize(),
+    options = {
+      scroll: '10m',
+      size: batchSize
+    };
```
benoitvidis: For the time being, shouldn't we include trash items too?

features/step_definitions/indexes.js
```diff
@@ -109,7 +109,7 @@ Then(/^I'm able to delete the index named "([^"]*)"$/, function (index, callback
     .catch(error => callback(error));
 });
 
-Then(/^I refresh the index( ".*?")?$/, function (index, callback) {
+Then(/^I refresh the index( "(.+)")?$/, function (index, callback) {
```
benoitvidis: I don't understand how this regex can match with the nested catch blocks. Also, why use a greedy catch?

```suggestion
Then(/^I refresh the index(?: "(.*?)")?$/, function (index, callback) {
```

bin/commands/indexDump.js
Aschen: I'm not sure since they are gonna be deleted by the garbage collector

features/step_definitions/indexes.js
Aschen: I didn't know but it does, maybe cucumber apply some magic here :upside_down_face: 
I use greedy catch because I want something between the double quote

bin/commands/indexDump.js
benoitvidis: What if the user has a very long or disabled garbage collector and wants to retrieve some data from the bin?

features/step_definitions/indexes.js
benoitvidis: I just tested and yes, it works while it should not (we should get 2 matches, the first one including the space and quotes).
About greedineess, I don't think this is what you are expecting:

```
> /"(.+)"/.exec('"somevar" then "some other var"')[1]
'somevar" then "some other var'
> /"(.*?)"/.exec('"somevar" then "some other var"')[1]
'somevar'
```

bin/commands/indexDump.js
Aschen: Yes you're right!

PR #1325, opened 2019-06-17
Fix typo in vault Decrypt Secrets command

ThomasF34:
## What does this PR do ?

Fix two typo errors in the decrypt secret command : The message was "Successfully encrypted" while it should be "Successfully decrypted"


Aschen:
I put theses fix in this PR on the vault encryption: https://github.com/kuzzleio/kuzzle/pull/1326

PR #1326, opened 2019-06-17
Fix vault encryption

Aschen:
## What does this PR do ?

When encrypting data with the vault, I only save the result of `cipher.final` but for data bigger than the key size we also have to save the result of `cipher.update` and concat it with `cipher.final`

## Other changes

  - use 16 random bytes as IV instead of hex representation of 8 random bytes
  - fix typo from https://github.com/kuzzleio/kuzzle/pull/1325

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ jobs:
         sonarqube: true
 
       cache:
-        directories: 
+        directories:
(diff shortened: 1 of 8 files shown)


Aschen:
> You have to update the version in package.json (execute `npm version patch`)

@scottinet done :+1:

PR #1331, opened 2019-06-20
Allow env or default value in vault cli

Aschen:
## What does this PR do ?

This PR allows user to set the secrets file with an environment variable when using the CLI to encrypt or decrypt files.  
It also use the default secret file if neither command line option or environment variable are provided.

  - Step 1 : Write secret in `config/secrets.json`

With the default file:
  - Step 2 : Encrypt it: `KUZZLE_VAULT_KEY=password ./bin/kuzzle encryptSecrets`
  - Step 3 : Decrypt it: `KUZZLE_VAULT_KEY=password ./bin/kuzzle decryptSecrets`

With the an environment variable:
  - Step 2 : Encrypt it: `KUZZLE_VAULT_KEY=password KUZZLE_SECRETS_FILE=./config/secrets.json ./bin/kuzzle encryptSecrets`
  - Step 3 : Decrypt it: `KUZZLE_VAULT_KEY=password KUZZLE_SECRETS_FILE=./config/secrets.enc.json`  ./bin/kuzzle decryptSecrets


bin/commands/decryptSecrets.js
```diff
@@ -28,8 +28,16 @@ const
...
 
-function commandDecryptSecrets (encryptedSecretsFile, options) {
-  const cout = new ColorOutput(options);
+function commandDecryptSecrets (file, options) {
+  const
+    vault = new Vault(),
+    encryptedSecretsFile = file || process.env.KUZZLE_SECRETS_FILE || vault.defaultEncryptedSecretsFile,
+    cout = new ColorOutput(options);
+
+  if (!encryptedSecretsFile) {
```
scottinet: This test is always false, since `Vault.defaultEncryptedSecretsFile`, being a default value, is guaranteed to always hold a value.

bin/commands/encryptSecrets.js
```diff
@@ -28,8 +28,16 @@ const
...
 
-function commandEncryptSecrets (secretsFile, options) {
-  const cout = new ColorOutput(options);
+function commandEncryptSecrets (file, options) {
+  const
+    vault = new Vault(),
+    secretsFile = file || process.env.KUZZLE_SECRETS_FILE || vault.defaultSecretsFile,
+    cout = new ColorOutput(options);
+
+  if (!secretsFile) {
```
scottinet: ditto

bin/kuzzle
```diff
@@ -152,7 +152,7 @@ bindOption(
 // $ kuzzle decryptSecrets
 bindOption(
   program
-    .command('decryptSecrets <file>')
+    .command('decryptSecrets [file]>')
```
Yoann-Abbes:
```suggestion
    .command('decryptSecrets [file]')
```

bin/commands/encryptSecrets.js
```diff
@@ -28,8 +28,11 @@ const
   readlineSync = require('readline-sync'),
   ColorOutput = require('./colorOutput');
 
-function commandEncryptSecrets (secretsFile, options) {
-  const cout = new ColorOutput(options);
+function commandEncryptSecrets (file, options) {
+  const
+    vault = new Vault(),
+    secretsFile = file || process.env.KUZZLE_CLEAR_SECRETS_FILE || vault.defaultSecretsFile,
```
scottinet: What is this new environment variable name?
Is this a manual test leftover?

Aschen: I introduce this variable because `KUZZLE_SECRETS_FILE` refers to the encrypted secrets file.  It can be confusing to have the same variable name for both clear and encrypted files.  
I added documentation for this variable in this PR: https://github.com/kuzzleio/documentation/pull/373

scottinet: As discussed, we should remove that environment variable for now.

PR #1338, opened 2019-06-27
Fix an error in Http Protocol when data content is null

ThomasF34:
## What does this PR do ?

When a controller action is returning a raw content but this content is `null`, an error is thrown since we are trying to access `content.type`. 

This PR transforms the verification `data.raw` into `data.raw && data.content`. Thus, if the content is null we will now check the type of it. (and directly go to the stringify step) 

- Step 1 : in a controller action : 
```js 
async anAction(request){
  request.response.raw = true 
  return null
}
```
- Step 2 : when you're calling this action, the error is thrown

### DIFF
--- a/lib/api/core/entrypoints/embedded/protocols/http.js
+++ b/lib/api/core/entrypoints/embedded/protocols/http.js
@@ -262,8 +262,11 @@ class HttpProtocol extends Protocol {
 
     debug('HTTP request response: %a', data);
 
-    if (data.raw && data.content) {
-      if (typeof data.content === 'object') {
+    if (data.raw) {
+      if (!data.content) {
+        request.response.setHeader('Content-Length', '0');
+        data = '';
+      } else if (typeof data.content === 'object') {
         /*
          This object can be either a Buffer object, a stringified Buffer object,
          or anything else.
--- a/lib/api/core/entrypoints/embedded/protocols/http.js
+++ b/lib/api/core/entrypoints/embedded/protocols/http.js
@@ -261,7 +261,7 @@ class HttpProtocol extends Protocol {
 
     debug('HTTP request response: %a', data);
 
-    if (data.raw) {
+    if (data.raw && data.content) {
       if (typeof data.content === 'object') {
         /*
          This object can be either a Buffer object, a stringified Buffer object,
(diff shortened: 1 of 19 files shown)


scottinet:
Also, I disagree with how this bug is fixed: the new condition makes Kuzzle think that the data is NOT raw, and so it embeds the null result in a standard Kuzzle API response, which is not the intended behavior.

Instead, it should send a 0-length content.

lib/api/core/entrypoints/embedded/protocols/http.js
```diff
@@ -262,7 +262,7 @@ class HttpProtocol extends Protocol {
 
     debug('HTTP request response: %a', data);
 
-    if (data.raw) {
+    if (data.raw && data.content) {
```
alexandrebouthinon:
```suggestion
    if (data.raw) {
```

lib/api/core/entrypoints/embedded/protocols/http.js
```diff
@@ -262,7 +262,7 @@ class HttpProtocol extends Protocol {
...
       if (typeof data.content === 'object') {
```
alexandrebouthinon:
```suggestion
      if (data.content && typeof data.content === 'object') {
```

scottinet: not enough: if you do that, you end up stringifying null and returning a string with the word "null" in it

you have to specifically return null if data.content is null, to truly send a 0-length response

lib/api/core/entrypoints/embedded/protocols/http.js
```diff
@@ -263,7 +263,10 @@ class HttpProtocol extends Protocol {
     debug('HTTP request response: %a', data);
 
     if (data.raw) {
-      if (typeof data.content === 'object') {
+      if (!data.content) {
```
scottinet: This test makes Kuzzle return a 0-length response if the content is falsey, such as  the number `0` or the `false` boolean value.

```suggestion
      if (data.content === null || data.content === undefined) {
```

ThomasF34:
Before checking the type of `data.content`, i'm checking if it's null. If it is, we set an header `Content Length` to `0` and set the data to an empty string. Thus, the `_getResponseData` will still return a `Buffer` which is empty if `data.content` is null. 
I also added a unit test to check if the content-length is correctly set to 0

PR #1350, opened 2019-07-15
Add public route to list api actions

Aschen:
## What does this PR do ?

Adds a new route to list only the API routes: 
```
  {
    controller: 'server',
    action: 'publicApi'
  }

// http
  GET /_publicApi
```

This route is needed by the Http protocols of the SDKs to construct URLs.

Internal: [KZL-1225](https://jira.kaliop.net/browse/KZL-1225)

SDK JS: https://github.com/kuzzleio/sdk-javascript/pull/421

  - Step 1 : Open http://localhost:7512/_publicApi

### Other changes

  - rename `controller` to `baseController`


features/kuzzle.feature
```diff
@@ -1,1969 +1,1974 @@
...
-          "stop": -1,
-          "options": ["withscores"]
-        }
-      }
-      """
-    Then The ms result should match the json ["one", "1", "uno", "1", "three", "3", "two", "4"]
-    Given I call the zadd method of the memory storage with arguments
-      """
-      {
-        "_id": "#p
```
benoitvidis: Should not be commented.

Tip: You can run a specific cucumber test by adding the line number to the scenario file:

```
./node_modules/.bin/cucumber-js --profile websocketEmbedded features/kuzzle.feature:350
```

lib/config/httpRoutes.js
```diff
@@ -85,6 +85,7 @@ module.exports = [
   {verb: 'get', url: '/_healthCheck', controller: 'server', action: 'healthCheck'},
   {verb: 'get', url: '/_serverInfo', controller: 'server', action: 'info'},
   {verb: 'get', url: '/_now', controller: 'server', action: 'now'},
+  {verb: 'get', url: '/public_api', controller: 'server', action: 'publicApi'},
```
benoitvidis: Should be renamed to `/_publicApi` to be consistent with other http routes.

Aschen:
@benoitvidis I added the documentation, you may review this PR again

doc/1/api/controllers/server/public-api/index.md
```diff
@@ -0,0 +1,82 @@
+---
+code: true
+type: page
+title: publicApi
+---
+
+# publicApi
+
+
+Returns available API, including Kuzzle API and plugins API.
```
scottinet:
```suggestion
Returns available API routes, including both Kuzzle's and plugins'.  
```

doc/1/api/controllers/server/public-api/index.md
```diff
@@ -0,0 +1,82 @@
...
+title: publicApi
+---
+
+# publicApi
+
+
+Returns available API, including Kuzzle API and plugins API.  
+
+::: warning
+This route is used by the HTTP protocols of the SDKs to build requests based on the controller and action names.
```
scottinet: (debate/optional) About plugins, shouldn't we have an option for that route allowing to either disable plugins routes, or maybe to hide some routes, or both?

I feel that might be needed for multi-tenant applications that need to hide routes created dynamically for their clients.

scottinet: Also, shouldn't the SDKs re-invoke that route upon a successful login attempt? So they can work if this route is closed for anonymous users, and opened for logged in ones, making this warning valid only if rights are too restricted, but not necessarily for anon users?

doc/1/api/controllers/server/public-api/index.md
```diff
@@ -0,0 +1,82 @@
...
+  "action": "publicApi"
+}
+```
+
+---
+
+## Response
+
+Returns an object containing the definition of the available API.  
+Each key correspond to a controller.
```
scottinet:
```suggestion
Each key matches an API controller.
```

features/kuzzle.feature
```diff
@@ -1,5 +1,9 @@
 Feature: Kuzzle functional tests
 
+  Scenario: API method server:publicApi
+    When I get public API
```
scottinet: (nitpicking) 
```suggestion
    When I get the public API
```

lib/api/controllers/serverController.js
```diff
@@ -264,6 +219,70 @@ class ServerController extends BaseController {
...
+        this.kuzzle.funnel.pluginsControllers,
+        this.kuzzle.pluginsManager.routes,
+        '_plugin/'
+      ),
+      apiDefinition = Object.assign({}, kuzzleApi, pluginsApi);
+
+    return Bluebird.resolve(apiDefinition);
+  }
+
+  _buildApiDefinition (controllers, httpRoutes, urlPrefix = '') {
```
scottinet: (optional) I wonder how much it would cost to get rid of that http routes configuration file and, instead, make each controller expose a list of HTTP routes for each action in their list of actions. That would greatly simplify routes retrieval.

doc/1/api/controllers/server/public-api/index.md
Aschen: 1) We discussed this point in the workshop and we agreed that if a plugin developer want to use additional routes in a SDK with the HTTP protocol, he should add theses routes himself. 
But As I said in the workshop, I agree with you on the need to hide some plugin routes.

2) Yes we can do that.

lib/api/controllers/serverController.js
Aschen: Yes we could totally do this, like for the plugin routes definition! Also we could think about an unified API for both internal controllers and plugin controllers so Kuzzle can handle them the same way.

But I think this is outside the scope of this PR.

Aschen:
> Nice job! We should think about fix Http protocol in JS and PHP SDKs then slightly_smiling_face

SDK JS: kuzzleio/sdk-javascript#421

doc/1/api/controllers/server/public-api/index.md
scottinet: Ah yes, I forgot about the privacy flag added to plugins http routes. I apologize.

lib/api/controllers/serverController.js
scottinet: I agree, but this would be a nice refactor to do :)

PR #1365, opened 2019-07-30
Do not check for a connection existence if none is set

scottinet:
# Description

The router "controller", amongst other tasks, manages connections, to allow skipping requests from dead connections.

This PR fixes 2 minor problems:
* some protocols cannot maintain a connection identifier (e.g. UDP, MODBUS, ...): the router now only checks the connection identifier if one is set
* the check always returns `true` for the HTTP protocol, because requests from Kuzzle Proxy cannot declare new connections beforehand. But this makes Kuzzle execute all HTTP requests, even useless ones from dropped HTTP connections. So this has been changed so that request coming from the proxy now have a special flag in the request.connection.misc object, and the router controller now relies on that flag to let pass proxy requests

It also fixes a leak occuring when requests are dropped by the funnel controller because its associated connection has been killed: the funnel now answers to protocols (embedded or not) with an appropriate error when that happens, allowing protocols to clean up resources if necessary.

# Other change

* Connections managed by the router controller are now stored in a map, which is a lot more suited for storing highly volatile keys (better performances)
* The documentation `doc.sh prepare` script fails if the framework repository is already there. This is fixed by force-pulling the repository instead of trying to clone it if it already exists.

### DIFF
--- a/doc/doc.sh
+++ b/doc/doc.sh
@@ -20,8 +20,14 @@ fi
 
 case $1 in
   prepare)
-    echo "Clone documentation framework"
-    git clone --depth 10 --single-branch --branch master https://github.com/kuzzleio/documentation.git framework/
+    if [ -d "framework" ]
+    then
+      echo "Pulling latest framework version"
+      bash -c "cd framework && git reset --hard HEAD~ && git pull"
+    else
+      echo "Clone documentation framework"
+      git clone --depth 10 --single-branch --branch master https://github.com/kuzzleio/documentation.git framework/
+    fi
 
     echo "Link local doc for dead links checking"
     rm framework/src$DOC_PATH
--- a/doc/doc.sh
+++ b/doc/doc.sh
@@ -20,14 +20,10 @@ fi
 
 case $1 in
   prepare)
-    if [ -d "framework" ]
-    then
-      echo "Pulling latest framework version"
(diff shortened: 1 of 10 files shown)


scottinet:
@Shiranuit @Aschen > due to unforeseen consequences of enforcing a live connection for HTTP requests with requests emanating from the (deprecated) Kuzzle Proxy, this PR got a bit more complex than I first anticipated. I'm afraid I need you to re-review it. :sweat:

doc/doc.sh
```diff
@@ -20,8 +20,14 @@ fi
 
 case $1 in
   prepare)
-    echo "Clone documentation framework"
-    git clone --depth 10 --single-branch --branch master https://github.com/kuzzleio/documentation.git framework/
+    if [ -d "framework" ]
+    then
+      echo "Pulling latest framework version"
+      bash -c "cd framework && git reset --hard HEAD~ && git pull"
```
Aschen:
```suggestion
      bash -c "cd framework && git reset --hard HEAD~ && git checkout master && git pull"
```
We have to switch back to master before pulling (see this failing build: https://app.netlify.com/sites/doc-sdk-javascript/deploys/5d40611fbfdf50000a98d5a7)

scottinet: Good call :+1:

scottinet: @Aschen > done

PR #1371, opened 2019-08-01
Fix erroneous protocols.execute documentation

scottinet:
# Description

The current documentation for the `execute` method exposed to custom protocols is a copy of the one for the same function name exposed to plugins. But in reality, these 2 functions have different behaviors:

- The `execute` exposed to protocols only takes a callback, it never returns a promise. This is a design choice, to keep the number of items in the event loop in check (whereas most plugins are executed after the funnel, which regulates the number of concurrent requests at any given time)
- The returned value is not an updated `Request` object, but a serialized JSON response, that a protocol must use to respond to its requesting clients


doc/1/protocols/api/entrypoint/execute/index.md
```diff
@@ -10,7 +10,7 @@ title: execute
 
 Executes a Kuzzle's [API action](/core/1/api/).
 
-The `execute` function main usage is to forward users API requests to Kuzzle.
+The `execute` function main usage is to forward users API requests to Kuzzle.
```
Yoann-Abbes:
```suggestion
The `execute` function main usage is to forward users API requests to Kuzzle.
```

doc/1/protocols/api/entrypoint/execute/index.md
```diff
@@ -24,38 +24,57 @@ execute(request, [callback]);
...
+This resulting object contains the following properties:
 
-- if it is: the `execute` function returns nothing, and the callback is called once the API call is finished, with the following arguments: `callback(error, request)`
-- otherwise: the `execute` function returns a promise, resolving to the updated request, or rejected with a KuzzleError object
+* `raw` (boolean, default `false`): if `false`, the response content is a JSON object following the [Kuzzle standard response format](core/1/api/essentials/kuzzle-response/). Otherwise, the content can be anything, depending on the executed API action.
+* `status` (integer, default `200`): request status, following the standard [HTTP status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes)
+* `requestId` (string, nullable): request identifier set by clients, and used to link asynchronous responses to emitted requests.
+* `content`: the response itself, this is the payload that should be sent to the requesting client following the implemented protocol format. This property type partlydepends the value of the `raw` flag (if `false`, then the content is always a JSON object).
```
Yoann-Abbes:
```suggestion
* `content`: the response itself, this is the payload that should be sent to the requesting client following the implemented protocol format. This property type partly depends on the value of the `raw` flag (if `false`, then the content is always a JSON object).
```

PR #1373, opened 2019-08-01
Add cli dump indexes doc

Aschen:
## What does this PR do ?

Add documentation for CLI action `indexDump` and `indexRestore`

https://deploy-preview-1373--doc-kuzzle.netlify.com/guides/essentials/cli/#indexdump

### DIFF
--- a/doc/1/guides/essentials/cli/index.md
+++ b/doc/1/guides/essentials/cli/index.md
@@ -603,10 +603,10 @@ Options:
       --batch-size <batchSize>  Maximum batch size (see limits.documentsFetchCount config)
 ```
 
-Dump the index `<index>` in the directory `<path>`.  
+Dumps the index `<index>` in the directory `<path>`.  
 Each collection is dumped in a separate file in [JSON lines](http://jsonlines.org/) format.  
 
-The dump directory have the following structure:
+The dump directory has the following structure:
 
 ```bash
 <path>
@@ -627,5 +627,5 @@ Options:
       --batch-size <batchSize>  Maximum batch size (see limits.documentsWriteCount config)
 ```
 
-Restore a previously dumped index from directory `<path>` into Kuzzle.  
+Restores a previously dumped index from directory `<path>` into Kuzzle.  
 The index and destination collections must exist.  
--- a/doc/1/guides/essentials/cli/index.md
+++ b/doc/1/guides/essentials/cli/index.md
@@ -33,19 +33,21 @@ To get a list of commands and options run the CLI:
 #
 #   Commands:
 #
-#     createFirstAdmin           create the first administrator user
-#     clearCache                 clear internal caches in Redis
-#     reset                      reset all users, profiles, roles and documents validation specifications
-#     resetSecurity              reset all users, profiles and roles
-#     resetDatabase              remove all data stored on Kuzzle
-#     shutdown                   gracefully exits after processing remaining requests
-#     start [options]            start a Kuzzle instance
-#     dump                       create a dump of current state of kuzzle
-#     loadMappings <file>        load database mappings into Kuzzle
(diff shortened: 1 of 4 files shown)


doc/1/guides/essentials/cli/index.md
```diff
@@ -589,3 +591,41 @@ The `file` can be provided in the command line, through the `KUZZLE_SECRETS_FILE
...
+<SinceBadge version="1.9.0" />
+
+```bash
+./bin/kuzzle indexDump <index> <path> [options]
+
+Options:
+      --batch-size <batchSize>  Maximum batch size (see limits.documentsFetchCount config)
+```
+
+Dump the index `<index>` in the directory `<path>`.
```
scottinet:
```suggestion
Dumps the index `<index>` in the directory `<path>`.  
```

doc/1/guides/essentials/cli/index.md
```diff
@@ -589,3 +591,41 @@ The `file` can be provided in the command line, through the `KUZZLE_SECRETS_FILE
...
+./bin/kuzzle indexDump <index> <path> [options]
+
+Options:
+      --batch-size <batchSize>  Maximum batch size (see limits.documentsFetchCount config)
+```
+
+Dump the index `<index>` in the directory `<path>`.  
+Each collection is dumped in a separate file in [JSON lines](http://jsonlines.org/) format.  
+
+The dump directory have the following structure:
```
scottinet:
```suggestion
The dump directory has the following structure:
```

doc/1/guides/essentials/cli/index.md
```diff
@@ -589,3 +591,41 @@ The `file` can be provided in the command line, through the `KUZZLE_SECRETS_FILE
...
+<SinceBadge version="1.9.0" />
+
+```bash
+./bin/kuzzle indexRestore <path> [options]
+
+Options:
+      --batch-size <batchSize>  Maximum batch size (see limits.documentsWriteCount config)
+```
+
+Restore a previously dumped index from directory `<path>` into Kuzzle.
```
scottinet:
```suggestion
Restores a previously dumped index from directory `<path>` into Kuzzle.  
```

doc/1/guides/essentials/cli/index.md
```diff
@@ -589,3 +591,41 @@ The `file` can be provided in the command line, through the `KUZZLE_SECRETS_FILE
...
+
+```bash
+./bin/kuzzle indexRestore <path> [options]
+
+Options:
+      --batch-size <batchSize>  Maximum batch size (see limits.documentsWriteCount config)
+```
+
+Restore a previously dumped index from directory `<path>` into Kuzzle.  
+The index and destination collections must exist.
```
scottinet: (debate / optional)

I didn't see that limitation when I reviewed the PR introducing these 2 commands. Is it intended to not make the `indexRestore` command automatically create the target index/collection?

Personnally I would have done the reverse: load only if the target index or collection doesn't exist, creating them in the process, and fail otherwise.
And I would have added a `--force` flag allowing to restore an index if it already exists.

What do you think?

Aschen: Yes it is intentional because for me the primary purpose of this command was to be able to change a collection mapping and for this we needed a way to easily dump/import collections.  

Otherwise, I planned to dump the collections mappings alongside the data, so the user can edit them before importing his collections again. It's not a big deal but the PR was big enough so I prefer to do this in another PR, what do you think ?

scottinet: I thought/expected that the mappings were already exported by the CLI :expressionless: 

IMO it doesn't seem practical for a dump utility to not have it recreate the dumped structure, either by default or by setting this behavior explicitly through options.

Anyway this is outside the scope of this PR, which is to document the already existing features.

PR #1333, opened 2019-06-24
Add subcodes to kuzzle core entrypoints

Yoann-Abbes:
## What does this PR do ?

Jira : [Subtask](https://jira.kaliop.net/browse/KZL-1200) from this [story](https://jira.kaliop.net/browse/KZL-1192)

This PR is to merge into the `KZL-1035-subcode-error` [(here)](https://github.com/kuzzleio/kuzzle/pull/1306) branch, such as the PRs to come.

Add subcodes to : 
`authController`
`adminController`
`bulkController`
`securityController`
`memoryStorage`
`funnelController`
`collectionController`
`indexController`
`documentController`

Add/adapt unit test.

Did not do for `PluginImplementationError`, it will be in another PR (it's another [subtask](https://jira.kaliop.net/browse/KZL-1199))

### How should this be manually tested?

unit tests

### Boyscout

80 columns

PR #1374, opened 2019-08-01
Add revokeTokens route to Security Controller

Yoann-Abbes:
## What does this PR do ?

Adds a `revokeTokens` route to the security controller, with tests and documentation page

https://jira.kaliop.net/browse/KZL-507

unit/functional tests


Aschen:
I'm thinking that we could both have revoke all tokens and revoke one token with this route.  
User could provide an array of token that need to be revoked to revoke specific token.  
If no array if provided, this route revoke all user tokens

scottinet:
@aschen That would mean that an administrator would know what tokens to revoke beforehand and that does not sound practical to me. How do you propose to make that work? :thinking:

Aschen:
> @Aschen That would mean that an administrator would know what tokens to revoke beforehand and that does not sound practical to me. How do you propose to make that work? thinking

I just thinking about and you're right, we would need a `security:listToken` before

scottinet:
@Aschen I'm absolutely and completely against any addition to API allowing tokens to be leaked to the outside.

scottinet:
@Aschen > check out JIRA KZL-509, I described how we could let users manage their sessions, without ever leaking tokens. I think this is the way to go, and it would be easy to add API routes in the security controller to allow an administrator to manage another user's sessions.

Aschen:
> @Aschen > check out JIRA KZL-509, I described how we could let users manage their sessions, without ever leaking tokens. I think this is the way to go, and it would be easy to add API routes in the security controller to allow an administrator to manage another user's sessions.

Yes you're right, I have been thinking about it and it's not possible to have this kind of route.  
Eventually the route could list token by another identifier but we would need to add metadata (like ip adress) to allow users to identify legitimate sessions

lib/config/httpRoutes.js
```diff
@@ -280,6 +280,7 @@ module.exports = [
   {verb: 'delete', url: '/roles/:_id', controller: 'security', action: 'deleteRole'},
   {verb: 'delete', url: '/users/:_id', controller: 'security', action: 'deleteUser'},
   {verb: 'delete', url: '/credentials/:strategy/:_id', controller: 'security', action: 'deleteCredentials'},
+  {verb: 'delete', url: '/_revokeTokens/:_id', controller: 'security', action: 'revokeTokens'},
```
benoitvidis: I would find 
`DELETE /users/:_id/tokens` more RESTful (not using gh suggestion here as it will break the functional tests & doc).

lib/api/controllers/securityController.js
```diff
@@ -1024,6 +1025,21 @@ class SecurityController extends BaseController {
...
+
+    const userId = request.input.resource._id;
+    return this.kuzzle.repositories.user.load(userId)
+      .then(user => {
+        if (!user) {
+          return Bluebird.reject(new NotFoundError(`User with id ${userId} not found`));
+        }
+        
+        this.kuzzle.repositories.token.deleteByUserId(userId);
+        return Bluebird.resolve();
```
benoitvidis: Useless promise (you are already in a promise context)

features/kuzzle.feature
```diff
@@ -604,6 +604,7 @@ Feature: Kuzzle functional tests
     Then I am able to perform a scrollUsers request
     Then I delete the user "user2-id"
     Then I search for {"ids":{"type": "users", "values":["#prefix#useradmin-id"]}} and find 1 users matching {"_id":"#prefix#useradmin-id","_source":{"name":{"first":"David","last":"Bowie"}}}
+    Then I revoke all tokens of the user "useradmin-id"
     When I log in as useradmin:testpwd expiring in 1h
```
benoitvidis: Shoudn't we test the the user's token are actually revoked?

something like:
- log in as user
- revoke user's token
- call any action
- get an error (which one btw?)

Probably deserves a dedicated scenario.

test/api/controllers/securityController/users.test.js
```diff
@@ -767,4 +767,16 @@ describe('Test: security controller - users', () => {
       });
     });
   });
+
+  describe('#revokeTokens', () => {
+    it('should revoke all tokens related to a given user', () => {
+      
+      return should(securityController.revokeTokens((new Request({ _id: 'test', })))).be.resolved();
```
benoitvidis: shouldn't we test the repository deletion method is called?

doc/1/api/controllers/security/revoke-tokens/index.md
```diff
@@ -0,0 +1,56 @@
...
+Revokes every token of a given user.
+
+---
+
+## Query Syntax
+
+### HTTP
+
+```http
+URL: http://kuzzle:7512/_revokeTokens_/<_id>
```
Aschen:
```suggestion
URL: http://kuzzle:7512/users/:_id/tokens
```

doc/1/api/controllers/security/revoke-tokens/index.md
```diff
@@ -0,0 +1,56 @@
...
+Method: DELETE
+Body:
```
Aschen:
```suggestion
```

doc/1/api/controllers/security/revoke-tokens/index.md
```diff
@@ -0,0 +1,56 @@
...
+---
+
+## Response
+
+```js
+{
+  "status": 200,
+  "error": null,
+  "result": {
+    "_id": "<kuid>"
```
Aschen: IMHO the route should return nothing (`null`)

features/kuzzle.feature
```diff
@@ -589,6 +589,16 @@ Feature: Kuzzle functional tests
...
 
+  @security
+  Scenario: revoke user's tokens
+    When I create a new role "role1" with id "role1"
+    And I create a new profile "profile1" with id "profile1"
+    And I create a user "user1" with id "user1-id"
+    Then I log in as user1:testpwd1 expiring in 1h
+    Then I'm able to find rights for user "user1-id"
+    Then I revoke all tokens of the user "user1-id"
+    Then I'm not able to find rights for user "user1-id"
```
Aschen: I think you should use the route `checkToken` with the revoked JWT instead

lib/api/controllers/securityController.js
```diff
@@ -1024,6 +1025,19 @@ class SecurityController extends BaseController {
...
   }
+
+  revokeTokens(request) {
+    assertHasId(request);
+
+    const userId = request.input.resource._id;
+    return this.kuzzle.repositories.user.load(userId)
+      .then(user => {
+        if (!user) {
+          return Bluebird.reject(new NotFoundError(`User with id ${userId} not found`));
```
Aschen: There is no need to use `Bluebird.reject` since you are already in a promise chain. If you throw an exception inside the promise chain, the entire promise will be rejected.
And also you could use the errorManger here no?

Yoann-Abbes: I prefere to wait https://github.com/kuzzleio/kuzzle/pull/1333 is merged before using the errorsManager here to avoid conflicts

lib/api/controllers/securityController.js
```diff
@@ -1024,6 +1025,19 @@ class SecurityController extends BaseController {
...
   }
+
+  revokeTokens(request) {
+    assertHasId(request);
+
+    const userId = request.input.resource._id;
+    return this.kuzzle.repositories.user.load(userId)
+      .then(user => {
+        if (!user) {
+          throw new NotFoundError(`User with id ${userId} not found`);
```
Aschen: Use the errorManager

lib/api/controllers/securityController.js
```diff
@@ -1212,6 +1213,19 @@ class SecurityController extends BaseController {
...
+  revokeTokens(request) {
+    assertHasId(request);
+
+    const userId = request.input.resource._id;
+    return this.kuzzle.repositories.user.load(userId)
+      .then(user => {
+        if (!user) {
+          errorsManager.throw('user_not_found', userId);
+        }
+        return this.kuzzle.repositories.token.deleteByUserId(userId);
```
Aschen: The documentation indicate that this route return `null` but here you return the result of `deleteByUserId`

PR #1375, opened 2019-08-02
Always set the content length for HTTP responses

scottinet:
# Description

Before this PR we never set the `Content-Length` header for HTTP responses. This wasn't a problem, until one of our users made their plugin set one, and when a requesting client asked for a compressed response: the content length set by our user plugin couldn't be the same than the compressed payload, and this resulted in errors with, for instance, the SPDY protocol.

This PR makes Kuzzle always set the `Content-Length` header, overwritting it if it's already set.

### DIFF
--- a/lib/api/core/entrypoints/embedded/protocols/http.js
+++ b/lib/api/core/entrypoints/embedded/protocols/http.js
@@ -201,7 +201,7 @@ class HttpProtocol extends Protocol {
           return;
         }
 
-        request.response.headers['Content-Length'] = String(deflated.length);
+        request.response.setHeader('Content-Length', String(deflated.length));
         response.writeHead(request.response.status, request.response.headers);
         response.end(deflated);
 
--- a/lib/api/core/entrypoints/embedded/protocols/http.js
+++ b/lib/api/core/entrypoints/embedded/protocols/http.js
@@ -201,6 +201,7 @@ class HttpProtocol extends Protocol {
           return;
         }
 
+        request.response.headers['Content-Length'] = String(deflated.length);
         response.writeHead(request.response.status, request.response.headers);
         response.end(deflated);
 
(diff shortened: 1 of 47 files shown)


test/api/core/entrypoints/embedded/protocols/http.test.js
```diff
@@ -718,8 +744,28 @@ describe('/lib/api/core/entrypoints/embedded/protocols/http', () => {
...
+        const
+          cb = kuzzle.router.http.route.firstCall.args[1],
+          request = new Request({}),
+          result = 'foobar';
+
+        request.setResult(result, {
+          raw : true
+        });
+
+        request.response.setHeader('content-length', 'one does not simply set a content length');
```
alexandrebouthinon: You can wrap this line
```suggestion
        request.response.setHeader(
          'content-length',
          'one does not simply set a content length');
```

lib/api/core/entrypoints/embedded/protocols/http.js
```diff
@@ -201,6 +201,7 @@ class HttpProtocol extends Protocol {
           return;
         }
 
+        request.response.headers['Content-Length'] = String(deflated.length);
```
benoitvidis: Here and for all other usages: shouldn't we rather use the response's `setHeader` method to handle the casing? (i.e. `content-Length`, `content-length`...)

scottinet: Forgot about that, good catch. I should also use getHeader when fetching a header (in unit tests).

test/api/core/entrypoints/embedded/protocols/http.test.js
```diff
@@ -615,7 +632,8 @@ describe('/lib/api/core/entrypoints/embedded/protocols/http', () => {
         cb(result);
 
         should(response.end).be.calledWith(Buffer.from(''));
-        should(result.response.headers['Content-Length']).be.eql('0');
+        should(result.response.getHeader('Content-Length'))
+          .eql(String('0'));
```
benoitvidis: nitpicking: useless cast

PR #1376, opened 2019-08-02
Add subcodes for Redis service

Yoann-Abbes:
## What does this PR do ?

Jira : [Story](https://jira.kaliop.net/browse/KZL-1188)

Add subcodes to Redis Service

Add/adapt unit test.

unit/functional tests

### Boyscout

format code

### DIFF
--- a/lib/services/redis.js
+++ b/lib/services/redis.js
@@ -23,7 +23,6 @@ const
   _ = require('lodash'),
   Bluebird = require('bluebird'),
   IORedis = require('ioredis'),
-  ServiceUnavailableError = require('kuzzle-common-objects').errors.ServiceUnavailableError,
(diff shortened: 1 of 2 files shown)


lib/services/redis.js
```diff
@@ -97,7 +97,8 @@ class Redis extends Service {
     return new Bluebird((resolve, reject) => {
       this._client.once('ready', () => {
         if (this._config.database > 0) {
-          return this._client.select(this._config.database, error => error ? reject(error) : resolve(this));
+          return this._client.select(
```
scottinet: (nitpicking) should be wrapped like this:

```js
          return this._client.select(
            this._config.database,
            error => error ? reject(error) : resolve(this));
```

PR #1382, opened 2019-08-07
Fix lost subscriptions when authentication tokens expire

benoitvidis:
## What does this PR do ?

This PR fixes the situation where a user would 
1. log in
2. subscribe to some rooms
3. after some time, log in again
4. see his subscriptions lost once his first login expires

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -11,6 +11,7 @@
     "new-cap": 1,
     "no-caller": 2,
     "no-catch-shadow": 2,
+    "no-console": 2,
     "no-else-return": 2,
     "no-extend-native": 2,
     "no-extra-bind": 2,
--- a/lib/api/core/auth/tokenManager.js
+++ b/lib/api/core/auth/tokenManager.js
@@ -23,7 +23,8 @@
 
 const
   { models: { RequestContext } } = require('kuzzle-common-objects'),
-  SortedArray = require('sorted-array');
+  SortedArray = require('sorted-array'),
+  Token = require('../models/security/token');
 
 /*
  Maximum delay of a setTimeout call. If larger than this value,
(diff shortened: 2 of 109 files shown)


.eslintrc.json
```diff
@@ -11,6 +11,7 @@
     "new-cap": 1,
     "no-caller": 2,
     "no-catch-shadow": 2,
+    "no-console": 2,
```
benoitvidis: removed from eslint:recommended

lib/api/core/auth/tokenManager.js
```diff
@@ -172,22 +169,20 @@ class TokenManager {
...
-        expiresAt: newToken.expiresAt,
-        userId: newToken.userId,
-        rooms: this.tokens.array[pos].rooms
-      });
-      this.tokens.remove({idx: oldIndex});
+      this._add(
+        newToken,
+        this.tokens.array[pos].connectionId,
+        [...this.tokens.array[pos].rooms]
+      );
```
Yoann-Abbes: (nit)
this should be on the line above according to the new wrapping rules

benoitvidis: :+1:

lib/api/core/auth/tokenManager.js
```diff
@@ -62,6 +63,7 @@ class TokenManager {
 
       return a.idx < b.idx ? -1 : 1;
     });
+    this.tokensByConnectedUser = {};
```
Aschen: Can't we use a `Map` instead?

benoitvidis: :+1:

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "kuzzle",
   "author": "The Kuzzle Team <support@kuzzle.io>",
-  "version": "1.9.1",
+  "version": "1.9.2",
```
scottinet: The package-lock.json file must also be updated.

benoitvidis: :+1:

PR #1383, opened 2019-08-07
Add more subcodes to controllers

Yoann-Abbes:
## What does this PR do ?

This PR complete the implementation of error subcodes in the controllers
It also adds them in the `repositories`

  - Step 1 :
  - Step 2 :
  - Step 3 :  
  ...

### Other changes

Adds a function who checks if each error code is unique

### DIFF
--- a/lib/api/controllers/routerController.js
+++ b/lib/api/controllers/routerController.js
@@ -75,7 +75,7 @@ class RouterController {
       return this.kuzzle.log.error(errorsManager.getError(
         'plugins',
         'runtime',
-        'unable_to_remove_connection_invalid_args',
+        'unable_to_remove_connection_unknown_connection_identifier',
         JSON.stringify(requestContext.connection.id)));
     }
 
--- a/lib/api/controllers/routerController.js
+++ b/lib/api/controllers/routerController.js
@@ -22,8 +22,8 @@
 'use strict';
 
 const
+  errorsManager = require('../../config/error-codes/throw'),
   HttpRouter = require('../core/httpRouter'),
-  { errors: { PluginImplementationError } } = require('kuzzle-common-objects'),
   generateSwagger = require('../core/swagger'),
   jsonToYaml = require('json2yaml');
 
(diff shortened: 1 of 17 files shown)


lib/api/controllers/routerController.js
```diff
@@ -60,11 +64,19 @@ class RouterController {
...
+        'unable_to_remove_connection_invalid_args',
+        JSON.stringify(requestContext.context)));
     }
 
     if (!this.connections.has(requestContext.connection.id)) {
-      return this.kuzzle.log.error(new PluginImplementationError(`Unable to remove connection - unknown connection identifier: ${JSON.stringify(requestContext.connection.id)}`));
+      return this.kuzzle.log.error(errorsManager.getError(
+        'plugins',
+        'runtime',
+        'unable_to_remove_connection_invalid_args',
```
Aschen:
```suggestion
        'unable_to_remove_connection_unknown_connection_identifier',
```

lib/config/error-codes/plugins.json
```diff
@@ -248,6 +248,41 @@
...
+          "class": "PluginImplementationError"
+        },
+        "user_creation": {
+          "code": 23,
+          "message": "An error occurred during the creation of user \"%s\":\n%s",
+          "class": "PluginImplementationError"
+        },
+        "unexpected_return_value": {
+          "code": 24,
+          "message": "Unexpected return value from action %s/%s: expected a Promise",
```
Aschen:
```suggestion
          "message": "Unexpected return value from action \"%s:%s\": expected a Promise",
```

lib/config/error-codes/api.json
```diff
@@ -160,9 +160,14 @@
           "class": "UnauthorizedError"
         },
         "forbidden_action": {
-          "code": 6,
+          "code": 7,
```
scottinet: I thought we had a check at startup ensuring that there were no code duplicates in:
* domain codes
* subdomain codes within a same domain
* error codes within a same subdomain

This change shows that we definitely MUST have one: we cannot change error codes on a whim, once they are released, they must be pretty much unchangeable.

Yoann-Abbes: The check at startup is about the format of the json file
I ll update it

Yoann-Abbes:
@scottinet 
I added to this PR subcodes models (`repositories` and `profile/user/role`...)
I putted that in `security` and `auth` controller for now but i'm not sure about this choice
Is it more relevant to have `models - repositories`, `models - user` etc...

Yoann-Abbes:
> I'm approving this PR but there won't be a Kuzzle release without some code guarding against error code duplicates.

https://github.com/kuzzleio/kuzzle/pull/1387

PR #1384, opened 2019-08-08
Fix dead link in plugin context introduction

Aschen:
## What does this PR do ?
Fix dead link in


Aschen:
The link was correct, the 404 come from the documentation framework bug

PR #1401, opened 2019-08-23
Drop support for Kuzzle Proxy

scottinet:
# Description

The Kuzzle Proxy was meant as a load balancer for Kuzzle cluster nodes. Since then, Kuzzle cluster has been re-architectured and is now masterless, rendering the Kuzzle Proxy project useless, and allowing users to use existing, more robust and more efficient proxies, such as nginx.

It was still supported throughout the entire v1 version of Kuzzle for backward-compatibility reasons, but it is too costly too maintain for too little a gain, so we decided to end its support.

Changes:

* Dropped the dual entrypoint handling. The previous "embedded" entrypoint is now the only remaining one (and the "embedded" name has thus been removed)
* HTTP objects emulating requests coming from the Proxy have been reworked to use a proper class (named `HttpMessage`), allowing for a better separation of concern, but also to remove superfluous JSON.stringify => JSON.parse processes when receiving pre-parsed JSON objects from HttpDataStream streams.
* Dropped the internal broker service, as it was used only for communications between Kuzzle nodes and Kuzzle Proxy

## Other changes

* Dropped handling of whitelists and blacklists for service initializations (obsolete and unused for a long time now)
* Bugfix: HTTP errors (not API errors) weren't correctly forwarded to users (this bugfix will also be implemented in Kuzzle v1)
* Fix unit tests on HTTP requests that were always passing


lib/api/core/httpRouter/index.js
```diff
@@ -135,99 +135,101 @@ class Router {
...
 
-      return;
+    if (!this.routes[message.method]) {
+      return this.routeUnhandledHttpMethod(message, cb);
     }
 
-    httpRequest.url = httpRequest.url.replace(LeadingSlashRegex, '');
+    message.url = message.url.replace(LeadingSlashRegex, '');
+
+    let request;
```
kuzzle: MAJOR Move the declaration of "request" to line 169. [rule](https://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS2392)

kuzzle:
* ![MAJOR][MAJOR] 1 major

scottinet:
@alexandrebouthinon > Adrien already did that. :)

lib/api/core/httpRouter/index.js
```diff
@@ -135,99 +135,105 @@ class Router {
...
+  }
+
+  /**
+   * Checks that an incoming HTTP message is well-formed
+   * @param  {HttpMessage} message
+   */
+  validateMessage (message) {
+    const contentType = message.headers['content-type'];
+
+    if (contentType && !contentType.startsWith('application/json')) {
```
Aschen: Nit (don't know if it's worth it here): `String.indexOf` is faster than `String.startsWith`

lib/api/core/httpRouter/routeHandler.js
```diff
@@ -34,34 +36,54 @@ const
...
     this.handler = null;
+    this._request = null;
     this.url = url;
-    this.request = null;
-    this.httpRequest = httpRequest;
+
     this.data = {
-      requestId: this.httpRequest.requestId
+      requestId: message.requestId,
+      body:message.content,
```
Aschen: Nit
```suggestion
      body: message.content,
```

lib/api/core/httpRouter/routeHandler.js
```diff
@@ -34,34 +36,54 @@ const
...
+        id: message.connection.id,
+        protocol: 'http',
+        url: message.url,
+        headers: message.headers,
+        ips: message.ips
+      }
+    };
+
+    for (const k of Object.keys(message.headers)) {
+      if ( k.toLowerCase() === 'authorization'
```
Aschen: Nit
```suggestion
      if (k.toLowerCase() === 'authorization'
```

lib/api/core/httpRouter/routeHandler.js
```diff
@@ -34,34 +36,54 @@ const
...
+        && message.headers[k].startsWith('Bearer ')
```
Aschen: Nit: same remark about indexOf (Not sure it's worth it either here)

lib/api/core/entrypoints/index.js
```diff
@@ -19,51 +19,529 @@
...
+  Context = require('./context'),
+  debug = require('../../../kuzzleDebug')('kuzzle:entry-point:embedded'),
+  fs = require('fs'),
+  http = require('http'),
+  { HttpProtocol } = require('./protocols/http'),
+  {
+    models: { RequestContext },
+    errors: { ServiceUnavailableError }
+  } = require('kuzzle-common-objects'),
+  MqttProtocol = require('./protocols/mqtt'),
```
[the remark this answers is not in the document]
scottinet: I don't see the point.

lib/api/core/entrypoints/index.js
```diff
@@ -19,51 +19,529 @@
...
+      }
+    }
+
+    // = apache combined
+    const protocol = connection.protocol.toUpperCase();
+    let
+      url,
+      verb = 'DO';
+
+    if (connection.protocol.startsWith('HTTP/')) {
```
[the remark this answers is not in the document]
scottinet: Surprising, but my own benchmarks show the same results. :o

Changing for indexOf. :+1:

lib/api/core/entrypoints/index.js
```diff
@@ -19,51 +19,529 @@
...
+      verb = extra.method;
+      url = extra.url;
+
+      // try to get plain user name, 1st form jwt token if present, then from
+      // basic auth
+      if (user === null && connection.headers.authorization) {
+        try {
+          if (bearerRegexp.test(connection.headers.authorization)) {
```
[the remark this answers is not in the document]
scottinet: Won't work because this can also be written `BeAREr`

But all you need is an indexOf on a lowercased string, which gives similar results to yours. :+1:

lib/api/core/httpRouter/index.js
scottinet: It does.

lib/api/core/httpRouter/routeHandler.js
scottinet: Check our wrapping rules: we add an extra space when conditions are split to align them.

scottinet:
@Aschen > Applied all your reviews, and then some.

lib/api/core/entrypoints/index.js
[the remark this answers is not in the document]
Aschen: As for the Javascript SDK where all the protocols are "requirable" within the same file
```
      {
        MqttProtocol,
        HttpProtocol,
        SocketIOProtocol,
        WebSocketProtocol
      } = require('./protocols');
```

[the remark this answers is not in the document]
scottinet: I would have found this useful if protocols were used in more than 1 place, but here we only need them in the entrypoint class.
I still don't see the point of creating a new intermediary JS file used only to regroup a few require calls that are used only by one other file: to me this adds an unneeded complexity.

PR #1404, opened 2019-08-23
Fix error codes generation

scottinet:
# Description

Error codes where improperly generated.

For instance, the following: domain=2, subdomain=8, error=245
gave the following code: `245`
instead of the expected (int32) 34078965 (hexadecimal: `0x020800f5`, can be verified with `Number(34078965).toString(16)`)

# Other changes

Moved config/throw.js unit tests from the test/api/kuzzle.test.js file to a new test/config/throw.test.js one


Yoann-Abbes:
When should them be int32 ?

PR #1407, opened 2019-08-26
Fix a crash occuring when trying to ping dead sockets

scottinet:
# Description

Fix a rare crash that might occur within the WebSocket entrypoint, when a socket is closed just before a heartbeat occurs. If that happens at just the "right" time, then a heartbeat can start before that socket "close" event is processed and before that socket is marked as dead.

If that happens, pinging that socket results in an exception being thrown. Since it is not caught, this results in an unhandled exception, trapped by Kuzzle and triggering a graceful shutdown.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -47,7 +47,7 @@
     "quotes": [2, "single"],
     "semi": [2, "always"],
     "space-before-blocks": 2,
-    "space-in-parens": [2, "never"],
+    "space-in-parens": [0, "never"],
     "vars-on-top": 2,
     "yoda": [2, "never"]
   },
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -47,7 +47,7 @@
     "quotes": [2, "single"],
     "semi": [2, "always"],
     "space-before-blocks": 2,
-    "space-in-parens": [2, "never"],
+    "space-in-parens": [0, "never"],
```
Aschen: Why ? It's because the following wrapping rule ? `if ( k === 'something')`

scottinet: yup

PR #1417, opened 2019-08-27
Add event on new/removed connection

Aschen:
## What does this PR do ?

PR in `2.x`: https://github.com/kuzzleio/kuzzle/pull/1418

Add two events on new/removed connections:
  - `connection:add`
  - `connection:remove`

The payload is the following:

| Properties   | Type              | Description                    |
| ------------ | ----------------- | ------------------------------ |
| `id`      | <pre>string</pre> | Connection unique ID              |
| `protocol` | <pre>string</pre> | Protocol name (eg: `websocket`, `http`, etc.)      |
| `headers` | <pre>object</pre> | Protocol specific headers                |
| `ips`      | <pre>array</pre> | Array of ips addresses              |

### DIFF
--- a/lib/api/core/entrypoints/embedded/index.js
+++ b/lib/api/core/entrypoints/embedded/index.js
@@ -488,11 +488,9 @@ class EmbeddedEntryPoint extends EntryPoint {
   newConnection (connection) {
     this.clients[connection.id] = connection;
 
-    this.kuzzle.emit('connection:new', JSON.parse(JSON.stringify(connection)));
+    this.kuzzle.emit('connection:new', connection);
 
-    this.kuzzle.router.newConnection(
-      new RequestContext({connection})
-    );
+    this.kuzzle.router.newConnection(new RequestContext({ connection }));
   }
 
   /**
@@ -502,8 +500,10 @@ class EmbeddedEntryPoint extends EntryPoint {
     const connection = this.clients[connectionId];
 
     if (connection) {
-      this.kuzzle.emit('connection:remove', JSON.parse(JSON.stringify(connection)));
-      this.kuzzle.router.removeConnection(new RequestContext({connection}));
+      this.kuzzle.emit('connection:remove', connection);
+
+      this.kuzzle.router.removeConnection(new RequestContext({ connection }));
+
       delete this.clients[connectionId];
     }
   }
--- a/lib/api/core/entrypoints/embedded/index.js
+++ b/lib/api/core/entrypoints/embedded/index.js
@@ -90,7 +90,7 @@ class EmbeddedEntryPoint extends EntryPoint {
   init () {
     // We need to verify the port ourselves, to make sure Node.js won't open
     // a named pipe if the provided port number is a string
-    if (!Number.isInteger(this.config.port)) {
+    if (! Number.isInteger(this.config.port)) {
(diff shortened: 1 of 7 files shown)


lib/api/core/entrypoints/embedded/index.js
```diff
@@ -487,6 +487,9 @@ class EmbeddedEntryPoint extends EntryPoint {
    */
   newConnection (connection) {
     this.clients[connection.id] = connection;
+
+    this.kuzzle.emit('connection:new', JSON.parse(JSON.stringify(connection)));
```
alexandrebouthinon:
```suggestion
    this.kuzzle.emit('connection:new', connection);
```

lib/api/core/entrypoints/embedded/index.js
```diff
@@ -499,6 +502,7 @@ class EmbeddedEntryPoint extends EntryPoint {
     const connection = this.clients[connectionId];
 
     if (connection) {
+      this.kuzzle.emit('connection:remove', JSON.parse(JSON.stringify(connection)));
```
alexandrebouthinon:
```suggestion
      this.kuzzle.emit('connection:remove', connection);
```

lib/api/core/entrypoints/embedded/index.js
scottinet: While I agree that we should make sure that the connection object cannot be altered, I feel that we should prevent unnecessary computation as much as possible.

I see two possibilities:
* make the ClientConnection constructor freezes itself (and its headers object). This needs a bit of investigation to make sure that this can be done without impacting the current code, but if possible, this would be the best option IMO
* emit only if there is [at least 1 listener)(https://nodejs.org/api/events.html#events_emitter_listenercount_eventname) to the connection:new event. This means extra-computation if there are at least one, unfortunately

test/api/core/entrypoints/embedded/index.test.js
```diff
@@ -633,6 +638,49 @@ describe('lib/core/api/core/entrypoints/embedded/index', () => {
         .be.calledOnce()
         .be.calledWithMatch(new RequestContext({connection}));
     });
+
+    it('should dispatch connection:new event', () => {
+      entrypoint.newConnection(connection);
+
+      should(kuzzle.emit).be.calledWithMatch('connection:new', {
```
Yoann-Abbes: wrong wrapping

lib/api/core/entrypoints/embedded/index.js
Aschen: I choose to froze the `ClientConnection` instance.

test/api/core/entrypoints/embedded/clientConnection.test.js
```diff
@@ -4,18 +4,34 @@ const
...
-      const
-        headers = {foo: 'bar'},
-        connection = new ClientConnection('protocol', ['ip1', 'ip2'], headers);
+      should(connection.headers).be.exactly(headers);
+    });
+
+    it('should be frozen', () => {
+      connection.id = 'not-set';
+      connection.protocol = 'not-set';
+      connection.headers.foo = 'not-set'
```
Yoann-Abbes:
```suggestion
      connection.headers.foo = 'not-set';
```

PR #1433, opened 2019-08-29
Fix cache roles and profiles

Yoann-Abbes:
What does this PR do ?

PR on 2-dev https://github.com/kuzzleio/kuzzle/pull/1435

This PR fixes the way that we store a role and a profile after writing on it.
Before that, what was stored was the content of the request, and so was different from what was stored in ES.
For instance, if you have a role
```json
{
  "controllers": {
    "index": {
      "actions": {
        "rez": true
      }
    },
    "document": {
      "actions": {
        "fireball": true
      }
    }
  }
}
```
that you update like
```json
{
  "controllers": {
    "document": {
      "actions": {
        "fireball": false
      }
    }
  }
}
```
getRole responses was
```json
{
  "controllers": {
    "document": {
      "actions": {
        "fireball": false
      }
    }
  }
}
```
And not
```json
{
  "controllers": {
    "document": {
      "actions": {
        "fireball": false
      }
    },
    "index": {
      "actions": {
        "rez": true
      }
    }
  }
}
```
expected

I noticed that while using `getRole` and `searchRoles` : result was different. `searchRoles` has the expected behavior since it don't load it from the cache. 

  - Step 1 :
  - Step 2 :
  - Step 3 :  
  ...

### DIFF
--- a/lib/api/core/models/repositories/profileRepository.js
+++ b/lib/api/core/models/repositories/profileRepository.js
@@ -220,13 +220,14 @@ class ProfileRepository extends Repository {
    * @returns {Promise<Profile>}
    **/
   validateAndSaveProfile (profile, options) {
+    
     if (!profile._id) {
       return Bluebird.reject(errorsManager.getError('missing_profile_id'));
     }
-
+    
     return profile.validateDefinition()
       .then(() => {
-
+            
         if (profile._id === 'anonymous'
           && profile.policies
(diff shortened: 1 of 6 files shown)


jenow:
> If this is a hotfix, you need to bump package.json version, generate new associated package-lock.json and create a new Kuzzle version

Just run `npm version patch` and push

PR #1437, opened 2019-08-30
Fix hydrating a profile with unexisting role

Yoann-Abbes:
## What does this PR do ?

PR on 2dev https://github.com/kuzzleio/kuzzle/pull/1438

This PR prevents to Update/Create/Replace a profile with a role who doesn't exist.
Before that, we could do it, and the route `searchProfiles` returned an error.
Since the list of the profiles in the admin console works with this route, it didn't display the list anymore.

Open the admin console and update a profile with a roleId who does not exist.
Or do it with curl directly.


test/api/core/models/repositories/profileRepository.test.js
```diff
@@ -408,6 +408,18 @@ describe('Test: repositories/profileRepository', () => {
...
 
+    it('should throw a NotFoundError when trying to write unexisting role in profile. ', () => {
+      const invalidProfile = new Profile();
+      invalidProfile._id = 'awesomeProfile';
+      invalidProfile.policies = [{roleId: 'notSoAwesomeRole'}];
+
+      kuzzle.repositories.role.loadRoles = sinon.stub().rejects();
+      return profileRepository.validateAndSaveProfile(invalidProfile)
+        .catch(e => {
+          should(e).be.instanceOf(NotFoundError, { message: 'Unable to hydrate the profile awesomeRole: missing role(s) in the database'});
```
scottinet: Shouldn't this be "unable to hydrate the profile awesomeProfile"?

scottinet: If so, I don't get why this test passes. You probably should add a `.then` handler throwing an exception, just in case the function resolves when it shouldn't

Aschen: Or you can use `rejectedWith`
```
const promise = profileRepository.validateAndSaveProfile(invalidProfile);

return should(promise).be.rejectedWith(NotFoundError, {
  message: 'Unable to hydrate the profile awesomeRole: missing role(s) in the database'
});
```

lib/api/core/models/repositories/profileRepository.js
```diff
@@ -234,7 +234,6 @@ class ProfileRepository extends Repository {
         ) {
           errorsManager.throw('missing_anonymous_role');
         }
-        this.profiles[profile._id] = profile;
```
Aschen: Why removing the RAM cache ?

Yoann-Abbes: I didnt removed it, look the entire file

PR #1455, opened 2019-09-09
fix init indices containing a dot

benoitvidis:
## What does this PR do ?

This PR fixes the case where Kuzzle would crash on start because of some existing elasticsearch indices containing a dot (.) in their name.

NB: does not apply to 2-dev

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,6 +98,8 @@ jobs:
       cache:
         directories:
           - $HOME/.gem/specs
+      env:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -98,6 +98,8 @@ jobs:
       cache:
         directories:
           - $HOME/.gem/specs
+      env:
+        - REPOSITORIES=kuzzle-how-to,sdk-javascript-5,sdk-javascript-6,plugin-s3-1,sdk-php-3,plugin-cloudinary-1,sdk-cpp-1,sdk-csharp-1,sdk-go-1,sdk-go-2,sdk-java-1,sdk-java-2,sdk-java-3,kuzzle-1,sdk-android-3,plugin-multi-tenancy-1
```
benoitvidis: Do not test 2-dev dead links on 1-dev branch

PR #1463, opened 2019-09-13
Fix an unhandled promise rejection

scottinet:
# Description

Most API routes in the admin controller can respond immediately if the `refresh` flag is not set. But in this case, the underlying promises created are unhandled if they are rejected, which makes Kuzzle crash in development environments.


scottinet:
This fix for Kuzzle v2: https://github.com/kuzzleio/kuzzle/pull/1464

PR #1468, opened 2019-09-17
Fix CLI start --enable-plugin command

alexandrebouthinon:
What this PR do?
================

In the `run.sh` script the `exec` command does not interpolate double
quote environment variable globbing.
This behavior results to this error on Kuzzle startup:

```
error: unknown option '--enable-plugins "kuzzle-plugin-cluster"'
```

### DIFF
--- a/bin/commands/start.js
+++ b/bin/commands/start.js
@@ -64,7 +64,10 @@ function commandStart (options = {}) {
   }
 
   if (options.enablePlugins) {
-    kuzzleParams.additionalPlugins = options.enablePlugins.split(',').map(x => x.trim());
+    kuzzleParams.additionalPlugins = options.enablePlugins
+      .trim()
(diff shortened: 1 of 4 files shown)


bin/commands/start.js
```diff
@@ -64,7 +64,10 @@ function commandStart (options = {}) {
   }
 
   if (options.enablePlugins) {
-    kuzzleParams.additionalPlugins = options.enablePlugins.split(',').map(x => x.trim());
+    kuzzleParams.additionalPlugins = options.enablePlugins
+      .trim()
+      .split(',')
+      .map(x => x.trim().replace (/(^")|("$)/g, ''));
```
benoitvidis:
```suggestion
      .map(x => x.trim().replace(/(^")|("$)/g, ''));
```

PR #1472, opened 2019-09-19
Remove CLI commands

Aschen:
## What does this PR do ?

Remove the CLI commands since the CLI is now externalized.  

Kuzzle starting script is now `start-kuzzle-server` and use only `yargs` for command line parsing.

  - Step 1 : Just start Kuzzle as usual with Docker compsoe

### Other changes

 - Move the Docker compose file to the directory root
 - Change the `commonMapping.dynamic` option to true for the tests


bin/start-kuzzle-server
```diff
@@ -0,0 +1,92 @@
...
+
+  if (options.fixtures) {
+    kuzzleParams.fixtures = loadJson(options.fixtures);
+  }
+
+  if (options.securities) {
+    kuzzleParams.securities = loadJson(options.securities);
+  }
+
+  if (options.enablePlugins) {
```
alexandrebouthinon: This block is outdated. (see: https://github.com/kuzzleio/kuzzle/blob/2-dev/bin/commands/start.js)

PR #1482, opened 2019-10-01
Fix socketio connections leak

scottinet:
# Description

Sockets closed by socket.io are correctly caught by our socketio handler, which starts by removing the connection identifier from the entrypoint, and then calls `entrypoint.removeConnection`.

Since the connection has already been removed, the entrypoint believes that it's an unmanaged one, and skips it. 

This means that resources attached to a connection (usage statistics, real-time filters, router connection context, ...) aren't freed.

Also, the `connection:remove` event isn't triggered, making the prometheus plugin thinks that socketio connections are never closed.

# Other changes

* socket.io layer: make the `sockets` cache a Map instead of a raw object (more adapted to that kind of very volatile data)
* entrypoint:
  * make the clients cache a Map instead of a raw object
  * rename `clients` into `_clients` as a warning that this shouldn't be trifled with 
  * new debug entries added

### DIFF
--- a/lib/api/core/entrypoints/embedded/protocols/socketio.js
+++ b/lib/api/core/entrypoints/embedded/protocols/socketio.js
@@ -26,11 +26,13 @@ const
   debug = require('../../../../../kuzzleDebug')('kuzzle:entry-point:protocols:socketio'),
   ClientConnection = require('../clientConnection'),
   Protocol = require('./protocol'),
-  Request = require('kuzzle-common-objects').Request,
   {
-    KuzzleError,
-    InternalError: KuzzleInternalError
-  } = require('kuzzle-common-objects').errors;
+    Request,
+    errors: {
+      KuzzleError,
+      InternalError
+    }
+  } = require('kuzzle-common-objects');
 
 /**
  * @class SocketIoProtocol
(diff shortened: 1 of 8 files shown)


lib/api/core/entrypoints/embedded/protocols/socketio.js
```diff
@@ -123,16 +123,15 @@ class SocketIoProtocol extends Protocol {
...
     debug('[%s] onClientDisconnection', clientId);
 
-    if (this.sockets[clientId]) {
-      delete this.sockets[clientId];
-      delete this.entryPoint.clients[clientId];
-
-      return this.entryPoint.removeConnection(clientId);
+    if (this.sockets.has(clientId)) {
+      this.sockets.delete(clientId);
+      this.entryPoint.removeConnection(clientId);
```
scottinet: The actual fix is here: we don't delete `this.entryPoint.clients` entries before invoking `removeConnection` anymore

lib/api/core/entrypoints/embedded/protocols/socketio.js
```diff
@@ -166,36 +165,40 @@ class SocketIoProtocol extends Protocol {
   }
 
   notify(data) {
-    debug('notify: %a', data);
+    const socket = this.sockets.get(data.connectionId);
 
-    if (this.sockets[data.connectionId]) {
+    if (socket) {
+      debug('notify: %a', data);
       data.channels.forEach(channel => {
```
Aschen: Using `for (let i..)` is better for performance and IMHO this can be considered as a critical section of code

scottinet: I took a look at the whole code and modernized this snippet and other parts as well (see my boyscout commit for a list of changes)

lib/api/core/entrypoints/embedded/protocols/socketio.js
```diff
@@ -26,11 +26,13 @@ const
...
   Protocol = require('./protocol'),
-  Request = require('kuzzle-common-objects').Request,
   {
-    KuzzleError,
-    InternalError: KuzzleInternalError
-  } = require('kuzzle-common-objects').errors;
+    Request,
+    errors: {
+      KuzzleError,
+      InternalError
```
kuzzle: MAJOR Remove this override of "InternalError". [rule](https://sonarqube.kaliop.net/coding_rules#rule_key=javascript%3AS2424)

kuzzle:
* ![MAJOR][MAJOR] 1 major

PR #1484, opened 2019-10-04
Upgrade script

scottinet:
# Description

Add a `bin/upgrade` script allowing to upgrade from a major version of Kuzzle to another.

This script lays the groundwork for version upgrades, and its architecture should be able to support multi-version migrations.

I've abundantly tested it... but manually. I'll work on unit tests if I find the time. The documentation is written, though.

Architecture principles:
* an `UpgradeContext` object is created by the main script entrypoint, and passed around to migration sub-programs: it log functions (all logged texts are duplicated on the console and in a file), configuration, and a customized version of [inquirer](https://www.npmjs.com/package/inquirer#methods), the CLI module I'm using 
* connectors are singletons meant to access to resources if needed, but without having to re-require informations from users (for now, there is only a connector to elasticsearch)
* the `bin/.upgrades/version` contains directories, each named `v???`, where `???` is the version number of Kuzzle that can be upgraded by the directory. For now, there is only a `v1` directory
* migration scripts should be runnable any number of times without error

# Specifics to the migration from v1 to v2

This migration is nearly all about upgrading from ES5 to ES7, and this means changing almost everything in the storage.

It also handles in-place migration vs. new stack migration.

This script upgrades:
* elasticsearch:
  * aliases
  * indexes not kuzzle-related that shouldn't be upgraded
  * collections in indexes that are not kuzzle-related
  * the destruction of the obsolete metadata mappings (active and deletedAt)
  * uppercases in collection names
* redis
* configuration files

# Other changes

The list of breaking changes on configuration files wasn't exhaustive. I've also reworked that list to make it clearer.

# How to test

Good question. I've tried to think of every scenario, but as usual, we'll likely come across clients with more imagination than us.

I'll try to duplicate client infrastructures and try the script on them to see how it fares. I invite you to do the same, and to try different, possibly intricated scenarios.


scottinet:
There is a mapping conflict when starting Kuzzle v2 after an upgrade from Kuzzle v1, probably when trying to upgrade either the %kuzzle.config or the %kuzzle.roles mappings.

scottinet:
Fixed: the migration now correctly acts as a Kuzzle v2 index bootstrapper, preventing the regular bootstrap to overwrite existing informations, and preventing mappings conflicts.

bin/.upgrades/versions/v1/upgradeStorage.js
```diff
@@ -0,0 +1,389 @@
...
+  const prefix = index[0] === '%' ? '' : PUBLIC_PREFIX;
+
+  return `${prefix}${index}${NAME_SEPARATOR}${collection}`;
+}
+
+async function moveData (source, target, index, collection, newIndex) {
+  let page = await source.search({
+    index,
+    type: collection,
+    body: { sort: [ '_doc' ] },
```
Aschen: Why sorting on `_doc` and not on `_id` ?

bin/.upgrades/versions/v1/upgradeStorage.js
```diff
@@ -0,0 +1,389 @@
...
+  // remove obsolete mapping properties
+  if (mappings.properties && mappings.properties._kuzzle_info) {
+    delete mappings.properties._kuzzle_info.properties.active;
+    delete mappings.properties._kuzzle_info.properties.deletedAt;
+  }
+
+  await target.indices.putMapping({
+    index: newIndex,
+    body: {
+      properties: mappings.properties,
```
Aschen: Here we should handle the `_kuzzle_info` separately because we have record of clients using an incorrect version of Kuzzle Metadata mappings (`text` instead of `keyword`).  
A solution could be to try to detect if Kuzzle Metadata mappings are different than the default ones and if so propose to the user to change mappings for the default ones

bin/.upgrades/versions/v1/upgradeStorage.js
```diff
@@ -0,0 +1,389 @@
...
+      dynamic: mappings.dynamic || false,
+      _meta: mappings._meta
+    }
+  });
+}
+
+async function createNewIndex (source, target, index, collection, newIndex) {
```
Aschen: `source`, `index` and `collection` are not used in this method

bin/.upgrades/versions/v1/upgradeStorage.js
```diff
@@ -0,0 +1,389 @@
...
+  const exists = await target.indices.exists({ index: newIndex });
+
+  if (exists.body) {
+    await target.indices.delete({ index:newIndex });
```
Aschen:
```suggestion
    await target.indices.delete({ index: newIndex });
```

bin/.upgrades/versions/v1/upgradeStorage.js
```diff
@@ -0,0 +1,389 @@
...
+  }
+
+  await target.indices.create({ index: newIndex });
+}
+
+async function upgrade (source, target, index, collection, newIndex) {
+  await createNewIndex(source, target, index, collection, newIndex);
+  await upgradeMappings(source, target, index, collection, newIndex);
+  return await moveData(source, target, index, collection, newIndex);
```
Aschen: IMHO, there is no need to `await` the promise here, the caller should decide if he want to `await` this method or not

bin/.upgrades/versions/v1/upgradeStorage.js
```diff
@@ -0,0 +1,389 @@
...
+async function upgradeInternalStorage (context, source, target) {
+  const
+    config = context.config.services.storageEngine.internalIndex,
+    index = `${INTERNAL_PREFIX}${config.name}`,
+    mapconfig = config.collections,
+    collections = {
+      validations: mapconfig.validations,
+      config: mapconfig.config,
+      roles: mapconfig.roles,
+      users: null,
```
Aschen: What about the users ? :thinking:

bin/.upgrades/versions/v1/upgradeStorage.js
```diff
@@ -0,0 +1,389 @@
...
+      profiles: mapconfig.profiles
+    };
+
+  for (const [collection, mappings] of Object.entries(collections)) {
+    const newIndex = getNewIndexName(index, collection);
+    let total;
+
+    if (mappings) {
+      await createNewIndex(source, target, index, collection, newIndex);
+      await target.indices.putMapping({ index: newIndex, body: mappings });
```
Aschen: Why don't you use `upgradeMappings` ?

bin/.upgrades/versions/v1/upgradeStorage.js
```diff
@@ -0,0 +1,389 @@
...
+
+  for (const index of indexes) {
+    const
+      plugin = index.split(':')[1],
+      newIndexBase = `%plugin-${plugin}${NAME_SEPARATOR}`,
+      mappings = await source.indices.getMapping({ index }),
+      collections = Object.keys(mappings.body[index].mappings);
+
+    for (const collection of collections) {
+      const newIndex = newIndexBase + collection;
```
Aschen:
```suggestion
      const newIndex = `${newIndexBase}${collection}`;
```
I know concatenation was enough but IMHO interpolation improve readability

bin/.upgrades/versions/v1/upgradeStorage.js
scottinet: This is not really a sort, rather a performance trick when scrollign through an entire collection.

See: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#request-body-search-scroll

(just before the "Keeping the search context alive" chapter)

bin/.upgrades/versions/v1/upgradeStorage.js
```diff
@@ -0,0 +1,389 @@
...
+  _ = require('lodash'),
+  getESConnector = require('../../connectors/es');
+
+const
+  INTERNAL_PREFIX = '%',
+  PUBLIC_PREFIX = '&',
+  NAME_SEPARATOR = '.';
+
+function getNewIndexName (index, collection) {
+  const prefix = index[0] === '%' ? '' : PUBLIC_PREFIX;
```
Aschen: Nit: wrong formatting for ternary expression

scottinet: How so?

bin/.upgrades/versions/v1/upgradeStorage.js
scottinet: This is one of the few examples where I personally find concatenation more readable. I guess this is a subjective matter.

bin/.upgrades/versions/v1/upgradeStorage.js
scottinet: I couldn't find it in Elasticsearch API, did I miss it?

bin/.upgrades/versions/v1/upgradeStorage.js
scottinet: Users mappings are customizable by administrators, to fit a business needs. This is the only internal index mapping that we import instead of bootstrapping it.

bin/.upgrades/versions/v1/upgradeStorage.js
scottinet: The problem is that this migation script is way faster than Elasticsearch, so if we don't wait, we saturate ES buffers very quickly. 
So I prefer to make this script slow but safe: I think that our clients would prefer that over a faster script that needs to be re-run with different options if their dataset are too large for their target ES server.

bin/.upgrades/versions/v1/upgradeStorage.js
scottinet: Ah, this is because I changed it last minute to solve the internal index mappings conflicts. Sorry :sweat: 

This will be fixed in an incoming commit (I'm working on the cache migration, I forgot it...)

bin/.upgrades/versions/v1/upgradeStorage.js
scottinet: Hm... Good catch. And that's a tricky one: I'll give it some thought after I finish the cache migration, but my first instinct is to automatically (and silently) convert the mapping: only the `author` field should be impacted, and in older versions of Kuzzle it was set to `text` because, well, we let ES set the mappings for kuzzle's metadata automatically. 
This time is as good as any to fix that error definitely.

bin/.upgrades/versions/v1/upgradeStorage.js
alexandrebouthinon: What if `newIndexBase` and `collection` are integers? :stuck_out_tongue:

bin/.upgrades/versions/v1/upgradeStorage.js
alexandrebouthinon: @Aschen: Are you talking about wrapping? If so, IMO this line is short enough

bin/.upgrades/versions/v1/upgradeStorage.js
scottinet: Fixed.

bin/.upgrades/versions/v1/upgradeStorage.js
Aschen: I though that the rule apply for every ternary expression

bin/.upgrades/versions/v1/upgradeStorage.js
Aschen: Yes, I didn't say it but it's nitpicking ;)

bin/.upgrades/versions/v1/upgradeStorage.js
Aschen: I was talking about the method you defined but.. nevermind :expressionless:

bin/.upgrades/versions/v1/upgradeStorage.js
Aschen: I didn't know this one!

bin/.upgrades/versions/v1/upgradeStorage.js
scottinet: Ah :)
I don't use it because `upgradeMappings`, as it name suggests, is meant to import a mapping from a Kuzzle v1 source, make it compatible with Kuzzle v2, and then push it to the v2 source.

But we don't want that for internal indexes: v1 mappings should always be considered obsolete, so I take the mappings from the bootstrap configuration directly, and I ignore the v1 mappings.

bin/.upgrades/versions/v1/upgradeStorage.js
scottinet: Well, it's defined in our "how to wrap code" document :wink:

scottinet:
@Aschen > I fixed the _kuzzle_info mappings issue by discarding those from Kuzzle v1 and replacing themp with the ones defined in the Kuzzle configuration files.

Also, since you last reviewed this PR, I added cache migration, so you might want to review that part too.

bin/.upgrades/lib/context.js
```diff
@@ -37,10 +37,19 @@ class Version {
 
 class UpgradeContext {
   constructor(args) {
-    this.config = null;
-    this.log = new Logger(args);
-    this.inquire = inquirer;
-    this.version = null;
+    // copy constructor
+    if (args instanceof UpgradeContext) {
```
Aschen: image

scottinet: homer

scottinet:
@Aschen @alexandrebouthinon > I've added a progress bar, to prevent frustration on long tasks (e.g. large quantity of users). You might want to review this commit.

scottinet:
New change: collections can have uppercase letters, in ES5, but those are forbidden from index names.

So the script now forces collections to lowercase. This change is also applied to the profiles policies, making sure that the security configuration stays functional after the upgrade.

bin/.upgrades/versions/v1/upgradeStorage.js
```diff
@@ -57,14 +63,16 @@ async function moveData (context, index, collection, newIndex) {
         delete doc._source._kuzzle_info.deletedAt;
       }
 
-      bulk.push({ index: { _index: newIndex, _id: doc._id } });
+      bulk.push({ create: { _index: newIndex, _id: doc._id } });
```
Aschen: Why changing this ? Aren't we supposed to replace existing data ?

scottinet:
Should be one of the last changes, I've almost finished my tests: in-place storage migration can now be made on an ES 5 instance, allowing an easier migration to ES 7 for our clients with only Kuzzle data who can't migrate to a new stack.

bin/.upgrades/versions/v1/upgradeStorage.js
scottinet: It doesn't matter: the [createNewIndex](https://github.com/kuzzleio/kuzzle/pull/1484/files#diff-9fcbc71bf9075c1d138dee6216312ec0R147) function deletes any previously existing index.

As for why I did change this... I made performance tests and thought it had an impact. It doesn't, and I forgot to change it back.... but then again, this makes no difference in our very specific case.

PR #1494, opened 2019-10-15
Check for SDK compatibility against current Kuzzle version

Aschen:
## What does this PR do ?

Now Kuzzle will check the `volatile.sdkName` header to ensure that the current SDK is compatible against the current Kuzzle version.  

For Kuzzle v1 SDKs, the `sdkName` property is not present. So when the property is missing, we assume that the SDK is compatible.

For Kuzzle v2 SDKs, a compatibility matrix has been added. It works with simplified semver ranges.

Example  compat matrix:
```
{
  "js": "<7" // this version of Kuzzle is compatible with the JS SDKs prior to the 7.x.x version
}
```

Example `sdkName` volatile property: `js@7.4.2`

PR #1495, opened 2019-10-15
Check for SDK compatibility against current Kuzzle version

Aschen:
## What does this PR do ?

v1: #1494 

Now Kuzzle will check the `volatile.sdkName` header to ensure that the current SDK is compatible against the current Kuzzle version.  

For Kuzzle v1 SDKs, the `sdkVersion` property is present. So when the property is present, we assume that the SDK is incompatible.

For Kuzzle v2 SDKs, a compatibility matrix has been added. It's checking the `volatile.sdkVersion` property and works with simplified semver ranges.

Example  compat matrix:
```
{
  "js": ">=7" // this version of Kuzzle is compatible with the JS SDKs beginning at the 7.x.x version
}
```

Example `sdkName` volatile property: `js@7.4.2`

### Other changes

 - upgrade Elasticsearch version to 7.4.0


lib/api/controllers/funnelController.js
```diff
@@ -710,4 +765,22 @@ function doAction(controller, request) {
...
+/**
+ * Very straightforward function to check only if the version satisfies
+ * semver range major version
+ *
+ * @param {String} version
+ * @param {String} range
+ *
+ * @returns {Boolean}
+ */
+function satisfiesMajor (version, range) {
```
scottinet: Why not using [semver](https://www.npmjs.com/package/semver), which is already included in our package.json?

The way it's currently coded, the compatibility matrix is misleading. I, for one, would have expected a semver range, and I expect to be able to enter something like `>=7 <9` for instance, which is a probable value.

Aschen: I choose to use some kind of simplified semver range for now because semver is really slow and this method is called whenever Kuzzle receive a request.
```
Semver x 176,414 ops/sec ±4.15% (85 runs sampled)
Manual: only major x 776,075,210 ops/sec ±2.42% (90 runs sampled)
```
We won't need support for this kind of range (`>=7 <9`) until v3.

Do you think it's unnecessary ?

scottinet: Indeed, this is slow.

My real problem is the syntax similarity with semver, I find that _very_ misleading for future maintenance.

Since the matrix is internal, I think that using a custom format that has nothing to do with semver might be a better idea.

Something like:

```
{
  "js": { "min": 7 },
  "csharp": { "max": 2},
  "cpp": { "max": 2},
  "java": { "max": 3},
  "android": { "max": 5},
  "go": { "max": 3},
  "php": { "max": 4}
}
```

Of course, we would be able to have both a min and a max version. This should make the version check even faster while unambiguously being different from semver.

What do you think?

Aschen:
> you need to remove the auth-local plugin update from this PR
> 
> apart from that: +1

Done @scottinet

PR #1505, opened 2019-10-24
Fix delete profile while being used by user

Yoann-Abbes:
## What does this PR do ?

This PR fixes delete profiles.
A typo error allowed to delete a profile while being used by some user.

It resolves https://jira.kaliop.net/browse/KZL-1399

  - Step 1 :
  - Step 2 :
  - Step 3 :  
  ...

### DIFF
--- a/features/kuzzle.feature
+++ b/features/kuzzle.feature
@@ -573,7 +573,6 @@ Feature: Kuzzle functional tests
     Then I create a user "user1" with id "user1-id"
     Then I'm not able to delete profile with id "profile1"
     Then I'm able to find the profile with id "profile1"
-    Then I log out
 
   @security
   Scenario: search and update profiles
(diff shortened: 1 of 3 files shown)


features/kuzzle.feature
```diff
@@ -564,6 +564,17 @@ Feature: Kuzzle functional tests
...
+  @security
+  Scenario: can't delete profile assigned to a user
+    When I create a user "useradmin" with id "useradmin-id"
+    Then I log in as useradmin:testpwd expiring in 1h
+    Then I create a new role "role1" with id "role1"
+    Then I create a new profile "profile1" with id "profile1"
+    Then I create a user "user1" with id "user1-id"
+    Then I'm not able to delete profile with id "profile1"
+    Then I'm able to find the profile with id "profile1"
+    Then I log out
```
Aschen: Is this necessary?

Yoann-Abbes: Since I have to login at the beginning of the scenario (in order to create a role/profile/user) I found it just more coherent to end it by logging out

Aschen: The logout step is handled by Cucumber hooks I think. Usually, we do all the cleaning stuff in hooks, can you check if the tests still pass without the logout action?

PR #1511, opened 2019-10-31
Hotfix/subscribe dead connection

benoitvidis:
Fixes an edge-case in which Kuzzle would crash when trying to subscribe a just-dead connection:

```
{ InternalError: Cannot read property 'roomId' of undefined at kuzzle.hotelClerk.addSubscription.then.result (/var/app/lib/api/controllers/realtimeController.js:63:18...
```

### Other

* settle `socket.io` version to `2.2.0` to keep node 6.x compatibility
* small enhancements to tests script


Aschen:
image

PR #1536, opened 2019-11-28
Prevent overloads caused by pre-funnel pipes

scottinet:
# Description

When plugin developers design pipe functions, they can either accept a callback, or return a promise.
The latter was added to make the plugin development experience better, but this can make Kuzzle brittle against flooding if a pipe is added to a pre-funnel event, meaning promises are created on the event loop without any sort of control (e.g. `http:*` events).

This PR solves the problem by making the following changes:

* the plugin's manager `pipe` method now only work with callbacks
* an overload-protection mechanism has been added to pipe executions, similar to the existing one in the funnel. It can be controlled by two new properties in Kuzzle's configuration file (`plugins.common.maxConcurrentPipes` and `plugins.common.pipesBufferSize`)
* the main entry point for triggering pipe events (`kuzzle.pipe(event, ...data)`) now accepts a callback argument, and creates+returns a promise only if NO callback has been provided
* `kuzzle.pipe` is now invoked with a callback by events triggered before the funnel's overload-protection mechanism

# Other changes

* Pipe events accepting multiple arguments now correctly pass all arguments from a pipe function to another

### DIFF
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -102,11 +102,25 @@
     //   * initTimeout:
     //        Maximum execution time (in milliseconds)
     //        of a plugin initialization
+    //   * maxConcurrentPipes:
+    //        Maximum number of pipes that can be executed in parallel
+    //        New pipes submitted while the maximum number of pipes is met are
+    //        delayed for later execution.
+    //        This parameter controls is used to limit the stress put on the
+    //        event loop, allowing for Kuzzle to process pipes faster, and to
+    //        protect it from performances degradation if an abnormal number of
+    //        pipes are submitted.
+    //        (timers do not start while a pipe is hold back)
+    //   * pipesBufferSize:
+    //        Maximum number of pipes that can be delayed. If full, new pipes
+    //        are rejected.
     "common": {
       "bootstrapLockTimeout": 5000,
       "pipeWarnTime": 40,
       "pipeTimeout": 250,
-      "initTimeout": 2000
+      "initTimeout": 2000,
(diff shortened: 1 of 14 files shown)


.kuzzlerc.sample
```diff
@@ -102,11 +102,25 @@
     //   * initTimeout:
     //        Maximum execution time (in milliseconds)
     //        of a plugin initialization
+    //   * maxConcurrentPipes:
+    //        Maximum number of pipes that can be executed in parallel
```
Aschen:
```suggestion
    //        Maximum number of pipes that can be executed in parallel.
```

lib/core/plugins/pipeRunner.js
```diff
@@ -0,0 +1,101 @@
...
+  Denque = require('denque'),
+  async = require('async'),
+  {errors: {KuzzleError}} = require('kuzzle-common-objects'),
+  errorsManager = require('../../util/errors').wrap('plugin', 'runtime');
+
+/**
+ * @private
+ * @class PipeChain
+ */
+class PipeChain {
```
Aschen: Does this class declaration have any impact on v8 performances or it's just for readability and maintainability?

scottinet: Both: declaring a class and not addind new properties to it on the fly allows turbofan to convert it to a C++ class with direct access (i.e. 0 cost) to its properties.

And it's also more readable to manipulate class instances for stored objects, because when doing maintenance you don't have to spend time wondering what properties that object might have

PR #1546, opened 2019-12-11
Requests rate limit per user

scottinet:
# Description

Introduce configurable requests rate limit:

* requests rate limit are per node, per second and per user. This means that a rate limit imposed to the anonymous profile will impact all anonymous requests, since those are all attached to the same user id (this behavior has been documented)
* the rate limit is a new profile property, meaning that it can easily target only a selection of users if needed
* unless specified otherwise, the rate limit of a profile is set to `0` (= no limit)
* when restricting rights (e.g. using the createFirstAdmin API route), rate limits are imposed as follows: admin: no limit, default: 10, anonymous: 200
* the `auth:logout` API route is never restricted when requested by authenticated users (since it has no sense that anonymous users execute that route, it stays limited for them)
* the `auth:login` API route is NOT impacted by the rate limit
* requests exceeding the rate limit are rejected with the new `TooManyRequestsError` error class introduced in kuzzle-common-objects@4.1.0 (error status: 429).

The `auth:login` API route is a special case: we do not want it to be unavailable to legitimate users if non-authenticated ones flood Kuzzle in the hope of preventing users to log in. So it is excluded from the rate limit system.
Instead, a new property has been added in the kuzzlerc configuration file: `loginsPerSecond`. This controls how many logins per second can be performed by any single network socket (not user accounts).
By default, it is set to `1`, as I don't see how allowing more than 1 login per second and per network connection is beneficial. 
This limit cannot be set to `0`, there is no "no limit" setting for that one. 

# Other changes

* The funnel now rejects requests without a controller or without an action immediately, instead of waiting for a slot to open and checking rights, as such requests could never be executed

# Boyscout

- asyncify methods


lib/core/models/repositories/repository.js
```diff
@@ -88,31 +88,34 @@ class Repository {
...
-          if (response._source) {
-            Object.assign(dto, response._source, {_id: response._id});
-          }
-          else {
-            Object.assign(dto, response);
-          }
+    try {
+      response = await this.indexStorage.get(this.collection, id);
+    }
+    catch(error) {
```
Aschen: Nit: 
```suggestion
    catch (error) {
```

lib/api/funnel.js
```diff
@@ -244,13 +260,25 @@ class Funnel {
...
       callback(processError.get('connection_dropped'), request);
-      return 0;
+      return 1;
     }
 
     debug('Starting request %s: %a', request.id, request.input);
 
+    let _request;
+
     this.checkRights(request)
```
Aschen: Why you didn't use `async/await` instead of a callback chain here?

lib/api/funnel.js
```diff
@@ -353,102 +382,87 @@ class Funnel {
...
+        this.getEventName(_request, 'before'),
+        _request);
+
+      const responseData = await doAction(controller, _request);
+
+      _request.setResult(
+        responseData,
+        { status: _request.status === 102 ? 200 : _request.status });
+
+      if ( !this.isNativeController(_request) && !_request.response.raw) {
```
Aschen: Nit
```suggestion
      if (!this.isNativeController(_request) && !_request.response.raw) {
```

lib/api/funnel.js
```diff
@@ -493,76 +508,55 @@ class Funnel {
...
-  executePluginRequest (request) {
-    return Bluebird.resolve()
-      .then(() => doAction(this.getController(request), request))
-      .catch(e => {
-        this.handleErrorDump(e);
-        return Bluebird.reject(e);
-      });
+  async executePluginRequest (request) {
+    try {
+      return await doAction(this.getController(request), request);
```
Aschen: Why do you await the promise before returning a new one ?

lib/api/funnel.js
```diff
@@ -493,76 +508,55 @@ class Funnel {
...
+    }
+    catch(e) {
```
Aschen: Nit:
```suggestion
    catch (e) {
```

doc/2/guides/essentials/security/index.md
```diff
@@ -107,7 +107,32 @@ curl -X GET 'http://localhost:7512/?pretty'
...
+
+### Rate limit
+
+<SinceBadge version="2.0.3" />
+
+The rate limit parameter controls how many API requests a user can send, **per second and per node**. Further requests made by a user that exceed the limit are rejected with a `429 Too Many Requests` error.  
+
+If no rate limit is defined, or if it is set to 0, then no limit is applied.
+
+If a user has multiple profiles with rate limits, the most permissive limit applies.
```
Yoann-Abbes:
```suggestion
If a user has several profiles with rate limits, the most permissive limit applies.
```
And, is the most permissive overriding the rest and not the most restrictive ?

lib/api/funnel.js
scottinet: Because the `execute` method is still behind the concurrent request limit wall, it mustn't create a promise until a slot is available.

This promise chain is created only after we request for a slot.

And we don't return it either because this method's callers don't expect a promise.

doc/2/guides/essentials/security/index.md
```diff
@@ -107,7 +107,32 @@ curl -X GET 'http://localhost:7512/?pretty'
...
-A `profile` definition is a JSON object that contains an array of policies, each composed of a roleId and an array of restrictions:
+A `profile` definition is a JSON object that contains an optional rate limit parameter, and an array of policies.
+
+### Rate limit
+
+<SinceBadge version="2.0.3" />
+
+The rate limit parameter controls how many API requests a user can send, **per second and per node**. Further requests made by a user that exceed the limit are rejected with a `429 Too Many Requests` error.  
+
+If no rate limit is defined, or if it is set to 0, then no limit is applied.
```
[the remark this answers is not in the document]
scottinet: Ah, good idea.

lib/api/rateLimiter.js
```diff
@@ -0,0 +1,88 @@
...
+      if (_id !== '-1' && controller === 'auth' && action === 'logout') {
+        return true;
+      }
+
+      const profiles = await this.profiles.loadProfiles(profileIds);
+
+      for (let i = 0; i < profiles.length; i++) {
+        const { rateLimit = 0 } = profiles[i];
+
+        if (limit === 0 || rateLimit === 0) {
```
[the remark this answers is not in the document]
scottinet: 1. we're inside a loop scanning a user's profiles, so `limit` might have been set to 0 by a previous iteration
2. to apply the "if more than 1 rateLimit is found, the most permissive applies" rule. If I set `limit = rateLimit` directly, then the last rateLimit found is applied instead.

lib/api/funnel.js
scottinet: to catch exceptions thrown by doAction

doc/2/guides/essentials/security/index.md
scottinet: Suggestion applied.

To answer your question: because in a whitelist strategy, no rights are granted by default, and roles/profiles give only the necessary rights. 
Following that line of thought, applying rate limits is, in fact, the same as giving the right to send a certain amount of requests per second.
So, for example, if a user has 2 profiles with rate limits of 200 and 50, then this is the same as  giving the right to send 200 reqs/s and the right to send 50 reqs/s. 
So I think it makes better sense to use the most permissive limit.

scottinet:
@Aschen > honestly I cannot tell if this is a new feature or an enhancement. I settled for the latter because I told myself that the "profile" feature already exists, and that the rateLimit is merely a new property added to it.

Aschen:
>  settled for the latter because I told myself that the "profile" feature already exists, and that the rate

Hmm yes we can just keep the `enhancement`, so we can wait a month before release a minor on Kuzzle v2 :)

doc/2/guides/essentials/security/index.md
Yoann-Abbes: Last question: why applying this limit to the profiles and not directly to the users ?

Aschen: Because in this way we can apply rate limits on group of users instead of individual users

Aschen:
The `Since` badges of this feature are outdated

PR #1549, opened 2019-12-13
Receive the document updated content when calling document:update

Yoann-Abbes:
## What does this PR do ?

Before this PR:
We could not retrieve the entire updated document in the request response.
Plus, Kuzzle was doing an extra request to ES in order to send notifications.

Now:
In Kuzzle, we passe the `source` option to ES when we update the document.
Doing that there is no more an extra request.
Finally, users can specify the option `source` in their request to receive the updated document in the request response.

Jira [Ticket](https://jira.kaliop.net/browse/KZL-1407) 

`http://localhost:7512/index/test/document/_update?source=true`
see the updated document in `result._source`

`http://localhost:7512/index/test/document/_update`
do not send `result._source`

### Boyscout
asyncify methods


doc/2/api/controllers/document/update/index.md
```diff
@@ -55,7 +55,7 @@ Body:
 
 - `refresh`: if set to `wait_for`, Kuzzle will not respond until the update is indexed
 - `retryOnConflict`: conflicts may occur if the same document gets updated multiple times within a short timespan, in a database cluster. You can set the `retryOnConflict` optional argument (with a retry count), to tell Kuzzle to retry the failing updates the specified amount of times before rejecting the request with an error.
-
+- `source`: if set to `true` Kuzzle will return the updated document in the response.
```
Aschen:
```suggestion
- `source`: if set to `true` Kuzzle will return the updated document body in the response.
```

doc/2/api/controllers/document/update/index.md
```diff
@@ -17,7 +17,7 @@ Updates a document content.
 ### HTTP
 
 ```http
-URL: http://kuzzle:7512/<index>/<collection>/<documentId>/_update[?refresh=wait_for][&retryOnConflict=<int>]
+URL: http://kuzzle:7512/<index>/<collection>/<documentId>/_update[?refresh=wait_for][&retryOnConflict=<int>][&source=<boolean>]
```
Aschen: Since it's a boolean flag, if it's in the query string then the flag is set 
```suggestion
URL: http://kuzzle:7512/<index>/<collection>/<documentId>/_update[?refresh=wait_for][&retryOnConflict=<int>][&source]
```

lib/api/controllers/base.js
```diff
@@ -598,6 +598,23 @@ class NativeController extends BaseController {
     return null;
   }
 
+  /**
+   * Returns source option
+   * @param {Request} request
+   * @returns {boolean}
+   */
+  getSource (request) {
```
Aschen: You should use `BaseController.getBoolean` method

lib/api/controllers/document.js
```diff
@@ -310,12 +311,16 @@ class DocumentController extends NativeController {
...
+        // response: { _id, _version, _source }
         response = _response;
-
-        return this.kuzzle.notifier.notifyDocumentUpdate(modifiedRequest);
+        return this.kuzzle.notifier.notifyDocumentUpdate(modifiedRequest, response);
       })
-      .then(() => response);
+      .then(() => {
+        if (!source) {
+          delete response._source;
```
Aschen: It's better to return a new object without the `_source` property than deleting a property in the existing `response`.  
The `delete` keyword will trigger a deoptimization of the code

lib/core/notifier.js
```diff
@@ -351,7 +344,13 @@ class NotifierController {
...
       promises.push(
         this.notifyDocument(
-          matchedRooms, request, 'out', 'delete', { _id: documents[i]._id }));
+          matchedRooms,
+          request,
+          'out',
+          'delete',
+          {
+            _id: documents[i]._id
+          }));
```
Aschen: Nitpicking: you can put this object on 1 line instead of 3

lib/core/notifier.js
```diff
@@ -277,30 +276,24 @@ class NotifierController {
...
+  notifyDocumentUpdate (request, updatedDocument) {
+
+    const matchedRooms = this._test(
+      request,
+      updatedDocument._source,
+      updatedDocument._id
+    );
+    const cacheId = getCachePrefix(request) + updatedDocument._id;
+
+    const updatedFields = Object.keys(request.input.body).filter(
```
Aschen: Nitpicking: according to our coding standards, the `.filter(...)`  should be on the new line

lib/core/notifier.js
```diff
@@ -277,30 +276,24 @@ class NotifierController {
...
-        const updatedFields = Object.keys(request.input.body)
-          .filter(_updatedFields => _updatedFields !== '_kuzzle_info');
-
-        return this.notifyDocument(matchedRooms, request, 'in', 'update', {
-          _id: updatedDocument._id,
-          _source: updatedDocument._source,
-          _updatedFields: updatedFields
-        });
-      })
+  notifyDocumentUpdate (request, updatedDocument) {
```
Aschen: You should use `async/await` instead of promise chain in this method.  
We try to progressively get ride of unnecessary promise chain

lib/core/notifier.js
```diff
@@ -375,7 +374,8 @@ class NotifierController {
    * @return {Promise}
    */
   _dispatch (channels, notification, connectionId) {
-    return this.kuzzle.pipe('notify:dispatch', notification)
+    return this.kuzzle
```
Aschen: You should use `async/await` instead of promise chain

lib/core/notifier.js
```diff
@@ -272,51 +269,42 @@ class NotifierController {
...
+      if (cachedRooms !== null) {
+        const stopListening = _.difference(
+          JSON.parse(cachedRooms), matchedRooms);
+
+        await this.notifyDocument(stopListening, request, 'out', 'update', {
+          _id: updatedDocument._id
+        });
+      }
+
+      if (matchedRooms.length > 0) {
```
Aschen: Nitpicking: 
IMHO it's more clearer to have all the method returns at the same place (inside or outside the try catch)

Yoann-Abbes:
Since the functional tests run with the JS SDK, you have to install this version https://github.com/kuzzleio/sdk-javascript/pull/479

Tests won't pass on travis until this is released

PR #1550, opened 2019-12-16
Allow to modify collection settings

Aschen:
## What does this PR do ?

This PR goal is mainly to allow users to modify a collection `settings` (which are [Elasticsearch index settings](https://www.elastic.co/guide/en/elasticsearch/reference/7.5/index-modules.html#index-modules-settings))

The route `collection:create` now accepts both `mappings` and `settings` in its body but it also still accepts directly the mappings properties (`dynamic`, '_meta`, `properties`) for compatibility reason.

I added a new route `collection:update` that allows to update both collection mappings and settings.


features-sdk/AuthController.feature
```diff
@@ -16,6 +16,7 @@ Feature: Auth Controller
     And The result should contain a property "_id" of type "string"
     And I can login with the previously created API key
     And I successfully call the route "auth":"searchApiKeys"
+    And I debug "result.hits"
```
scottinet: debug leftover?

lib/services/elasticsearch.js
```diff
@@ -812,7 +816,7 @@ class ElasticSearch extends Service {
     const exists = await this.collectionExists(index, collection);
 
     if (exists) {
-      return this.updateMapping(index, collection, mappings);
+      return this.updateCollection(index, collection, { mappings, settings });
```
scottinet: Shouldn't we throw an "already exists" error instead?

lib/services/elasticsearch.js
```diff
@@ -874,7 +879,28 @@ class ElasticSearch extends Service {
...
-   * Adds a mapping definition to a specific type
+   * Updates a collection mappings and settings
+   *
+   * @param {String} index - Index name
+   * @param {String} collection - Collection name
+   * @param {Object} body - mappings ({}), settings ({})
+   *
+   * @return {Promise}
+   */
+  async updateCollection (index, collection, { mappings={}, settings={}} = {}) {
```
scottinet: The index is left in an inconsistent state (from a user's viewpoint) if applying the mapping succeeds, but applying the settings fail.

I didn't see an easy way to play these 2 operations as 1 in ES, nor a way to easily rollback either of them.

Since mappings cannot be rolled back at all, maybe we could instead get the current settings, apply the new ones, then apply the mappings, and if that fails, re-apply the previous settings before throwing an error.

What do you think?

lib/services/elasticsearch.js
Aschen: I'm afraid it will be a breaking change since we allow subsequent calls to `collection:create` to modify the mappings :-/
 (I fact we use this behavior in the core also in the [bootstrap sequence](https://github.com/kuzzleio/kuzzle/blob/master/lib/core/storage/bootstrap/internalIndexBootstrap.js#L96) )

lib/services/elasticsearch.js
Aschen: Yes you're right, I'm gonna change this

PR #1557, opened 2020-01-11
Closure optimization for pipes

Aschen:
## What does this PR do ?

Avoid useless allocation of closures and closures context in the pipe management system.  (See [this article](https://www.reaktor.com/blog/javascript-performance-fundamentals-make-bluebird-fast/) from Bluebird creator)

We do not use `async.waterfall` anymore because this library was very slow and inefficient (~6-7 closures allocated at each call of `async.waterfall`).

```
Before: 657
After: 464
~30% faster
```


Aschen:
> Damn I was certain this would be the end of async.js... I guess we'll have to code a async.retry function to be used in functional tests to get rid of it. expressionless

At least `async` is now in the devDependencies :sweat_smile:

PR #1558, opened 2020-01-15
More precise server:healthCheck route

Yoann-Abbes:
## What does this PR do ?

Before this PR:
Hitting `server:healthCheck` returned you the status of each services.

Now:
You can specify which service you want the status of.

[Jira Ticket](https://jira.kaliop.net/browse/KZL-1316)

`http://localhost:7512/_healthCheck`

`http://localhost:7512/_healthCheck?services=storageEngine`

`http://localhost:7512/_healthCheck?services=storageEngine,memoryStorage`

`http://localhost:7512/_healthCheck?services=storageEngine,memoryStorage,internalCache`

### Other changes

Document the route

### DIFF
--- a/doc/2/api/controllers/server/health-check/index.md
+++ b/doc/2/api/controllers/server/health-check/index.md
@@ -8,7 +8,7 @@ title: healthCheck
 
 
 
-Returns status of the different services.
+Returns the status of the different services.
 
 ---
 
--- a/doc/2/api/controllers/server/health-check/index.md
+++ b/doc/2/api/controllers/server/health-check/index.md
@@ -1,0 +1,69 @@
+---
+code: true
+type: page
+title: healthCheck
+---
+
+# healthCheck
+
+
+
+Returns status of the different services.
+
+---
+
+## Query Syntax
+
+### HTTP
+
+```http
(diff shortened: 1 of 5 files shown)


doc/2/api/controllers/server/health-check/index.md
```diff
@@ -0,0 +1,69 @@
...
+code: true
+type: page
+title: healthCheck
+---
+
+# healthCheck
+
+
+
+Returns the status of the different services.
```
Aschen:
```suggestion
Returns the status of Kuzzle and it's internal services.
```

lib/api/controllers/server.js
```diff
@@ -122,41 +122,48 @@ class ServerController extends NativeController {
         err => errorsManager.get('core', 'fatal', 'service_unavailable', err),
       result = { status: 'green', services: {} };
 
-    try {
-      await this.kuzzle.cacheEngine.internal.info();
-      result.services.internalCache = 'green';
+    let services;
+    if (request.input.args && request.input.args.services) {
```
Aschen: `request.input.args` will always be present, and you should check the type of `services` instead to avoid an error if somebody send an array for example

PR #1559, opened 2020-01-17
Add document:updateByQuery

Yoann-Abbes:
## What does this PR do ?

This PR adds a new route `document:updateByQuery` which does not use the ES api since changes are only possible by `script`
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update-by-query.html

So it fetches documents matching the `query`, and `mUpdate` them with `changes` field from `updateByQuery` request as `body` for `mUpdate` request.

[Jira Ticket](https://jira.kaliop.net/browse/KZL-1358)

`POST http://localhost:7512/index/collection/_query`

```
{
    query: {
         match: {
             name: "lame"
         }
     },
    changes: {
         name: "awesome"
     }
}
```

Every document having a field `name` matching `lame` value will be updated with `awesome` value.

### Other changes

Remove useless `result` field

replace `::: warn` by  `::: warning` on the doc

Correct `mUpdate` doc


lib/services/elasticsearch.js
```diff
@@ -687,6 +687,65 @@ class ElasticSearch extends Service {
...
+        { refresh }
+      );
+
+      debug('Update by query: %o', esRequest);
+
+      await this.refreshCollection(index, collection);
+
+      return {
+        documents: response.items,
+        total: response.items.length,
```
Aschen: I don't see the point to return `total` and `updated` if it's the same value

Yoann-Abbes: Well, i'm not sure about that, see the ES doc
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update-by-query.html

features-sdk/step_definitions/documents-steps.js
```diff
@@ -56,7 +56,7 @@ Then('I {string} the following documents:', async function (action, dataTable) {
   this.props.result = await this.sdk.document[action](
     this.props.index,
     this.props.collection,
-    documents);
+    documents, { refresh: 'wait_for' });
```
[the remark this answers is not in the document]
Yoann-Abbes: Yes it is. Without, some tests fail

doc/2/api/controllers/document/update-by-query/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+---
+
+## Query Syntax
+
+### HTTP
+
+```http
+URL: http://kuzzle:7512/<index>/<collection>/_queryUpdate
+Method: POST
```
alexandrebouthinon: Shouldn't this route use the HTTP PUT verb as the `document:update` method?

lib/config/httpRoutes.js
```diff
@@ -308,6 +308,7 @@ module.exports = [
   {verb: 'put', url: '/:index/:collection/_mReplace', controller: 'document', action: 'mReplace'},
   {verb: 'put', url: '/:index/:collection/_mUpdate', controller: 'document', action: 'mUpdate'},
   {verb: 'put', url: '/:index/:collection/:_id/_update', controller: 'document', action: 'update'},
+  {verb: 'post', url: '/:index/:collection/_queryUpdate', controller: 'document', action: 'updateByQuery'},
```
alexandrebouthinon: Shouldn't this route use the HTTP PUT verb as the document:update method?

features-sdk/step_definitions/documents-steps.js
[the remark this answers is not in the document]
Aschen: You should use the `I refresh the collection` step instead, in this way the collection is only refreshed when needed and not within every other scenario using this step

doc/2/api/controllers/document/update-by-query/index.md
```diff
@@ -0,0 +1,109 @@
...
+code: true
+type: page
+title: updateByQuery
+---
+
+# updateByQuery
+
+Updates documents matching the provided search query. 
+
+Documents updated that way trigger real-time notifications.
```
Aschen: IMHO it's not necessary to specify that this method will trigger realtime notifications because all API method should trigger notifications

doc/2/api/controllers/document/update-by-query/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+---
+
+### Optional:
+
+- `source`: if set to `true` Kuzzle will return the updated documents body in the response.
+
+## Body properties
+
+- `query`: documents matching this search query will be updated. Uses the [ElasticSearch Query DSL](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/query-dsl.html) syntax.
```
Aschen: Missing the `changes` body property

doc/2/api/controllers/document/update-by-query/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+---
+
+## Response
+
+Returns a `result` object containing the a `documents` array of the updated documents.
```
Aschen:
```suggestion
Returns a `result` object containing a `documents` array representing updated documents.
```

doc/2/api/controllers/document/update-by-query/index.md
```diff
@@ -0,0 +1,109 @@
...
+  "index": "<index>",
+  "collection": "<collection>",
+  "controller": "document",
+  "action": "updateByQuery",
+  "requestId": "<unique request identifier>",
+  "result": {
+    "documents": [
+      {
+        _id: "document-1",
+        _source: [Object],  // If `source` option is set to true
```
Aschen: Nitpicking (same format as for the `document:update` route)
```suggestion
        _source: "<updated document>" // If `source` option is set to true
```

doc/2/api/controllers/document/update-by-query/index.md
```diff
@@ -0,0 +1,109 @@
...
+  "error": null,
+  "index": "<index>",
+  "collection": "<collection>",
+  "controller": "document",
+  "action": "updateByQuery",
+  "requestId": "<unique request identifier>",
+  "result": {
+    "documents": [
+      {
+        _id: "document-1",
```
Aschen:
```suggestion
        "_id": "document-1",
```
All properties must be enclosed with double quote in valid JSON

doc/2/api/controllers/document/update-by-query/index.md
```diff
@@ -0,0 +1,109 @@
...
+        _source: [Object],  // If `source` option is set to true
+        status: 200,
+        result: "updated"
```
Aschen:
```suggestion
        result: "updated"
```
Seems unnecessary to me since this method can only update document (not create them) and failures will end in the `failures` array

doc/2/api/controllers/document/update-by-query/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+- `source`: if set to `true` Kuzzle will return the updated documents body in the response.
+
+## Body properties
+
+- `query`: documents matching this search query will be updated. Uses the [ElasticSearch Query DSL](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/query-dsl.html) syntax.
+
+---
+
+## Response
```
Aschen: I think it will be better to names the 2 arrays `successes` and `errors` like for the `m*` methods of the API.  

See https://docs.kuzzle.io/core/2/api/controllers/document/m-update/


Also, the `total` property is useless since you just have to compute the `successes` array length.

lib/api/controllers/document.js
```diff
@@ -466,6 +467,35 @@ class DocumentController extends NativeController {
...
+      index,
+      collection,
+      query,
+      changes,
+      { refresh });
+
+    await this.kuzzle.notifier.notifyDocumentMChanges(request, result.documents);
+
+    if (!source) {
+      result.documents.forEach(d => (d._source = null));
```
Aschen: A property set to `null` will be serialized by the JSON

```suggestion
      result.documents.forEach(d => (d._source = undefined));
```

lib/services/elasticsearch.js
```diff
@@ -687,6 +687,62 @@ class ElasticSearch extends Service {
...
+        index: this._getESIndex(index, collection),
+        body: this._sanitizeSearchBody({ query }),
+        scroll: '5s',
+        from,
+        size
+      };
+
+      const documents = await this._getAllDocumentsFromQuery(esRequest);
+      documents.forEach(document => {
+        document._source = null;
```
Aschen: `undefined` mean no value and `null` mean no object value, it's better to set the value to `undefined` when you want to get ride of it

```suggestion
        document._source = undefined
```

lib/services/elasticsearch.js
```diff
@@ -687,6 +687,62 @@ class ElasticSearch extends Service {
...
+        document.body = changes;
+        return document;
```
Aschen: Unnecessary return

lib/services/elasticsearch.js
```diff
@@ -687,6 +687,62 @@ class ElasticSearch extends Service {
...
+        { refresh }
+      );
+
+      debug('Update by query: %o', esRequest);
+
+      await this.refreshCollection(index, collection);
+
+      return {
+        documents: response.items,
+        total: response.items.length,
```
Aschen: This property is useless since we have the `documents` array, so we can use the `length` property

lib/services/elasticsearch.js
```diff
@@ -687,6 +687,62 @@ class ElasticSearch extends Service {
...
+      const response = await this.mUpdate(
+        index,
+        collection,
+        documents,
+        { refresh }
+      );
+
+      debug('Update by query: %o', esRequest);
+
+      await this.refreshCollection(index, collection);
```
Aschen: Why refreshing the collection here? The `mUpdate` method will refresh the collection if `refresh` is set to `wait_for`

lib/services/elasticsearch.js
Yoann-Abbes: Sonarqube does not allow it

lib/api/controllers/document.js
Yoann-Abbes: Sonarqube does not allow it

doc/2/api/controllers/document/update-by-query/index.md
Yoann-Abbes: It's a reminder, and I wrote it to be consistent with other method like
https://docs.kuzzle.io/core/2/api/controllers/document/delete-by-query/

lib/api/controllers/document.js
Yoann-Abbes: Rule deactivated

lib/services/elasticsearch.js
Yoann-Abbes: Rule deactivated

lib/services/elasticsearch.js
```diff
@@ -687,6 +687,58 @@ class ElasticSearch extends Service {
...
+      const esRequest = {
+        index: this._getESIndex(index, collection),
+        body: this._sanitizeSearchBody({ query }),
+        scroll: '5s',
+        from,
+        size
+      };
+
+      const documents = await this._getAllDocumentsFromQuery(esRequest);
+      documents.forEach(document => {
```
alexandrebouthinon: NP: You have an array as input and array as output. Why don't you use `map` function?

Aschen: Recreate a new array with map is slower and also consume a lot of memory, it's better to use `Array.forEach` to mutate the original array since this method could handle very large array (up to 10k elem)

doc/2/api/controllers/document/update-by-query/index.md
```diff
@@ -0,0 +1,131 @@
...
+To update a greater number of documents, either change the server configuration, or split the search query.
+
+---
+
+## Query Syntax
+
+### HTTP
+
+```http
+URL: http://kuzzle:7512/<index>/<collection>/_queryUpdate
```
Aschen: After thinking, just a `PUT` request on `/_query` would be better to keep consistency with deleteByQuery 
```suggestion
URL: http://kuzzle:7512/<index>/<collection>/_query
```

lib/config/httpRoutes.js
```diff
@@ -308,6 +308,7 @@ module.exports = [
   {verb: 'put', url: '/:index/:collection/_mReplace', controller: 'document', action: 'mReplace'},
   {verb: 'put', url: '/:index/:collection/_mUpdate', controller: 'document', action: 'mUpdate'},
   {verb: 'put', url: '/:index/:collection/:_id/_update', controller: 'document', action: 'update'},
+  {verb: 'put', url: '/:index/:collection/_queryUpdate', controller: 'document', action: 'updateByQuery'},
```
Aschen:
```suggestion
  {verb: 'put', url: '/:index/:collection/_query', controller: 'document', action: 'updateByQuery'},
```

doc/2/api/controllers/document/update-by-query/index.md
```diff
@@ -0,0 +1,131 @@
...
+---
+
+## Arguments
+
+- `collection`: collection name
+- `index`: index name
+
+---
+
+### Optional:
```
scottinet: Missing `refresh` option

lib/services/elasticsearch.js
```diff
@@ -687,6 +687,58 @@ class ElasticSearch extends Service {
...
+  * @param {Object} options - from (undefined), size (undefined), refresh (undefined), timeout (undefined), userId (undefined)
+  *
+  * @returns {Promise.<Object>} { documents, total, failures: [ _shardId, reason ] }
+  */
+  async updateByQuery(
+    index,
+    collection,
+    query,
+    changes,
+    { refresh, from, size } = {})
```
scottinet: `from` and `size` are undocumented and unused by this function callers

lib/services/elasticsearch.js
```diff
@@ -687,6 +687,58 @@ class ElasticSearch extends Service {
...
   }
 
+  /**
+  * Updates all documents matching the provided filters
+  *
+  * @param {String} index - Index name
+  * @param {String} collection - Collection name
+  * @param {Object} query - Query to match documents
+  * @param {Object} changes - Changes wanted on documents
+  * @param {Object} options - from (undefined), size (undefined), refresh (undefined), timeout (undefined), userId (undefined)
```
scottinet: wrong options list

lib/services/elasticsearch.js
```diff
@@ -687,6 +687,58 @@ class ElasticSearch extends Service {
...
+  *
+  * @returns {Promise.<Object>} { documents, total, failures: [ _shardId, reason ] }
```
scottinet: this isn't the returned object I see in the code

test/api/controllers/document.test.js
```diff
@@ -540,6 +540,61 @@ describe('DocumentController', () => {
     });
   });
 
+  describe('#updateByQuery', () => {
```
scottinet: missing tests verifying that the controller catch badly formatted requests, that the "refresh" option is correctly forwarded to the ES layer, and that the "source" option correctly removes document contents if set to false (while ensuring that the notifier module still receives them)

lib/services/elasticsearch.js
Aschen: Actually `size` is used by `_getAllDocumentsFromQuery`, and in the `deleteByQuery` method it has a default value of `1000`, so IMHO this method should have the same default value and no `from` argument as @scottinet says.   
See https://github.com/kuzzleio/kuzzle/blob/master/lib/services/elasticsearch.js#L629  
(The `from` option should be removed from the `deleteByQuery` method also)

test/api/controllers/document.test.js
Yoann-Abbes: Since this implementation of `updateByQuery` calls `mUpdate`, the test for the `refresh` option already exists no ?
https://github.com/kuzzleio/kuzzle/blob/master/test/services/elasticsearch.test.js#L2519
+
https://github.com/kuzzleio/kuzzle/pull/1559/files#diff-19c0f27b8ad55ff3c5a5fdf9bd1247dbR919

Aschen: So you should test that the `mUpdate` method is called with the right arguments

Yoann-Abbes: @scottinet for the test about `source` correctly removes document contents if set to `false`, the test is confusing: 
- I first call `updateByQuery` with `_source` at false, so tthe documents contents are not in the response
- Then It should call `notifyDocumentMChanges` with the document contents
- But it feels like it works with a reference of the result of `updateByQuery`, so for mocha, `notifyDocumentMChanges` is called without the document contents

Yoann-Abbes: @Aschen it's done here https://github.com/kuzzleio/kuzzle/pull/1559/files#diff-19c0f27b8ad55ff3c5a5fdf9bd1247dbR919

scottinet: You can stub notifyDocumentMChanges and make it clone the received object. You can then use the cloned object to verify whether it correctly receives the complete documents despite having a "_source = false" option passed to the API route.

test/api/controllers/document.test.js
```diff
@@ -615,13 +624,6 @@ describe('DocumentController', () => {
         { bar: 'foo' },
         { refresh: 'wait_for' });
 
-      should(kuzzle.notifier.notifyDocumentMChanges).be.calledWith(
```
Aschen: You should ensure that the method is called

PR #1560, opened 2020-01-23
Add the http verb to the requestContext object

scottinet:
# Description

Add the HTTP verb used to invoke an API route to the Request object, in the `request.context.connection.misc` object.

There are use cases from clients where they expose a single route that can be invoked with multiple verbs to simplify their API, and they need to be able to differenciate with what verb that route is called.

### DIFF
--- a/lib/core/httpRouter/routeHandler.js
+++ b/lib/core/httpRouter/routeHandler.js
@@ -56,7 +56,8 @@ class RouteHandler {
         protocol: 'http',
         url: message.url,
         headers: message.headers,
-        ips: message.ips
+        ips: message.ips,
+        verb: message.method
       }
     };
 
--- a/test/core/httpRouter/httpRouter.test.js
+++ b/test/core/httpRouter/httpRouter.test.js
@@ -133,7 +133,7 @@ describe('core/httpRouter', () => {
       });
     });
 
-    it('should init request.context with the good values', done => {
+    it('should init request.context with the right values', done => {
       router.post('/foo/bar', handler);
 
       rq.url = '/foo/bar';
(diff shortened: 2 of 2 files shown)


test/core/httpRouter/httpRouter.test.js
```diff
@@ -149,12 +149,13 @@ describe('core/httpRouter', () => {
           const apiRequest = handler.firstCall.args[0];
 
           should(apiRequest).be.instanceOf(Request);
-          should(apiRequest.context.protocol).be.exactly('http');
-          should(apiRequest.context.connectionId).be.exactly('requestId');
-          should(apiRequest.input.headers).be.eql({
+          should(apiRequest.context.connection.protocol).be.exactly('http');
```
scottinet: That unit test was using deprecated properties. Those still exist, but only as links to the correct location.

lib/core/httpRouter/routeHandler.js
```diff
@@ -56,7 +56,8 @@ class RouteHandler {
         protocol: 'http',
         url: message.url,
         headers: message.headers,
-        ips: message.ips
+        ips: message.ips,
+        verb: message.method
```
Aschen: IMHO we should lowercase the verb to make it case insensitive

scottinet: I asked myself that question and, truth is, I don't know what to do with it.
On one hand, lowercases won't make it case-insensitive, and it feels more standard to use the exact same value as the one provided by Node.js' HTTP interface. On the other hand, other values we provide are usually in lowercases...

Aschen: I would rather put the verb in lowercase to follow "our" convention but I won't block this PR for that detail

PR #1561, opened 2020-01-23
Ensure properties are always declared in the same order (eslint rule)

Aschen:
## What does this PR do ?

Add the `sort-keys` rule to ensure each object properties is declared in alphabetical order.  
This rule allow to reduce the number of internal hidden classes used to represent objects.

The codebase has been updated with the eslint plugin: https://www.npmjs.com/package/eslint-plugin-sort-keys-fix

:warning: Only the code in `lib/` and `bin/` is affected by the rule and has been modified


Yoann-Abbes:
It seems like it does not apply the rule inside the units test :(

.eslintrc.json
```diff
@@ -50,8 +50,22 @@
...
+    "yoda": [2, "never"],
+    "sort-keys": ["error", "asc"]
   },
+  "overrides": [
+    {
+      "files": [
+        "test/**/*.js",
+        "features/**/*.js",
+        "features-sdk/**/*.js",
+        "lib/api/controllers/memoryStorage.js"
```
Aschen: Memory storage controller use object to declare it's methods, so the property order is important

test/core/janitor.test.js
```diff
@@ -290,8 +290,8 @@ describe('Test: core/janitor', () => {
           .be.exactly(baseDumpPath.concat('/kuzzle.json'));
         should(fsStub.writeFileSync.getCall(0).args[1])
           .be.exactly(JSON.stringify({
-            version: require('../../package.json').version,
-            config: kuzzle.config
+            config: kuzzle.config,
```
Aschen: One of the two unit test modified

test/core/storage/bootstrap/internalIndexBootstrap.test.js
```diff
@@ -104,8 +104,8 @@ describe('InternalIndexBootstrap', () => {
       should(internalIndexStorage.createOrReplace.getCall(3).args[0])
         .be.eql('profiles');
       should(ret).be.eql({
-        profileIds: ['admin', 'default', 'anonymous'],
-        roleIds: ['admin', 'default', 'anonymous']
+        profileIds: ['admin', 'anonymous', 'default'],
```
Aschen: One of the two unit test modified

lib/config/httpRoutes.js
```diff
@@ -23,300 +23,300 @@
 
 module.exports = [
   // GET (idempotent)
-  {verb: 'get', url: '/users/_me', controller: 'auth', action: 'getCurrentUser'},
```
alexandrebouthinon: I think we should make an exception with this file. The  old properties ordering method was more developer friendly IMHO :slightly_smiling_face:

lib/util/esWrapper.js
```diff
@@ -31,97 +31,97 @@ const
 
 const errorMessagesMapping = [
   {
+    getPlaceholders: matches => [matches[1]],
     regex: /^\[es_rejected_execution_exception] rejected execution .*? on EsThreadPoolExecutor\[(.*?), .*$/,
-    subCode: 'too_many_operations',
```
alexandrebouthinon: Same as http routes file

.eslintrc.json
```diff
@@ -50,8 +50,24 @@
...
+    "yoda": [2, "never"],
+    "sort-keys": ["error", "asc"]
   },
+  "overrides": [
+    {
+      "files": [
+        "test/**/*.js",
+        "features/**/*.js",
+        "features-sdk/**/*.js",
+        "lib/api/controllers/memoryStorage.js",
```
scottinet: I agree about overriding test files in eslint's configuration file, but I'm not sure about overrides for code files. Chances are we are going to forget about those pretty fast.

I think deactivating the sort-keys rule for these 3 files with a comment directive at the beginning of each file would be safer.

lib/config/httpRoutes.js
```diff
@@ -2,7 +2,7 @@
  * Kuzzle, a backend software, self-hostable and ready to use
  * to power modern apps
  *
- * Copyright 2015-2020 Kuzzle
+ * Copyright 2015-2018 Kuzzle
```
scottinet: ?

lib/config/httpRoutes.js
```diff
@@ -263,6 +257,8 @@ module.exports = [
   // DELETE
   {verb: 'delete', url: '/credentials/:strategy/_me', controller: 'auth', action: 'deleteMyCredentials'},
 
+  {verb: 'post', url: '/:index/:collection/_refresh', controller: 'collection', action: 'refresh' },
```
scottinet: should be moved further up, in the "post" section

PR #1563, opened 2020-01-24
Add more security actions logs

Yoann-Abbes:
## What does this PR do ?

This PR adds logs on security actions who modify either a role, a profile or a user.

[Jira Ticket](https://jira.kaliop.net/browse/KZL-1357)

  - Step 1 :
  - Step 2 :
  - Step 3 :  
  ...

### Other changes

Add a new `getUser` getter in the base controller that returns the context user object

Asyncify `mDelete` function
Move `mDelete` and `updateMetadata` function inside class SecurityController


lib/api/funnel.js
```diff
@@ -167,7 +167,7 @@ class Funnel {
     // resolves the callback immediately if a slot is available
     if (this.concurrentRequests < this.kuzzle.config.limits.concurrentRequests) {
       if (this.requestsCacheById[request.internalId]) {
-        delete this.requestsCacheById[request.internalId];
+        this.requestsCacheById[request.internalId] = undefined;
```
Aschen: We need to keep the `delete` keyword here since the `requestsCacheById` will continue to grow with properties set to `undefined`.  
```
    const user = {
      name: 'aschen',
      age: 26
    };
    user.age = undefined;
    Object.keys(user) //=> ['name', 'age']
```

lib/core/entrypoints/protocols/mqtt.js
```diff
@@ -187,7 +187,7 @@ class MqttProtocol extends Protocol {
         const connection = this.connections.get(client);
 
         this.connections.delete(client);
-        delete this.connectionsById[connection.id];
+        this.connectionsById[connection.id] = undefined;
```
Aschen: Same remark about `value = undefined` vs `delete value`

lib/core/statistics.js
```diff
@@ -142,7 +142,7 @@ class StatisticsController {
     }
 
     if (this.currentStats.connections[requestContext.connection.protocol] === 1) {
-      delete this.currentStats.connections[requestContext.connection.protocol];
+      this.currentStats.connections[requestContext.connection.protocol] = undefined;
```
Aschen: Same remark about value = undefined vs delete value

lib/api/controllers/security.js
```diff
@@ -1442,6 +1479,12 @@ function mDelete (kuzzle, type, request) {
           errorsManager.get('services', 'storage', 'incomplete_delete', errors));
       }
 
+      if (ids.length > 1000) {
+        kuzzle.log.info(`[SECURITY] User "${request.context.user}" deleted the following ${type}s ${ids.join(', ')}... (${ids.length - 1000} more users deleted)."`);
```
alexandrebouthinon: As it is, all the ids will be displayed in the log line.
Try this:
```suggestion
        kuzzle.log.info(`[SECURITY] User "${request.context.user}" deleted the following ${type}s ${ids.slice(0, 999).join(', ')}... (${ids.length - 1000} more users deleted)."`);
```

Aschen:
Also, you should add a PR label

lib/api/funnel.js
scottinet: or we can use a Map instead
either way that object will end up being treated as a b-tree, which is exactly what we want here

lib/api/controllers/security.js
```diff
@@ -304,6 +304,7 @@ class SecurityController extends NativeController {
       description,
       { creatorId, refresh, apiKeyId });
 
+    this.kuzzle.log.info(`[SECURITY] User "${request.context.user}" applied action "${request.input.action}" on user ${userId}."`);
```
scottinet: wrong number of double quotes (same goes for all other messages below)

lib/api/controllers/security.js
```diff
@@ -1247,7 +1281,10 @@ class SecurityController extends NativeController {
       'delete');
 
     return deleteMethod(request, request.input.resource._id, request.input.args.strategy)
-      .then(() => ({acknowledged: true}));
+      .then(() => {
+        this.kuzzle.log.info(`[SECURITY] User "${request.context.user}" applied action "${request.input.action}" on user ${request.input.resource._id}."`);
+        return ({acknowledged: true});
```
scottinet:
```suggestion
        return {acknowledged: true};
```

lib/core/plugins/manager.js
```diff
@@ -648,7 +648,7 @@ class PluginsManager {
         throw strategyError.get('unauthorized_removal', strategyName);
       }
 
-      delete this.strategies[strategyName];
+      this.strategies[strategyName] = undefined;
```
scottinet: :warning: this causes a subtle bug

If you register a strategy dynamically, then unregister it, this will set `this.strategies[strategyName]` to undefined. But if you try to register a strategy again with the same name, the following check will be truthy: https://github.com/kuzzleio/kuzzle/blob/master/lib/core/plugins/manager.js#L522

Because even if set to `undefined`, the object `this.strategies` has a property named with the strategy name. This will trigger an attempt to unregister the strategy, but because it is undefined, the manager will throw a `strategy_not_found` error.

Thus you can never again re-register a strategy with the same name after it has been unregistered.

Aschen:
Considering the work necessary to remove the `delete` keyword (use `Map` instead), I advise to make these change in a separate PR

lib/api/controllers/security.js
```diff
@@ -304,6 +304,7 @@ class SecurityController extends NativeController {
       description,
       { creatorId, refresh, apiKeyId });
 
+    this.kuzzle.log.info(`[SECURITY] User "${request.context.user}" applied action "${request.input.action}" on user "${userId}."`);
```
Aschen: Here `request.context.user` is an object, IMHO it's better to print only the ID
(This remark apply everywhere)

Yoann-Abbes:
> Considering the work necessary to remove the `delete` keyword (use `Map` instead), I advise to make these change in a separate PR

https://github.com/kuzzleio/kuzzle/pull/1568

lib/core/janitor.js
```diff
@@ -50,12 +50,12 @@ class Janitor {
    * @param {object} securities
    * @returns {Promise}
    */
-  async loadSecurities (securities = {}) {
+  async loadSecurities (securities = {}, userId = '-1') {
```
scottinet: This parameter is not currently used, as `this.createSecurity` still takes only 2 arguments.

Also, I'm not sure about this default value. The "-1" kuid is reserved for the anonymous user, i.e. a non-authenticated user invoking the API. I'm not sure that this should be the default value when loadSecurities is invoked by Kuzzle itself upon startup.

Last remark: you should probably receive a nullable User object instead of a user kuid, for the same reason as the one stated in one of my previous remarks: if `admin:loadSecurities` is executed by a plugin, you won't get any user object at all, so you cannot have a valid user id.

lib/api/controllers/security.js
```diff
@@ -305,6 +305,7 @@ class SecurityController extends NativeController {
       description,
       { apiKeyId, creatorId, refresh });
 
+    this.kuzzle.log.info(`[SECURITY] User "${request.context.user._id}" applied action "${request.input.action}" on user "${userId}."`);
```
scottinet: :warning: `request.context.user` can be null, for instance if the request has been instantiated by a plugin (either with the SDK or with the execute method).

After giving it some thoughts, my guess is that you should create a dedicated `getUserId` getter in the `BaseController` class. 
This getter should return  `<null>` or something like that (to be debated) if the user object is null.
It should return neither `-1` nor `anonymous` because we don't know if the request is not a consequence of some other API call made by an authenticated user.

That new getter should be used everywhere.

lib/core/janitor.js
```diff
@@ -374,7 +375,7 @@ class Janitor {
         body: objects[_id],
         controller: 'security',
         refresh: 'wait_for'
-      });
+      }, {user: { _id: 'User' } });
```
scottinet: "User" is not a valid kuid. You should revert that change and instead let the new getter I propose above handle the null user object.

lib/core/janitor.js
```diff
@@ -392,7 +393,7 @@ class Janitor {
         body: { ids },
         controller: 'security',
         refresh: 'wait_for'
-      });
+      }, { user: { _id: 'User' } });
```
scottinet: ditto

lib/api/controllers/security.js
```diff
@@ -1505,11 +1505,15 @@ function mDelete (kuzzle, type, request) {
           errorsManager.get('services', 'storage', 'incomplete_delete', errors));
       }
 
+      const userId = request.context && request.context.user && request.context.user._id
```
Aschen: Why don't use `getUserId` here?

lib/core/janitor.js
```diff
@@ -50,18 +50,18 @@ class Janitor {
    * @param {object} securities
    * @returns {Promise}
    */
-  async loadSecurities (securities = {}, userId = '-1') {
+  async loadSecurities (securities = {}, user = null) {
```
Aschen: The method `admin:loadSecurities` should forward the current user to this method

lib/api/controllers/security.js
Yoann-Abbes: Because this function is outside the class

lib/core/janitor.js
Yoann-Abbes: It's already done https://github.com/kuzzleio/kuzzle/pull/1563/files#diff-8be4ba2ed5b9aab3d1b100866dcc323aR176

lib/api/controllers/security.js
Aschen: The security controller inherits from the NativeController which inherits from the BaseController which contain this method

Aschen: Oh yes you're right!
This function should be a method of the controller instead so you won't need to pass the kuzzle object to it and you can use the `getUserId` method

lib/api/controllers/admin.js
```diff
@@ -172,8 +172,8 @@ class AdminController extends NativeController {
 
   loadSecurities (request) {
     const securities = this.getBody(request);
-
-    const promise = this.kuzzle.janitor.loadSecurities(securities, request.context.user._id);
+    const user = this.getUser(request);
+    const promise = this.kuzzle.janitor.loadSecurities(securities, { user });
```
scottinet: `janitor.loadSecurities` seems to expect a User object, not an anonymous `{user: <User>}` object

lib/api/controllers/security.js
```diff
@@ -305,7 +305,7 @@ class SecurityController extends NativeController {
       description,
       { apiKeyId, creatorId, refresh });
 
-    this.kuzzle.log.info(`[SECURITY] User "${request.context.user._id}" applied action "${request.input.action}" on user "${userId}."`);
+    this.kuzzle.log.info(`[SECURITY] User "${this.getUserId(request)}" applied action "${request.input.action}" on user "${userId}."`);
```
scottinet: you should use the creatorId variable instead (which should use that new getUserId function to prevent an error if invoked by a plugin)

lib/api/controllers/security.js
```diff
@@ -1010,7 +1010,7 @@ class SecurityController extends NativeController {
         updatedUser,
         options);
 
-    this.kuzzle.log.info(`[SECURITY] User "${request.context.user._id}" applied action "${request.input.action}" on user "${createdUser._id}."`);
+    this.kuzzle.log.info(`[SECURITY] User "${this.getUserId(request)}" applied action "${request.input.action}" on user "${createdUser._id}."`);
```
scottinet: I know this isn't your code, but the metadata object built by this function (l:983) should use the new, safer method.

Same goes for the `updateMetadata` method

lib/api/controllers/security.js
```diff
@@ -1399,6 +1399,67 @@ class SecurityController extends NativeController {
...
+        controller: 'security'
+      }, request.context);
+
+
+      deleteRequest.input.args.refresh = getRefresh(request);
+
+      return Bluebird.promisify(kuzzle.funnel.mExecute, {context: kuzzle.funnel})(deleteRequest);
+    });
+
+    return Bluebird.all(promises)
```
scottinet: (nitpicking) would benefit from being converted to async/await calls

test/api/controllers/security/index.js
```diff
@@ -170,18 +167,18 @@ describe('/api/controllers/security', () => {
...
-        .then(ids => {
-          should(ids)
-            .match([
-              'foo',
-              'bar',
-              'baz'
-            ]);
-        });
+      const ids = await securityController.mDelete(kuzzle, 'type', request);
+      // .then(ids => {
```
scottinet: remove commented code

test/api/controllers/security/index.js
```diff
@@ -170,18 +167,18 @@ describe('/api/controllers/security', () => {
...
+      should(ids)
+        .match([
+          'foo',
+          'bar',
+          'baz'
+        ]);
+      // });
```
scottinet: remove commented code

test/api/controllers/security/index.js
```diff
@@ -198,16 +195,16 @@ describe('/api/controllers/security', () => {
...
-          should(request.error)
-            .be.an.instanceOf(PartialError);
-          should(request.error.errors)
-            .be.an.Array()
-            .and.have.length(1);
-          should(request.error.errors[0])
-            .match(new KuzzleInternalError(error));
-        });
+      await securityController.mDelete(kuzzle, 'type', request);
+      //  .then(() => {
```
scottinet: remove commented code

test/api/controllers/security/index.js
```diff
@@ -198,16 +195,16 @@ describe('/api/controllers/security', () => {
...
+      should(request.error)
+        .be.an.instanceOf(PartialError);
+      should(request.error.errors)
+        .be.an.Array()
+        .and.have.length(1);
+      should(request.error.errors[0])
+        .match(new KuzzleInternalError(error));
     });
+    //});
```
scottinet: remove commented code

PR #1568, opened 2020-02-05
Optimization : get rid of the "delete" keyword on objects

Yoann-Abbes:
## What does this PR do ?

This PR intends to bring a little performance optimization by getting rid of the `delete` keyword that was used on objects.

For static objects as `requestsCacheById` (funnel), `currentStats` objects (statistics) and `connectionsById` (mqtt), they are now `Map`.
Those objects are often mutated, and being Maps increase performance on a large volume of addition/deletion 

  - Step 1 :
  - Step 2 :
  - Step 3 :  
  ...

### DIFF
--- a/test/services/elasticsearch.test.js
+++ b/test/services/elasticsearch.test.js
@@ -2701,7 +2701,7 @@ describe('Test: ElasticSearch service', () => {
           ];
           const rejected = [
             {
-              document: { _id: undefined, body: { _kuzzle_info:undefined, city: 'Ho Chi Minh City' } },
+              document: { _id: undefined, body: { _kuzzle_info: undefined, city: 'Ho Chi Minh City' } },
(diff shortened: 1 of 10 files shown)


test/services/elasticsearch.test.js
```diff
@@ -2701,7 +2701,7 @@ describe('Test: ElasticSearch service', () => {
           ];
           const rejected = [
             {
-              document: { _id: undefined, body: { city: 'Ho Chi Minh City' } },
+              document: { _id: undefined, body: { _kuzzle_info:undefined, city: 'Ho Chi Minh City' } },
```
alexandrebouthinon:
```suggestion
              document: { _id: undefined, body: { _kuzzle_info: undefined, city: 'Ho Chi Minh City' } },
```

PR #1570, opened 2020-02-11
Fix realtime:publish from embedded SDK

Aschen:
## What does this PR do ?

Fix a bug preventing to use `realtime.publish` method from the embedded SDK without an user in the request context.  

Also, the actual check about the user passed to be used in the request context is overly restrictive.  
For example, I may want to get an user with the `security.getUser` method and then use it as context.  
Now this check only ensure the presence of an `_id` string property.

### Other changes

 - asyncify


lib/api/controllers/realtime.js
```diff
@@ -130,23 +129,20 @@ class RealtimeController extends NativeController {
...
-          createdAt: Date.now()
-        };
-
-        return this.kuzzle.notifier.publish(newRequest);
-      })
-      .then(() => ({published: true}));
+    const newRequest = await this.kuzzle.validation.validate(request, false);
+
+    newRequest.input.body._kuzzle_info = {
+      author: this.getUserId(request),
```
Aschen: Actual fix

test/api/controllers/realtime.test.js
```diff
@@ -116,15 +116,13 @@ describe('RealtimeController', () => {
     it('should throw an error if body is not provided',() => {
       request.input.body = null;
 
-      should(() => {
-        realtimeController.unsubscribe(request);
-      }).throw(BadRequestError, { id: 'api.assert.body_required' });
+      should(realtimeController.unsubscribe(request))
```
scottinet: must be either returned or awaited, otherwise test will always pass even if the condition isn't verified

test/api/controllers/realtime.test.js
```diff
@@ -116,15 +116,13 @@ describe('RealtimeController', () => {
...
+        .rejectedWith(BadRequestError, { id: 'api.assert.body_required' });
     });
 
     it('should throw an error if roomId is not provided',() => {
-      should(() => {
-        realtimeController.unsubscribe(request);
-      }).throw(BadRequestError, { id: 'api.assert.missing_argument' });
+      should(realtimeController.unsubscribe(request))
```
scottinet: must be either returned or awaited, otherwise test will always pass even if the condition isn't verified

test/api/controllers/realtime.test.js
```diff
@@ -143,15 +141,13 @@ describe('RealtimeController', () => {
...
-      }).throw(BadRequestError, { id: 'api.assert.body_required' });
+      should(realtimeController.count(request))
+        .rejectedWith(BadRequestError, { id: 'api.assert.body_required' });
     });
 
     it('should throw an error if roomId is not provided',() => {
-      should(() => {
-        realtimeController.count(request);
-      }).throw(BadRequestError, { id: 'api.assert.missing_argument' });
+      should(realtimeController.count(request))
```
scottinet: must be either returned or awaited, otherwise test will always pass even if the condition isn't verified

PR #1571, opened 2020-02-12
Add plugin name in log

Aschen:
## What does this PR do ?

Just add the plugin name to the log method exposed in the context.  
```
// template
[plugin-${pluginName} ${message}

// example
[plugin-omniscient] Hello from omniscient plugin
```

### DIFF
--- a/lib/core/plugins/context.js
+++ b/lib/core/plugins/context.js
@@ -124,12 +124,12 @@ class PluginContext {
         enumerable: true,
         get: () => {
           return {
-            debug: (...args) => kuzzle.log.debug(...addPluginName(pluginName, args)),
-            error: (...args) => kuzzle.log.error(...addPluginName(pluginName, args)),
-            info: (...args) => kuzzle.log.info(...addPluginName(pluginName, args)),
-            silly: (...args) => kuzzle.log.silly(...addPluginName(pluginName, args)),
-            verbose: (...args) => kuzzle.log.verbose(...addPluginName(pluginName, args)),
-            warn: (...args) => kuzzle.log.warn(...addPluginName(pluginName, args))
+            debug: msg => kuzzle.log.debug(`[plugin-${pluginName}] ${msg}`),
+            error: msg => kuzzle.log.error(`[plugin-${pluginName}] ${msg}`),
+            info: msg => kuzzle.log.info(`[plugin-${pluginName}] ${msg}`),
+            silly: msg => kuzzle.log.silly(`[plugin-${pluginName}] ${msg}`),
+            verbose: msg => kuzzle.log.verbose(`[plugin-${pluginName}] ${msg}`),
+            warn: msg => kuzzle.log.warn(`[plugin-${pluginName}] ${msg}`)
           };
         }
       });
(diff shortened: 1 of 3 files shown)


lib/core/plugins/context.js
```diff
@@ -409,4 +408,16 @@ function curryRemoveStrategy(kuzzle, pluginName) {
...
 
+/**
+ * Adds the plugin name to log lines
+ *
+ * @param {String} pluginName
+ * @param {String[]} lines
+
+ * @returns {String[]}
+ */
+function addPluginName (pluginName, lines = []) {
```
scottinet: I'm not sure about what happens when you send multiple arguments to log functions. 
If `console` is invoked (before Kuzzle logger has started), then those aren't lines. And if this is winston (99% of the time), I think other arguments after the 1st one are ignored (to be verified).

Either way, I don't think you have to worry about multiple arguments, especially considering that our documentation says nothing of the sort: https://docs.kuzzle.io/core/2/plugins/plugin-context/log/error/

This means that you could simplify the code by replacing this:
`warn: (...args) => kuzzle.log.warn(...addPluginName(pluginName, args))`

with that:
`warn: message => kuzzle.log.warn(`[${pluginName}] - ${message}`)`

lib/core/plugins/context.js
```diff
@@ -120,17 +120,16 @@ class PluginContext {
...
         enumerable: true,
         get: () => {
           return {
-            debug: (...args) => kuzzle.log.debug(...args),
-            error: (...args) => kuzzle.log.error(...args),
-            info: (...args) => kuzzle.log.info(...args),
-            silly: (...args) => kuzzle.log.silly(...args),
-            verbose: (...args) => kuzzle.log.verbose(...args),
-            warn: (...args) => kuzzle.log.warn(...args)
+            debug: msg => kuzzle.log.debug(`[plugin-${pluginName}] ${msg}`),
```
alexandrebouthinon: This implementation break the logger public API since this is not a variadic funtion anymore. Maybe something like this could do the trick:
```suggestion
            debug: (...args) => kuzzle.log.debug(`[plugin-${pluginName}] `, args),
```

test/core/plugins/context/context.test.js
```diff
@@ -217,14 +217,26 @@ describe('Plugin Context', () => {
 
         should(kuzzle.log[level])
           .calledOnce()
-          .calledWithExactly('test');
+          .calledWithExactly('[plugin-pluginName] test');
```
alexandrebouthinon:
```suggestion
          .calledWithExactly('[plugin-pluginName] `, 'test');
```

test/core/plugins/context/context.test.js
```diff
@@ -217,14 +217,26 @@ describe('Plugin Context', () => {
...
         ['execute', 'validation', 'storage', 'trigger', 'strategies', 'sdk']);
     });
 
+    it('should adds the plugin name in logs', done => {
+      context.log.info('foobar');
+
+      process.nextTick(() => {
+        should(kuzzle.log.info)
+          .be.calledOnce()
+          .be.calledWith('[plugin-pluginName] foobar');
```
alexandrebouthinon:
```suggestion
          .be.calledWith('[plugin-pluginName] `, 'foobar');
```

test/core/plugins/context/context.test.js
alexandrebouthinon: To make the test match my previous suggestion

test/core/plugins/context/context.test.js
alexandrebouthinon: Same

lib/core/plugins/context.js
Aschen: In the current documentation, the log methods only takes one argument so it's not a breaking change.
See @scottinet comments here: https://github.com/kuzzleio/kuzzle/pull/1571#discussion_r378295795

PR #1575, opened 2020-02-25
Split error codes documentation

scottinet:
# Description

Splits the error codes documentation to make it compatible with algolia (the current page is too large for it).

Current page: https://docs.kuzzle.io/core/2/api/essentials/errors/codes/

Changes details:

* The "Error handling" page is now directly available under the "Essentials" group: https://docs.kuzzle.io/core/2/api/essentials/error-handling
* The "Error codes" is now directly available under the "Essentials" group: https://docs.kuzzle.io/core/2/api/essentials/error-codes
  * The "Error codes" now contains 1 section per error domain: "api", "core", "services", ...
  * Error codes pages are sorted by domain code
* A new CI script has been added to make sure that the documentation is up-to-date (replacing the current check which only had a diff on a single file to do)

# How to review

Most of the changes are from auto-generated documentation. 
You should only review non-markdown files. For markdown ones, sync the branch and run `npm run doc-dev` to have a better view of what changes in the documentation.


doc/2/api/essentials/codes/index.md
```diff
@@ -0,0 +1,8 @@
+---
```
Aschen: Why this file name is not `error-codes` instead? IMHO, it will be more explicit in the URL

doc/2/api/essentials/handling/index.md
```diff
@@ -3,7 +3,7 @@ code: false
 type: page
```
Aschen: Same remark as for the error code page, IMHO the url will be nicer like this `api/essentials/error-handling`

scottinet:
@Aschen > good point, URL renamed.

PR #1576, opened 2020-02-26
Add meaningful error message on strict mapping rejection

Aschen:
## What does this PR do ?

Adds a meaningful error message when a document is rejected because it contain a field not defined in a collection mapping when the `dynamic` property is `strict`.  

The message contain now the collection name.

### How to review

You can review only the `.js` files and also the `DocumentController.feature` file.  
The other changes on feature files are just indentation fix

:rocket:  Kourou ROCKS :muscle:

```bash
kourou query index:create --arg index:index
kourou query collection:create --arg index:index --arg collection:collection --body '{ "dynamic": "strict" }'
kourou query document:create --arg index:index --arg collection:collection --body '{ "a": 1 }'
=>  Cannot create document. Field "a" is not present in collection "index:collection" strict mapping
```

### Other changes

 - Do not trigger errors when loading default securities
 - Adds a better message when a custom protocol initialization or a plugin initialization timeout

### Boyscout

 - format all feature files


features-sdk/DocumentController.feature
```diff
@@ -1,31 +1,50 @@
 Feature: Document Controller
 
+  # document:create ============================================================
```
Aschen: New functional test

PR #1577, opened 2020-03-03
Allow plugins to listen to document changes with Koncorde filter

Aschen:

  - Step 1 :
  - Step 2 :
  - Step 3 :  
  ...


doc/2/plugins/guides/hooks/index.md
```diff
@@ -15,6 +15,116 @@ Hooks can only listen: the received information cannot be changed. And Kuzzle do
...
+    callback: (document, request) => {
+      // this callback will be called only when a document contains
+      // a 'name' property equals to 'Saigon'
+    },
+    filter: {
+      equals: { '_source.city': 'Saigon' }
+    }
+  }
+}
+```
```
scottinet: Wouldn't it be better to invoke the callback with a filtered array containing only objects matching the filters?

```suggestion
The callback is invoked if at least one object of the array matches the filters. The callback payload will be an array containing only matching objects.

Example with a [Document generic event](/core/2/plugins/guides/events/generic-document-events/):
```js
this.hooks = {
  'document:generic:beforeWrite': {
    callback: (document, request) => {
      // this callback will be called only when at least one document contains
      // a 'name' property equal to 'Saigon'
      // "document" will be an array of objects containing only a "name" property
      // equal to "Saigon"
    },
    filter: {
      equals: { '_source.city': 'Saigon' }
    }
  }
}
```
```

File: doc/2/plugins/guides/hooks/index.md [hunk 2]
```diff
@@ -15,6 +15,116 @@ Hooks can only listen: the received information cannot be changed. And Kuzzle do
...
+      // the first argument is an integer so Kuzzle will throw an error and shutdown
+    },
+    filter: {
+      // any filter
+    }
+  }
+}
+```
+
+<h2>Deprecated Usage <DeprecatedBadge version="2.2.0" /></h2>
```
scottinet: I'm not sure we should deprecate other syntaxes, they can be viewed as shorthands for the new, more complete syntax. What do you think?

doc/2/plugins/guides/hooks/index.md
scottinet: ... I think I know why you worded it like you did. Because pipes/hooks can be chained, with payloads being passed from a plugin to another. :expressionless:

Aschen: Yes for the hooks it could be a filtered array but for pipes you have to return an array which will be passed to the next pipe :disappointed:

doc/2/plugins/guides/hooks/index.md
Aschen: Yes I agree this syntax is more convenient when you don't need filters but I'm not comfortable with having different possible syntax for the same feature

User could have difficulties to switch to one syntax to another, for example if I'm working on a project and the dev only use the short syntax so I may didn't even know the existence of the other syntax. Then if I start to contribute on another project using the complete syntax it could be difficult to adapt

plugins/available/functional-test-plugin/index.js
```diff
@@ -42,9 +42,21 @@ class FunctionalTestPlugin {
...
-  init (config, context) {
+  async init (config, context) {
     this.config = config;
     this.context = context;
+
+    // context.onDocument* related declarations ================================
+
+    const filters = { exists: 'name' };
+
+    await this.context.accessors.beforeDocumentWrite('test', 'test', filters, async (document, request) => {
```
alexandrebouthinon:
```suggestion
    await this.context.accessors.beforeDocumentWrite('test', 'test', filters, (document, request) => {
```

Aschen: This code is obsolete sorry, it was my first draft trying to implement the feature

doc/2/plugins/guides/hooks/index.md
scottinet: This can still be done I think, and it would make this feature really nice.

Imagine our pipe runner having a pipe wrapper for object arrays, behaving like this (rough approximation, I didn't study our current code in details):

```js
function wrapObjectArrays (arr, cb, next) {
  const filtered = [];
  const indices = [];

  for (let i = 0; i < arr.length; i++) {
    // let's just assume that there is a koncorde filter in the current context
    if (this.koncorde.test('pipes', '<event name>', arr[i]).length > 0) {
      filtered.push(arr[i]);
      indices.push(i);
    }
  }

  if (filtered.length === 0) {
    next(null, arr);
    return;
  }

  cb(filtered, (err, res) => {
    if (err)  {
      next(err);
    }
    else {
      // we should test that "res" is an array of the same length than filtered
      for (let i = 0; i < indices.length; i++) {
        arr[i] = res.shift();
      }
    }
  }
}
```

doc/2/plugins/guides/hooks/index.md
```diff
@@ -13,7 +13,189 @@ Hooks can only listen: the received information cannot be changed. And Kuzzle do
...
+    }
+  }
+};
+```
+
+## Filter on Request
+
+If a filter is set in the `withRequest` property, then the callback will be executed only if the filter match the request.  
+
+This allow you to easily decide when you want to execute a precise callback:
```
scottinet:
```suggestion
This allows you to easily decide when you want to execute a specific callback:
```

doc/2/plugins/guides/hooks/index.md
```diff
@@ -13,7 +13,189 @@ Hooks can only listen: the received information cannot be changed. And Kuzzle do
...
+    },
+    withRequest: {
+      equals: { 'input.resource.index': 'nyc-open-data' }
+    }
+  }
+}
+```
+
+::: warning
+The `withRequest` filter is only available for events that receive a Request object, with other events Kuzzle will throw an error and interrupt its initialization sequence. (eg: [core:overload](/core/2/plugins/guides/events/core:overload))
```
scottinet:
```suggestion
The `withRequest` filter is only available for events that provide a Request object.  
If set on other events (e.g. [core:overload](/core/2/plugins/guides/events/core:overload)), Kuzzle throws an error and interrupts its initialization sequence.
```

doc/2/plugins/guides/hooks/index.md
```diff
@@ -13,7 +13,189 @@ Hooks can only listen: the received information cannot be changed. And Kuzzle do
...
+      else {
+        return null;
+      }
+    },
+    withDocument: {
+      exists: 'age'
+    }
+  }
+}
+```
```
scottinet: Why don't we use a [range filter](https://docs.kuzzle.io/core/2/guides/cookbooks/realtime-api/terms/#range) instead?

```suggestion
this.hooks = {
  'generic:document:beforeWrite': {
    callback: (document, request) => {
      // This callback will be executed only on documents 
      // containing an "age" property greater or equal than 18
    },
    withDocument: {
      range: { age: { gte: 18 } }
    }
  }
}
```
```

File: doc/2/plugins/guides/hooks/index.md [hunk 7]
```diff
@@ -13,7 +13,189 @@ Hooks can only listen: the received information cannot be changed. And Kuzzle do
...
+
+::: warning
+The `withDocument` filter is only available for [Generic Document events](/core/2/plugins/guides/events/generic-document-events), if a `withDocument` filter is set on another event, Kuzzle will throw an error and interrupt its initialization sequence.
```
scottinet:
```suggestion
The `withDocument` filter is only available for [Generic Document events](/core/2/plugins/guides/events/generic-document-events). If a `withDocument` filter is set on any other event, Kuzzle throws an error and interrupts its initialization sequence.
```

lib/core/plugins/manager.js
```diff
@@ -84,6 +83,17 @@ class PluginsManager {
     this.authenticators = {};
 
     this.config = kuzzle.config.plugins;
+
+    this.__documentPipes = null;
+  }
+
+  // Lazy loading because DocumentPipes embed a Koncorde engine
```
scottinet: (nitpicking)

```suggestion
  // Lazy loading because DocumentPipes embeds a Koncorde engine
```

lib/core/plugins/manager.js
```diff
@@ -709,25 +716,63 @@ class PluginsManager {
...
+    const count = await this._documentPipes.registerOnWrite(
+      pluginName,
+      index,
+      collection,
+      filters,
+      callback);
 
-    if (plugin.config && plugin.config.pipeWarnTime !== undefined) {
-      _warnTime = plugin.config.pipeWarnTime;
+    if (count === 1) {
```
scottinet: If a similar conditional pipe is registered by another plugin (e.g. a condition on the index name), then the room id returned by Koncorde will be identical for that plugin, and the count returned will be > 1.

doc/2/plugins/guides/hooks/index.md
Aschen: With the range filter, the callback will be called only for document with `age` greater than 18, and the callback will not be called document with `age` less than 18 so we cannot filter them.  
The filter allow to call the callback only with matching document, he is not actually filtering the documents

lib/core/plugins/manager.js
Aschen: Good catch :+1:

doc/2/plugins/guides/hooks/index.md
scottinet: In any case, your callback is wrong: the received payload from that event is an array of documents.

Aschen: Yes but my point is that the callback is unitary called with each matching document and not with a filtered list of documents

scottinet: I didn't get that. And I don't like it, because this means that large queries such as mCreate and the like, cannot be handled by plugins unless they process documents 1 by 1.

Aschen: I agree but in this case we cannot use this pipe to continue the processing only with a subset of the original documents.  

For example, I want my action to be executed only on document where `age > 18` but that didn't mean that I don't want the document with `age <= 18` to be excluded of the rest of the processing.

If you want to process the array of documents, you can only provide the `withRequest` filter and you will have to manually filter documents that you don't want.

scottinet: First, there is another thing that I don't like, and it is that your solution changes the callback payload. Usually, that event sends an array of documents, but if filtered, it receives a single document object instead. This change is made silently, and it isn't obvious in the documentation (hell, I didn't get it either with the documentation under my nose).

And as I said, the solution you propose makes that feature unusable with `m*` API routes, because you lose the ability to perform operations on whole sets of documents, with far better performances than performing these operations times the number of documents.

The solution I proposed initially was a hard filter only for the callback with the filters, if you read it carefully, you'll see that the original documents array is mutated with the result of the pipe function, to report all changes to their corresponding, original array indexes. Before invoking the next pipe.

To me, there are 2 ways of handling these filters more or less gracefully:
1. filter the documents sent to the callback according to the provided Koncorde filter. This is the solution I proposed, and basically, this means 1/ building a subarray containing only the matching documents, 2/ provide this subarray to the pipe, 3/ mutate the original array with the pipe's results, and 4/ continue with the next pipe function
1. do not alter the documents array in any way. Instead, make the Koncorde filter act as a filter on the whole array: if at least one document matches, then the pipe function is executed with the whole set of documents

Solution 1 is the nicest for plugin developers.
Solution 2 is the safest and the most maintenable for us.

Aschen:
IMHO this is obsolete because of https://github.com/kuzzleio/kuzzle/pull/1645

PR #1578, opened 2020-03-03
Allow to limit scroll usage

Aschen:
## What does this PR do ?

When searching documents with the `scroll` option, Elasticsearch has to duplicate the transaction log to ensure the results will not change between 2 requests.  
This can lead to memory leak if too many scroll sessions are open.  

This PR introduce 1 new configuration variables: 
 - `services.storage.maxScrollDuration` : maximum scroll duration (default `1m`)

```bash
kourou query index:create --arg index:test
kourou query collection:create --arg index:test --arg collection:test
kourou query document:search --arg index:test --arg collection:test --arg scroll:42m
# Scroll duration is too great. (42m > 15m)
```


.kuzzlerc.sample
```diff
@@ -585,6 +585,10 @@
           }
         }
       },
+      // Maximum number of simultaneous scroll session
+      "maxScroll": 300,
```
alexandrebouthinon: This parameter is already configurable directly in Elasticsearch. Moreover, IMHO, it is not the role of Kuzzle to manage this.

doc/2/api/controllers/document/search/index.md
```diff
@@ -9,10 +9,21 @@ title: search
...
 
 There is a limit to how many documents can be returned by a single search query.
-That limit is by default set at 10000 documents, and you can't get over it even with the `from` and `size` pagination options.
+That limit is by default set at 10000 documents (See `limits.documentsFetchCount`), and you can't get over it even with the `from` and `size` pagination options.
 
 To handle larger result sets, you have to either create a cursor by providing a value to the `scroll` option or, if you sort the results, by using the Elasticsearch [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/search-request-body.html#request-body-search-search-after) command.
 
+::: warning
+When using a cursor with the `scroll` option, Elasticsearch has to duplicate the transaction log to keep the same result during the entire scroll session.  
+It can lead to memory leaks in a too great scroll duration is provided or if too many scroll session are open simultaneously.
```
Yoann-Abbes:
```suggestion
It can lead to memory leaks if a too great scroll duration is provided or if too many scroll sessions are open simultaneously.  
```

doc/2/api/controllers/document/search/index.md
```diff
@@ -9,10 +9,21 @@ title: search
...
 
+::: warning
+When using a cursor with the `scroll` option, Elasticsearch has to duplicate the transaction log to keep the same result during the entire scroll session.  
+It can lead to memory leaks in a too great scroll duration is provided or if too many scroll session are open simultaneously.  
+:::
+
+
+::: info
+<SinceBadge version="2.2.0"/>
+You can restrict the scroll session maximum duration under the `services.storage.maxScrollDuration` configuration key and the maximum number of simultaneously opened scroll session under the `services.storage.maxScroll` configuration key.
```
Yoann-Abbes:
```suggestion
You can restrict the scroll session maximum duration under the `services.storage.maxScrollDuration` configuration key and the maximum number of simultaneously opened scroll sessions under the `services.storage.maxScroll` configuration key.
```

test/services/elasticsearch.test.js
```diff
@@ -112,14 +112,15 @@ describe('Test: ElasticSearch service', () => {
         });
     });
 
-    it('should return a rejected promise if a scroll fails', () => {
+    it('should rejects promise if a scroll fails', () => {
```
Yoann-Abbes:
```suggestion
    it('should reject promise if a scroll fails', () => {
```
:gb:

.kuzzlerc.sample
Aschen: Yes I know but it's more convenient to configure this value inside Kuzzle 
configuration file than in Elasticsearch.  

Actually we could also restrict the `maxScrollDuration` directly with ES but I see at least 2 reasons to keep theses restrictions inside Kuzzle:
 - less links between ES and Kuzzle: Kuzzle handle everything
 - better error message: ES errors are known to be a mess

scottinet: What happens if the value set in that configuration is greater than the elasticsearch one?

doc/2/api/controllers/document/search/index.md
```diff
@@ -9,10 +9,21 @@ title: search
...
 
 There is a limit to how many documents can be returned by a single search query.
-That limit is by default set at 10000 documents, and you can't get over it even with the `from` and `size` pagination options.
+That limit is by default set at 10000 documents (See `limits.documentsFetchCount`), and you can't get over it even with the `from` and `size` pagination options.
 
 To handle larger result sets, you have to either create a cursor by providing a value to the `scroll` option or, if you sort the results, by using the Elasticsearch [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/search-request-body.html#request-body-search-search-after) command.
 
+::: warning
+When using a cursor with the `scroll` option, Elasticsearch has to duplicate the transaction log to keep the same result during the entire scroll session.  
+It can lead to memory leaks if a too great scroll duration is provided or if too many scroll sessions are open simultaneously.
```
scottinet:
```suggestion
It can lead to memory leaks if a scroll duration too large is provided, or if too many scroll sessions are open simultaneously.  
```

lib/config/error-codes/services.json
```diff
@@ -238,6 +238,18 @@
           "code": 42,
           "message": "The collection name \"%s\" is invalid",
           "class": "BadRequestError"
+        },
+        "too_many_scroll": {
+          "description": "The maximum number of concurrent scroll requests as been reached. (See config.services.storageEngine.maxScroll)",
+          "code": 44,
+          "message": "Maximum number of concurrent scroll requests reached. (%s)",
```
scottinet: We shouldn't leak the configuration value for the maximum number of concurrent scroll requests: if  administrators want that information they can check the configuration files, normal users don't need it, and malicious users might find that information useful.

default.config.js
```diff
@@ -302,6 +302,8 @@ module.exports = {
           }
         }
       },
+      maxScroll: 300,
+      maxScrollDuration: '15m',
```
scottinet: 15m is a pretty long time for a default value, 1m should be largely sufficient for most use cases. What do you think?

lib/services/elasticsearch.js
```diff
@@ -242,15 +253,28 @@ class ElasticSearch extends Service {
 
     debug('Search: %j', esRequest);
 
+    if (scroll) {
+      const ongoingScrolls = await this._kuzzle.cacheEngine.internal.keys(
```
scottinet: Never use KEYS: it blocks other commands during its execution and can be harmful. Use scan instead.
(see https://redis.io/commands/keys)

default.config.js
Aschen: I'm thinking about the `kourou index:dump` action which could take more than 1m for large collections.  
Maybe the admin shouldn't be restrained by theses limitation?

scottinet: Or maybe should make that limit configurable per profile? :thinking:

scottinet: Also, scroll durations can be refreshed each time a new page is requested (see the scrollTTL option), so 1m should really be enough.

.kuzzlerc.sample
alexandrebouthinon: That's why I'd prefer we leave it to Elasticsearch to handle it.

lib/services/elasticsearch.js
```diff
@@ -196,25 +194,38 @@ class ElasticSearch extends Service {
...
-      .then(exists => {
-        if (exists === 0) {
-          throw errorsManager.get('unknown_scroll_id');
-        }
+    if (_scrollTTL) {
+      const scrollDuration = ms(_scrollTTL);
 
-        // ms(scroll) may return undefined if in microseconds or in nanoseconds
-        const ttl = ms(_scrollTTL) || ms(this._config.defaults.scrollTTL);
+      if (scrollDuration > ms(this._config.maxScrollDuration)) {
```
scottinet: We should probably add an assertion tested during this service's initialization, to make sure that the configuration is correct (i.e. that `ms(...)` does not return `undefined`).

Also, we might want to compute that value once and for all during the initialization, to prevent calling `ms` everytime we need it.

lib/services/elasticsearch.js
```diff
@@ -196,25 +194,38 @@ class ElasticSearch extends Service {
...
-      .then(({ body }) => this._formatSearchResult(body))
-      .catch(error => this._esWrapper.reject(error));
+    const exists = await this._kuzzle.cacheEngine.internal.exists(cacheKey);
+
+    if (exists === 0) {
+      throw errorsManager.get('unknown_scroll_id');
+    }
+
+    // ms(scroll) may return undefined if in microseconds or in nanoseconds
+    const ttl = ms(_scrollTTL) || ms(this._config.defaults.scrollTTL);
```
scottinet: I know this code was already there, but we should add an assert for this configuration too (and precompute the value).

.kuzzlerc.sample
Aschen: Just for you to know, that not the only limit duplicated into Kuzzle. Actually the `documentsWriteCount` limit is also a duplicate of an ES limit. See [Threadpool.write](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-threadpool.html) 

If you upgrade this limit or send to many requests (like 300 call to `document.update` in a short period of time), you will hit the ES limit which is 200 by default:
`EsThreadPoolExecutor[name = alyx/write, queue capacity = 200, org.elasticsearch.common.util.concurrent.EsThreadPoolExecutor@59828a6b[Running, pool size = 2, active threads = 2, queued tasks = 200, completed tasks = 1326511]]`

scottinet: Good point. But is is really the same thing? It seems that the thread pool is more similar to our funnel request buffer than to a hard limit about the number of updates to perform.

If I'm right, then this is not exactly the same thing and we shouldn't worry about it.
But if I'm wrong, then we should at lest, during the ES service init, fetch the current Threadpool.write limit from ES, and fail to start Kuzzle if our WriteLimit is greater than this value.

What do you think?

Aschen: The thread pool is more similar to the funnel request buffer yes so we don't need to read this value from ES. But I think we could use a better error message

PR #1584, opened 2020-03-18
Adds admin:reloadIndexCache

Aschen:
## Use case / Pain point

After loading a dump directly in Elasticsearch, Kuzzle index cache is obsolete and the new index/collections could not be used before Kuzzle restarts. 

## Feature

Adds a new API method: `admin:refreshIndexCache` to force a refresh of the internal index cache.  

I didn't adds functional tests on purpose since the feature is already tested when Kuzzle starts and it will be very costly to test this new API route (elasticdump, dump ES, restore ES..)

  - Step 1 :
  - Step 2 :
  - Step 3 :  

### Other changes

 - improve error message when collection does not exists

### Boyscout

 - some nitpicking on the documentation

### DIFF
--- a/doc/2/api/controllers/admin/load-fixtures/index.md
+++ b/doc/2/api/controllers/admin/load-fixtures/index.md
@@ -48,6 +48,7 @@ Body:
 {
   "controller": "admin",
   "action": "loadFixtures",
+  
   "body": {
     "index-name": {
       "collection-name": [
--- a/doc/2/api/controllers/admin/load-mappings/index.md
+++ b/doc/2/api/controllers/admin/load-mappings/index.md
@@ -50,6 +50,7 @@ Body:
 {
   "controller": "admin",
   "action": "loadMappings",
+  
   "body": {
     "index-name": {
       "collection-name": {
--- a/doc/2/api/controllers/admin/load-securities/index.md
+++ b/doc/2/api/controllers/admin/load-securities/index.md
@@ -62,6 +62,7 @@ Body:
 {
   "controller": "admin",
   "action": "loadSecurities",
+  
   "body": {
     "roles": {
       "role-id": {
--- a/doc/2/api/controllers/admin/refresh-index-cache/index.md
+++ b/doc/2/api/controllers/admin/refresh-index-cache/index.md
@@ -1,0 +1,42 @@
+---
+type: page
+
+code: true
+title: refreshIndexCache
+---
+
+# refreshIndexCache
+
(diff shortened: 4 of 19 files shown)


scottinet:
Hm...  I don't know if this is a good feature. 

Loading a dump is a heavy procedure. You cannot know what changes exactly: if you need to refresh kuzzle indexes, chances are that this is because kuzzle indexes were in the dump. 
That means that, for all you know, everything may change: configuration, token seed, plugin storages (e.g. hashed passwords), and so on. 

And what about the Koncorde cache for exiting documents? 

It seems to me that this is not the kind of operation that can be safely performed while Kuzzle is running. And this route might give the impression that this can be done while in reality, I doubt it does.

scottinet:
Also, this kind of operation might warrant a dedicated guide, where we would explain what operations should be done to safely reload a dump: depending on what information are loaded, you might also want to wipe Redis as well for instance.

Aschen:
> Hm... I don't know if this is a good feature.
> 
> Loading a dump is a heavy procedure. You cannot know what changes exactly: if you need to refresh kuzzle indexes, chances are that this is because kuzzle indexes were in the dump.
> That means that, for all you know, everything may change: configuration, token seed, plugin storages (e.g. hashed passwords), and so on.
> 
> It seems to me that this is not the kind of operation that can be safely performed while Kuzzle is running. And this route might give the impression that this can be done while in reality, I doubt it does.

When I said Kuzzle indexes, I was talking about public indexes and not internal one.  
Public indexes can be safely loaded while Kuzzle is running

Concerning the internal Kuzzle index, I agree that it could be more complicated but we could advise users to reboot Kuzzle when they want to load the internal index from a dump?

scottinet:
Perhaps you chould leave a notice in this new route documentation stating that it can only be used when public indexes have been reloaded? To make sure that no one will expect Kuzzle to continue to work after a complete ES restore.

Also, to answer your question, I think that Kuzzle should be shut down if private indexes are reloaded from a dump. Otherwise you cannot guarantee that operations such as a new password being created won't be submitted immediately after the dump, and before Kuzzle is restart, meaning you can end up with inconsistencies.

Aschen:
Be redacting a Guide on how to export and import data, I figured out that using Elasticdump is not the better way to import public collections so I close this PR because the `admin:refreshIndexCache` route does not seems useful for me now.  

If people want to restore a dump from Elasticdump, they should shutdown Kuzzle.  

If they want to import public data, they should use Kourou and then they don't need to refresh the index cache since Kourou use the public API

PR #1591, opened 2020-04-03
Better error message on 401 and 403

Aschen:
## What does this PR do ?

Better error message on 401 or 403 error.  

The former message was not helpful for the user:
image

Restrain anonymous user right on your Kuzzle and then run a query `kourou query server:now`

### Other changes
 - Also improve the document not found message with index and collection

### DIFF
--- a/doc/2/api/essentials/error-codes/security/index.md
+++ b/doc/2/api/essentials/error-codes/security/index.md
@@ -42,7 +42,7 @@ description: error codes definitions
 
 | Id | Error Type (Status Code)             | Message           |
 | ------ | -----------------| ------------------ | ------------------ |
-| security.rights.unauthorized<br/><pre>0x07030001</pre> | [UnauthorizedError](/core/2/api/essentials/error-handling#unauthorizederror) <pre>(401)</pre> | Authentication required |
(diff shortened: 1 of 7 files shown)


scottinet:
I understand the problem and it definitely needs solving, but I'm not entirely sure this is the right way to do it (won't block this PR for that, though).

Here is why I think that: API errors are part of the whole Response object (which is detailed here: https://docs.kuzzle.io/core/2/api/essentials/kuzzle-response/). And that object contains a lot of information about the request of origin, amongst them the invoked API controller+action.

It seems to me that the problem lies in SDKs, and more specifically their error objects: they probably should contain a few more information coming from the received response.

Aschen:
Yes, I agree that the error thrown by the SDK should contain the entire response object somewhere.  (I put a ticket on our product workshop)

But even with that, most of the time developers only print the error message so that's what we are gonna have most of the time to debug client. That is why I want those message to be as explicit as possible.

I don't have the time to go through all the error at once but I intend to do it incrementally when I encounter an error with a message that could use a little more context. What do you think?

scottinet:
I do  not see this as a solution. But more explicit error messages cannot hurt either.

PR #1592, opened 2020-04-06
Change and deprecate default pagination for collection:list

Aschen:
## What does this PR do ?

The collection:list method can be paginated with `from` and `size`.   
There is no reason to do that since it's only a list of string and also because an index will not contain thousands of collection.

This PR puts the default value very high and deprecate this usage

### Other changes

  - improve unknown collection error message

### DIFF
--- a/doc/2/api/controllers/collection/list/index.md
+++ b/doc/2/api/controllers/collection/list/index.md
@@ -39,9 +39,10 @@ Method: GET
 - `collection`: collection name
 - `index`: index name
 
-<DeprecatedBadge since="2.1.4" />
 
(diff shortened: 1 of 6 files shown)


doc/2/api/controllers/collection/list/index.md
```diff
@@ -43,6 +39,8 @@ Method: GET
 - `collection`: collection name
 - `index`: index name
 
+<DeprecatedBadge since="2.1.4" />
```
scottinet, line 42: should be put under the `Optional` chapter, a couple lines below

PR #1594, opened 2020-04-14
Lighter production dependencies

Aschen:
## What does this PR do ?

Remove some non-essentials Kuzzle dependencies:
 - `bytes`: used to parse string like `1 kb`, replaced with a simple method
 - `fs-extra`: used to make `mkdir -m` and `cp`, use native `fs` package instead
 - `glob`: used to list files starting by `core*`, use native JS instead
 - `kuzdoc`: moved to devDependencies
 - `ndjson`: used by the old cli to dump collection, removed
 - `readline-sync`: unused
 - `stack-chain`: unused
 - `triple-beam`: unused

Before: 809 total dependencies for 322 Mo

After: 794 dependencies for 319 Mo


scottinet:
Also, cli-color is nice for the upgrade script, it makes information much more readable...

Aschen:
> Also, cli-color is nice for the upgrade script, it makes information much more readable...

IYes you're right, I put it back until the upgrade script is externalized into Kourou

scottinet:
cli-colors is still removed from our dependencies :thinking:

Aschen:
> cli-colors is still removed from our dependencies thinking

:man_facepalming:

PR #1595, opened 2020-04-17
Install dependencies when starting a new development stack

Leodau:
## What does this PR do ?

Makes the _run-dev.sh_ script which is executed inside docker, install all dependencies (core & plugins) if they haven't been installed yet.

To run the stack, you would only need to clone and launch with docker-compose:

`git clone -b feature/automatic-dependencies-install https://github.com/kuzzleio/kuzzle.git && cd kuzzle && docker-compose up`

### DIFF
--- a/docker-compose/scripts/run-dev.sh
+++ b/docker-compose/scripts/run-dev.sh
@@ -8,11 +8,23 @@ fi
 
 elastic_host=${kuzzle_services__storageEngine__client__node:-http://elasticsearch:9200}
 
-if [ ! -z "$TRAVIS" ] || [ ! -z "$REBUILD" ] || [ -d "node_modules/" ]; then
+if [ ! -d "./node_modules/" ]; then
+    git submodule init
+    git submodule update
+    npm ci --unsafe-perm
+    ./docker-compose/scripts/install-plugins.sh
+fi
+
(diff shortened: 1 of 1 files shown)


Aschen:
Oh was is still WIP?

docker-compose/scripts/run-dev.sh
```diff
@@ -8,11 +8,23 @@ fi
 
 elastic_host=${kuzzle_services__storageEngine__client__node:-http://elasticsearch:9200}
 
+if [ ! -d "./node_modules/" ]; then
+    git submodule init
+    git submodule update
+    npm ci --unsafe-perm
+    ./docker-compose/scripts/install-plugins.sh
+fi
```
scottinet, line 16: These lines seem redundant: the `elif` condition below seems sufficient. What's their added value?

scottinet:
I've taken the liberty of changing the PR title (it's used as our changes history so it needs to be explicit), and the label: since this only affects development stacks, I believe that this shouldn't appear in our changelogs.

Let me know if you disagree.

docker-compose/scripts/run-dev.sh
Leodau: Yep :+1: , test/old chunk went through. Fixed

PR #1596, opened 2020-04-20
Keep the number of opened scroll contexts in check

scottinet:
# Description

Elasticsearch's scroll contexts are very expensive to maintain. So much in fact that ES impose a (configurable) limit on how many scroll contexts can exist at any given time([500 by default](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-request-body.html#scroll-search-context)).

ES API contains the [clearScroll](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-request-body.html#_clear_scroll_api) route, allowing to destroy an active scroll context.

This is especially useful when Kuzzle performs large numbers of search operations, such as done by `admin:resetSecurity`: I ran into a problem running that route with thousands of users in ES, each one containing 1 API Key. The command always failed with an ES error because there were too many scroll contexts opened. All by Kuzzle itself.

# The actual fix

There are multiple fixes:
1. Limit the number of scroll contexts opened by `repositories.user.truncate` (invoked by `admin:resetSecurity`) and in particular, get rid of `batchExecute` for removing API keys. Instead, use `deleteByQuery`. Also, `batchExecute` now clears the scroll context it creates.
1. Keep track of items fetched by scroll invocations: we already store a Redis key whenever a scroll ID is created to track what scroll contexts are used, and with what Kuzzle API route. Instead of having only the key existing or not, it now stores the number of items fetched. Once all items have been fetched through scroll calls, Kuzzle now destroys the scroll context immediately

All search/scroll queries executed by Kuzzle use their ES service counterpart, so all scroll contexts created by Kuzzle should now automatically be destroyed once all results have been fetched.

# How to reproduce

Load thousands of users, each one with an API key, into Kuzzle. Then run `admin:resetSecurity`.

Before this fix, the route fails (quite quickly) on an ES error because too many scroll contexts are opened.
This doesn't happen anymore with this fix.

# A word about tests

Normally, we would need test reproducing the bug, to prove that it is fixed. Due to the nature of this bug, only unit tests are ensuring that scroll contexts are correctly managed, but there is no actual test reproducing this specific bug.

### DIFF
--- a/lib/api/controllers/admin.js
+++ b/lib/api/controllers/admin.js
@@ -73,43 +73,36 @@ class AdminController extends NativeController {
   /**
    * Reset all roles, profiles and users
    */
-  resetSecurity (request) {
+  async resetSecurity (request) {
     this._lockAction(
       request,
       'Kuzzle is already reseting roles, profiles and users.');
 
-    const
-      options = { refresh: 'wait_for' },
-      result = {};
+    const result = {};
 
-    return this.kuzzle.repositories.user.truncate(options)
-      .then(deletedUsers => {
-        result.deletedUsers = deletedUsers;
-        return this.kuzzle.repositories.profile.truncate(options);
-      })
-      .then(deletedProfiles => {
-        result.deletedProfiles = deletedProfiles;
-        return this.kuzzle.repositories.role.truncate(options);
-      })
-      .then(deletedRoles => {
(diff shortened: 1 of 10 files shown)


scottinet:
Well... this is more about stability than performance: we were flooding ES' buffers :-(

PR #1598, opened 2020-04-22
Add token hash in API key

Aschen:
## What does this PR do ?

This PR adds an `hash` property corresponding to the SHA256 hash of the associated authentication token into the API key documents.  

This will allows users to search for an API key when they only got the authentication token.

```
$ kourou query auth:createApiKey --body '{ description: "API Key" }'
# Keep the token value 
$ echo -n <token> | sha256sum
# Then search
$ kourou query auth:searchApiKeys --body '{ term: { hash: "<hash>" }}'
```

### Other changes

 - fix `services.storage.not_found` error message (add index and collection)

### DIFF
--- a/doc/2/api/controllers/auth/create-api-key/index.md
+++ b/doc/2/api/controllers/auth/create-api-key/index.md
@@ -74,6 +74,7 @@ Returns an object containing the newly created API key:
   - `expiresAt`: expiration date in UNIX micro-timestamp format (`-1` if the token never expires)
   - `ttl`: original ttl
   - `description`: description
+  - `hash`: SHA256 hash of the authentication token
   - `token`: authentication token associated with this API key
 
 ::: warning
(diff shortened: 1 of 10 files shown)


scottinet:
If I get it right, this means that this feature allows searches returning 0 or 1 result: the searched API token.

And to search for it, then we have to produce a sha256 hash of the current token.

Am I right?

If so, API methods returning 0 or 1 return are usually `get` methods... wouldn't it be simpler for users to have an `auth:getApiToken` returning infos on the current API token (or a 404 - NotFound if no API token is associated to the current jwt)?

Aschen:
Yes that's right. But the `search` method is also useful to search for API key that will expires soon or matching content in the description ("list me sigfox api keys").

Users can already checks if an API key is still available with `auth:checkToken` but the hash is still convenient to identify API key in a list for example.
(Most of API key provided I saw give a fingerprint of the key for this purpose.)

Token exposure is still a risk but since the API key content is only returned from the [serialize](https://github.com/kuzzleio/kuzzle/blob/master/lib/core/storage/models/apiKey.js#L46) method who does not include the token by default. I think we should just be careful when we modify this code

scottinet:
Ok, I think I understand better what this feature is. I like the "fingerprint" term, in fact I find it far mor understandable than "hash". What do you think? 

Also, does that mean that we have convenient ways of getting that fingerprint, especially for the current token?

Aschen:
> Ok, I think I understand better what this feature is. I like the "fingerprint" term, in fact I find it far mor understandable than "hash". What do you think?
> 
> Also, does that mean that we have convenient ways of getting that fingerprint, especially for the current token?

Yes actually I also think that `fingerprint` is better :-) 

Not really but since it's "just" a SHA256, users can figure it out (eg `echo -n <token> | sha256sum`)
We could also provide a command in Kourou. 
Did you had something else in mind?

scottinet:
Nah, a bit of documentation and a kourou command should be enough I think :+1:

Aschen:
@scottinet I added a fix on an error message, you may want to review it as well

PR #1602, opened 2020-04-22
Add an empty updateByQuery after collection:update

Yoann-Abbes:
## What does this PR do ?
This PR adds an empty `updateByQuery` request after a `collection:update` in order to be able to perform `search` queries on properties that was not included in the mapping before being updated.

Functional test.
Create a collection with a mapping having `dynamic:false`
Create a document with a property not present in the mapping.
Update the mapping with a the same property/type

Before: The search query did not retrieve those documents.
Now : The search index is updated, the search query retrieve documents.

### DIFF
--- a/lib/services/elasticsearch.js
+++ b/lib/services/elasticsearch.js
@@ -1004,6 +1004,7 @@ class ElasticSearch extends Service {
     try {
       if (!_.isEmpty(mappings)) {
         await this.updateMapping(index, collection, mappings);
+        await this.updateSearchIndex(index, collection);
       }
     }
     catch (error) {
@@ -1022,6 +1023,36 @@ class ElasticSearch extends Service {
     }
 
     return null;
+  }
+
+  /**
+  * Sends an empty UpdateByQuery request to update the search index
+  *
+  * @param {String} index - Index name
+  * @param {String} collection - Collection name
+  * @returns {Promise.<Object>} {}
+  */
+  async updateSearchIndex(
+    index,
+    collection) {
+
+    let esRequest;
+    try {
+      esRequest = {
+        body: {},
+        index: this._getESIndex(index, collection),
+        refresh: true
+      };
+    }
+    catch (error) {
+      throw this._esWrapper.formatESError(error);
+    }
+
+    debug('UpdateByQuery: %o', esRequest);
+
+    await this._client.update_by_query(esRequest);
+
+    return;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


lib/services/elasticsearch.js
```diff
@@ -1004,6 +1004,7 @@ class ElasticSearch extends Service {
     try {
       if (!_.isEmpty(mappings)) {
         await this.updateMapping(index, collection, mappings);
+        await this.updateSearchIndex(index, collection);
```
scottinet, line 1007: This might be costly. When we discussed this item with @Aschen , we wondered if it wouldn't be better to only perform this operation if there is at least one `dynamic: 'false'` property set in the mappings (prior to update)

lib/services/elasticsearch.js
```diff
@@ -1024,6 +1025,34 @@ class ElasticSearch extends Service {
...
+  * @param {String} index - Index name
+  * @param {String} collection - Collection name
+  * @returns {Promise.<Object>} {}
+  */
+  async updateSearchIndex(
+    index,
+    collection) {
+
+    let esRequest;
+    try {
```
scottinet, line 1040: as it is, this `try..catch` statement is useless

did you mean to include `_client.update_by_query` in it?

Yoann-Abbes: in deed

lib/services/elasticsearch.js
```diff
@@ -1003,7 +1003,11 @@ class ElasticSearch extends Service {
 
     try {
       if (!_.isEmpty(mappings)) {
+        const previousMappings = await this.getMapping(index, collection);
         await this.updateMapping(index, collection, mappings);
+        if (_.isMatch(previousMappings, { dynamic: 'false'})) {
```
Aschen, line 1008: You should walk the entire previous mapping content to search for former `dynamic: false` property since it could have been set on any level.  
Example:
```
{
  dynamic: true,
  properties: {
    name: { type: 'keyword' },
    metadata: {
      dynamic: false,
      properties: {  }
    }
  }
}
```

scottinet: dang, I read the lodash documentation wrong, I saw the word "deep" and I assumed it made a deep seach... turns out, it "only" performs a deep comparison, which is a bit different :sweat_smile: 

Good catch :+1:

Yoann-Abbes: :disappointed: In deed, I missed it too

PR #1606, opened 2020-04-24
Lighter and faster MQTT server

scottinet:
# Description

Migrate Kuzzle's MQTT server to [Aedes](https://www.npmjs.com/package/aedes) instead of Mosca.
The former is a pure MQTT server, while the latter is more of a third-party MQTT server wrapper with an in-memory driver.

This means that Aedes has less dependencies (a lot less), is lighter, and is faster than Mosca.

Also, I'll add TLS support in another PR (to keep things easily readable)


lib/core/entrypoints/protocols/mqtt.js
```diff
@@ -198,50 +175,52 @@ class MqttProtocol extends Protocol {
...
   }
 
   _respond (client, response) {
+    debug('sending response: %o', response.content);
+
     if (process.env.NODE_ENV === 'development' && this.config.developmentMode) {
       this.broadcast({
         channels: [this.config.responseTopic],
-        payload: response.content
+        payload: response.content,
```
Aschen, line 215: I agree that we should always put a trailing comma :+1:

PR #1608, opened 2020-04-26
Update vault guide with Kourou

Aschen:
## What does this PR do ?

Add references to Kourou in the Kuzzle Vault guide

### DIFF
--- a/doc/2/guides/essentials/secrets-vault/index.md
+++ b/doc/2/guides/essentials/secrets-vault/index.md
@@ -24,6 +24,8 @@ Kuzzle offers a secure storage system for these secrets, the operation is as fol
 
 Thus, the only secret that it is necessary to communicate to the rest of a team is the encryption password for this file.
 
+See this project on [Github](https://github.com/kuzzleio/kuzzle-vault).
+
 ## Secrets file format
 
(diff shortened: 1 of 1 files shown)


doc/2/guides/essentials/secrets-vault/index.md
```diff
@@ -74,15 +76,17 @@ kuzzle decryptSecrets config/secrets.enc.json --vault-key strongpassword
 You can also specify the vault key in the `KUZZLE_VAULT_KEY` environment variable.
 :::
 
+You can see the complete list of Kuzzle Vault related command in Kourou on [Github](https://github.com/kuzzleio/kourou/#kourou-vaultadd-secrets-file-key-value).
```
scottinet, line 79:
```suggestion
You can see the complete list of Kuzzle Vault related commands in Kourou on [Github](https://github.com/kuzzleio/kourou/#kourou-vaultadd-secrets-file-key-value).
```

Aschen:
> I think we should deprecate the old kuzzle CLI before remove documentation about it. A good solution could be to have two sub part in the `Encrypt and decrypt with the CLI` such as `with Kuzzle legacy CLI` and `with Kourou CLI` slightly_smiling_face

IMHO we could only give a link to the old CLI on Github, what do you think?

Issue #1609, opened 2020-04-27
Documentation for plugin errorsManager is broken


benoitvidis:
https://docs.kuzzle.io/core/1/plugins/plugin-context/errors-manager/

The documentation states the `errorsManager` exposes 2 methods: `get` and `throw`, while, actually, the object returned is a wrapper around this module: https://github.com/kuzzleio/kuzzle/blob/master/lib/util/errors.js

The actually exposed methods are:

1. `get(errorId, ...placeholders)`
2. `getFrom(error, errorId, ...placeholders)`
3. `reject(errorId, ...placeholders)`
4. `rejectFrom(error, errorId, ...placeholders)`

Further, I've had a hard time figuring out the `rejectXX` methods need to be returned from a plugin to avoid an unhandled exception.

```javascript
async crashAction (request) {
  this.context.errorsManager.reject('my_error')  // <- this will cause an unhandled exception
}

async noCrashAction (request) {
  return this.context.errorsManager.reject('my_error')
}
```

Suggestion:
The behaviour of the `rejectXX` method look too dangerous to me.
I would actually replace the errorsManager object by a single `getError` method, equivalent to `errorsManager.get`.  
Exposing a dedicated method to throw an error brings very little added value IMO

PR #1610, opened 2020-04-27
Fix errors manager doc

Aschen:
## What does this PR do ?

Fix https://github.com/kuzzleio/kuzzle/issues/1609

### DIFF
--- a/doc/2/plugins/plugin-context/errors-manager/index.md
+++ b/doc/2/plugins/plugin-context/errors-manager/index.md
@@ -31,7 +31,7 @@ Example:
 
 The `errorsManager` exposes 2 functions:
   - `get(errorId, ...placeholders)`: Returns the corresponding error
-  - `getFrom(error, errorId, ...placeholders)`: Returns the corresponding error but keep the stacktrace from `error`
+  - `getFrom(error, errorId, ...placeholders)`: Returns the corresponding error with its stacktrace taken from the provided `error` argument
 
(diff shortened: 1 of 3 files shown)


Aschen:
> I am sorry, I just figured out the `rejectXX` methods were causing problem too.
How are they causing problem?

About the `getFrom`, we use this method to keep the original stacktrace

scottinet:
I agree with that suggestion: in the core I already planned to remove reject* methods because they do not make much sense, as they give the impression that they can interrupt the current flow.

In the end, only get* methods are relevant.

doc/2/plugins/plugin-context/errors-manager/index.md
```diff
@@ -13,26 +13,25 @@ Custom errors have to be specified in the [manifest.json](/core/2/plugins/guides
...
-    }
 }
 ```
 
-The `errorsManager` provides two functions:
-- To throw : `context.errorsManager.throw(error, placeholders);`.
-- To get the built error: `context.errorsManager.get(error, placeholders);`
+The `errorsManager` exposes 2 functions:
+  - `get(errorId, ...placeholders)`: Returns the corresponding error
+  - `getFrom(error, errorId, ...placeholders)`: Returns the corresponding error but keep the stacktrace from `error`
```
scottinet, line 34:
```suggestion
  - `getFrom(error, errorId, ...placeholders)`: Returns the corresponding error with its stacktrace taken from the provided `error` argument
```

PR #1611, opened 2020-04-28
create user before credentials

benoitvidis:
# Description

Authentication strategies may rely on kuzzle users information to (in)validate credentials.

This PR forces the Kuzzle user creation before the credentials are so that an auth strategy can safely fetch it, even on creation.

NB: no test added as the only functional change is possibly the Error rejected type in case the credentials creation would fail and throw a KuzzleError.


lib/api/controllers/security.js
```diff
@@ -1402,12 +1255,12 @@ class SecurityController extends NativeController {
...
 
   /**
- *
- * @param {Kuzzle} kuzzle
- * @param {string.<profile|role|user>} type
- * @param {Request} request
- * @returns {Promise<*>}
- */
+   *
+   * @param {Kuzzle} kuzzle
```
Aschen, line 1259: I know you didn't change this method but can you remove the `kuzzle` parameter and use the `this.kuzzle` member instead please?

lib/api/controllers/security.js
```diff
@@ -1480,6 +1333,176 @@ class SecurityController extends NativeController {
...
+      createdAt: Date.now(),
+      updatedAt: null,
+      updater: null
+    };
+
+    try {
+      loadedUser = await this.kuzzle.repositories.user.load(pojoUser._id);
+    }
+    catch (error) {
+      if (!(error instanceof NotFoundError)) {
```
Aschen, line 1358: It's better to use error codes now
```suggestion
      if (error.id !== 'services.storage.not_found') {
```

lib/api/controllers/security.js
```diff
@@ -1480,6 +1333,176 @@ class SecurityController extends NativeController {
...
+      .fromDTO(pojoUser)
+      .then(modifiedUser =>
+        this.kuzzle.repositories.user.persist(modifiedUser, options)
+      )
+      .then(modifiedUser => formatProcessing.serializeUser(modifiedUser));
+
+    // Creating credentials
+    let creationFailure = null;
+
+    for (const strategy of strategies) {
```
Aschen, line 1411: Nitpicking: this could be replaced by
```suggestion
    for (const [strategyName, StrategieContent] of Object.entries(strategies)) {
```

lib/api/controllers/security.js
```diff
@@ -1480,6 +1333,176 @@ class SecurityController extends NativeController {
...
+    if (loadedUser !== null) {
+      throw errorsManager.get('security', 'user', 'already_exists', pojoUser._id);
+    }
+
+    const strategies = request.input.body.credentials
+      ? Object.keys(request.input.body.credentials)
+      : [];
+
+    // Early checks before the user is created
+    for (const strategy of strategies) {
```
Aschen, line 1373: Nitpicking: this could be replaced by
```suggestion
    for (const strategy of Object.keys(strategies)) {
```

scottinet: @Aschen > I don't get it, what do you mean?

lib/api/controllers/security.js
scottinet: Same here: `strategies` is the result of an `Object.keys` so I don't think your suggestion does what you want it to do.

scottinet:
@benoitvidis @Aschen > I badly need this PR reviewed for my work about loose coupling, so I solved your issues @Aschen. Can you review it again?

lib/api/controllers/security.js
Aschen: Here honestly I don't remember :sweat_smile:

lib/api/controllers/security.js
Aschen: My point was to replace `strategies` by the content of `request.input.body.credentials` or an empty object.
But I'm not gonna block this PR for that :+1:

Aschen:
@scottinet merged :+1:

PR #1614, opened 2020-05-04
Add bulk:deleteByQuery

Aschen:
## What does this PR do ?

Adds `bulk:deleteByQuery` method.

This method is more directly linked to ES original `delete_by_query` method.  
The purpose is to bypass limitations of `document:deleteByQuery` about the limit of deleted documents.

You may want to deletes millions of documents matching a query but since Kuzzle has to trigger realtime notifications for every method of the document controller, the `document:deleteByQuery` method has to fetch document before delete them. That's why there is a limit.

But if you only want to delete a lot of documents matching a query you will have to make calls to `document:search` and then `document:delete`. For 200 000 documents and the default limit (200 documents), you will have to make 1000 calls to both methods which is unacceptable


doc/2/api/controllers/document/delete-by-query/index.md
```diff
@@ -14,7 +14,13 @@ Documents removed that way trigger real-time notifications.
 
 The request fails if the number of documents returned by the search query exceeds the `documentsWriteCount` server configuration (see the [Configuring Kuzzle](/core/2/guides/essentials/configuration) guide).
 
-To remove a greater number of documents, either change the server configuration, or split the search query.
+This behavior is to prevent Kuzzle from Out of Memory error or DOS because every deleted document has to be fetched first to trigger realtime notifications.
```
scottinet, line 17: I don't think using `deleteByQuery` can be considered a DoS attack.

```suggestion
This behavior aims at limiting the pressure on memory and on real-time notifications.
```

doc/2/api/controllers/document/delete-by-query/index.md
```diff
@@ -61,6 +68,10 @@ Body:
 - `collection`: collection name
 - `index`: index name
 
+### Optional:
```
scottinet, line 71:
```suggestion
### Optional
```

doc/2/api/controllers/document/delete-by-query/index.md
```diff
@@ -61,6 +68,10 @@ Body:
...
+
+- `refresh`: if set to `wait_for`, Kuzzle will not respond until the created/replaced documents are indexed
```
scottinet, line 73:
```suggestion
- `refresh`: if set to `wait_for`, Kuzzle will not respond until the deleted documents are removed from the search indexes
```

doc/2/api/controllers/bulk/delete-by-query/index.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+---
+
+## Arguments
+
+- `collection`: collection name
+- `index`: index name
+
+### Optional:
```
scottinet, line 59:
```suggestion
### Optional
```

doc/2/api/controllers/bulk/delete-by-query/index.md
```diff
@@ -0,0 +1,88 @@
...
+
+- `refresh`: if set to `wait_for`, Kuzzle will not respond until the deleted documents are indexed
```
scottinet, line 61:
```suggestion
- `refresh`: if set to `wait_for`, Kuzzle will not respond until the deleted documents are removed from the search indexes
```

test/api/controllers/bulk.test.js
```diff
@@ -188,4 +188,34 @@ describe('Test the bulk controller', () => {
         request, mCreateOrReplaceResult, true);
     });
   });
+
+  describe.only('#deleteByQuery', async () => {
```
scottinet, line 192:
```suggestion
  describe('#deleteByQuery', async () => {
```

doc/2/api/controllers/document/delete-by-query/index.md
Aschen: I meant that the OOM can lead to DOS but I agree it could be confusing

PR #1617, opened 2020-05-05
Allows inter-plugins method calls with side-effect

Aschen:
## What does this PR do ?

The [context.accessors.trigger](https://docs.kuzzle.io/core/2/plugins/plugin-context/accessors/trigger/)  method trigger an event with `kuzzle.pipe` but does not return the result.  

Because of this we can't use properly use this method for plugin communication because if the plugin A want to call some method from plugin B, it will not be able to apply any side effect from plugin B method.

This PR allows the plugin context to returns the pipe chain result in plugin context trigger method.

You can have a look on the functional test to see an example.

_PR name is open to debate_

IMHO we should expose the new event system to plugin so they can have a proper inter-plugin communication system but in the meantime this quickwin allow to do the same

### DIFF
--- a/doc/2/plugins/plugin-context/accessors/trigger/index.md
+++ b/doc/2/plugins/plugin-context/accessors/trigger/index.md
@@ -12,6 +12,7 @@ This allows interactions with other plugins using [hooks](/core/2/plugins/guides
 
 ::: info
 If the event is listened by pipes, the result of the pipe chain will be returned in a promise.
+If there is no listener, the result will be the same payload.
 This behavior can be used for a Remote Procedure Call (RPC) system between plugins.
 :::
 
--- a/doc/2/plugins/plugin-context/accessors/trigger/index.md
+++ b/doc/2/plugins/plugin-context/accessors/trigger/index.md
@@ -6,11 +6,14 @@ title: trigger
 
 # trigger
 
-
-
 Triggers a custom event.
 
 This allows interactions with other plugins using [hooks](/core/2/plugins/guides/hooks) or [pipes](/core/2/plugins/guides/pipes).
+
+::: info
+If the event is listened by pipes, the result of the pipe chain will be returned in a promise.
+This behavior can be used for a Remote Procedure Call (RPC) system between plugins.
+:::
 
 ## Arguments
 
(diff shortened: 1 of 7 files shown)


doc/2/plugins/plugin-context/accessors/trigger/index.md
```diff
@@ -6,12 +6,15 @@ title: trigger
...
 # trigger
 
-
-
 Triggers a custom event.
 
 This allows interactions with other plugins using [hooks](/core/2/plugins/guides/hooks) or [pipes](/core/2/plugins/guides/pipes).
 
+::: info
+If the event is listened by pipes, the result of the pipe chain will be returned in a promise.
```
scottinet, line 14: The result of that call if no pipe is listening to an event should be documented (and it should probably resolve to an undefined value)

test/core/plugins/context/context.test.js
```diff
@@ -347,24 +347,25 @@ describe('Plugin Context', () => {
     });
 
     describe('#trigger', () => {
-      it('should log an error if the event name contains a colon', () => {
-        context.accessors.trigger('event:with:colons');
-        should(kuzzle.log.error).be.calledOnce();
+      it('should rejects if the event name contains a colon', () => {
```
scottinet, line 350:
```suggestion
      it('should reject if the event name contains a colon', () => {
```

lib/core/plugins/context.js
```diff
@@ -282,15 +282,16 @@ function curryTrigger (kuzzle, pluginName) {
   /**
    * @this   {Kuzzle}
    * @param  {String} eventName The name of the custom event to trigger.
-   * @param  {Object} payload   The payload of the event.
+   * @param  {Any} payload      The payload of the event.
    */
   return function trigger (eventName, payload) {
     if (eventName.indexOf(':') !== -1) {
```
scottinet, line 288: I can't recall why we have such a limitation. And I wonder if that doesn't make triggering events impractical... Or even usable. What do you think?

Aschen: Yes, I removed it :+1:

doc/2/plugins/plugin-context/accessors/trigger/index.md
Aschen: I'm not sure about this one because pipes are meant to take a variable and may or not apply transformation to it and returning it.  
So IMHO it's normal if the pipe return the variable if there is no listener.

I made this to be able to have some kind of communication between plugins but with the new loosely coupling system I think `ask` event will be more appropriated to do some kind of RPC between plugins.   
And those events will return `undefined` if there is no listener right?

scottinet: Ah, you're right, it makes much more sense to return the first argument untouched if there is no pipe listening. :+1:

PR #1620, opened 2020-05-06
Prepare core directory for loose coupling

scottinet:
# Description

Move files around in the `lib/` directory to prepare for loose coupling.

Each entry in the `lib/` directory will soon be considered as independant modules, and they will progressively be converted to loose coupling.
As discussed, this conversion will be done module per module, starting only with modules necessary for the absorption of the cluster "plugin".

Philosophy: entries in the `lib/` directory will be considered as independant modules, "world" that can communicate between them only using events, once loose coupling is in place.
But objects in the same "world" can freely address eachothers directly. For instance, the "hotelClerk" and the "notifier" modules can invoke eachothers functions directly because they both belong to the "core/realtime" module, but in the future, that module can be called by other modules only using events.

# Changes 

(all directory names are relative to the `lib/` directory)

Unit test files/directories have been updated accordingly.

## lib/kuzzle

Before we had a (big) `kuzzle.js` file in `lib/` directory. Because we'll want to trigger `kuzzle:*` events, I created a new `lib/kuzzle` directory, with multiple files in it:

* `kuzzle/kuzzle.js`: the main kuzzle object we all know and love (minus a few methods)
* `kuzzle/dumpGenerator.js`: generate dumps for client support. Previously a function in `core/janitor.js`
* `kuzzle/events` (moved from `core/events`)
* `kuzzle/log.js` (moved from `util/`)
* `kuzzle/shutdown.js` (moved from `util/`)
* `kuzzle/signal.js`: signal handling code, previously in `kuzzle.js`
* `kuzzle/vault.js`: vault handling code, previously in `kuzzle.hs`

## lib/core

* Converted standalone files to modules:
  * `statistics.js` => `statistics/statistics.js`
* Renamed `entrypoints/` into `network/`:
  * moved `httpRouter/` and `router.js` in it
* New folder: `realtime/`:
  * contains the notifier module, previously located in the `core/notifier.js` file
  * contains the hotelClerk module, previously located in the `core/hotelClerk.js` file
* New folder: `security/`: contains the previous content of `models/repository` minus the `repository.js` file (see below)
* New folder: `shared/`: classes shared by multiple **independant** modules
  * contains the `repository.js` file, previously located in `models/`. This file is used by both the `security` module and the `plugins` one
  * contains the `abstractManifest.js` file, shared by the `plugins` and the `network` modules
* Moved `swagger.js` ini `lib/api/` (as this file is meant to generate an API documentation for swagger)
* Removed `janitor.js` and moved the code it contained in other modules:
  * `janitor.loadSecurities` => `repositories.loadSecurities`
  * `janitor.loadMappings` => `storageEngine.[public|internal].loadMappings`
  * `janitor.loadFixtures` => `storageEngine.[public|internal].loadFixtures`
  * `janitor.dump` => `kuzzle.dump`
  * `janitor.shutdown` => `kuzzle.shutdown`
* Deleted folder:
  * `models/` 

## lib/models

Contains classes modelizing standard API responses or internal documents. Can be used anywhere in the code.

* `models/notification`: contains the previous content of `core/models/notification`
* `models/security`: contains the previous content of `core/models/security`
* `models/storage`: contains the previous content of `core/storage/models`

## lib/services

Sometimes we need to have multiple files for the same service (see the `lib/util` section below).
So I added an extra level. This will also help in building event names:

* `services/elasticsearch.js` => `services/storage/elasticsearch.js`
* `services/redis.js` => `services/cache/redis.js`

## lib/util

Removed a few files from that directory: it should now only contains generic classes meant to be used by any module (e.g. debug, promback, memoize, ...)

* `util/DocumentExtractor.js` => `api/documentExtractor.js`
* `util/esWrapper.js` => `services/storage/esWrapper.js`
* `util/log.js` => `kuzzle/log.js`
* `util/runShutdown.js` => `kuzzle/shutdown.js`

# Kuzzle properties did NOT change

For instance, the new `network/` module is still referred as `this.kuzzle.entrypoints`.
This is intentional: there is no point modifying these entries and generate a HUGE PR, because we'll progressively loosely couple those modules. Soon, we'll refer to them with event names such as `core:network:...` so there is no need to update the Kuzzle object's properties.

# How to review

This should be fairly quick, as there is virtually no code change. 

Here is the exhaustive list of modifications:

* require paths have been converted
* when applicable, list of variable declarations have been converted to comply with eslint's [one-var rule](https://eslint.org/docs/rules/one-var)
* some functions have been converted to async/await (mainly in unit tests, but sometimes also in the code)
* some code parts have been removed from files, and isolated in files of their own
* new event: `kuzzle:start`. It deprecates `core:kuzzleStart`: proper event nomenclature, and this new event now accepts pipes, allowing plugins to insert init processes of their own before Kuzzle accepts new requests (documented)
* new event `kuzzle:shutdown`: it deprecates `core:shutdown` (fixed event nomenclature) (documented)

What should be reviewed: the new files distribution. Don't hesitate to comment and to discuss if you disagree with the choices I made.


Aschen:
Few points:
 - the `core/realtime/document/` directory contain the notifications class, did you mean `core/realtime/notifications/` ? Those classes does not represents documents (unlike the securities)
 - `statistics/index.js` I'm not very comfortable about naming `index.js` a file containing a class. IMHO it's more easier to retrieve files and classes when the filename correspond to the class name.
 - `storage/models` if we follow your logic this directory should be `storage/document`, we should choose either `document` or `models` I think
 - `validation/index.js` same as for `statistics.index.js`
 - `kuzzle/event/emitter` IMHO the file name should match the class name

scottinet:
@Aschen > 

* About "document" directories (`core/realtime/notifications/`, `core/storage/models`, `core/security/document`): these classes are instantiated by their corresponding modules, but they are used potentially anywhere in the code, since they are really class representations of standard messages or internal stored data. I wonder if a `lib/models` directory wouldn't be better, regrouping all of these classes in a single directory, with subdirectories auto-documenting the purpose for each class: `lib/models/security`, `lib/models/realtime`, `lib/models/storage`, ... What do you think?

* `stastistics/index.js`... I know you don't like this one, but I'm drawing a blank here. The best "solution" I can come up with is a `core/statistics/stastistics.js` class, and perhaps a `core/statistics/index.js` helper to help requiring that module. Same for `validation/index.js` obviously.

* `lib/kuzzle/event/emitter` => I think you're right, I'll rename that one right away.

Aschen:
> @Aschen >
> 
>     * About "document" directories (`core/realtime/notifications/`, `core/storage/models`, `core/security/document`): these classes are instantiated by their corresponding modules, but they are used potentially anywhere in the code, since they are really class representations of standard messages or internal stored data. I wonder if a `lib/models` directory wouldn't be better, regrouping all of these classes in a single directory, with subdirectories auto-documenting the purpose for each class: `lib/models/security`, `lib/models/realtime`, `lib/models/storage`, ... What do you think?
> 
>     * `stastistics/index.js`... I know you don't like this one, but I'm drawing a blank here. The best "solution" I can come up with is a `core/statistics/stastistics.js` class, and perhaps a `core/statistics/index.js` helper to help requiring that module. Same for `validation/index.js` obviously. I find this only mildly satisfactory, so I'd be glad if you have a better solution
> 
>     * `lib/kuzzle/event/emitter` => I'll rename that one right away +1

I agree for the `lib/models` directory and sub directory by modules, I would be more comfortable with that :+1: 

About statistics and validation, I have no better idea for now. We will see when we will have to refactor those modules.

scottinet:
@Aschen > I applied all your suggestions :+1:

Leodau:
Shouldn't we maintain folder suffix consistently?
'lib/model**s**, lib/service**s** & **lib/utils**' or 'lib/model, lib/service & **lib/util**'

Aschen:
> Shouldn't we maintain folder suffix consistency?
> 'lib/model**s**, lib/service**s** & **lib/utils**' or 'lib/model, lib/service & **lib/util**'

Good point :+1: (I prefer the plural version)

Yoann-Abbes:
> Shouldn't we maintain folder suffix consistency?
> 'lib/model**s**, lib/service**s** & **lib/utils**' or 'lib/model, lib/service & **lib/util**'

I also agree with that

scottinet:
@Aschen > I don't :sweat_smile: 
I just started the change and opted for the singular form, because I don't want to require `core/securities` or `core/storages/indexesStorage`, it just doesn't make sense to me.

scottinet:
@Leodau > I think you're right for most of directories, and I'm going a bit further than the examples you posted, but I think there are some directories that will stay plural because they contain "a collection of things". So far I've only one: the directory containing error codes (I'm moving it to `lib/error/codes`), using the singular form just seem wrong in that particular case.
WDYT?

PR #1624, opened 2020-05-13
Loose coupling: rename directories using plural form

scottinet:
# Description

This PR is a direct follow-up of https://github.com/kuzzleio/kuzzle/pull/1620#issuecomment-627372062

Directory names in plural form have been renamed for the sake of consistency (all directory paths below are relative to `lib/`):

* `core/api/controllers` => `core/api/controller`
* `core/plugins` => `core/plugin`
* `kuzzle/events` => `kuzzle/event`
* `models` => `model`
* `services` => `service`

Also, I moved the standardized error code and stored it in a new `kerror/` directory:
* `util/errors.js` => `kerror/index.js` (this is not a class, there are only a few getter functions in here)
* `config/error-codes/` => `kerror/codes/`

I also renamed all instances of `errorsManager` to `kerror` in the code: 
* shorter name so less prone to line wrapping
* more exact: there is no error management code in these functions, this directory contains only a convenient way to instantiate standard Kuzzle errors


Yoann-Abbes:
Did you update https://github.com/kuzzleio/kuzzle/blob/2-dev/doc/2/plugins/plugin-context/errors-manager/index.md ?

scottinet:
@Yoann-Abbes > wow good catch, I completely forgot about that one. Fixed.

PR #1629, opened 2020-05-15
Provides a HTTP GET endpoint for document:search 

Yoann-Abbes:
## What does this PR do ?
[Jira Ticket](https://jira.kaliop.net/browse/KZL-1493)

This PR provides HTTP GET endpoint for document:search, and retrieves properly the query string designated to ES 

Run tests with this version of the SDK JS
`npm install git://github.com/kuzzleio/sdk-javascript#KZL-1493-document-search-get --save`

### Other changes

update codecov to 3.7.0

### Boyscout
Remove unused and undocumented route `{verb: 'get', url: '/:index/:collection', controller: 'document', action: 'search'},`
Fix deadlinks on the documentation


lib/api/controller/base.js
```diff
@@ -632,14 +632,50 @@ class NativeController extends BaseController {
     return null;
   }
 
+  getSearchBody (request) {
```
Aschen, line 635: Looks like a duplicate of https://github.com/kuzzleio/kuzzle/blob/master/lib/service/storage/elasticsearch.js#L95

You should get the allowed search body key list with an `ask` event and the storage service must respond to this event

lib/api/controller/base.js
```diff
@@ -632,14 +632,50 @@ class NativeController extends BaseController {
...
+      'sort',
+      '_name',
+      '_source',
+      '_source_excludes',
+      '_source_includes'
+    ];
+    
+    let searchBody = {};
+    if ( request.context.connection.protocol === 'http'
+      && request.input.args.verb
```
Aschen, line 654: Useless condition since you check that `request.input.args.verb` is equal to `GET` after`.
Also IMHO this should be case insensitive

lib/api/controller/base.js
```diff
@@ -632,14 +632,50 @@ class NativeController extends BaseController {
...
+      && request.input.args.verb === 'GET'
+    ) {
+      searchBodyKeys.forEach(key => {
```
Aschen, line 657: Nitpicking: usually we use `for (const ... of ...)` when we are not in a critical code section

lib/config/httpRoutes.js
```diff
@@ -50,7 +51,6 @@ module.exports = [
   {verb: 'post', url: '/admin/_loadMappings', controller: 'admin', action: 'loadMappings'},
   {verb: 'post', url: '/admin/_loadSecurities', controller: 'admin', action: 'loadSecurities'},
 
-  {verb: 'get', url: '/:index/:collection', controller: 'document', action: 'search'},
```
Aschen, line 53: This looks like a breaking change to me

doc/2/api/controllers/document/search/index.md
```diff
@@ -31,7 +31,7 @@ You can restrict the scroll session maximum duration under the `services.storage
 ### HTTP
 
 ```http
-URL: http://kuzzle:7512/<index>/<collection>/_search[?from=<int>][&size=<int>][&scroll=<time to live>]
+URL: http://kuzzle:7512/<index>/<collection>/_search[?from=<int>][&size=<int>][&scroll=<time to live>][&verb=string]
```
Aschen, line 34: I don't see the point of passing the `verb` argument in the query string for Http since this information is already present in the native verb used for the request (GET or POST`

doc/2/api/controllers/document/search/index.md
```diff
@@ -50,6 +50,15 @@ Body:
 }
 ```
 
+You can also access this route with the `GET` verb:
+
+```http
+URL: http://kuzzle:7512/<index>/<collection>/_search[?query=<string>][?from=<int>][&size=<int>][&scroll=<time to live>][&verb=<string>]
```
Aschen, line 56: Ditto for the `verb` query string argument

doc/2/api/controllers/document/search/index.md
```diff
@@ -50,6 +50,15 @@ Body:
...
+Method: GET
+```
+
+'/nyc-open-data/yellow-taxi/
```
Aschen, line 60: Leftover?

doc/2/api/controllers/document/search/index.md
Yoann-Abbes: When you use the SDK and pass the verb option. the sdk will construct the query string like this. That is why I wrote it this way in the doc.
I let it because it is still an option, even if it is not really necessary...

doc/2/api/controllers/document/search/index.md
Yoann-Abbes: same as above

lib/config/httpRoutes.js
Yoann-Abbes: This route was not documented, is it a breaking change?

doc/2/api/controllers/document/search/index.md
Aschen: The SDK should not pass the `verb` in the query string but it should change the HTTP verb used for the query.  
Having a `verb` option is useless in HTTP since we already have the verb (also cache system as in webworkers don't care about a `verb` argument in the query string for the cache policiy, they only rely on the http verb)

lib/config/httpRoutes.js
Aschen: Technically not but why changing the route?

Yoann-Abbes: It has a conflict with another route when I hit it (with collection update. le SDK chose the last one)
and imho it's less confusing to have the same url for GET as for POST

lib/api/controller/base.js
scottinet: It doesn't look like a duplicate to me. The purpose of this method is to fetch the search body either in the request's body, or in its optional arguments, depending on whether it's a HTTP:GET request or not.
The responsibility of the search API is to parse the provided request and pass it to the ES service afterwards.

The lines you linked are about parsing the search body, and verifying if it's correct. And that's clearly the ES service's responsibility to me.

lib/api/controller/base.js
```diff
@@ -632,13 +632,31 @@ class NativeController extends BaseController {
...
+  * @param {Request} request
+  *
+  * @returns {Object}
+  */
+  getSearchBody (request) {
+    if ( request.context.connection.protocol === 'http'
+      && request.context.connection.misc.verb === 'GET'
+      && request.input.args.searchBody
+    ) {
+      return JSON.parse(request.input.args.searchBody);
```
scottinet, line 647: Parsing exceptions should be caught and converted into `api.assert.invalid_argument` errors.

test/api/controller/base.test.js
```diff
@@ -633,9 +634,38 @@ describe('#base/native controller', () => {
...
+
+    it('should extract searchBody param', () => {
+      request.context = { connection: { protocol: 'http', misc: { verb: 'GET' } } };
+      request.input.args.query = '{"query":{"foo":"bar"}}';
+      const searchBody = nativeController.getSearchBody(request);
+
+      should(searchBody).be.eql({ query: { foo: 'bar' } });
+    });
+
+    it('should extract searchBody param', () => {
```
scottinet, line 658: missing tests on what happens if:
* an invalid search body is provided
* no search body is provided and the route is invoked using HTTP:GET

scottinet:
This PR's description seems outdated, it seems that it doesn't depend on https://github.com/kuzzleio/sdk-javascript/pull/514 anymore

test/api/controller/base.test.js
Yoann-Abbes: shouldnt we provide an empty searchBody by default since that's what we do when we use POST ?

Yoann-Abbes:
> This PR's description seems outdated, it seems that it doesn't depend on [kuzzleio/sdk-javascript#514](https://github.com/kuzzleio/sdk-javascript/pull/514) anymore

It still depends on it for the functional tests, since it uses the sdk

lib/api/controller/base.js
```diff
@@ -632,13 +632,35 @@ class NativeController extends BaseController {
...
+    }
+    return this.getBody(request, {});
+  }
+
   getSearchParams (request) {
     const
       from = this.getInteger(request, 'from', 0),
       size = this.getInteger(request, 'size', 10),
       scrollTTL = this.getScrollTTLParam(request),
       query = this.getBodyObject(request, 'query', {}),
```
Yoann-Abbes, line 662: I can't see where this is used, why it exists ?

test/api/controller/base.test.js
scottinet: Of course. But tests don't check if that happens.

scottinet:
> It still depends on it for the functional tests, since it uses the sdk

Then it still cannot be deployed. You CANNOT have a circular dependency between repositories. 
Use `sdk.query` instead.

test/api/controller/base.test.js
```diff
@@ -633,9 +635,58 @@ describe('#base/native controller', () => {
...
+      request.context = { connection: { protocol: 'http', misc: { verb: 'GET' } } };
+      request.input.body = null;
+      request.input.args.searchBody = {};
+
+      should(() => {
+        nativeController.getSearchBody(request);
+      }).throw({ id: 'api.assert.invalid_type', message: 'Wrong type for argument "searchBody" (expected: string)' });
+    });
+
+    it('should throw when the route is invoked with GET and no search body is provided', () => {
```
scottinet, line 676: Should it? When I send a `document:search` without a null body with the current API, the search matches all documents, same as with an empty body. We should keep the same behavior with the `GET` verb.

lib/api/controller/base.js
Aschen: Hmm I don't know what happen because initially I put this comment on a list of allowed keys in the search body

PR #1631, opened 2020-05-19
Fix: invalid context on plugin pipes declared with a function name

scottinet:
# Description

This is an emergency patch, fixing a behavior broken by #1613 
Plugins can attach either functions or function names to pipe events. 

When using function names, plugin developpers have no way of controlling the context used to invoke their functions, and Kuzzle expects the function to exist within the plugin main object (this is verified with assertions when plugins are loaded by Kuzzle).

With Kuzzle 2.2.0, plugin pipes defined using function names are invoked with an undefined context, breaking existing implementations.

This PR fixes that by binding the plugin pipe function to its object of origin, when loading the plugin. 

This is only applied to plugin pipes defined with function names: when using functions, plugin developpers are responsible of the function's context, as Kuzzle has no way of knowing in what object the function belongs.

# How to test

Load a plugin with a pipe function declared using a function name, and use `this` in the function.

Example:

```js
'use strict';

class Foo {
  constructor() {
    this.foo = 'bar';
  }

  async init (config, context) {
    this.config = config;
    this.context = context;
    this.pipes = {
      'server:beforeNow': '_lol'
    };
  }

  async _lol (rq) {
    console.log('============= LOL: ', this.foo);
    return rq;
  }
}

module.exports = Foo;
```

Without this fix, this fails with a `PluginImplementationError` stating that the `foo` property cannot be read from `undefined`.
With the fix, this pipe runs as expected.

# Other changes

* Simplify unit tests a bit

### DIFF
--- a/lib/core/plugin/manager.js
+++ b/lib/core/plugin/manager.js
@@ -401,10 +401,10 @@ class PluginsManager {
    * @param {string|function} fn - function to attach
    */
   registerPipe (plugin, warnDelay, event, fn) {
+    // function names need to be bound to their object of origin to prevent
+    // invalid context execution
     const _fn = typeof fn === 'function'
       ? fn
-      // function names need to be bound to their object of origin to prevent
-      // invalid context execution
       : plugin.object[fn].bind(plugin.object);
     const name = plugin.manifest.name;
 
--- a/lib/core/plugin/manager.js
+++ b/lib/core/plugin/manager.js
@@ -401,7 +401,11 @@ class PluginsManager {
    * @param {string|function} fn - function to attach
    */
   registerPipe (plugin, warnDelay, event, fn) {
-    const _fn = typeof fn === 'function' ? fn : plugin.object[fn];
+    const _fn = typeof fn === 'function'
+      ? fn
+      // function names need to be bound to their object of origin to prevent
+      // invalid context execution
+      : plugin.object[fn].bind(plugin.object);
     const name = plugin.manifest.name;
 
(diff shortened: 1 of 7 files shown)


lib/core/plugin/manager.js
```diff
@@ -401,7 +401,11 @@ class PluginsManager {
    * @param {string|function} fn - function to attach
    */
   registerPipe (plugin, warnDelay, event, fn) {
-    const _fn = typeof fn === 'function' ? fn : plugin.object[fn];
+    const _fn = typeof fn === 'function'
+      ? fn
+      // function names need to be bound to their object of origin to prevent
+      // invalid context execution
+      : plugin.object[fn].bind(plugin.object);
```
Leodau, line 408: Comment inside a ternary operator may be a little too unorthodox?
```suggestion
      // function names need to be bound to their object of origin to prevent
      // invalid context execution
    const _fn = typeof fn === 'function' ? fn : plugin.object[fn].bind(plugin.object);
```

scottinet: Probably.

scottinet: Fixed. :+1:

Aschen:
A functional test in the functional test plugin would be better 👌

scottinet:
@Aschen > oh wow... we have a plugin dedicated to functional tests :astonished: 
I'll add this test right away.

scottinet:
Functional test added (and verified). I'll merge this PR once the CI passes.

PR #1632, opened 2020-05-19
Implement source argument on document:delete & document:deleteByQuery

Leodau:
## What does this PR do ?

Prevents the user in some situations from having to get a document before removing it.
Adds support for the optional 'source' argument, returns the already retrieved sources if required.

### Boyscout

- Added a base functional test for the **document:deleteByQuery**
- One-variable-per-declaration on the [document controller](https://github.com/kuzzleio/kuzzle/pull/1632/commits/263d94c00013cd36ba0b0c14b5fced89b557e758)
- Deprecates _array of ids_ in **document:deleteByQuery** response.

### DIFF
--- a/doc/2/api/controllers/document/delete-by-query/index.md
+++ b/doc/2/api/controllers/document/delete-by-query/index.md
@@ -6,7 +6,7 @@ title: deleteByQuery
 
 # deleteByQuery
 
-Deletes documents matching the provided search query. 
+Deletes documents matching the provided search query.
 
 Documents removed that way trigger real-time notifications.
 
@@ -31,7 +31,7 @@ To remove all documents from a collection, use [collection:truncate](/core/2/api
 ### HTTP
 
 ```http
-URL: http://kuzzle:7512/<index>/<collection>/_query[?refresh=wait_for]
+URL: http://kuzzle:7512/<index>/<collection>/_query[?refresh=wait_for][&source]
 Method: DELETE
 Body:
 ```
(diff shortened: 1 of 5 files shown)


doc/2/api/controllers/document/delete-by-query/index.md
```diff
@@ -94,11 +97,21 @@ Returns a `ids` array containing the list of deleted document identifiers.
   "action": "deleteByQuery",
   "requestId": "<unique request identifier>",
   "result": {
+    // Present only if 'source' parameter is not set, or set to false.
```
Aschen, line 100: I'm not comfortable to have two completely different response regarding the arguments.  
Also, we always returns array of documents in the API so maybe we can deprecate the `ids` response and returns an array of document with empty `_source` when the `source` parameter is not present.
What do you think?
(ping @scottinet )

Leodau: Yep, I was going to propose that, felt like having two different responses went against every principle, but i figured it was a breaking change.

Aschen: We can keep both the `ids` and `documents` arrays when the `source` option is not present so we avoid the breaking change.  
Then we deprecate the `ids` array in the documentation so people know they should stop using it

Leodau: Agreed :+1:

test/api/controller/document.test.js
```diff
@@ -814,7 +834,7 @@ describe('DocumentController', () => {
     });
   });
 
-  describe('#deleteByQuery', () => {
+  describe.only('#deleteByQuery', () => {
```
scottinet, line 837: debugging leftover: must be removed

PR #1635, opened 2020-05-24
Use nodemon instead of PM2 for core-dev and plugin-dev images

Aschen:
## What does this PR do ?

:warning: Depends on https://github.com/kuzzleio/kuzzle-containers/pull/54 :warning: 

Use Nodemon instead of PM2:
 - remove pm2 from scripts
 - remove pm2 log from core dump
 - remove pm2 from plugin-dev image
 - install nodemon in plugin-dev image
 - use nodemon in scripts
 - update [documentation](https://deploy-preview-1635--doc-kuzzle.netlify.app/guides/essentials/installing-kuzzle/)

Why ?
 - PM2 is annoying because it keep restarting when the application crash instead of waiting for changes before trying to restart
 - It has a configuration file
 - It is huge

Clone the repository (you really should have a fresh start) then run `docker-compose up`

### Other changes

 - Bump Node.js version to 12.16.3
 - Deprecate AWS marketplace install since it's only Kuzzle v1
 - remove unused ` docker-compose/scripts/devtools.sh`


doc/2/guides/essentials/installing-kuzzle/index.md
```diff
@@ -208,79 +212,47 @@ Kuzzle uses Elasticsearch and Redis as a persistent and key-value store, respect
 
 Please refer to the [configuration section](/core/2/guides/essentials/configuration) for more details.
 
-### Setup PM2
+### Setup Nodemon
 
-Now that you have installed Kuzzle and loaded its plugins, lets install pm2. Run the following command to install pm2 using NPM as a global package so that it can be run from anywhere on your machine:
+Now that you have installed Kuzzle and loaded its plugins, lets install nodemon. Run the following command to install nodemon using NPM as a global package so that it can be run from anywhere on your machine:
```
Leodau, line 217:
```suggestion
Now that you have installed Kuzzle and loaded its plugins, let's install nodemon. Run the following command to install nodemon using NPM as a global package so that it can be run from anywhere on your machine:
```

PR #1636, opened 2020-05-24
Reduce images size

Aschen:
## Reduce image size

~:warning: Depends on https://github.com/kuzzleio/kuzzle/pull/1635~

Reduce Kuzzle images size:
 - `kuzzleio/kuzzle:2`: from 563 Mo to 165 Mo
 - `kuzzleio/plugin-dev:2`: from 968 Mo to 410 Mo

This PR also includes 2 new tags:
 - `kuzzleio/kuzzle:2-alpine`: 140 Mo
 - `kuzzleio/kuzzle:2-scratch`: 94 Mo

I followed the tricks explained here: https://medium.com/trendyol-tech/how-we-reduce-node-docker-image-size-in-3-steps-ff2762b51d5a

Also I made few other changes for the production image:
 - remove all object and build files (`.o`, `.h`)
 - add only relevant files (`lib`, `config`, `plugins`, etc)
 - remove `.git`
 - remove useless npm dependencies
 - minify JS
 - strip `re2` and `node` binaries

The `scratch` image contains only Node.js (with libstdc++), musl and Kuzzle source code.

## Scripts

All docker related scripts are in `docker/scripts/`:

 - `entrypoint.sh`: the entrypoint
 - `clean-node.sh`: remove every unnecessary file from a directory. If `NODE_ENV=production` then the JS and JSON files will be minified
 - `run-dev.sh`: used by the `core-dev` image to run a core-developer stack
 - `run-plugin.sh`: used by `plugin-dev` to run an application composed of plugins (like we do now)
 - `run-prod.sh`: used by the `kuzzle` images to run kuzzle in production

Build the images and play the functional tests

### Other changes

 - Elasticsearch initialization is now waited with the `bin/wait-elasticsearch` script instead of curl
 - Functionals tests are run in parallel in 2 travis jobs


Aschen:
I really struggled with the `.dockerignore` file, it does not worked for me

.ci/privatebuild/Dockerfile
```diff
@@ -10,8 +10,8 @@ ARG SSH_KEY
 ENV GIT_SSH_COMMAND="ssh -o StrictHostKeyChecking=no"
 ENV NODE_ENV=production
 
-WORKDIR /var/app
-COPY . /var/app
+WORKDIR /app
```
alexandrebouthinon, line 13: This is a breaking change

bin/wait-elasticsearch
```diff
@@ -0,0 +1,80 @@
...
+  try {
+    await waitForEs(options.timeout);
+  }
+  catch (error) {
+    console.error(`[x] ${error.message}`);
+    process.exit(1);
+  }
+};
+
+run();
```
benoitvidis, line 78:
```suggestion
if (require.main === module) {
  run();
}
```
Else, will be called on require (i.e. from kuzzle-start)

docker/images/kuzzle.alpine/Dockerfile
```diff
@@ -0,0 +1,120 @@
...
+ADD ./plugins/available/ /app/plugins/available/
+
+RUN  set -x \
+  # Link default plugins
+  && mkdir -p /app/plugins/enabled/ \
+  && ln -s /app/plugins/available/kuzzle-plugin-auth-passport-local /app/plugins/enabled/kuzzle-plugin-auth-passport-local \
+  && ln -s /app/plugins/available/kuzzle-plugin-logger /app/plugins/enabled/kuzzle-plugin-logger \
+  # Install plugins dependencies
+  && for plugin in plugins/enabled/*; do cd "$plugin"; npm install --production; cd /app; done \
+  && for plugin in plugins/available/*; do cd "$plugin"; npm install --production; cd /app; done \
```
benoitvidis, line 43:
```suggestion
  && for plugin in plugins/*/*; do npm install --production --prefix "$plugin"; done \
```

docker/scripts/everest-clean-node.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Remove useless files
+# @todo remove when https://github.com/tj/node-prune/pull/68 is merged
+rm -rf .git/
+find . -type d -name "man" | xargs rm -rf
+find . -type f \( -name "*.o"  -o -name "*.h" -o -name "package-lock.json" -o -name "*.mk" -o -name "*Makefile*" -o -name "*.c" -o -name "*.cpp" -o -name "*.hpp" \) -exec rm {} \;
+
+# Minify JS & JSON
+# I wasn't able to find a portable way to do string comparison with if
+( echo $NODE_ENV | grep production ) || exit 0
```
benoitvidis, line 19:
```suggestion
[ "${NODE_ENV}" = "production" ] || exit 0
```
This is posix

Aschen:
I will wait for Kaaf before finishing this PR

benoitvidis:
I would definitely not go for the tar compression.

1. the time gained in downloading the image (once) is lost on each run
2. makes it difficult for the end-user to use custom entrypoints
3. makes it difficult to use custom volumes

Side node: As already mentioned by @alexandrebouthinon , moving the root dir is a breaking change. If you were to go for it, I would rather use /opt/kuzzle, which sounds more "natural" to me.

Aschen:
@benoitvidis About the compression, the decompression is very fast (~4-5s) and it allows to have only 19 mo transiting through the network instead of 94. It could be very useful for K8S environment where the pods have to download the new image before running for example.

Also the way I copy Kuzzle still allows developers to add plugins in the `plugins/*` directories

I will leave the `kuzzleio/kuzzle`, `kuzzleio/plugin-dev` and `kuzzleio/kuzzle:*-alpine` without compression and I will create a new `kuzzleio/kuzzle:*-everest` image with the compression for those who want extra small images, what do you think?

.ci/scripts/install-plugins.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+set -e
+set -x
```
alexandrebouthinon, line 4:
```suggestion
set -ex
```

.ci/scripts/run-test-arm.sh
```diff
@@ -1,8 +1,7 @@
 #!/bin/sh
 
 set -e
-
-elastic_host=${kuzzle_services__storageEngine__client__node:-http://elasticsearch:9200}
+set -x
```
alexandrebouthinon, line 4: `set` arguments could be combined

.ci/scripts/run-test.sh
```diff
@@ -1,57 +1,21 @@
 #!/bin/sh
 
 set -e
-
-elastic_host=${kuzzle_services__storageEngine__client__node:-http://elasticsearch:9200}
+set -x
```
alexandrebouthinon, line 4: Same

docker/scripts/run-dev.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+
+set -e
+
+if [ ! -z "$WITHOUT_KUZZLE" ]; then
```
alexandrebouthinon, line 5:
```suggestion
if [ -n "$WITHOUT_KUZZLE" ]; then
```

docker/scripts/run-dev.sh
```diff
@@ -0,0 +1,35 @@
...
+  exit 0
+fi
+
+if [ ! -z "$TRAVIS" ] || [ ! -z "$REBUILD" ]; then
```
alexandrebouthinon, line 9:
```suggestion
if [ -n "$TRAVIS" ] || [ -n "$REBUILD" ]; then
```

docker/scripts/everest-extract.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+
```
alexandrebouthinon, line 2: Missing `set -e`?

benoitvidis:
@Aschen , I am still not sure about everest image.

1. docker layers are already compressed. May be worth checking but my guess is there is gain to extra compress the layer for download time.
2. because of the mergefs system, you will get 2 layers on the disk: the one with the compressed tar + the diff once is is extracted, which is also counter productive considering the goal. The docker image ls here needs to be compared vs the real size of the corresponding layers in /var/lib/docker/overlay/merged|diff 
3. the rollback issues you are mentioning should not have much impact in most  cases, as only the application layer should be updated.

These 2 points + the already discussed UX issues make me really dubious about this image.

Aschen:
@benoitvidis 

1. Yes I know they are compressed but only when they are sended/received and with gzip which is not as efficient as lzma
image
The lzma compressed image is still 2 time less big as the gziped one.

2. You are right about this one, a running container will use more space on the filesystem but the stored previous images won't so you still gain a lot of space. Especially because images are not stored gzipped.

3. The thing is with Node.js the application layer is the bigger one
image

Aschen:
@benoitvidis I replaced the image by a `scratch` image without the compression (size 94 Mo)

I still think the compression is useful in a K8S environment because the pods has to download the image whenever they are spawned so it allows to save bandwidth, which is quite expensive in most of the cloud providers.

Docker compress layers with Gzip which is not as efficient as Lzma, especially for binary files:
image

But I don't want to argue anymore and we need those smaller image so LGTM

PR #1638, opened 2020-06-03
Allows external user sdk impersonation

Aschen:
## What does this PR do ?

Allows to use external user (any POJO with a string `_id` property) in SDK impersonation (`as` method)

This fix allows to pass an arbitrary user from the SDK for example:
```
const user = await this.context.accessors.sdk.security.getUser('gordon')
await this.context.accessors.sdk.as(user).query(...)
```

### Other changes

 - plugin submodules where on Kuzzle v2, I put them back on kuzzle v1

### DIFF
--- a/lib/api/core/plugins/pluginContext.js
+++ b/lib/api/core/plugins/pluginContext.js
@@ -24,7 +24,6 @@
 const
   errorsManager = require('../../../util/errors'),
   Bluebird = require('bluebird'),
-  User = require('../models/security/user'),
   Koncorde = require('koncorde'),
   { Kuzzle: KuzzleSDK } = require('kuzzle-sdk'),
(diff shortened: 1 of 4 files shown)


scottinet:
Also, I don't know if that's necessary, but pojos returned by getUser and User instances have slightly different forms: additional properties of a user are stored in _source in pojos, and at the root level of User instances. Does that matter? Should we add some kind of cast to use the same kind of objects when impersonating?

Aschen:
> Also, I don't know if that's necessary, but pojos returned by getUser and User instances have slightly different forms: additional properties of a user are stored in _source in pojos, and at the root level of User instances. Does that matter? Should we add some kind of cast to use the same kind of objects when impersonating?

For this fix I don't think it matters but for the upcoming complete impersonation (with rights) for Kuzzle v2 it will

PR #1641, opened 2020-06-05
Kuzzle as a framework

Aschen:
## What does this PR do ?

This PR introduce a new way to develop application with Kuzzle.  
You can now develop an application in the same way you do with express-like framework instead of developing a business plugin injected in the core from the filesystem.

This is the first step, I will introduce new base class for plugin development in a subsequent PR.

Everything is backward compatible with the old plugin API (loading from the filesystem).

The `kuzzle` package now expose a `Backend` class. You can instantiate this class and define features before starting your kuzzle application.   
This class has been written in Typescript to improve Developer Experience (DX).

Internally the `Backend` class is just a wrapper to create a `Plugin` (like any other plugin). The main difference is that plugin has the `application` flag.  
Kuzzle ensure that only one application plugin can be loaded.

Example:
```js
import { Backend } from 'kuzzle;

const app = new Backend('black-mesa');

app.version = '1.42.21';

app.plugin.use(new FunctionalTestPlugin());

app.controller.register('email', {
  actions: {
    get: {
      handler: async request => console.log(request.input)
      // default http route will be generated: GET /email/get
    },
    send: {
      handler: async request => console.log(request.input),
      http: [
        { verb: 'post', url: '/email/send' }
      ]
    },
  }
});

app.pipe.register('server:afterNow', async request => {
  app.context.log.info(`Server INFO: ${app.name}`);

  request.result.now = (new Date()).toUTCString();

  return request;
});

app.hook.register('server:beforeNow', () => {
  console.log('server before now');
});

app.config.set('services.foobar.host', 'redisHost');

app.config.merge({
  realtime: {
    pcreSupport: true
  }
});

app.start()
  .then(() => {
    console.log(`Application ${app.name} started`)
  })
  .catch(error => {
    console.error(error);
    process.exitCode = 1;
  });
```

### Notable changes

**New API definition**
To extend Kuzzle API, developers can use the `app.controller.register` method.  The new API for controller definition take an object containing actions and http routes.
Default http routes will be generated if the `http` property is not specified

```js
app.controller.register('email', {
  actions: {
    get: {
      handler: async request => console.log(request.input)
      // default http route will be generated: POST /email/get
    },
    send: {
      handler: async request => console.log(request.input),
      http: [
        { verb: 'post', url: '/email/send' }
      ]
    },
  }
});
```

**Url prefix**

Application URL are prefixed by `/_/` only.  
Plugin URL are prefixed by `/_/` if they start with a `/`, otherwise they are prefixed with `/_/<plugin-name>/`
Old plugin URL prefixed with `/_plugin/<plugin-name>` has been kept for backward compatibility.

### Other changes

 - Add two examples of a Kuzzle app with JS and TS in `examples/` 
 - Fix vault policy on missing key or encrypted secrets file

### Boyscout

 - fix link to JS SDK 6
 - use `Map` in the pluginManager

### Glossary

Application - Program that uses the Backend class to run a Kuzzle application and declare new features
Action (API) - Available action in the API, belongs to a controller (IMHO we should avoid the `API method` term because it could be confusing with function, class method, etc)


lib/core/framework/backend.ts
```diff
@@ -0,0 +1,490 @@
...
+import * as Plugin from '../plugin/plugin';
+import * as kerror from '../../kerror';
+import { kebabCase } from '../../util/inflector';
+
+import { ObjectWithStringKey } from '../../util/interfaces';
+
+const assertionError = kerror.wrap('plugin', 'assert');
+const runtimeError = kerror.wrap('plugin', 'runtime');
+
+/* PipeManager class ======================================================== */
```
scottinet, line 36: (open to debate) I'm not fond of that kind of comment. In my mind, this raises an alarm saying "this needs to be split in multiple files"

lib/core/plugin/plugin.js
```diff
@@ -0,0 +1,318 @@
...
+      controllers: [],
+      hooks: [],
+      manifest: this.manifest,
+      pipes: [],
+      routes: [],
+      strategies: [],
+      version: this.version
+    };
+
+    if (_.has(this.instance, 'hooks')) {
```
scottinet, line 102: (nitpicking) use `lib/util/safeObject` instead of lodash (much faster methods)
This is a nitpicking because speed isn't relevant here, but using the same methods for all our code is better I think

lib/kerror/codes/plugin.json
```diff
@@ -57,6 +57,24 @@
           "code": 9,
           "message": "A plugin named %s already exists",
           "class": "PluginImplementationError"
+        },
+        "invalid_plugin_name": {
+          "description": "The plugin name in is invalid. Plugin can only contain lowercase letters and dash.",
```
scottinet, line 62:
```suggestion
          "description": "The plugin name is invalid. Plugin can only contain lowercase letters and dash.",
```

scottinet: Also, why this restriction?
I see this as a breaking change.

lib/kerror/codes/plugin.json
```diff
@@ -87,6 +105,18 @@
           "code": 4,
           "message": "Request discarded: maximum number of executing pipe functions reached.",
           "class": "ServiceUnavailableError"
+        },
+        "already_started": {
+          "description": "Features definition can not be changed after startup.",
```
scottinet, line 110:
```suggestion
          "description": "Features definition cannot be changed after startup.",
```

lib/kerror/index.js
```diff
@@ -62,11 +62,46 @@ function get (domain, subdomain, error, ...placeholders) {
...
+  let errorMessage = '';
+  let i = error.name === 'PluginImplementationError' ? 2 : 1;
+  while (i--) {
+    const idx = error.stack.indexOf('\n') + 1;
+    errorMessage += error.stack.slice(0, idx);
+    error.stack = error.stack.slice(idx);
   }
 
-  return new errors[kuzzleError.class](message, id, code);
+  // Remove the first 3 lines of the stacktrace
```
scottinet, line 99: Instead of removing them, can we replace them with a single line telling that some lines have been removed?
Something like `[...Kuzzle internal calls deleted...]`

This is not entirely for aesthetics: I'm wondering if removing lines from the stacktrace is entirely safe, and if there won't be cases were these lines shouldn't be removed. In these cases, having some kind of information telling that the stacktrace is incomplete can help immensely when debugging.

Also, I think that `InternalError` errors shouldn't be cleaned up.

doc/2/api/essentials/error-codes/plugin/index.md
```diff
@@ -25,6 +25,7 @@ description: error codes definitions
 | plugin.assert.privileged_not_set<br/><pre>0x04010007</pre> | [PluginImplementationError](/core/2/api/essentials/error-handling#pluginimplementationerror) <pre>(500)</pre> | The 'privileged' flag has been set in the plugin's manifest file, but it needs also to be added in Kuzzle's configuration |
 | plugin.assert.not_a_constructor<br/><pre>0x04010008</pre> | [PluginImplementationError](/core/2/api/essentials/error-handling#pluginimplementationerror) <pre>(500)</pre> | The exposed plugin is not a constructor |
 | plugin.assert.name_already_exists<br/><pre>0x04010009</pre> | [PluginImplementationError](/core/2/api/essentials/error-handling#pluginimplementationerror) <pre>(500)</pre> | Another plugin with the same name has already been loaded |
+| plugin.assert.invalid_plugin_name<br/><pre>0x0401000a</pre> | [PluginImplementationError](/core/2/api/essentials/error-handling#pluginimplementationerror) <pre>(500)</pre> | The plugin name in the manifest is invalid. Plugin can only contain lowercase letters and dash. |
```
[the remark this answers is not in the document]
Aschen: We use the lowercase representation everywhere, even if the original name contained uppercase.
But I agree that it may cause a breaking change so I will just print a warning instead

lib/kerror/index.js
Aschen: I remove the lines only when we use our error wrapper, and every time we use it we have those 3 additional lines because the error object is instantiated here and not on the `throw` line.

lib/core/framework/backend.ts
Aschen: I didn't wanted to create new classes at the beginning because those classes are only kind of set to edit internal properties of the `Backend` class but I decided to create classes so we can benefits from the strong typing.
I can split them in file but we will have 6 additional files containing a class with only one or 2 methods

docker-compose.yml
```diff
@@ -39,7 +39,7 @@ services:
       - kuzzle_server__protocols__mqtt__developmentMode=false
       - kuzzle_limits__loginsPerSecond=50
       - NODE_ENV=${NODE_ENV:-development}
-      - DEBUG=kuzzle:*,-kuzzle:entry-point:protocols:websocket
+      - DEBUG=${DEBUG:-kuzzle:*,-kuzzle:entry-point:protocols:websocket}
```
alexandrebouthinon, line 42: This does not work, try this:
```suggestion
      - DEBUG=$DEBUG,-kuzzle:*,-kuzzle:entry-point:protocols:websocket
```

Aschen: I used this syntax because I want to use the value of `DEBUG` is set and otherwise I want `kuzzle:*,-kuzzle:entry-point:protocols:websocket`

doc/2/api/controllers/admin/load-securities/index.md
```diff
@@ -10,11 +10,11 @@ title: loadSecurities
 
 Load roles, profiles and users into the storage layer.
 
-The roles, profiles and users definitions follow the same structure as in the body parameter of these corresponding API routes:
+The roles, profiles and users definitions follow the same structure as in the body parameter of these corresponding API actions:
```
Leodau, line 13:
```suggestion
The roles, profiles and users definitions follow the same structure as in the body parameter of their corresponding API actions:
```

doc/2/api/controllers/document/delete-by-query/index.md
```diff
@@ -84,9 +84,7 @@ Body:
 
 Returns an object containing information about the deleted documents:
 
-<DeprecatedBadge version="2.2.1">
-- `ids`: an array containing the list of each deleted documents identifier.
-</DeprecatedBadge>
+- `ids`: an array containing the list of each deleted documents identifier <DeprecatedBadge version="2.2.1"/>.
```
Leodau, line 87:
```suggestion
- `ids`: an array containing the list of each deleted document identifier <DeprecatedBadge version="2.2.1"/>.
```

package.json
```diff
@@ -90,10 +98,26 @@
...
+    "strip-json-comments": "^3.1.1",
+    "ts-node": "^8.10.2",
+    "typescript": "^3.9.3"
   },
   "engines": {
     "node": ">= 12.13.0"
   },
-  "license": "Apache-2.0"
+  "license": "Apache-2.0",
+  "files": [
```
scottinet, line 109: You should probably include the following files:
* `.kuzzlerc.sample`
- `*.md`

And I wonder if we shouldn't include those either:
- `Vagrantfile`
- `vagrant/**/*`

Lastly: I cannot recall what's the use for the `config/` and `run/` directories.

lib/api/controller/server.js
```diff
@@ -171,6 +171,7 @@ class ServerController extends NativeController {
    * @returns {Promise<Object>}
    */
   async info () {
+    /* eslint-disable sort-keys */
```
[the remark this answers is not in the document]
Aschen: I wanted to put the `application` definition next to the `plugins` but you are right it's better to keep alphabetic order

package.json
Aschen: I'm not sure about Vagrant, those files are very old and I don't even know if it's still working.. 

In `config` Kuzzle try to look for `secrets.enc.json` but I don't know for `run/`

scottinet: I believe it was to host an [unix socket](https://en.wikipedia.org/wiki/Unix_file_types#Socket), used in a past now forgotten when we had our own kuzzle proxy. 
This needs to be verified but if I'm right, it's now unused.

PR #1644, opened 2020-06-08
Update swagger to OpenAPI3

Leodau:
## What does this PR do ?
Updates the old swagger route, to [OpenApi 3.0.1](http://spec.openapis.org/oas/v3.0.1) and a few more tweaks.

openapi

- Includes a link to Kuzzle Documentation
- Groups routes by controller using tags (Greatly improves readability)
- Includes custom metadata **If** an OpenApi 'info' object is detected in the route object (Set's defaults data if object is incomplete)
- Supports request body definition

## **Route.info** example
```js
  {
    verb: 'get',
    url: '/credentials/:strategy/_me',
    controller: 'auth',
    action: 'getMyCredentials',
    info: {
      description: 'Custom Route description',
      parameters: [{
        name: 'strategy',
        in: 'path',
        required: true,
        description: 'Parameter description',
        schema: {type: 'string'}}
      ]
    }
  }
```
_More info about parameters [here](https://swagger.io/docs/specification/describing-parameters/)._

Having Kuzzle running, go to [/swagger.json](http://localhost:7512/swagger.json), then copy/paste into [https://editor.swagger.io/](https://editor.swagger.io/)

### What's next?
Another PR would update **each** Kuzzle's API routes to **fully** support open api

### DIFF
--- a/lib/api/swagger.js
+++ b/lib/api/swagger.js
@@ -66,7 +66,8 @@ module.exports = function generateSwagger (kuzzle) {
       {
         'url': '/'
       }
-    ]
+    ],
+    paths: {}
   };
 
   routes.forEach(route => {
--- a/lib/api/swagger.js
+++ b/lib/api/swagger.js
@@ -93,9 +93,9 @@ module.exports = function generateSwagger (kuzzle) {
         route.infos.description += `\nController: ${route.controller}. Action: ${route.action}.`;
       }
 
-      if (route.infos.produces === undefined) {
-        route.infos.produces = ['applications/json'];
-      }
+      //   if (route.infos.produces === undefined) {
+      //     route.infos.produces = ['applications/json'];
+      //   }
 
       if (route.infos.parameters === undefined) {
         route.infos.parameters = [];
--- a/lib/api/swagger.js
+++ b/lib/api/swagger.js
@@ -19,6 +19,8 @@
  * limitations under the License.
  */
 
+/* eslint sort-keys: 0 */
+
 'use strict';
 
 const _ = require('lodash');
--- a/lib/api/swagger.js
+++ b/lib/api/swagger.js
@@ -45,7 +45,7 @@ module.exports = function generateSwagger (kuzzle) {
   });
 
   const swagger = {
-    openApi: '3.0.1',
+    openapi: '3.0.1',
     info: {
       title: 'Kuzzle API',
       description: 'The Kuzzle HTTP API',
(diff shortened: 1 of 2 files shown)


Aschen:
> Another PR would update each Kuzzle's API routes to support open api

I didn't understand this part? Each Kuzzle API route isn't already present in the OA3 definition?

Leodau:
> > Another PR would update each Kuzzle's API routes to support open api
> 
> I didn't understand this part? Each Kuzzle API route isn't already present in the OA3 definition?

Yes, all the routes are present in the output, but they are not **fully** compatible, given that as you can see in the example, routes need to have an info object containing more information about the parameters & response...

lib/api/swagger.js
```diff
@@ -19,6 +19,8 @@
  * limitations under the License.
  */
 
+/* eslint sort-keys: 0 */
```
Aschen, line 22: Is this valid only for this file or until the rule is re-activated?

lib/api/swagger.js
```diff
@@ -36,97 +38,115 @@ module.exports = function generateSwagger (kuzzle) {
...
+      route.info.parameters = [];
+
+      let m = routeUrlMatch.exec(route.url);
+      while (m !== null) {
+        routeUrlMatch.lastIndex++;
+        route.info.parameters.push({
+          name: m[1],
+          in: 'path',
+          required: true,
+          schema: {type: 'integer'}
```
Aschen, line 130: Why integer? Isnt always string in URL?

lib/api/swagger.js
Leodau: If the rule is in the outer scope, it works for the entire file. BUT, I should have used _disable_ instead, as suggested in the [docs nitpicky](https://eslint.org/docs/user-guide/configuring.html#disabling-rules-with-inline-comments)

```suggestion
/* eslint-disable sort-keys */
```

lib/api/swagger.js
Leodau: Ops, slipped test. Should be _string_ as default.

```suggestion
          schema: {type: 'string'}
```

Btw, even thought url's contain strings params, the whole idea of having a full api documentation is to be able to _also_ precisely define parameter types, this may help us in many ways (using the api-ui, generating code from this doc..). 
[Schema.type is mandatory in OAS, and it is quite complete (required, defaults, enums, integer ranges...).](https://swagger.io/docs/specification/describing-parameters/)

jenow:
What's wrong with codecov?

PR #1645, opened 2020-06-08
Allows to use realtime subscriptions in plugins

Aschen:
## What does this PR do

This PR allows to use realtime subscriptions in plugin with the embedded SDK.

Subscription should be done in the `init` method or in a hook to the `core:start:before` event.  

:book: https://deploy-preview-1645--doc-kuzzle.netlify.app/plugins/plugin-context/accessors/sdk/

```js
class MyPlugin {
  async init (config, context) {
    this.sdk = context.accessors.sdk;


    await this.sdk.realtime.subscribe('index', 'collection', {}, notif => {
      console.log(notif.result);
    });
  }
}
```

### Behavior

When you subscribe to notification in the `init` method or in the `core:kuzzleStart` event, then every node make a subscription.

#### with `replicate: false`
   
When a request triggering a notification is received by one of the node, only his own subscription is triggered.  
It allows to send an email from a real-time subscription (you don't want each node to send an email obviously)

#### with `replicate: true`

When a request triggering a notification is received, then every node trigger his own subscription.  
It allows to maintain RAM cache for example. (eg: indexCache, repositories) 

### How it works

A new `InternalProtocol` has been created. This protocol only have one connection with the router.  

The `FunnelProtocol` used by the plugin SDK send every request with the `InternalProtocol` connection ID so when the notifier want to transmit realtime notification for a subscription made with the internal SDK it's sended to the `InternalProtocol`.

The `InternalProtocol` then send the notification through Kuzzle internal event system and the `FunnelProtocol` is listening to it and then it dispatch it to the SDK object with the `roomId` using the same mechanism as for every SDK protocol.

To avoid replicating the notification on every cluster node, a `cluster` option has been introduced in the channels description (and as an argument of the `realtime:subscribe` method). This argument is `true` by default.

The `HotelClerc` will not emit cluster subscription synchronization related events for channels that have the `cluster` option set to `false`.   
So a subscription made in a plugin with `cluster: false` will not be propagated to other node (each node will only be aware of his own local subscription).

The `Notifier` methods `_notifyDocument` and `_notifyUser` can be called either from the node that is actually triggering the notification (from `notifyDocument` and `notifyUser`) or from the cluster code when the `core:notify:document` or `core:notify:user` are emitted.  
When the methods are called from the primary node then an additional parameter is passed to `_notify*` methods and then it will also trigger channels that have the `cluster: false` option.  
When they are called from the cluster, this additional parameter is not present and so channels with the `cluster: false` option are not triggered.

### Known limitations

If a subscription is made inside a controller method without the `replicate: true` option, then the subscription will only be triggered when this node receive a request meant to trigger a realtime notification.  

In fact it could be considered harmful to subscribe to notifications at runtime because the notification will only be present on one node so if this node is shutdown or unavailable then the subscription will never be triggered.

IMHO runtime subscription should be avoided until we don't have a subscription propagation mechanism for app/plugins

First, you will have to declare a plugin who subscribe in the `init` method.

Then you can run a cluster with the following commands in three different terminals:
```bash
# term 1
$ WITHOUT_KUZZLE docker-compose up

# term 2
$ kuzzle_plugins__cluster__privileged=true kuzzle_server__port=7512 NODE_ENV=development DEBUG=kuzzle:*,-kuzzle:events*,-kuzzle:entry-point:protocols:websocket,-kuzzle:funnel,-kuzzle:entry-point:protocols:http,-kuzzle:http:router,-kuzzle:plugins,-kuzzle:services:elasticsearch nodemon bin/start-kuzzle-server --enable-plugins functional-test-plugin,kuzzle-plugin-cluster

# term 3
$ kuzzle_plugins__cluster__privileged=true kuzzle_server__port=8512 kuzzle_plugins__cluster__bindings__router="tcp://[_site_:ipv4]:8510" kuzzle_plugins__cluster__bindings__pub="tcp://[_site_:ipv4]:8511" NODE_ENV=development DEBUG=kuzzle:*,-kuzzle:events*,-kuzzle:entry-point:protocols:websocket,-kuzzle:funnel,-kuzzle:entry-point:protocols:http,-kuzzle:http:router,-kuzzle:plugins,-kuzzle:services:elasticsearch nodemon bin/start-kuzzle-server --enable-plugins functional-test-plugin,kuzzle-plugin-cluster
```

You can then send request to a specific node with Kourou:
```bash
# Send to the first node on port 7512
$ kourou sdk:query realtime:publish -i index -c collection --body '{ hello: "adrien" }' --port 7512

# Send to the second node on port 8512
$ kourou sdk:query realtime:publish -i index -c collection --body '{ hello: "adrien" }' --port 8512
```

### Other changes

 - rename `entryPoints` to `entryPoint`
 - protocol names should be passed as parent constructor argument
 - deprecate `kuzzle:start` and add `kuzzle:start:before` and `kuzzle:start:after`
 - add new `Channel` class

### Boyscout

 - one const per line
 - asyncify some methods
 - use Map instead of object for RAM caches
 - use `Bluebird.all` instead of `Promise.all`


lib/core/network/protocols/internal.js
```diff
@@ -0,0 +1,102 @@
...
+    debug('notify: %a', data);
+
+    this._send(data);
+  }
+
+  _send (data) {
+    for (let i = 0; i < data.channels.length; i++) {
+      const message = { ...data.payload, room: data.channels[i] };
+
+      this.kuzzle.emit('core:network:internal:message', message);
```
Aschen, line 97: This event is listened by the `FunnelProtocol` to receive realtime notifications

lib/core/realtime/hotelClerk.js
```diff
@@ -158,21 +175,23 @@ class HotelClerk {
...
-    return this._subscribeToRoom(roomId, request)
-      .then(response => {
-        result = response.data;
-        return this.kuzzle.pipe('core:hotelClerk:join', response.diff);
-      })
-      .then(() => result);
+    const response = await this._subscribeToRoom(roomId, request);
+
+    if (response.cluster) {
+      // @todo should be an emit since we don't use the result?
```
Aschen, line 190: Can this be an emit since we don't use the result?

lib/core/network/entryPoint.js
```diff
@@ -153,8 +186,9 @@ class EntryPoint {
     }
 
     try {
-      this.protocols[client.protocol].leaveChannel(channel, connectionId);
-    } catch (e) {
+      this.protocols.get(client.protocol).leaveChannel(channel, connectionId);
+    }
+    catch (e) {
```
Aschen, line 191: Shouldn't this be a runtime error instead?

lib/api/controller/realtime.js
```diff
@@ -54,9 +53,16 @@ class RealtimeController extends NativeController {
     assertHasIndexAndCollection(request);
     assertHasBody(request);
 
+    if (request.input.args.cluster === undefined) {
```
Aschen, line 56: Unclear, `replicate` could be better

Leodau: `replicate` or `clusterReplicate` may be more explicit.

doc/2/api/essentials/error-codes/plugin/index.md
```diff
@@ -108,7 +108,7 @@ description: error codes definitions
 | Id | Error Type (Status Code)             | Message           |
 | ------ | -----------------| ------------------ | ------------------ |
 | plugin.context.missing_collection<br/><pre>0x04060001</pre> | [PluginImplementationError](/core/2/api/essentials/error-handling#pluginimplementationerror) <pre>(500)</pre> | The 'collection' argument is required, but none was provided |
-| plugin.context.unavailable_realtime<br/><pre>0x04060002</pre> | [PluginImplementationError](/core/2/api/essentials/error-handling#pluginimplementationerror) <pre>(500)</pre> | Due to technical limitations, plugins have limited access to the realtime API controller |
+| plugin.context.unavailable_realtime<br/><pre>0x04060002</pre> | [PluginImplementationError](/core/2/api/essentials/error-handling#pluginimplementationerror) <pre>(500)</pre> | Due to technical limitations, subscribe and unsubscribe method are only available in the embedded SDK. |
```
Leodau, line 111:
```suggestion
| plugin.context.unavailable_realtime<br/><pre>0x04060002</pre> | [PluginImplementationError](/core/2/api/essentials/error-handling#pluginimplementationerror) <pre>(500)</pre> | Due to technical limitations, subscribe and unsubscribe methods are only available in the embedded SDK. |
```

Aschen: I changed it in`lib/kerror/codes/plugin.json`

doc/2/api/controllers/realtime/subscribe/index.md
```diff
@@ -33,7 +33,8 @@ Due to the synchronous nature of the HTTP protocol, real-time notifications are
     // query volatile data
   },
   "scope": "<all|in|out|none>",
-  "users": "<all|in|out|none>"
+  "users": "<all|in|out|none>",
+  "cluster": true
```
alexandrebouthinon, line 37:
```suggestion
  "replicate": true
```

doc/2/plugins/plugin-context/accessors/sdk/index.md
```diff
@@ -69,13 +69,72 @@ async myAwesomePipe (request) {
...
+  context.accessors.sdk.realtime.subscribe(
+    'nyc-open-data',
+    'yellow-taxi',
+    {},
+    notification => {
+      // this callback will be executed only on one node
+    });
+
+
+  // the default value for the "cluster" option is "true"
```
alexandrebouthinon, line 119:
```suggestion
  // the default value for the "replicate" option is "true"
```

lib/core/realtime/notifier.js
```diff
@@ -372,72 +390,80 @@ class NotifierController {
...
   }
 
   /**
    * Broadcasts a notification about a document change or a
-   * real-time message
+   * real-time message.
+   *
+   * When this method is called from the node who received
+   * the request triggering notification, then every channel is notified
+   * regardless the "cluster" option.
```
alexandrebouthinon, line 410:
```suggestion
   * regardless the "replicate" option.
```

lib/core/realtime/notifier.js
```diff
@@ -372,72 +390,80 @@ class NotifierController {
...
+   *
+   * When this method is called from the cluster synchronization,
+   * then only the channel having the "cluster: true" option are notified
```
alexandrebouthinon, line 413:
```suggestion
   * then only the channel having the "replicate: true" option are notified
```

test/core/realtime/hotelClerk/addSubscription.test.js
```diff
@@ -169,12 +169,36 @@ describe('Test: hotelClerk.addSubscription', () => {
...
-    return should(() => hotelClerk.join(joinRequest)).throw(NotFoundError, {
-      id: 'core.realtime.room_not_found'
-    });
+    return should(hotelClerk.join(joinRequest))
+      .be.rejectedWith(NotFoundError, {
+        id: 'core.realtime.room_not_found'
+      });
+  });
+
+  it('#join should propagate notification only with "cluster" option', async () => {
```
alexandrebouthinon, line 181:
```suggestion
  it('#join should propagate notification only with "replicate" option', async () => {
```

test/core/realtime/hotelClerk/removeRoomForCustomer.test.js
```diff
@@ -128,27 +128,49 @@ describe ('lib/core/hotelclerk:removeRoomForCustomer', () => {
...
+    should(hotelClerk.roomsCount).be.eql(1);
+    should(kuzzle.notifier.notifyUser).calledWithMatch(
+      'roomId', sinon.match.instanceOf(Request), 'out', {count: 1});
+
+    should(kuzzle.pipe).be.calledWith(
+      'core:hotelClerk:removeRoomForCustomer',
+      { requestContext, room: hotelClerk.rooms.get('roomId') });
+  });
+
+  it('should only propagate channels with the "cluster" option', async () => {
```
alexandrebouthinon, line 157:
```suggestion
  it('should only propagate channels with the "replicate" option', async () => {
```

lib/core/network/entryPoint.js
Leodau: Maybe the service shouldn't crash if the protocol fails to do any action, should it keep trying though?

Aschen: Yes that's why I think.
(ping @scottinet )

lib/core/plugin/sdk/embeddedSdk.js
```diff
@@ -0,0 +1,96 @@
...
+   */
+  query (request, options = {}) {
+    // By default, do not propagate realtime notification accross cluster nodes
+    if ( isPlainObject(request)
+      && request.controller === 'realtime'
+      && request.action === 'subscribe'
+    ) {
+      request.replicate = options.replicate === undefined
+        ? false
+        : options.replicate;
```
Yoann-Abbes, line 85: maybe you could ave a getReplicate function ? If its called to be reused in other context ? Just a question

Aschen: I could but this is never gonna be reused somewhere else since it's very specific

lib/core/realtime/notifier.js
Aschen: The option is named `cluster` for channel because it has more meaning to have "cluster" channels and not "replicate" channels

test/core/realtime/hotelClerk/removeRoomForCustomer.test.js
Aschen: The option is named `cluster` for channel because it has more meaning to have "cluster" channels and not "replicate" channels

test/core/realtime/hotelClerk/addSubscription.test.js
Aschen: The option is named `cluster` for channel because it has more meaning to have "cluster" channels and not "replicate" channels

doc/2/plugins/plugin-context/accessors/sdk/index.md
Aschen: Good catch ;)

lib/core/realtime/notifier.js
Aschen: Ditto

doc/2/plugins/plugin-context/accessors/sdk/index.md
```diff
@@ -69,13 +69,72 @@ async myAwesomePipe (request) {
...
+## Realtime notifications
+
+<SinceBadge version="change-me" />
+
+The [realtime](/sdk/js/7/controllers/realtime) is entirely available with the [subscribe](/sdk/js/7/controllers/realtime/subscribe) and [unsubscribe](/sdk/js/7/controllers/realtime/unsubscribe) methods.  
+
+Realtime subscription should be made in the plugin [init](core/2/plugins/guides/manual-setup/init-function) method or in a hook on the [kuzzle:start:before](/core/2/plugins/guides/events/kuzzle-start) event.
+
+::: warning
+If you want to make subscription in hooks, pipes, or controller actions then you should use the `replicate: true` option otherwise your callback will not be executed with every notification.
```
scottinet, line 81: Hard to understand, and possibly a source of many support tickets.

I have multiple possible suggestions to improve the wording of this warning, but first I need to understand something that bothers me:
* when `replicate: true` is set, what is replicated? The subscription, or notifications resulting from it?
* in the latter case, will a request submitted on a different node:
  * trigger the notification on the node where the plugin made the subscription?
  * trigger the notification on the same node than where the notification is generated (the subscription is replicated but the notification is not propagated)?
  * don't trigger the notification

doc/2/plugins/plugin-context/accessors/sdk/index.md
```diff
@@ -69,13 +69,72 @@ async myAwesomePipe (request) {
...
+:::
+
+When you receive a notification, only one node will execute the associated callback.  
+
+::: info
+  You can use the `replicate` option if you want your callback function to be executed on each nodes.
```
scottinet, line 87: I don't get it. If plugins subscribe before kuzzle starts, then the probability that they will register the same filters is close to 1. And that means that all plugin instances should receive the notification, independantly to the value of the `replicate` option.

scottinet: I wonder if this is an architecture problem, or a semantic one: could `propagate` be a better keyword for this behavior?

doc/2/plugins/plugin-context/accessors/sdk/index.md
```diff
@@ -69,13 +69,72 @@ async myAwesomePipe (request) {
...
+:::
+
+#### replicate: false (default)
+
+With `replicate: false`, the callback function will be executed only on the node receiving a request who triggers a notification. (only one execution)
```
scottinet, line 92: Impractical: subscriptions should always be kept in sync in all nodes. The question is if a notification is propagated/duplicated or not. Am I right?

doc/2/plugins/guides/events/kuzzle-start/index.md
```diff
@@ -6,6 +6,24 @@ title: kuzzle:start
...
 # kuzzle:start
 
+<DeprecatedBadge version="change-me" />
+
+Triggered when Kuzzle has finished its startup sequence, and is about to accept network connections.
+
+Kuzzle will wait for pipe listeners to return before accepting user requests.
+
+
+## kuzzle:start:before
```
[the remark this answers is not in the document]
Aschen: Since Kubernetes since to became a real things now, maybe we can copy the probe event names indicating the system state?
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

doc/2/plugins/guides/events/kuzzle-start/index.md
```diff
@@ -6,6 +6,24 @@ title: kuzzle:start
...
+
+<SinceBadge version="change-me" />
+
 Triggered when Kuzzle has finished its startup sequence, and is about to accept network connections.
 
 Kuzzle will wait for pipe listeners to return before accepting user requests.
+
+## kuzzle:start:after
+
+<SinceBadge version="change-me" />
```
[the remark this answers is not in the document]
Aschen: Now we are sure it's gonna be a new feature but in general you don't know exactly the next Kuzzle version when you make your PR because it could be an enhancement but then a PR with a new feature is merged and the next release will be a minor and not a patch.
We should change those tag in the release tool IMHO

doc/2/plugins/plugin-context/accessors/sdk/index.md
Aschen: Proposal: `executionStrategy: cluster|node`

scottinet: Also hard to understand. I might want a "node" behavior in a cluster, and this new proposal can lead to the assumption that this should not be used that way.

If I think about how to document the feature to users, I think that `propagate` is simpler to understand. We could write something like this:

```
The `propagate` option allows to define if, for that subscription, notifications should be propagated to all cluster nodes, or if only the node having received the triggering event should execute it.
Values: true, false
```

wdyt?

doc/2/plugins/guides/events/kuzzle-start/index.md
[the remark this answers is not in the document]
scottinet: That documentation page tells a lot of things, and I'm not really sure about what you're proposing, could you be a bit more specific on that one?

They do speak of `probe readiness" however... "ready" is really the usual way of telling that a system is operational.

doc/2/plugins/guides/events/kuzzle-start/index.md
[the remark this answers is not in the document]
scottinet: I like that idea, and that shouldn't be too hard to code either.

lib/core/network/entryPoint.js
scottinet: Should not be a crash, and it should keep trying. Don't forget that protocols can be user-made, and that kind of error might happen if, for instance, this action is triggered when cleaning real-time rooms _after_ the connection has been dropped, so we're asking a protocol to register that a non-existing connection has left a channel. It might trigger an exception on that one, and that doesn't mean it's worth to stop sending further events of that kind to it.

doc/2/plugins/guides/events/kuzzle-start/index.md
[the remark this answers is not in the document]
Aschen: I will modify the release-tool with the next release :+1:

doc/2/plugins/plugin-context/accessors/sdk/index.md
Aschen: LGTML :+1:

doc/2/plugins/guides/events/kuzzle-start/index.md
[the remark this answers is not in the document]
Aschen: K8S use 3 types of probes to define application state:
 - startup: the application start booting sequence, useful to kill/restart (or emit an error) if the container is to slow to boot
 - liveness: the application is successfully running (will reboot the container of this probe fail)
 - readyness: the application is ready to receive requests

So we could trigger a first event when Kuzzle starts it's booting sequence, another one when the booting sequence is over and the last one when Kuzzle can accepts requests

(ping @alexandrebouthinon @rolljee )

doc/2/plugins/plugin-context/accessors/sdk/index.md
Aschen: Do you still need insight on this one since we discuss the matter ?

doc/2/plugins/guides/events/kuzzle-start/index.md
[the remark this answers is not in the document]
scottinet: Ok so start, live, ready? 

Why not.

doc/2/plugins/plugin-context/accessors/sdk/index.md
scottinet: Nope I'm fine now. Thanks.

doc/2/plugins/plugin-context/accessors/sdk/index.md
```diff
@@ -69,13 +69,70 @@ async myAwesomePipe (request) {
...
+
+With `propagate: false`, the callback function will be executed only on the node receiving a request who triggers a notification. (only one execution) 
+
+::: info 
+This behavior is suitable for most usage like sending emails, write in the database, call an external API, etc.
+:::
+
+#### propagate: true
+
+With `propagate: true`, the callback function will be executed one each node of the cluster. (n executions)
```
scottinet, line 98:
```suggestion
With `propagate: true`, notifications are propagated to all nodes of a cluster, executing all callback functions.
```

doc/2/plugins/plugin-context/accessors/sdk/index.md
```diff
@@ -69,13 +69,70 @@ async myAwesomePipe (request) {
...
+#### Example
+
+```js
+async init (config, context) {
+  context.accessors.sdk.realtime.subscribe(
+    'nyc-open-data',
+    'yellow-taxi',
+    {},
+    notification => {
+      // this callback will be executed only on one node
```
scottinet, line 113:
```suggestion
      // this callback will be executed only on the node generating a notification
```

doc/2/plugins/plugin-context/accessors/sdk/index.md
```diff
@@ -69,13 +69,70 @@ async myAwesomePipe (request) {
...
+    });
+
+
+  // the default value for the "propagate" option is "true"
```
scottinet, line 117: Isn't it `false` by default?

lib/api/controller/realtime.js
```diff
@@ -54,9 +53,18 @@ class RealtimeController extends NativeController {
     assertHasIndexAndCollection(request);
     assertHasBody(request);
 
+    if ( request.input.args.propagate === undefined
+      || request.input.args.propagate === null
+    ) {
+      request.input.args.propagate = true;
```
scottinet, line 59: Shouldn't we allow that option only for subscriptions made by plugins?
(we can limit this by using the value of the `request.context.connection.protocol` property)

lib/core/network/protocols/internal.js
```diff
@@ -0,0 +1,106 @@
...
+  leaveChannel (channel, connectionId) {
+    debug('leaveChannel: %s', channel, connectionId);
+
+    this.channels.delete(channel);
+  }
+
+  disconnect (connectionId) {
+    debug('disconnect: %s', connectionId);
+
+    // Never happen, the InternalProtocol always keep his only connection open
```
scottinet, line 82:
```suggestion
    // Never happens, the InternalProtocol always keep his only connection open
```

lib/core/network/protocols/protocol.js
```diff
@@ -21,47 +21,62 @@
...
     this.entryPoint = entryPoint;
 
-    return Promise.resolve()
-      .then(() => {
-        this.name = name;
-        this.maxRequestSize = bytes(entryPoint.config.maxRequestSize);
+    // name should be passed in the constructor
+    if (! this.name) {
+      this.name = name;
+    }
```
scottinet, line 61: (nitpicking) possible candidate for an assertion test, failing init if a `name` value is provided both to the constructor and to the init methods.

lib/core/plugin/sdk/embeddedSdk.js
```diff
@@ -0,0 +1,96 @@
...
+    if ( isPlainObject(request)
+      && request.controller === 'realtime'
+      && request.action === 'subscribe'
+    ) {
+      request.propagate = options.propagate === undefined || options.propagate === null
+        ? false
+        : options.propagate;
+
+      if (this.kuzzle.started && process.env.NODE_ENV !== 'production') {
+        this.kuzzle.log.warn('A realtime subscription has been made at runtime.\nRealtime subscription should only be made ar Kuzzle initialization to ensure a correct cluster replication.\nSee https://docs.kuzzle.io/core/2/plugins/plugin-context/accessors/sdk/#realtime-notifications for more infos.');
```
scottinet, line 88:
```suggestion
        this.kuzzle.log.warn('A realtime subscription has been made at runtime.\nRealtime subscription should only be made during Kuzzle initialization to ensure correct cluster replication.\nSee https://docs.kuzzle.io/core/2/plugins/plugin-context/accessors/sdk/#realtime-notifications for more information.');
```

lib/core/realtime/hotelClerk.js
```diff
@@ -24,9 +24,30 @@
 const Bluebird = require('bluebird');
 const { Request } = require('kuzzle-common-objects');
 const kerror = require('../../kerror');
+const debug = require('../../util/debug')('kuzzle:realtime:hotelClerc');
 
 const realtimeError = kerror.wrap('core', 'realtime');
 
+const CHANNEL_ALLOWED_VALUES = ['all', 'in', 'out', 'none'];
```
scottinet, line 31: (nitpicking) could be a static getter in the `Channel` class, as this appear to be a dedicated constant to that class.

lib/core/realtime/hotelClerk.js
scottinet: Nope: the cluster uses this event to sync other nodes. We need to keep the request slot occupied until the sync finishes,  to prevent event loop overloads, so this needs to be a blocking operation.

lib/kerror/codes/plugin.json
```diff
@@ -349,9 +349,9 @@
           "class": "PluginImplementationError"
         },
         "unavailable_realtime": {
-          "description": "Due to technical limitations, plugins have limited access to the realtime API controller",
+          "description": "Due to technical limitations, subscribe and unsubscribe methods are only available in the embedded SDK.",
```
scottinet, line 352:
```suggestion
          "description": "Due to technical limitations, subscribe and unsubscribe methods are only available through the embedded SDK.",
```

lib/kuzzle/kuzzle.js
```diff
@@ -28,7 +28,7 @@ const stringify = require('json-stable-stringify');
 const Koncorde = require('koncorde');
 
 const config = require('../config');
-const EntryPoints = require('../core/network');
+const EntryPoint = require('../core/network/entryPoint');
```
scottinet, line 31: (note / no action required) This change was unnecessary: soon this will be loosely coupled ;-)

test/core/plugin/sdk/funnelProtocol.test.js
```diff
@@ -13,28 +14,42 @@ const {
...
+
+    it('should forward messages received from the internalProtocol to the SDK', done => {
+      const payload = { room: 'room-id', hello: 'Gordon' };
+
+      funnelProtocol.on('room-id', () => {
+        should(funnelProtocol.kuzzle.on).be.calledOnce();
+        should(funnelProtocol.kuzzle.on.getCall(0).args[0])
+          .be.eql('core:network:internal:message');
+
+        done();
```
scottinet, line 48: If a `should` assertion fails, the callback won't be invoked, potentially making it painful to debug.

```suggestion
        try {
          should(funnelProtocol.kuzzle.on).be.calledOnce();
          should(funnelProtocol.kuzzle.on.getCall(0).args[0])
            .be.eql('core:network:internal:message');

          done();
        }
        catch (e) {
          done(e);
        }
```

test/core/realtime/hotelClerk/addSubscription.test.js
```diff
@@ -169,12 +169,36 @@ describe('Test: hotelClerk.addSubscription', () => {
       collection: 'bar',
       controller: 'realtime',
       action: 'join',
-      body: {roomId: 'no way I can exist'}
+      body: {roomId: 'i-exists'}
```
scottinet, line 172:
```suggestion
      body: {roomId: 'i-exist'}
```

test/core/realtime/hotelClerk/addSubscription.test.js
```diff
@@ -169,12 +169,36 @@ describe('Test: hotelClerk.addSubscription', () => {
...
+      });
+  });
+
+  it('#join should propagate notification only with "cluster" option', async () => {
+    const joinRequest = new Request({
+      index: 'foo',
+      collection: 'bar',
+      controller: 'realtime',
+      action: 'join',
+      body: {roomId: 'i-exists'}
```
scottinet, line 187:
```suggestion
      body: {roomId: 'i-exist'}
```

test/core/realtime/hotelClerk/addSubscription.test.js
```diff
@@ -169,12 +169,36 @@ describe('Test: hotelClerk.addSubscription', () => {
...
+    }, context);
+    const response = { cluster: false, diff: 'diff', data: 'data'};
+    hotelClerk.rooms.set('i-exists', {});
```
scottinet, line 190:
```suggestion
    hotelClerk.rooms.set('i-exist', {});
```

doc/2/plugins/plugin-context/accessors/sdk/index.md
Aschen: That's not true, the notifications are always propagated but they will not trigger notifications but since it's an internal behavior I'm OK with this sentence if it's more easy to understand for the users

lib/core/realtime/hotelClerk.js
Aschen: That's what I've done in the first place but eslint/vscode complained.. I will try to bypass them

scottinet: Don't spend time on it, it's just a nitpicking.

doc/2/plugins/guides/events/kuzzle-state/index.md
```diff
@@ -0,0 +1,23 @@
...
+code: true
+type: page
+title: kuzzle:state
+---
+
+# kuzzle:state
+
+<SinceBadge version="change-me" />
+
+These events allows to hooks on Kuzzle startup sequence. They follow [Kubernetes probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/).
```
scottinet, line 11: These states names are pretty much standard and used by a lot of systems. 

```suggestion
These events allow to hook on Kuzzle startup sequence. 
```

doc/2/plugins/guides/events/kuzzle-state/index.md
```diff
@@ -0,0 +1,23 @@
...
+
+# kuzzle:state:start
+
+This event is triggered when Kuzzle begin it's startup sequence.
```
scottinet, line 15:
```suggestion
This event is triggered when Kuzzle begins its startup sequence.
```

doc/2/plugins/guides/events/kuzzle-state/index.md
```diff
@@ -0,0 +1,23 @@
...
+
+# kuzzle:state:live
+
+This event is triggered when Kuzzle has loaded every internal components but it's not accepting incoming requests.
```
scottinet, line 19:
```suggestion
This event is triggered when Kuzzle has loaded every internal components but is not yet accepting user requests.
```

Dockerfile
```diff
@@ -35,7 +35,10 @@ RUN  set -x \
   && rm -rf /var/lib/apt/lists/* \
   && echo "alias ll=\"ls -lahF --color\"" >> ~/.bashrc
 
-ADD . /var/app
+ADD ./package.json /var/app/package.json
+ADD ./package-lock.json /var/app/package-lock.json
```
alexandrebouthinon, line 39: There is no need to add this file since it will be overwritten by using `npm install`. However, you could keep it and use `npm this` for installing dependencies.

Dockerfile
```diff
@@ -44,6 +47,13 @@ RUN  npm install --unsafe-perm \
   && for plugin in plugins/enabled/*; do cd "$plugin"; npm install --unsafe-perm; cd /var/app; done \
   && for plugin in plugins/available/*; do cd "$plugin"; npm install --unsafe-perm; cd /var/app; done
 
+ADD ./bin /var/app/bin
+ADD ./config /var/app/config
+ADD ./docker-compose /var/app/docker-compose
+ADD ./lib /var/app/lib
+ADD ./protocols /var/app/protocols
+ADD ./default.config.js /var/app/default.config.js
+
```
alexandrebouthinon, line 56: I don't get why you change those `ADD` lines. It will result in a breaking change on `kuzzlerc` possible location https://github.com/dominictarr/rc/blob/master/README.md#standards .

Dockerfile
alexandrebouthinon: IMO, you should revert changes in this file

Dockerfile
alexandrebouthinon: IMO, you should revert changes from this file

Aschen: Yes, it's a mistake

Aschen:
LGTM :+1:

PR #1647, opened 2020-06-12
Use latest version of auth plugin for password policies

Aschen:
## What does this PR do ?

Update the auth-local plugin to have password policies

https://github.com/kuzzleio/kuzzle-plugin-auth-passport-local/releases/tag/6.2.0


jenow:
I think there is no need for a pull request in this case ^^'

PR #1648, opened 2020-06-12
Fix cluster race conditions

Aschen:
## What does this PR do ?

_I will wait until the end of the cluster feature I'm developing before merging this one_

This PR fix some race conditions I encounter by using the cluster.

IMHO we should start to run every application always in cluster mode to spot this kind of bug more often.

I also added a docker-compose file who can be used to run a 2 nodes cluster. But there is not round-robin, you have to target each node individually if you want to target different nodes.

```bash
$ docker-compose -f docker-compose/docker-compose-cluster.yml up -d

$ kourou sdk:query server:now --port 7512
# target the other node
$ kourou sdk:query server:now --port 6512
```

### Other changes

 - fix empty `_source` property on `generic:document:afterUpdate`


docker-compose/docker-compose-cluster.yml
```diff
@@ -0,0 +1,96 @@
+version: '3.4'
```
alexandrebouthinon, line 1: I don't get the point of this file, you could achieve the same enabling cluster plugin and using docker-compose `scale` capacity (https://github.com/kuzzleio/kuzzle-plugin-prometheus/blob/master/demo/docker-compose.yml). To make the Kuzzle service scale to 3 nodes for example use this:
```
docker-compose up --scale kuzzle=3
```

alexandrebouthinon: Docker Compose services provides native round robin between service container(s)

Aschen: Oh I didn't know that in fact I thought that cluster port will conflict if you scale the docker compose.
But the purpose of this file is to be able to target each node independently for testing purpose, which you can easily achieve by changing the destination port when you use Kourou for example

Aschen: The best thing would be to have nginx doing the round Robin on each node and still exposing each node port.
in this way we can have a real cluster endpoint and also the possibility to target each node independently

alexandrebouthinon: With `scale`, you have capabilities to request each node with its docker local IP

alexandrebouthinon: IMO, the project root Docker Compose file is sufficient to tests all the possible scenario. Maybe the only thing to add is a complete documentation in README or CONTRIBUTING to know how to start Kuzzle cluster or Kuzzle with HTTPS for example

Aschen: Yes but then I have to loop into the IP of each Kuzzle node and then type the 12 numbers of the IP to target a node which is not very convenient compare to only change the port which is fixed.
The root docker compose it's sufficient but not convenient IMHO

lib/api/controller/document.js
```diff
@@ -295,27 +292,26 @@ class DocumentController extends NativeController {
...
+
+    if (source) {
+      return updatedDocument;
+    }
+
+    return {
+      _id: updatedDocument._id,
+      _source: content,
+      _version: updatedDocument._version,
+    };
```
Yoann-Abbes, line 314:
```suggestion
   return source
      ? updatedDocument
      : {
      _id: updatedDocument._id,
      _source: content,
      _version: updatedDocument._version,
    };
```

scottinet: The ternary operator is only useful when it doesn't make the code less readable, and this doesn't appear to be the case here.

Yoann-Abbes: well I wont fight for this, the only thing I see is a gain of space, not less or more readable code

test/api/controller/document.test.js
```diff
@@ -398,6 +398,16 @@ describe('DocumentController', () => {
       });
     });
 
+    it('should reject if users give document with "_source" property', async () => {
+      request.input.body.documents = [
+        { _id: 'doc-1', body: {} },
+        { _id: 'doc-2', _source: {} },
+      ];
+
+      should(documentController._mChanges(request, 'mCreate', true))
```
scottinet, line 407: This test always succeeds: missing await or return statement

docker-compose/scripts/run-dev.sh
```diff
@@ -46,10 +46,17 @@ fi
 
 echo "[$(date --rfc-3339 seconds)] - Starting Kuzzle..."
 
+if [ ! -z "$KUZZLE_PLUGINS" ];
```
alexandrebouthinon, line 49: NP - Shellcheck suggestion:
```suggestion
if [ -n "$KUZZLE_PLUGINS" ];
```

docker-compose/docker-compose-cluster.yml
Aschen: About the cluster explanation I prefer to wait the cluster integration into the core to update the CONTRIBUTING and README

lib/api/controller/document.js
scottinet: Gaining space doesn't really matter, as it may or may not have an impact on maintenability.
When 2 syntaxes are strictly equivalent, the most important factor of choice should always be "what's the easiest to read?"

The fact that the indentation of the returned object is off hints that using a ternary here might not be a good choice.

PR #1651, opened 2020-06-17
Stick deleteRole/Profile behavior to the documentation

Yoann-Abbes:
## What does this PR do ?

`deleteRole` and `deleteProfile` had a different behavior than `deleteUser` : they didnt returned the delete object's `id` while it is documented that way

### Boyscout

Asyncify deleteRole and deleteProfile
Adapt unit test

### DIFF
--- a/lib/api/controller/security.js
+++ b/lib/api/controller/security.js
@@ -425,17 +425,20 @@ class SecurityController extends NativeController {
    * @param {Request} request
    * @returns {Promise<Object>}
    */
-  deleteRole(request) {
+  async deleteRole(request) {
     assertHasId(request);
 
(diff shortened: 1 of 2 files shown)


Aschen:
Actually I don't think it's a good idea to do this PR now because @scottinet is doing the loose coupling on the security module so they will be a lot of conflict.
You should at least base your PR on https://github.com/kuzzleio/kuzzle/pull/1630

Yoann-Abbes:
> Actually I don't think it's a good idea to do this PR now because @scottinet is doing the loose coupling on the security module so they will be a lot of conflict.
> You should at least base your PR on #1630

It's been regularized in another pull request, I close this one

PR #1695, opened 2020-06-26
Fix dump generation

Aschen:
## What does this PR do ?

Fix the call to `dumpGenerator.dump` method


test/mocks/kuzzle.mock.js
```diff
@@ -98,7 +98,11 @@ class KuzzleMock extends Kuzzle {
       listSubscriptions: sinon.stub().resolves(foo),
     };
 
-    this.dump = sinon.stub().resolves();
+    // this.dump = sinon.stub().resolves()
```
Leodau, line 101:
```suggestion
```

PR #1698, opened 2020-06-29
Fix generic events on search queries

scottinet:
# Description

Generic events on the following API routes were handled by a default document extractor, providing incorrect results to plugin developers: `document:search`, `document:deleteByQuery`, `document:updateByQuery`

# Changes

* Remove the `default` document extractor. Instead, if Kuzzle is tasked to extract documents from an unknown request action, it throws an `InternalError` with this new error code: `core.fatal.assertion_failed`. This new error is meant to be used for runtime assertions, making sure that the code performs according to specifications. Here, a `default` document extractor is clearly not what we want, as it hides potential errors or omissions (and it did exactly that here)
* Add the missing document extractors for `document:search`, `document:deleteByQuery`, `document:updateByQuery`
* Blacklist `document:updateByQuery` from "before" document events, as documents cannot be extracted from requests made to that API route
* Update the documentation accordingly
* Reorganize the document extractor unit tests, for better maintenability
* Add a lot of missing unit tests


lib/api/documentExtractor.js
```diff
@@ -154,7 +157,27 @@ const extractors = ([
...
+          request.result.hits = documents;
+        }
+        else if (request.input.action === 'deleteByQuery') {
+          request.result.documents = documents;
+        }
+
+        return request;
+      },
+    },
+    targets: ['search', 'deleteByQuery'],
```
Aschen, line 180: I thought `updateByQuery` was not eligible either?

scottinet: `updateByQuery` is handled by the same extractor than `mCreate`, `mReplace`, etc. because they share the same format.

The blacklist is done in the `documentEventAliases.js` file: https://github.com/kuzzleio/kuzzle/pull/1698/files#diff-cda9b5cb3d677fb4f72bd365d3318910

So the `extractFromRequest` and `insertInRequest` extractor methods for `updateByQuery` cannot be invoked.

PR #1700, opened 2020-07-03
Fix: handling of Elasticsearch aliases

scottinet:
# Description

Kuzzle should be able to handle index aliases using its formalism and pointing to another place.

Example/How to reproduce the bug:
* create an index named `foobar`
* alias it to `&foo.bar` 

Expected result: Kuzzle should correctly map the alias to `{ index: 'foo', collection: 'bar'}` and use it as any other non-alias indexes

Without this patch, Kuzzle incorrectly maps the alias to: `{ index: 'foo', collection: '&foo.bar'}`

# Changes

The `listAliases` method in the `elasticsearch` service now returns the following array of objects: 
```js
[ 
  {name: '<origin ES index>', index: '<inferred index name>', collection: '<inferred index collection>'},
  {name: '<origin ES index>', index: '<inferred index name>', collection: '<inferred index collection>'},
  // ...
]
```

The `name` property is not currently used, but I found it interesting to keep it for later uses.

Kuzzle's index cache bootstrap now correctly uses that `listAliases` method to get non-listed indexes by `cat.indices` and complete its cache with its results.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "kuzzle",
   "author": "The Kuzzle Team <support@kuzzle.io>",
-  "version": "2.3.0",
+  "version": "2.3.1",
   "description": "Kuzzle is an open-source solution that handles all the data management through a secured API, with a large choice of protocols.",
   "main": "./lib/kuzzle",
   "bin": {
--- a/lib/core/storage/storageEngine.js
+++ b/lib/core/storage/storageEngine.js
@@ -21,11 +21,12 @@
 
 'use strict';
 
+const Bluebird = require('bluebird');
+
 const Elasticsearch = require('../../service/storage/elasticsearch');
 const kerror = require('../../kerror').wrap('services', 'storage');
 const ClientAdapter = require('./clientAdapter');
 const BaseModel = require('../../model/storage/baseModel');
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "kuzzle",
   "author": "The Kuzzle Team <support@kuzzle.io>",
-  "version": "2.3.0",
+  "version": "2.3.1",
```
alexandrebouthinon, line 4: You should maybe update your local master branch :slightly_smiling_face: 
```suggestion
  "version": "2.3.2",
```

alexandrebouthinon: And generate again the package-lock.json

alexandrebouthinon: Never mind we choose what fix we want to release first

scottinet: dang

scottinet: @alexandrebouthinon > err... are you sure?
https://github.com/kuzzleio/kuzzle/blob/master/package.json

Aschen: Talking about this on Slack https://kaliop.slack.com/archives/G4LN18NLQ/p1593763809492900?thread_ts=1593762619.490900&cid=G4LN18NLQ

PR #1701, opened 2020-07-03
Fix: description should not be required for plugin custom errors

### LINKED ISSUE #1603: Either make custom errors description field optional or fix the documentation
# Expected Behavior

The configuration sample given on https://docs.kuzzle.io/core/2/plugins/plugin-context/errors-manager/ should  work out of the box (BTW, the code in the sample at the bottom of this page seems incorrect too).

## Current Behavior

```
PluginImplementationError: [kuzzle-plugin-auth-passport-local] errors field in manifest.json badly formatted: Error configuration file : Field 'description' is required and must be a string (domain: 4, subdomain: 0, error: expired-password). 
```

## Possible Solution

I would personally remove all description fields from the API errors as this information does not seem to be used anywhere and, if it is, does not provide much added-value compared to the error message.

Alternatively, the documentation should be fixed.

scottinet:
# Description

Fix #1603:

The error loader now makes the difference between native and plugin errors, and applies fewer checks to the latter.

# Changes

* add unit tests to the errors loader
* enforce non-empty strings, when applicable
* fix a few error messages
* check property types correctly (e.g. does not handle false booleans or nul numbers as non-existing properties)
* does not allow subdomain codes to be equal to 0, except for plugin errors (0 is the default plugin subdomain, if none is set)

# Other changes

* update eslint to v7 to allow comments when disabling rules

### DIFF
--- a/doc/2/plugins/plugin-context/kerror/index.md
+++ b/doc/2/plugins/plugin-context/kerror/index.md
@@ -20,11 +20,12 @@ Example:
       "code": 1,
       "message": "An error occurred: %s",
       "class": "BadRequestError"
-	},
+	  },
     "some_other_error": {
       "code": 2,
       "message": "An other error occurred: %s",
       "class": "ForbiddenError"
+    }
 	}
 }
 ```
--- a/doc/2/plugins/plugin-context/kerror/index.md
+++ b/doc/2/plugins/plugin-context/kerror/index.md
@@ -20,13 +20,13 @@ Example:
       "code": 1,
       "message": "An error occurred: %s",
       "class": "BadRequestError"
-	  },
+    },
     "some_other_error": {
       "code": 2,
       "message": "An other error occurred: %s",
       "class": "ForbiddenError"
     }
-	}
+  }
 }
 ```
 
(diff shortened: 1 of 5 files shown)


doc/2/plugins/plugin-context/kerror/index.md
```diff
@@ -20,11 +20,12 @@ Example:
       "code": 1,
       "message": "An error occurred: %s",
       "class": "BadRequestError"
-	},
+	  },
```
alexandrebouthinon, line 23:
```suggestion
    },
```

alexandrebouthinon: Bad indent, edit directly the file I think my suggestion is wrong too :grin:

doc/2/plugins/plugin-context/kerror/index.md
```diff
@@ -20,11 +20,12 @@ Example:
...
     "some_other_error": {
       "code": 2,
       "message": "An other error occurred: %s",
       "class": "ForbiddenError"
+    }
 	}
```
alexandrebouthinon, line 29: Same here

doc/2/plugins/plugin-context/kerror/index.md
scottinet: I blame tabulations :angry:

PR #1702, opened 2020-07-06
Updates search index only when dynamic changes from false to true

Aschen:
## What does this PR do ?

Force the search index update only if at least one dynamic field value went from `false` to `true`

### DIFF
--- a/lib/service/storage/elasticsearch.js
+++ b/lib/service/storage/elasticsearch.js
@@ -1034,7 +1034,7 @@ class ElasticSearch extends Service {
 
         await this.updateMapping(index, collection, mappings);
 
-        if (JSON.stringify(previousMappings).includes('"dynamic":"false"')) {
+        if (this._dynamicChanges(previousMappings, mappings)) {
           await this.updateSearchIndex(index, collection);
         }
       }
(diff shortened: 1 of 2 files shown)


lib/service/storage/elasticsearch.js
```diff
@@ -2362,10 +2362,49 @@ class ElasticSearch extends Service {
...
+/**
+ * Finds paths and values of mappings dynamic properties
+ */
+function findDynamic (mappings, path = [], results = {}) {
+  if (mappings.dynamic !== undefined) {
+    results[path.concat('dynamic').join('.')] = mappings.dynamic;
+  }
+
+  for (const [key, value] of Object.entries(mappings)) {
+    if (_.isPlainObject(value)) {
```
scottinet, line 2400: (nitpicking) you shoud use our own `isPlainObject` method from `util/safeObject.js`

lib/service/storage/elasticsearch.js
```diff
@@ -2362,10 +2362,49 @@ class ElasticSearch extends Service {
...
+    const previousValues = findDynamic(previousMappings);
+
+    for (const [path, previousValue] of Object.entries(previousValues)) {
+      if (previousValue.toString() !== 'false') {
+        continue;
+      }
+
+      const newValue = _.get(newMappings, path);
+
+      if (newValue && newValue.toString() === 'true') {
```
scottinet, line 2380: `dynamic` can be one of the following values: `false, true, strict`

I think that we should also trigger a reindex if the dynamic value goes from false to strict.

PR #1711, opened 2020-07-09
Prevent timeouts when refreshing large indexes

scottinet:
# Description

When updating a collection mapping, if a `dynamic` property is set to true or strict, and if it was "false" before, then we trigger an index refresh using [updateByQuery](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/api-reference.html#_updatebyquery).

When processing large indexes, this refresh can take some time and exceeds the default request timeout.

Rather than extending the timeout with some arbitrary large value, I've opted to ask ES to process this refresh as a background task: this refresh should be rare, and I don't see the value of waiting for it in our request response.


etrousset:
👍

PR #1720, opened 2020-07-13
Allow to delete a still-used profile by removing it from assigned users

### LINKED ISSUE #1683: As an end user, I'd like to be able to remove a profile and it's associated users
Add an option to update users associated with the profile when deleting a profile

https://trello.com/c/qaIPCFZf/218-nij-allow-to-remove-a-profile-even-if-it-is-used-by-users

Leodau:
## What does this PR do ?
Adds an option to **security:deleteProfile** so it can update associated users (remove the required profile from their _profileIds_ list) when removing a profile.

If a user only had one profile, and it happens to be the one that is getting removed: the _Anonymous_ profile will be assigned.

closes #1683 

- Create a user and give it a custom profile
- Remove the profile
- Check that the user got updated


doc/2/api/controllers/security/delete-profile/index.md
```diff
@@ -42,6 +45,7 @@ Method: DELETE
 ### Optional:
 
 - `refresh`: if set to `wait_for`, Kuzzle will not respond until the profile deletion is indexed (default: `"wait_for"`)
+- `onAssignedUsers`: if set to `remove`, Kuzzle will remove this profile from assigned users. If this was a user's sole profile, then the `Annonymous` profile will be assigned. (default: `"fail"`) <SinceBadge version="auto-version" />
```
scottinet, line 48: Shouldn't we apply the `default` profile instead?

btw:

```suggestion
- `onAssignedUsers`: if set to `remove`, Kuzzle will remove this profile from assigned users. If this was a user's sole profile, then the `anonymous` profile will be assigned. (default: `"fail"`) <SinceBadge version="auto-version" />
```

lib/core/security/profileRepository.js
```diff
@@ -361,10 +365,45 @@ class ProfileRepository extends Repository {
...
+        batch.length = 0;
+
+        for (const user of userPage.hits) {
+          user.profileIds = user.profileIds.filter(e => {
+            return (e !== profile._id);
+          });
+          if (user.profileIds.length === 0) {
+            user.profileIds.push('anonymous');
+          }
+          batch.push(this.module.user.update(
```
scottinet, line 385: Each page contains at most 1000 users. This means we can send up to 1000 concurrent updates, I'm not sure that ES won't complain.

You probably should use smaller pages. Or implement a `mUpdate` method, if you feel like it.
Or perhaps try an [updateByQuery](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update-by-query.html) query using a script to remove the profile value, something like that:

```js
{
  script: {
    source: `ctx._source.profileIds.remove('${profile._id}')`,
    lang: 'painless',
  },
  query,
}
```

Leodau: Gonna try the **updateByQuery** method, makes way more sense.

doc/2/api/controllers/security/delete-profile/index.md
Leodau: In some situations it could be better to have the `default` instead ? @Aschen

Aschen: I'm not comfortable with the `default` profile since this profile could have more right than the previous user profile.  
At least with the anonymous user we are 100% sure that we are not gonna have an unwanted privilege escalation

lib/core/security/profileRepository.js
Leodau: I'll bring the page size to 100,
Liked the updateByQuery idea, will do with a mUpdate feature in the future.

PR #1727, opened 2020-07-17
Loosely couple the realtime core module

scottinet:
# Description

Promote realtime methods to an independant and loosely coupled core module (hotel clerk, notifier).

## Changes

* `core:realtime:list` (previously `hotelClerk.listSubscriptions()`):
  * remove the "sorted" option, which was supposed to return index/collection items in a sorted manner: the option was unused, and that's cool. Because it cannot work: Kuzzle responses are sent as JSON objects to clients. And object keys in JSON objects are unordered (see [JSON specs](https://www.json.org)).
  * the `realtime:list` API action now returns the list of subscriptions on all the index/collection pairs the requesting user is allowed to **subscribe to** (API action: `realtime:subscribe`). Before this, the `document:search` API action was used as the reference to check those rights
* remove dead code
* simplify hotel clerk method names
* moved notification classes from `lib/core/model/notification/` to `lib/core/realtime/notification/`
* this is no longer the responsibility of the callers to decide if notifications might have cached keys or not. This is now handled solely by the notifier core module (removal of the `cached` arguments in document notifying events)
* this is now the responsibility of the callers to tell the notifier module if the notifications are about document creations, updates, replacements, and so on. To do that, a new enum has been added, with the following available values: CREATE, DELETE, REPLACE, UPDATE, WRITE (= a mix of document creations and replacements)
* the `_updatedFields` property (list of updated document keys, sent in update notifications) was only sent for `document:update` calls. It is now also sent for `document:updateByQuery`. We still need to add that prop for `document:mUpdate`, and that will be done in another PR (we need to modify the ES service to be able to add this feature)
* greatly simplified the notifier core module. Now all the cache logic is centralized in `notifyDocuments`, and the `notifyDocument[Create|Delete|Replace|Update]` methods only handle the creation and the sending of the right notification. Also, dead code was removed.

# Events

New `ask` events:

| name | Description |
|-------|-----------------|
| `core:realtime:join` | Makes a user join an existing room |
| `core:realtime:list` | Returns the list of index, collection, rooms and user counts per room a user is allowed to see |
| `core:realtime:collections:get` | Given an index, returns the list of collections having subscriptions on them |
| `core:realtime:disconnect` | Removes a user and all their associated subscriptions |
| `core:realtime:publish` | Publishes a realtime message |
| `core:realtime:subscribe` | Subscribes a user to a room |
| `core:realtime:unsubscribe` | Unsubscribes a user from a room |
| `core:realtime:document:dispatch` | Sends a document notification to a precomputed list of rooms (low-level method, used by the cluster) |
| `core:realtime:user:sendMessage` | Sends a user notification to the given room (low-level method, used by the cluster) |
| `core:realtime:document:notify` | Sends a single document notification |
| `core:realtime:document:mNotify` | Sends multiple document notifications in one go |
| `core:realtime:tokenExpired:notify` | Sends a "token expired" server notification |
| `core:realtime:room:create` | Create an empty room using the provided index, collection and room ID (used by the cluster) |
| `core:realtime:room:size:get` | Returns the number of connections listening to a given room |

The following pipe events have been renamed: 

| Old name | New name |
|-------|-----------------|
| `core:hotelClerk:addSubscription` | `core:realtime:user:subscribe:after` |
| `core:hotelClerk:removeRoomForCustomer` | `core:realtime:user:unsubscribe:after` |
| `core:hotelClerk:join` | `core:realtime:user:join:after` |
| `room:remove` | `core:realtime:room:remove:before` |
| `room:new` | `core:realtime:room:create:after` |


scottinet:
Switching this PR to draft: I need to add more events to allow the cluster plugin to interact with the realtime module.

scottinet:
It should be ok now.
@Leodau > you might want to review my latest commit. This should be quick, I only added a couple events, renamed a few existing ones, and moved files around.

lib/api/controller/document.js
```diff
@@ -221,40 +225,35 @@ class DocumentController extends NativeController {
...
-            modifiedRequest,
-            response);
-        }
-
-        return this.kuzzle.notifier.notifyDocumentReplace(modifiedRequest);
-      })
-      .then(() => response);
+    const modifiedRequest = await this.kuzzle.validation.validate(request, false);
+
+    // response: { _id, _version, _source, created }
```
alexandrebouthinon, line 237: I don't get the purpose of this comment

lib/api/controller/realtime.js
```diff
@@ -50,10 +45,12 @@ class RealtimeController extends NativeController {
...
   async subscribe (request) {
-    assertHasIndexAndCollection(request);
-    assertHasBody(request);
+    // assertion checks
+    this.getIndexAndCollection(request);
+    this.getBody(request);
 
-    // Only funnel protocol (and thus the embedded SDK) can use "propagate" option
+    // Only funnel protocol (and thus the embedded SDK) can use "propagate"
+    // option
```
alexandrebouthinon, line 53:
```suggestion
    // Only funnel protocol (and thus the embedded SDK) can use "propagate" option
```

scottinet: This line exceeds 80 chars.

lib/api/controller/document.js
scottinet: That's an old one. Probably a reminder of what's returned by the ES request :thinking: 
Should I remove it?

alexandrebouthinon: I don't think so but current format is not very useful IMHO 🙂

lib/api/controller/realtime.js
alexandrebouthinon: I didn't see ESLint rule about line length. I think we can do this on a case-by-case basis (at least as far as comments are concerned).

scottinet: Even if there was wrapping rules in eslint, I'm not sure we would want to enforce this, because our set of rules contain exceptions (such as message strings, or when wrapping would be more harmful to the code lisibility).

Here are our own wrapping rules: https://github.com/kuzzleio/meeting-notes/blob/master/team/code%20wrap%20rules.md

lib/core/realtime/hotelClerk.js
```diff
@@ -109,20 +114,104 @@ class HotelClerk {
...
+    this.kuzzle.onAsk(
+      'core:realtime:listCollections',
+      index => this.listCollections(index));
+
+    /**
+     * Removes a user and all their subscriptions.
+     * @param  {RequestContext} requestContext
+     */
+    this.kuzzle.onAsk(
+      'core:realtime:disconnect',
```
xbill82, line 177: I'd rather call this `core:realtime:unsubscribeAll`, since `disconnect` would imply its connection is closed, which is not.

scottinet: resolved in #1735: this event has been renamed `core:realtime:user:remove`

lib/core/realtime/hotelClerk.js
```diff
@@ -619,30 +569,46 @@ class HotelClerk {
           }, {});
       }
 
-      await this.kuzzle.pipe(
-        'core:hotelClerk:removeRoomForCustomer',
-        { requestContext, room: roomPayload });
-    }
+      await this.kuzzle.pipe('core:realtime:user:unsubscribe:after', {
+        requestContext,
```
Aschen, line 573: Why sending the whole `requestContext` if we need only the `connectionId` ? (that's the only information inside)

lib/core/realtime/hotelClerk.js
```diff
@@ -424,7 +455,18 @@ class HotelClerk {
     const response = this.kuzzle.koncorde.store(normalized);
     const roomId = response.id;
 
-    this.kuzzle.emit('room:new', {collection, index, roomId: response.id});
+    this.kuzzle.emit('core:realtime:room:create:after', {
+      collection,
+      index,
+      roomId: response.id,
```
Aschen, line 461:
```suggestion
      roomId,
```

lib/core/realtime/hotelClerk.js
```diff
@@ -424,7 +455,18 @@ class HotelClerk {
...
+    });
+
+    // @deprecated -- to be removed in the next major version of kuzzle
+    this.kuzzle.emit('room:new', {
+      collection,
+      index,
+      roomId: response.id,
+    });
```
Aschen, line 469:
```suggestion
    this.kuzzle.emit('room:new', { collection, index, roomId });
```

lib/core/realtime/notifier.js
```diff
@@ -99,276 +173,206 @@ class NotifierController {
...
-    if (rooms.length > 0) {
-      for (const room of rooms) {
-        const hotelClerkRoom = this.kuzzle.hotelClerk.rooms.get(room);
+    for (const room of rooms) {
+      const hotelClerkRoom = this.module.hotelClerk.rooms.get(room);
 
-        if (hotelClerkRoom !== undefined) {
-          channels.push(...Object.keys(hotelClerkRoom.channels));
-        }
+      if (hotelClerkRoom !== undefined) {
```
Aschen, line 199: Nitpicking
```suggestion
      if (hotelClerkRoom) {
```

lib/core/realtime/hotelClerk.js
scottinet: Honestly? I'm not sure why. I had some logic behind it, and I was about to explain it here, but then it occured to me that it just doesn't hold water. Especially considering other events I wrote that take a connectionId, such as `core:realtime:user:remove` or `core:realtime:unsubscribe` :thinking: 

I'll change it :+1:

lib/core/realtime/notifier.js
scottinet: Since we get a value from a map, I think it's perfectly valid to check against `undefined` to test whether the value exists.
I don't think I'll apply this nitpicking, unless you've a strong argument in its favor.

Aschen: It's just because it's more concise and we are not suppose to store `false` or `null` values inside this map but it was really a small nitpicking and also I agree that this is the actual pattern when checking a value get from a map :+1:

alexandrebouthinon (approved):

Aschen (approved):

PR #1729, opened 2020-07-21
Register subscription

### LINKED ISSUE #1667: As a plugin developer, I'd like to use (un)subscribe in a controller action
Allow plugin developers to create controller actions that subscribe the user to a realtime subscription.

We could pass the Request Context to the context.sdk.as method, so the hotel clerk will be able to register a new subscription for the connection.

The exposed subscribe method should be different from the original one because it won't accept a callback anymore.
Actually the subscribe method should not use the original SDK method but rather make a direct call to Kuzzle API.

The context.sdk.as() method should accept both User and Context object to not introduce breaking change.

{{}}

context.sdk.as(request.context).subscribe('index', 'collection', filters);
context.sdk.as(request.context).unsubscribe('index', 'collection', filters);

context.sdk.subscribe('index', 'collection', filters); // throw error

ttps://trello.com/c/VNtKDy5q/169-seb-autoriser-les-plugins-%C3%A0-faire-sdkasuserrealtimeunsubscribe

xbill82:
## What does this PR do ?
Closes  #1667 by introducing the `registerSubscription`and `unregisterSubscription` methods in the Plugin Context.

Write a plugin that exposes a custom action. In this action, use `registerSubscription` to subscribe to a given collection with a given set of filters. Then, connect to the backend via `wscat` and trigger the custom action. From now on, the `wscat` session should receive notifications for the subscription defined by the call to `registerSubscription`.

### TODO
- [ ] Documentation
- [ ] `unregisterSubscription` method
- [ ] put the methods into a more semantic namespace in the Plugin Context (like `realtime`)
- [ ] tests

### Other changes

--

### Boyscout

--


doc/2/api/controllers/document/delete-by-query/index.md
```diff
@@ -84,7 +84,7 @@ Body:
 
 Returns an object containing information about the deleted documents:
 
-- `ids`: an array containing the list of each deleted document identifier <DeprecatedBadge version="2.2.1"/>.
+- `ids`: an array containing the list of each deleted document identifier.
```
Leodau, line 87: Do this suddenly got un-deprecated ?

xbill82: ouch?

Aschen:
Regarding the naming I would prefer something like `susbcription.register` instead of `realtime.registerSubscription`.
IMHO it's less verbose and more precise.
Finally we could imagine other methods in this `subscription` module, to list all current subscriptions for example

PR #1732, opened 2020-07-23
Add support for security:getUserStrategies

Leodau:
## What does this PR do ?
Implements a feature that retrieves all available authentication **strategies** for a user.

- [x] Functional tests
- [x] Unit tests
- [x] Documentation


scottinet (approved):

rolljee (approved):

doc/2/api/controllers/security/get-user-strategies/index.md
```diff
@@ -0,0 +1,58 @@
...
+
+- `_id`: user [kuid](/core/2/guides/kuzzle-depth/authentication#the-kuzzle-user-identifier)
+
+---
+
+## Response
+
+Returns an object with the following properties:
+
+- `hits`: array containing all the available authentication strategies on the requested user.
```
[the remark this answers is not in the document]
Leodau: Oh yep makes sense, I followed what I thought was a convention in our responses (hits), we could simply name it `strategies` indeed.

lib/api/controller/security.js
```diff
@@ -611,6 +612,45 @@ class SecurityController extends NativeController {
...
+      for (const strategy of availableStrategies) {
+        const existStrategy = this.getStrategyMethod(strategy, 'exists');
+
+        promises.push(
+          existStrategy(request, id, strategy)
+            .then(exists => exists ? strategy : null));
+      }
+    }
+
+    return Bluebird.all(promises)
```
Aschen, line 643: Prefer `async/await`

Aschen (changes requested):

lib/api/controller/security.js
```diff
@@ -619,36 +619,28 @@ class SecurityController extends NativeController {
...
-    return Bluebird.all(promises)
-      .then(strategies => {
-        const userStrategies = strategies.filter(item => item !== null);
-
-        return {
-          strategies: userStrategies,
-          total: userStrategies.length
-        };
-      });
+    return {
```
Aschen, line 640: I meant to use `await` here only instead of `.then`

lib/api/controller/security.js
```diff
@@ -619,36 +619,28 @@ class SecurityController extends NativeController {
...
       const availableStrategies = global.kuzzle.pluginsManager.listStrategies();
 
       for (const strategy of availableStrategies) {
-        const existStrategy = this.getStrategyMethod(strategy, 'exists');
+        const existMethod = this.getStrategyMethod(strategy, 'exists');
 
-        promises.push(
-          existStrategy(request, id, strategy)
-            .then(exists => exists ? strategy : null));
+        if (await existMethod(request, userId, strategy)) {
```
Aschen, line 634: Here you first code was better because it was a parallel processing and now it's sequential.
`async/await` should be used unless you can/need parallel processing

Aschen (changes requested):

lib/api/controller/security.js
Leodau: Ofc.

Aschen (approved):

PR #1735, opened 2020-07-29
overhauling the event overhaul -_-

scottinet:
# Description

So... after discussing with the team about a convention to name events, I decided to overhaul the new system that has been developed in #1727 

This PR is an addendum to the already validated other PR.

# Changes

Here we go:

* this is no longer the responsibility of the callers to decide if notifications might have cached keys or not. This is now handled solely by the notifier core module (removal of the `cached` arguments in document notifying events)
* this is now the responsibility of the callers to tell the notifier module if the notifications are about document creations, updates, replacements, and so on. To do that, a new enum has been added, with the following available values: CREATE, DELETE, REPLACE, UPDATE, WRITE (= a mix of document creations and replacements)
* the `_updatedFields` property (list of updated document keys, sent in update notifications) was only sent for `document:update` calls. It is now also sent for `document:updateByQuery`. We still need to add that prop for `document:mUpdate`, and that will be done in another PR (we need to modify the ES service to be able to add this feature)
* greatly simplified the notifier core module. Now all the cache logic is centralized in `notifyDocuments`, and the `notifyDocument[Create|Delete|Replace|Update]` methods only handle the creation and the sending of the right notification. Also, dead code was removed.

## Events

This is the exhaustive list of realtime events, and will be used as a reference in the original PR.

New `ask` events:

| name | Description |
|-------|-----------------|
| `core:realtime:join` | Makes a user join an existing room |
| `core:realtime:list` | Returns the list of index, collection, rooms and user counts per room a user is allowed to see |
| `core:realtime:collections:get` | Given an index, returns the list of collections having subscriptions on them |
| `core:realtime:disconnect` | Removes a user and all their associated subscriptions |
| `core:realtime:publish` | Publishes a realtime message |
| `core:realtime:subscribe` | Subscribes a user to a room |
| `core:realtime:unsubscribe` | Unsubscribes a user from a room |
| `core:realtime:document:dispatch` | Sends a document notification to a precomputed list of rooms (low-level method, used by the cluster) |
| `core:realtime:user:sendMessage` | Sends a user notification to the given room (low-level method, used by the cluster) |
| `core:realtime:document:notify` | Sends a single document notification |
| `core:realtime:document:mNotify` | Sends multiple document notifications in one go |
| `core:realtime:tokenExpired:notify` | Sends a "token expired" server notification |
| `core:realtime:room:create` | Create an empty room using the provided index, collection and room ID (used by the cluster) |
| `core:realtime:room:size:get` | Returns the number of connections listening to a given room |

The following pipe events have been renamed: 

| Old name | New name |
|-------|-----------------|
| `core:hotelClerk:addSubscription` | `core:realtime:user:subscribe:after` |
| `core:hotelClerk:removeRoomForCustomer` | `core:realtime:user:unsubscribe:after` |
| `core:hotelClerk:join` | `core:realtime:user:join:after` |
| `room:remove` | `core:realtime:room:remove:before` |
| `room:new` | `core:realtime:room:create:after` |

# Other changes

A few of the renamed events were documented. That's unfortunate: although I'm pretty sure noone use them, this forces me to duplicate event emissions and to document the new ones.

Also, to prevent cluttering, I regrouped all realtime-related events in a single "Realtime Events" documentation, making it easier for users to fetch the right event for their needs.


scottinet:
There is one, in the cluster plugin. It starts Kuzzle's functional tests in a cluster context, so it's pretty thorough.

I created a PR to upgrade the cluster plugin to support the changes made by the event overhaul (https://github.com/kuzzleio/kuzzle-plugin-cluster/pull/108) and I checked those against that branch.

It's absolutely not ideal, but it does the job until the cluster plugin is killed in favor of a cluster module directly inside kuzzle.

doc/2/plugins/guides/events/realtime/index.md
```diff
@@ -0,0 +1,257 @@
...
+
+| Properties     | Type                 | Description             |
+| -------------- | -------------------- | ----------------------- |
+| `id`           | <pre>string</pre>    | Room unique identifier  |
+| `index`        | <pre>string</pre>    | Index                   |
+| `collection`   | <pre>string</pre>    | Collection              |
+
+---
+
+## notify:dispatch
```
Aschen, line 164: According to the new naming, this event and the following shouldn't be `notify:dispatch:before` ?

scottinet: That would be a breaking change, unfortunately. 
Same goes for all our API events named `controller:[before|after]Action`, which should really be named: `api:controller:action:[before|action]`

That'll be for a next major version.

lib/api/controller/document.js
```diff
@@ -523,11 +541,12 @@ class DocumentController extends NativeController {
       documents,
       { refresh, userId });
 
+    // @todo inject changes for mUpdate requests
```
Aschen, line 544: I don't understand, what changes are you talking about?

lib/core/realtime/notifier.js
```diff
@@ -254,209 +221,158 @@ class NotifierController {
   publish (request) {
     const rooms = this._test(request);
 
-    return this.notifyDocument(rooms, request, 'in', 'publish', {
-      _id: request.input.resource._id,
-      _source: request.input.body
-    });
+    if (rooms.length > 0) {
+      return this.notifyDocument(rooms, request, 'in', 'publish', {
```
Aschen, line 225: Nitpicking: use a guard instead of indenting the business code

lib/api/controller/document.js
scottinet: Detailed in the PR description:

> the _updatedFields property (list of updated document keys, sent in update notifications) was only sent for document:update calls. It is now also sent for document:updateByQuery. We still need to add that prop for document:mUpdate, and that will be done in another PR (we need to modify the ES service to be able to add this feature)

scottinet: I've updated the comment to make it clearer.

xbill82:
I relay [my remark](https://github.com/kuzzleio/kuzzle/pull/1727/files#r467955496) about `core:realtime:disconnect` here. To me, the naming of this event is not precise.

scottinet:
@xbill82 > I agree that "disconnect" is not a good verb for that event... but I do not like "unsubscribeAll", and I don't have a better proposal. :thinking: 

Since it's really about cleaning up the hotel clerk when a user disconnects, I wonder if, instead, it wouldn't be better to remove that event altogether, and to listen to pipe events triggered by the network layer and when the token expires. :thinking: 

I'm not sure that's a good idea though... I'll think about it. If I don't find anything better, I'll apply your suggestion to that event name.

xbill82:
@scottinet I agree that `unsubscribeAll` is _pretti ugli_... But rather meaningful indeed. When I'm in these situations, I really try to formulate a phrase telling about the contract of a block of code and see what names come out of it. In this case it's: "Delete all of a users' subscriptions (because the dude is about to be disconnected, probably)", which gives

* unsubscribeUser
* deleteUserSubscriptions
* leaveUser
* kickUser
* forgetUser

Ok, some of them are rather... "creative"! 😅 

My 2 cents 💰

scottinet:
@xbill82 > what about `core:realtime:user:remove`?

PR #1736, opened 2020-07-29
Add index and collection property inside documents

ChimiDEV:
## What does this PR do ?

Hello,
Working with Kuzzle motivated me to check out some issues with the goal in mind to contribute something to the project.
I saw the  open issue, that specified to add informations about the index and collection into the documents.

So I thought I might give it a go. Since this is my first contribution to Kuzzle, I'm looking forward to any feedback you have.
Hopefully I did understand the issue correct. Simply put:
This PR will add an `index` and a `collection` property to the `_kuzzle_info` object stored inside documents.

I tried to update all docs and unit tests. I did not change `_kuzzle_info` of other features (roles, subscription), as I was not sure, if this would be wanted as well.

  - Step 1 : Create an index and a collection
  - Step 2 : Create, read and update a document


Aschen:
Hello and thanks for this contribution, it's very much appreciated!
About this feature we should have been more specific on the issue because if I remember well we talked about adding those properties at the same level as `_id` and `_source`.
But putting them in Kuzzle metadata seems to be a good idea also, for example it will allows to do search or sort result on those properties when the search on multi index feature will be implemented.

Also we talked about having a kind of urn property that uniquely identify a document with index, collection and _id and this property should be at the root level. Having index and collection at the root level in addition with the urn would have feel like a duplication.
I'm gonna wait the other team member opinions about this

scottinet:
Hi,

Thanks A LOT for your contribution, this is highly appreciated.

But...

I'm not fond of storing data that can be easily retrieved otherwise, as it might weight a lot on storage consumption: even if optimized (ES won't _really_ duplicate the strings, but a string id will still be stored), we're talking about adding redundant information in millions and more documents.

About the future multi-index search features, ES supports that natively. Kuzzle forbids it because of rights management: until now, we only enforce rights using a single index/collection pair, and we'd need to (carefully) amend our API to allow multiple index/collection to be safely queried.

And this brings me to API responses. When querying a single index/collection pair, I don't see the added value of repeating an information we already have at the root of the request response in every documents returned.
My opinion is mitigated by the following considerations:
* I think there is value in having a URN field for every document returned, but this will only be useful if our API is able to use that URN fluently)
* In the future, when we'll support them (IF we support them, there are performances impacts to be considered), queries impacting multiple index/collection pairs in one go won't have "index" and "collection" fields in the response obviously, and in that case, we'll have to put that information in all documents returned. But this'll have to be done dynamically when building the response: we shouldn't need to rely on additional stored information

Aschen:
@scottinet You forget the use case involving aliases. If you have an alias to many ES indexes then it could be useful to know in which index / collection belongs every returned documents.

Apart from those 2 use cases (multi-index search and aliases), I agree that it's not necessary to returns index and collection name

scottinet:
About injecting the original index name when a document comes from an alias:

pros
* when retrieving documents from a rotating index, this information could be useful

cons
* I wonder if that's an information we want to leak to the outside, as it provides inner details about the storage
* aliases can point to indexes that do not follow Kuzzle's naming format, so that information could be unusable without access to ES... and for anyone with direct access to ES, that information is readily available

I might miss something, but right now I can't find a use case where this information would be useful (from a client's point of view, admins are another matter). Do you have one?

Aschen:
I don't see the problem about leaking Elasticsearch indexes names, I don't see how it could be used for a malicious usage.

You're right about the 2nd point, so maybe we should have an `esIndex` property instead of having `index` and `collection` for aliases.

If I remember well Benjamin from ffly4u had an example where multiple indexes where referenced by one alias and he wanted to know from which one the results came from.

Maybe we could try to infer Kuzzle index and collection from index name and also add the original elasticsearch index name when using `document:search` on an index alias

scottinet:
I would only put the `_index` property in search hits, unmodified, and only if it differs from the requested index.

Aschen:
I'm closing this PR since the feature description was unclear. Sorry for that.

PR #1745, opened 2020-08-06
Ensure correct values for mappings dynamic properties

### LINKED ISSUE #1743: Cannot create a mapping with dynamic set to false
Observed on 1.x but looking at the code, most likely still is the case on v2.

cf https://github.com/kuzzleio/kuzzle/blob/master/lib/service/storage/elasticsearch.js#L944
As long as the common mapping is set to allow dynamic mapping, users cannot create a non-dynamic one.

Aschen:
##  What does this PR do ?

The mappings dynamic property accept 3 values `true`, `false` and `strict`. The value should be a string.

This PR checks the dynamic properties values and throw an error if it's not a string. 
Boolean are also accepted to correct common mistakes.
An error is also thrown if the value if different than one of the 3 accepted values. 

fix #1743 

### Other changes

 - throw an error when trying to trigger an event before app start


Aschen:
> The PR description states that boolean values are accepted to fix common mistakes, but I don't see that in the code.
> 
> Am I missing something?

The actual code is here: https://github.com/kuzzleio/kuzzle/pull/1745/files#diff-0a506d167f9f96c040374af2280e4b55R2409

scottinet:
Ah, yes, it was added a few minutes after my review, you were probably already on it if I read the underlying issue correctly. :+1:

lib/service/storage/elasticsearch.js
```diff
@@ -2388,12 +2398,46 @@ class ElasticSearch extends Service {
...
+    for (const [path, value] of Object.entries(dynamicProperties)) {
+      // Prevent common mistake
+      if (typeof value === 'boolean') {
+        _.set(mappings, path, value.toString());
+      }
+      else if (typeof value !== 'string') {
+        throw kerror.get(
+          'invalid_mapping',
+          path,
+          'Dynamic property value should be a string.');
```
rolljee, line 2416:
```suggestion
          'Dynamic property value must be a string.');
```

scottinet: well... not exactly since booleans are also accepter.

Perhaps this message would be better: `'The value of the "dynamic" property must be either a string or a boolean (accepted values: "true", "false", "strict")'`

Aschen: Actually even if we accept boolean as a common mistake, users should use the string version otherwise sometime they will send boolean for `true` and `false` and sometimes string for `strict`.  
So if they send a boolean it will be silently converted but any other value with result in an error

scottinet: Personally, I like unsurprising and exhaustive APIs. So if we accept boolean values, we should state so. If we don't, then we should reject them.

IMO, as a user, if I submit a boolean by mistake and Kuzzle doesn't complain, I'll go check the documentation. If I don't find anything about booleans being allowed, I'd be very confused... and think that the API is probably wonky.

Aschen:
> I'm not comfortable accepting a Boolean for this property. Given that there can be three possible values, I find it doesn't make much sense. What do you think about it?

People use boolean for true and false because it's more logical than string so we have to accept both to prevent frustration.

PR #1750, opened 2020-08-10
Use latest version of kuzzle-common-objects for Typescript

Aschen:
## What does this PR do ?

Use the latest (5.0.0) version of `kuzzle-common-objects` to use the `Request` type in controller definition


lib/util/interfaces.ts
```diff
@@ -30,7 +31,7 @@ export interface JSONObject {
 export interface ControllerDefinition {
   actions: {
     [action: string]: {
-      handler: (request: any) => Promise<any>,
+      handler: (request: Request) => Promise<any>,
```
Aschen, line 34: Actual change

PR #1751, opened 2020-08-11
Add a "remaining" property to scroll search results

scottinet:
# Description

Add a new `remaining` property to scroll search results, showing the number of remaining documents that can still be fetched.

This should make easier for our API clients to know when they have fetched the last page of scroll searches without having to track  it using the number of documents already retrieved, and comparing that to the total number of hits.

# Other changes

* fix a couple unit tests about `document:updateByQuery` that were erroneously successful

### DIFF
--- a/doc/2/api/controllers/document/scroll/index.md
+++ b/doc/2/api/controllers/document/scroll/index.md
@@ -60,6 +60,7 @@ Returns a paginated search result set, with the following properties:
   - `_id`: document unique identifier
   - `_score`: [relevance score](https://www.elastic.co/guide/en/elasticsearch/guide/current/relevance-intro.html)
   - `_source`: new document content
+- `remaining`: remaining documents that can be fetched
 - `scrollId`: identifier to the next page of result. Can be different than the previous one(s)
 - `total`: total number of found documents. Usually greater than the number of documents in a result page
 
--- a/doc/2/api/controllers/document/scroll/index.md
+++ b/doc/2/api/controllers/document/scroll/index.md
@@ -60,11 +60,9 @@ Returns a paginated search result set, with the following properties:
   - `_id`: document unique identifier
   - `_score`: [relevance score](https://www.elastic.co/guide/en/elasticsearch/guide/current/relevance-intro.html)
   - `_source`: new document content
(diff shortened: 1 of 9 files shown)


features-sdk/run.sh
```diff
@@ -8,5 +8,5 @@ export SCOPE=${1:-features-sdk}
 
 for protocol in websocket http; do
   # profiles are defined in the cucumber.js file at the root of this project
-  KUZZLE_PROTOCOL=$protocol ./node_modules/.bin/cucumber-js --profile $protocol $SCOPE
+  KUZZLE_PROTOCOL=$protocol npx cucumber-js --profile $protocol $SCOPE
```
alexandrebouthinon, line 11: Why switching to `npx` now?

Leodau, line 11: :+1: :100: Yessss

scottinet: All it does is to fetch module binaries in the right place, this prevents us to type the exact path by ourselves. I find this a bit cleaner.

alexandrebouthinon: Ok great 👍

doc/2/api/controllers/document/scroll/index.md
```diff
@@ -60,10 +60,12 @@ Returns a paginated search result set, with the following properties:
   - `_id`: document unique identifier
   - `_score`: [relevance score](https://www.elastic.co/guide/en/elasticsearch/guide/current/relevance-intro.html)
   - `_source`: new document content
-- `remaining`: remaining documents that can be fetched
 - `scrollId`: identifier to the next page of result. Can be different than the previous one(s)
 - `total`: total number of found documents. Usually greater than the number of documents in a result page
 
+<SinceBadge version="auto-version"/>
+- `remaining`: remaining documents that can be fetched
```
Aschen, line 67: It's better to put the badge on the same line
```suggestion
- `remaining`: remaining documents that can be fetched <SinceBadge version="auto-version"/>
```
image

scottinet: Ah, it works if put at the end of the line, interesting... :+1: 

This looks a tad better than having a separate list, I agree.

PR #1754, opened 2020-08-12
Make the dev image compatible with kaaf functional tests

scottinet:
# Description

This is a dev experience PR: it allows kuzzle devs to locally run all functional tests, including those involving the kaaf functional test plugin.

# Changes

* Kuzzle's starter script no longer waits for Elasticsearch. This task is now performed by Kuzzle's Elasticsearch service itself, upon instantiating a new ES client. It's done in such a way that only 1 ES wait will ever be performed whatever the number of ES clients built. This change is necessary since developper images now don't use the starter script
* Remove the `wait-elasticsearch` script
* The default timeout for services init has been increased to 2min, to give time to ES to start

# Boyscout

* fix lint errors and simplify the `wait-kuzzle` script


scottinet:
/ping @Leodau, I think you'll like this one.

PR #1758, opened 2020-08-14
Fix manifest loading for cluster

Aschen:
## What does this PR do ?

Fix plugin cluster loading with the `KUZZLE_PLUGINS` env variable

### Other changes

 - add error handler on `httpServer` to exit startup sequence if the port is already used for example
 - allows number in plugin name

### DIFF
--- a/bin/wait-elasticsearch
+++ b/bin/wait-elasticsearch
@@ -26,12 +26,13 @@
 const yargs = require('yargs');
 const { Client: ESClient } = require('@elastic/elasticsearch');
 
+const config = require('../lib/config');
 const Kuzzle = require('../lib/kuzzle');
 
 const sleep = seconds => new Promise(resolve => setTimeout(() => resolve(), seconds * 1000))
 
 async function waitForEs (maxSeconds = 120) {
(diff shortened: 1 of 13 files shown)


bin/wait-elasticsearch
```diff
@@ -26,12 +26,13 @@
...
 const { Client: ESClient } = require('@elastic/elasticsearch');
 
+const config = require('../lib/config');
 const Kuzzle = require('../lib/kuzzle');
 
 const sleep = seconds => new Promise(resolve => setTimeout(() => resolve(), seconds * 1000))
 
 async function waitForEs (maxSeconds = 120) {
-  const kuzzle = new Kuzzle();
+  const kuzzle = new Kuzzle(config);
```
scottinet, line 35: The Kuzzle constructor doesn't take arguments.

Aschen: Ah yes not on master :expressionless:

scottinet: And on 2-dev the wait-elasticsearch script will soon be removed too (see #1754) :grin:

PR #1759, opened 2020-08-14
Expose Controller class for Kaaf 

Aschen:
## What does this PR do ?

This PR exposes an abstract class `Controller` which can be used to define controller classes and use them with Kaaf.  

This is intended to allows users to have a better code organization for their controllers.

Example:

```js
import { Controller, Request } from 'kuzzle';

class EmailsController extends Controller {
  constructor (app) {
    super(app);

    this.definition = {
      actions: {
        send: {
          handler: this.send
        }
      }
    };
  }

  async send (request: Request) {
    /* ... */
  }
}

// Then use the controller 

app.controller.use(new EmailsController(app));
```

The controller name is inferred from the class name. The class name is kebab-cased and the word `controller` is removed.
e.g. `PaymentSolutionController` will became `payment-solution` 

### Other changes

 - exposes `kuzzle-common-objects` in `index.ts`

<details><summary>Original post</summary>

This PR is an open debate to talk about the next `KuzzleController` class that will be available to extends the API instead of using the `app.controller.register` declarative method.

The purpose is to give users a base class so they can extends the API with POO by inheriting from this class.  
The `KuzzleController` class will contains the helpers to extract and validate user input. ([those ones](https://github.com/kuzzleio/kuzzle/blob/master/lib/api/controller/base.js#L97))

The pain point is the routes declaration. Since we want to allow people to develop in both JS and TS, we cannot use decorator (only available in TS).
We will have a default route generation mechanism but we need to allows users to define custom routes for their actions.

Proposals:
```js
class AppController extends KuzzleController {
  constructor () {
    super('app', [
      'greet',
      'slapMe'
    ]);

    // propal 1
    this.routes = {
      greet: [
        // GET /_/app/greet
        { verb: 'GET', path: '/app/greet' },

        // POST /_/app/greet
        { verb: 'POST', path: 'greet' }
      ]
    };

    // propal 2
    this.routesBis = [
      // GET /_/app/greet
      { action: 'greet', verb: 'GET', path: '/app/greet' },

      // POST /_/app/greet
      { action: 'greet', verb: 'POST', path: 'greet' }
    ];
  }
  
  // this is an available API action
  async greet (request) {

  }
  greetRoutes = [ // propal 3
    // GET /_/app/greet
    { verb: 'GET', path: '/app/greet' },

    // POST /_/app/greet
    { verb: 'POST', path: 'greet' }
  ];


  // this is an available API action
  async slapMe (request) {
    // default route generated: GET /_/app/slap-me
  }

  // this method is not an API action
  async otherMethod () {
  }
 }

app.controller.use(AppController);
```

Among Node.js framework there is 2 way of extending the API: declaration DSL or class.

The declarative way is the one we already have in Kaaf:
```js
app.controller.register('email', {
  actions: {
    get: {
      handler: async request => console.log(request.input)
    },
    send: {
      handler: async request => console.log(request.input),
      http: [
        { verb: 'post', url: '/email/send' }
      ]
    },
  }
});
```

My proposal is to allow both way of extending the API because the declaration DSL is useful when you want to quickly prototype something or just building a small application and the classes are better when you have a lot of controllers because they allows to have a better organization of your code.

Example of declaration DSL: [fastify](https://www.fastify.io/docs/latest/Routes/), [hapi.js](https://hapi.dev/tutorials/gettingstarted/?lang=en_US#-adding-routes), [sails.js](https://sailsjs.com/documentation/concepts/actions-and-controllers), [derby.js](https://derbyjs.com/docs/derby-0.10/apps/routes)

Examples of classes: [nest.js](https://docs.nestjs.com/controllers#full-resource-sample), [adonis.js](https://adonisjs.com/docs/4.1/controllers), [loopback 4](https://loopback.io/doc/en/lb4/Getting-started.html)

</details>


scottinet:
I really don't like having 2 different ways of doing the exact same thing, especially when there are subtle differences between them.
This is difficult to document, thus it can be counter-intuitive for users.

I have several proposal to allow having helpers without changing (much) how controllers are declared using kaaf

## 1. Return a KuzzleController instance when registering a new controller

```js
// Extend Kuzzle API with an empty controller
// Return a KuzzleController instance
const emailController = app.controller.register('email');

// Add a single controller action
// The handler has access to all KuzzleController helpers
emailController.addAction('send' {
  handler: async function (request) {
    console.log(`Title: ${this.getBodyString(request, 'title')}`);
    return 'email sent';
  },
  http: [
    { verb: 'post', url: '/email/send' },
  ],
});
```

## 2. Pass a preset context containing helpers (and more):

For instance, we could have a context with the following objects
```
{
  helpers: {
    // request helpers
  },
 pluginContext, // ?
 // other properties?
}
```

Declaring a new controller would look like this:

```js
app.controller.register('email', {
  actions: {
    send: {
      handler: async function (request) => console.log(`Title: ${this.getBodyString(request, 'title')}`)
      http: [
        { verb: 'post', url: '/email/send' },
      ],
    },
    // ...other actions...
  }
});
```

## 3. Simply export helpers as a simple object

That would give something like this:

```js
import { Backend, requestHelpers as helpers } from 'kuzzle';

const app = new Backend('foo');

app.controller.register('email', {
  actions: {
    send: {
      handler: async request => console.log(`Title: ${helpers.getBodyString(request, 'title')}`)
      http: [
        { verb: 'post', url: '/email/send' },
      ],
    },
    // ...other actions...
  }
});
```

Aschen:
@scottinet Every Node.js framework has made a choice between declaration and classes.  
I think both are useful and it could be an advantage to allows users use both.  

Regarding your proposals for the helpers:

1) I don't like this solution because it's forbid the usage of arrow function for handler.
2) Same remark as yours, hidden context is difficult to understand for developers and will lead to many bugs
3) It could be a solution but I don't really like the API. It's like a functional API instead of a OO API.

## Other proposal: put the helpers methods in the Request object

We could simply put the helpers in the Request object so the Request is also responsible for the methods that allows to extract parameters (and throw errors) from it.

```js
// example proposal
request.input.getBoolean('waitForRefresh`);
request.input.getBodyBoolean('flag');
```

scottinet:
>  Every Node.js framework has made a choice between declaration and classes.

I think this summarizes my concern well: we have to choose too, we cannot have both.

And I like your proposal about having those helpers in the Request object.

Aschen:
### Retained solution

Expose a `IController` interface and a `app.controller.use` method to allows user to declare controller classes:
```js
interface IController {
  name: string;
  definition: ControllerDefinition;
}
class Controller implements IController {
  public name: string;
  public definition: ControllerDefinition;
  constructor () {
    this.name = 'email';
    this.definition = {
      actions: {
        get: {
          handler: this.get
        },
      }
    };
  }
  async get (request: Request) {
    return 'hello';
  }
}

const controller = new Controller();

this.app.controller.use(controller);
```

lib/core/application/backend.ts
```diff
@@ -179,6 +180,70 @@ class ControllerManager extends ApplicationManager {
...
+   * Uses a new controller class.
+   *
+   * The controller class must:
+   *  - call the super constructor with the application instance
+   *  - extend the "Controller" class
+   *  - define the "definition" property
+   *  - (optional) define the "name" property
+   *
+   * The controller name will be inferred from the class name.
+   *   e.g. "PaymentSolutionController" will become "payment-solution"
```
scottinet, line 196: I'm not too fond of inferring controller names. Especially considering that the link between the controller class name, and the exposed API entrypoint isn't clear to me.

What about requiring a `name` property in the definition object, and rejecting if it's not there with a meaningful error?

The definition would look like this:

```js
this.definition = {
  name: 'payment',
  actions: {
    // ...
  }
};
```

Aschen: Since it's a class I wanted to have a coherence between the class name and the plugin name. Especially since class name are really important with the new `export` and `import` syntax.

For example our auth plugin class is called `AuthenticationPlugin` (which is a bad generic name) and the plugin name is `kuzzle-plugin-auth-local`. ([see here](https://github.com/kuzzleio/kuzzle-plugin-auth-passport-local/blob/master/lib/index.js#L91))

That's why I wanted to infer the controller name from the class name by default, it will enforce developers to choose meaningful names for their classes.

But for some usecase you may want to set the name by yourself, in this case you use the `name` property:
```
class EmailsController extends Controller {
  constructor (app) {
    super(app);

    this.name = 'emails';

    this.definition = { ... }
  }
}
```

scottinet: The way I see this: when its API is extended with new controllers, Kuzzle needs to know the controller name, its actions, what to do when a controller action is queried, and how those actions can be accessed through HTTP.
With this proposal, 3 of these 4 constraints are regrouped in the `definition` object, and the last one is inferred after the class name. That's why I proposed to also put the name in the definition object.

Your approach might be a valid one if the link between the class name and the controller name is transparent, but I'm not sure this is the case, especially considering that controllers can be added on- the-fly with the `register` method.
Speaking of which, the `register` method takes the controller name as an argument, a closer approach to that behavior with classes might be to pass the controller name to the parent constructor?

One last argument: what about template controllers? Say, an application needs to extend Kuzzle's API, with almost the same controller, with only a few configuration changing. I don't know, might be for multi-tenancy purposes, or to expose the same API for multiple storage indexes with the same structures... if you infer the controller name after the class one, this cannot be done, since Kuzzle will (rightfully) reject new controllers as duplicates.

Aschen: The controller name is inferred from the class name which should most of the use cases.  
About the usecase you are talking about, as I said developers can declare manually the controller name with the `name` property.

Plugins are no longer loaded from the filesystem but they are required/imported from a file or a package. What is imported is a class describing the plugin functionalities so there is no more need for manifest file.

I still didn't had time to work properly on plugins with Kaaf but the manifest file content could be just other properties defined on the plugin class

scottinet: Ah, you're right, it can be done with the `name` property. Ok so I don't see technical reasons to not do it that way. I still don't like it though but, oh, well...

scottinet (approved):

Leodau (approved):

PR #1763, opened 2020-08-18
Use the latest SDK compatible with Typescript

Aschen:
## What does this PR do ?

Use the latest SDK (7.4.0) which is compatible with Typescript


lib/util/interfaces.ts
```diff
@@ -76,3 +76,44 @@ export interface ControllerDefinition {
...
+  collection?: string;
+  error?: {
+    id: string;
+    code: number;
+    message: string;
+    status: number;
+    stack?: string;
+  };
+  requestId: string;
+  result: JSONObject | null;
```
scottinet, line 115: Request results can be of any type. For instance, [this route](https://docs.kuzzle.io/core/2/api/controllers/memory-storage/bitcount/) returns an integer result, and anyone can define plugin APIs returning raw results, which are processed as-is by Kuzzle (could be anything, buffers,  string, json object, ...)

```suggestion
  result: any;
```

scottinet (changes requested):

scottinet (approved):

test/core/application/Backend.test.js
```diff
@@ -43,7 +43,7 @@ describe('Backend', () => {
   });
 
   describe('#sdk', () => {
-    it('should returns the embedded sdk', async () => {
+    it.only('should returns the embedded sdk', async () => {
```
scottinet: debug leftover

scottinet (changes requested):

scottinet (approved):

PR #1764, opened 2020-08-19
Add current commit to server:info

Aschen:
## What does this PR do ?

Try to retrieve the current commit and expose it in the `server:info` action.

image

I know that not every project will be able to use this feature because people can use another CVS (very unlikely) or they will deploy Kuzzle without the `.git` directory (likely).  
But at least people that deploy Kuzzle with the `.git` directory will be able to see the commit hash.


scottinet (approved):
Should be tremendously useful when doing client support :+1:

Aschen:
> Should be tremendously useful when doing client support +1

It's not Kuzzle commit, it's the client application commit. Is that what you understood ?

scottinet:
Of course (we already have the kuzzle "commit" because server:info gives the current kuzzle version)

alexandrebouthinon (commented):
the code LGTM but you need to update Dockerfile to copy the `.git` directory in the builded image

Aschen:
> the code LGTM but you need to update Dockerfile to copy the `.git` directory in the builded image

Why ? The purpose is not to display Kuzzle commit but the current commit of the application built with Kuzzle

alexandrebouthinon (approved):

PR #1765, opened 2020-08-19
Make getRequestSlot accept a function in place of a string

xbill82:
Make the `getRequestSlot` accept a reference to a function in place of a string representing a function name.
The signature of the function changed from this

```javascript
getRequestSlot (executor, request, executeCallback)
```

to this

```javascript
throttle (fn, context, requestId)
```

The previous implementation assumed that `getRequestSlot` could be called only by methods of the Funnel. The `executor` string represented the name of the method. In fact, `getRequestSlot` was kind of aware that `funnel.execute` was the only function that actually called it.

The new implementation takes a reference to a function and its execution context, as well as a unique id that the funnel uses to keep track of it (the name `requestId` is used because, most of the time, these functions are executed to process incoming requests).

I use the functional tests to check that everything works like before and also take a tour of the server with the Admin Console.

### Other changes
* `CacheItem` has been renamed to `PendingRequest`
* `requestsCacheQueue` 👉 `pendingRequestsQueue`
* `requestsCacheById` 👉 `pendingRequestsById`

Again, I think these names are more suitable to the purpose of these lists. It's not a cache we can hit or miss, but rather a queue of functions that wait to be executed (thus, _pending_). Since these functions are often associated with incoming requests, I kept the word `Requests`, but it would be more precise to replace it with `Functions` (ore something else).

* `askRequestSlotAnd` now throws an error if the server is overloaded, whereas it returns true or false indicating whether the function has been executed immediately or is pending in a slot.

### DIFF
--- a/lib/api/funnel.js
+++ b/lib/api/funnel.js
@@ -46,13 +46,13 @@ const processError = kerror.wrap('api', 'process');
 
 /**
  * @class PendingRequest
- * @param requestId
- * @param fn
- * @param context
+ * @param {Request} request
+ * @param {Function} fn
+ * @param {Object} context
  */
 class PendingRequest {
-  constructor(requestId, fn, context) {
-    this.requestId = requestId;
+  constructor(request, fn, context) {
+    this.request = request;
     this.fn = fn;
     this.context = context;
   }
@@ -118,12 +118,12 @@ class Funnel {
   }
 
   /**
-   * Asks the overload-protection system for a request slot.
+   * Asks the overload-protection system for an execution slot.
    *
-   * Returns immediately a truthy value if the request can be
+   * Returns immediately true if the request can be
    * executed.
    *
-   * Otherwise, a falsey value is returned, and the caller MUST
+   * Otherwise, false is returned, and the caller MUST
    * stop the request execution.
    * In this case:
    *   - if it can be bufferized, then the request is left untouched
@@ -137,13 +137,12 @@ class Funnel {
    * @param {Function} fn - The function to be executed as soon as the slot is
    *                        available.
    * @param {Object} context - The execution context of the `fn` param.
-   * @param {String} requestId - A unique ID associated with the execution. It
-   *                             handy to use `request.id` since this operation
-   *                             is often performed upon an incoming request.
+   * @param {String} request - The request object that will be passed as
+   *                           argument to `fn`.
    * @returns {Boolean} - A boolean telling whether the request has been
    *                      immediately executed or not.
    */
-  askRequestSlotAnd (fn, context, requestId) {
+  throttle (fn, context, request) {
     if (this.shuttingDown) {
       throw processError.get('shutting_down');
     }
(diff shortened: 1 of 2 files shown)


lib/api/funnel.js
```diff
@@ -241,64 +246,64 @@ class Funnel {
       return 1;
     }
 
-    const processNow = this.getRequestSlot('execute', request, callback);
+    try {
+      const executing = this.askRequestSlotAnd(() => {
```
Aschen, line 250: I find this method name very verbose, what do you think of `executeOnSlot` ?

Aschen: The new name `throttle` is good to me :+1:

lib/api/funnel.js
```diff
@@ -241,64 +246,65 @@ class Funnel {
...
+      }, this, request);
 
-    // if the connection is closed there is no need to execute the request
-    // => discarding it
-    if (!this.kuzzle.router.isConnectionAlive(request.context)) {
-      debug('Client connection dead: dropping request: %a', request.input);
-      callback(processError.get('connection_dropped'), request);
+      return executing ? 0 : -1;
+    } catch (error) {
+      request.setError(error);
```
scottinet, line 304: `throttle` only throws when refusing a request with a 503 error. When it does, it already sets the right error object in the request, so doing so here is redundant. 
You have to either remove that line, or the `request.setError` one performed by `throttle`.

lib/api/funnel.js
```diff
@@ -241,64 +246,65 @@ class Funnel {
...
     }
 
-    const processNow = this.getRequestSlot('execute', request, callback);
+    try {
+      const executing = this.throttle(req => {
+        if (req.error) {
+          // "handleErrorDump" shouldn't need to be called for 503 errors
+          callback(req.error, req);
+          return;
+        }
```
scottinet, line 255: This test is now obsolete: before it was outside the executor callback. In your new version of the code, this case is handled by the try..catch statement.

Furthermore, the callback is invoked by `throttle`, and it never invokes it if there is an error, thus this test is always false.

scottinet (changes requested):

lib/api/funnel.js
xbill82: This point tinkers me a bit (which is why I left this redundant behavior).
* On one side, I think `throttle` should be responsible of erroring a request if it is discarded,
* On the other side, I find it a bit obscure to let `throttle` mutate an object passed as argument: how do you make it explicit? 

In other words, throwing or returning is the most explicit we can do, but we'll have to remember to manually `request.setError` upon rejection. Mutating the request inside `throttle` does the job for us but we'll have to remember this implicit "side-effect".

What do you think?

xbill82: For the moment, I've choosen the explicit option.

scottinet: I like explicit behaviors.

scottinet (approved):

Aschen (approved):

PR #1766, opened 2020-08-19
Add KuzzleContext base class

Aschen:
## What does this PR do ?

Add a `KuzzleContext` class meant to be extended by users custom classes (services, models, etc.).

This class contain a reference to the instantiated `Backend` so users can access the application context

```
import { Backend, KuzzleContext } from 'kuzzle';

class Service extends KuzzleContext {
  // user defined class
}

const service = new Service();

// here service.app property is not available since the Backend is not instantiated

const app = new Backend('black-mesa');

app.start()
  .then(async () => {
    console.log(await service.app.sdk.server.info())
  })
  .catch(error => {
    console.error(error);
    process.exitCode = 1;
  });
```

### Other changes

  - ensure only one `Backend` can be instantiated 
  - clean travis.ci: parallelize unit tests and linters, play functional tests at the first stage

### DIFF
--- a/lib/core/application/kuzzleContext.ts
+++ b/lib/core/application/kuzzleContext.ts
@@ -1,0 +1,45 @@
+/*
+ * Kuzzle, a backend software, self-hostable and ready to use
+ * to power modern apps
+ *
+ * Copyright 2015-2020 Kuzzle
+ * mailto: support AT kuzzle.io
+ * website: http://kuzzle.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


lib/core/application/kuzzleContext.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { Backend } from './backend';
+import kerror from '../../kerror';
+
+const runtimeError = kerror.wrap('plugin', 'runtime');
+
+/**
+ * Base class to access the application context
+ */
+export class KuzzleContext {
```
scottinet, line 30: I don't get why this should be an object that must be instantiated or inherited to be used. Also, I don't get the purpose of this object at all: what's the use, since all it provides is an app that users already have full control on?

They can already do this if they really want to have a local "app" variable:

```js
class Service {
  constructor (app) {
    this.app = app;
  }
}
```

What problem does this new class solve? :thinking:

Aschen: Yes user can declare their own class and pass the `app` in the constructor but IMHO there is 2 limits to this
 - it's an hassle to have to instantiate every class with the `app` variable (controllers, models, services, etc)
 - class method cannot access the `app` variable

Since the `app` variable it's some kind of global context we could provide an access to it everywhere through this class without having to think about how to pass it across the application and instantiate everything with it

PR #1769, opened 2020-09-01
Avoid printing ES errors in development

Aschen:
## What does this PR do ?

When ES returns an error, with print it for production debugging purpose.  
The problem is that some errors are "normal" and handled by plugins.  

For example the auth plugin use this mechanism to [create the reset secret](https://github.com/kuzzleio/kuzzle-plugin-auth-passport-local/blob/master/lib/index.js#L212) 

This result in a scary message printed by Kuzzle at startup:
image

This PR ensure that ES errors are only logged in production environments.  
The error is also printed as a single JSON string otherwise some information may not be available because they are too deep


lib/service/storage/esWrapper.js
```diff
@@ -160,12 +160,14 @@ class ESWrapper {
       return error;
     }
 
-    this.kuzzle.log.info({
```
Aschen, line 163: Actual change (I don't know why all the cluster files appear on this PR..)

alexandrebouthinon: You should try to rebase your branch on latest 2-dev commit or merge 2-dev on your branch 🙂

Aschen: I already merged 2-dev :(

alexandrebouthinon (approved):

PR #1775, opened 2020-09-03
Allows to load rights for non-existent API actions at startup

### LINKED ISSUE #1771: Cannot load a securities file that contain plugin rights at startup
## Expected Behavior

Trying to load plugin rights with the `--securities` flag at startup fail with an error indicating that the controller/actions does not exists

## Current Behavior

Should load the rights even if the controller/action did not exists for now

## Steps to Reproduce

1.
2.
3.
4.

## Context (Environment)

Kuzzle version:
Node.js version:
SDK version:

Aschen:
## What does this PR do ?

A regression prevent to set rights for API action if those actions does not exists.  

It's problematic because application and plugins can declare new API actions and the only way to prevent unauthorized usage of those action after startup is to set the rights with the `--securities <securities file>` flag with the command line.

Now only a warning will be printed.

Fix #1771

### DIFF
--- a/lib/core/security/roleRepository.js
+++ b/lib/core/security/roleRepository.js
@@ -358,7 +358,7 @@ class RoleRepository extends Repository {
     });
 
     this.checkRoleNativeRights(role);
-    await this.checkRolePluginsRights(role, options);
+    this.checkRolePluginsRights(role, options);
     await this.persistToDatabase(role, options);
 
     const updatedRole = await this.loadOneFromDatabase(role._id);
@@ -411,7 +411,7 @@ class RoleRepository extends Repository {
    * @param {Role} role
    * @param {Force} force
    */
-  async checkRolePluginsRights (role, { force=false, forceWarn=false } = {}) {
+  checkRolePluginsRights (role, { force=false, forceWarn=false } = {}) {
     const plugins = this.kuzzle.pluginsManager;
 
     for (const roleController of Object.keys(role.controllers)) {
@@ -472,7 +472,7 @@ class RoleRepository extends Repository {
     const roles = await this.search({}, {});
 
     for (const role of roles.hits) {
-      await this.checkRolePluginsRights(role, { force: true, forceWarn: true });
+      this.checkRolePluginsRights(role, { force: true, forceWarn: true });
     }
   }
 
--- a/lib/core/security/roleRepository.js
+++ b/lib/core/security/roleRepository.js
@@ -414,8 +414,6 @@ class RoleRepository extends Repository {
   async checkRolePluginsRights (role, { force=false, forceWarn=false } = {}) {
     const plugins = this.kuzzle.pluginsManager;
 
-    const kuzzleStarted = await this.kuzzle.ask('kuzzle:started:get');
-
     for (const roleController of Object.keys(role.controllers)) {
       if ( roleController === '*'
         || this.kuzzle.funnel.isNativeController(roleController)
@@ -437,7 +435,7 @@ class RoleRepository extends Repository {
         // because plugins controllers are loaded after default roles
         // then we need to display non-existing controllers with the sanity check
         // made after plugins controllers loading.
-        if (kuzzleStarted || forceWarn) {
+        if (this.kuzzle.running || forceWarn) {
           this.kuzzle.log.warn(`The role "${role._id}" gives access to the non-existing controller "${roleController}".`);
         }
 
@@ -457,7 +455,7 @@ class RoleRepository extends Repository {
           }
 
           // see the other comment
-          if (kuzzleStarted || forceWarn) {
+          if (this.kuzzle.running || forceWarn) {
             this.kuzzle.log.warn(`The role "${role._id}" gives access to the non-existing action "${action}" for the controller "${roleController}".`);
           }
         }
--- a/lib/core/security/roleRepository.js
+++ b/lib/core/security/roleRepository.js
@@ -358,7 +358,7 @@ class RoleRepository extends Repository {
     });
 
     this.checkRoleNativeRights(role);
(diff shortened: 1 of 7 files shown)


alexandrebouthinon (approved):

lib/core/security/roleRepository.js
```diff
@@ -446,25 +441,12 @@ class RoleRepository extends Repository {
...
       });
     });
   }
 
-  /**
-   * Fetching roles and check for each of them for invalid plugin rights.
-   * If there are some, Kuzzle will log a warning.
-   */
-
-  async sanityCheck () {
```
scottinet, line 460: Why do you remove the sanity check? It prints only warnings about potentially unused plugin rights, to allow maintainers to clean up their projects?

scottinet (changes requested):

lib/core/security/roleRepository.js
Aschen: Because when we register the rights we already print warnings about rights that uses non-existent controllers.

But you're right this is useful so I will modify this PR to not print warning when registering rights and I will run the sanity check after plugins loading :+1:

scottinet (approved):

xbill82 (approved):

Aschen:
@scottinet and @alexandrebouthinon you may want to review this PR again since I changed almost everything

alexandrebouthinon (approved):

lib/kuzzle/kuzzle.js
```diff
@@ -107,6 +107,11 @@ class Kuzzle extends KuzzleEventEmitter {
 
     // Kuzzle version
     this.version = require('../../package.json').version;
+
+    /**
+     * Returns true if Kuzzle has started and is ready to handle requests
+     */
+    this.onAsk('kuzzle:started:get', () => this.started);
```
scottinet, line 114: Instead of an event, wouldn't it better to have a state machine using an enum, and available directly as a property under the kuzzle object?
Could be named `kuzzle.state`.

For now I have the following states in mind: starting, running, shutting down

test/core/security/roleRepository.test.js
```diff
@@ -488,6 +488,8 @@ describe('Test: security/roleRepository', () => {
     const { NativeController } = require('../../../lib/api/controller/base');
 
     beforeEach(() => {
+      kuzzle.ask = sinon.stub();
```
scottinet, line 491: Always stub with `sinon.stub(<object>, "<method name>")`: that way, if the method doesn't exist anymore, or if it is already stubbed, sinon will complain and tests will fail, which might prevent false positive and is more secure.
Also, this allows restoring the stubbed method with `<stubbed method>.restore()`

And in this case, `kuzzle.ask` is already stubbed in KuzzleMock :wink:

test/core/security/roleRepository.test.js
```diff
@@ -591,22 +595,22 @@ describe('Test: security/roleRepository', () => {
...
-    it('should skip non-plugins or wildcarded controllers', () => {
+    it('should skip non-plugins or wildcarded controllers', async () => {
       kuzzle.pluginsManager.isController.returns(false);
 
       const role = new Role();
       role.controllers = { '*': 123 };
 
       kuzzle.funnel.isNativeController.returns(false);
-      should(() => roleRepository.checkRolePluginsRights(role)).not.throw();
+      await should(async () => await roleRepository.checkRolePluginsRights(role)).not.throw();
```
scottinet, line 605: async/await methods don't throw, they reject their promises

So it's simpler to write:

```js
      await should(roleRepository.checkRolePluginsRights(role)).be.fulfilled();
```

Or better yet: 

```suggestion
      await roleRepository.checkRolePluginsRights(role);
```

test/core/security/roleRepository.test.js
```diff
@@ -591,22 +595,22 @@ describe('Test: security/roleRepository', () => {
...
 
       role.controllers = { 'foo': 0 };
 
       kuzzle.funnel.isNativeController.returns(true);
-      should(() => roleRepository.checkRolePluginsRights(role)).not.throw();
+      await should(async () => await roleRepository.checkRolePluginsRights(role)).not.throw();
```
scottinet, line 610: ditto

```suggestion
      await roleRepository.checkRolePluginsRights(role);
```

test/core/security/roleRepository.test.js
```diff
@@ -695,7 +722,8 @@ describe('Test: security/roleRepository', () => {
         }
       };
 
-      should(() => roleRepository.checkRolePluginsRights(role)).not.throw();
+      should(async () => await roleRepository.checkRolePluginsRights(role)).not.throw();
```
scottinet, line 725: possible false positive, you have to "await" the should method since you pass it an async function

but same as before, this is simpler: 

```suggestion
      await roleRepository.checkRolePluginsRights(role);
```

scottinet (changes requested):

test/core/security/roleRepository.test.js
Aschen: Yes you're right :sweat_smile: I just adapted the test quickly without thinking a lot about it..

lib/core/security/roleRepository.js
```diff
@@ -411,10 +411,10 @@ class RoleRepository extends Repository {
    * @param {Role} role
    * @param {Force} force
    */
-  checkRolePluginsRights (role, { force=false } = {}) {
+  async checkRolePluginsRights (role, { force=false, forceWarn=false } = {}) {
```
scottinet, line 414: Should no longer be async, and shouldn't be awaited either (l. 361)

scottinet (changes requested):

scottinet (approved):

PR #1783, opened 2020-09-17
Avoid useless allocations in plugin context

Aschen:
## What does this PR do ?

Most of the objects we expose in the plugin context are re-instantiated each time they are accessed.  
This put useless pressure on the GC (particularly with the Embedded SDK).

Objects are now instantiated only once.


Aschen:
This PR was already merged :facepalm:

PR #1790, opened 2020-09-23
Allow users to disable/configure the Swagger specifications HTTP endpoint

alexandrebouthinon:
## What does this PR do ?

Allow users to disable/configure the Swagger specifications HTTP endpoint from the KuzzleRC configuration file.

* Start a Kuzzle stack
* test the two URL `http://localhost:7512/swagger.json` and `http://localhost:7512/swagger.yml` while playing with configuration values in kuzzlerc file or those environment variables in your Docker Compose file:

```shell
    - kuzzle_http__swagger__enabled=false
    - kuzzle_http__swagger__yamlOutput=true
```

### Other changes

* DRY on old Swagger HTTP routes declarations

### DIFF
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -19,7 +19,17 @@
     // (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS)
     "accessControlAllowOrigin": "*",
     "accessControlAllowMethods": "GET,POST,PUT,DELETE,OPTIONS,HEAD",
-    "accessControlAllowHeaders": "Content-Type, Access-Control-Allow-Headers, Authorization, X-Requested-With, Content-Encoding, Content-Length, X-Kuzzle-Volatile"
+    "accessControlAllowHeaders": "Content-Type, Access-Control-Allow-Headers, Authorization, X-Requested-With, Content-Encoding, Content-Length, X-Kuzzle-Volatile",
+    
+    // [Swagger]
+    //   * enabled: 
+    //      enable the Swagger specifications endpoint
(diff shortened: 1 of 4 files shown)


.kuzzlerc.sample
```diff
@@ -19,7 +19,17 @@
...
+    "accessControlAllowHeaders": "Content-Type, Access-Control-Allow-Headers, Authorization, X-Requested-With, Content-Encoding, Content-Length, X-Kuzzle-Volatile",
+    
+    // [Swagger]
+    //   * enabled: 
+    //      enable the Swagger specifications endpoint
+    //   * yamlOutput:
+    //      use YAML format to print Swagger specifications instead of JSON
+    "swagger": {
+      "enabled": true,
+      "yamlOutput": false
```
scottinet, line 31: Personal preference here, to be debated: what about an `output` property taking either "json" or "yaml" as values?

```js
      "output": "json"
```

This would need an assertion check when using the configuration, but otherwise I find that easier to read/understand. Here, one could assume that this enables YAML output on top of the JSON one.
WDYT?

scottinet:
When we discussed this issue, we were more inclined to make the swagger route a true API action (e.g. `server:swagger`), rather than an option to activate/deactivate the feature.

This would allow clients to configure control rights on this route, and we could expose the same URL to prevent breaking changes.
And we would finally be able to properly document it (this is still an undocumented feature).

I think that an activation toggle is mutually exclusive with adding this route to the API, so we should consider this matter carefully before approving that PR.

alexandrebouthinon:
We agreed to make swagger specifications available via a new server controller API route. So I close this PR

PR #1792, opened 2020-09-24
Add (un)subscription to PluginContext

xbill82:
Introduce `accessors.realtime.subscribe` and `accessors.realtime.unsubscribe` to the `PluginContext`.

See the functions `registerSubscription` and `unregisterSubscription` in the functional-test Plugin.


Aschen:
Proposal: instead of having those method under `accessors.realtime.registerSubscription` we could have `accessors.subscriptions.register`.

In this way we could also imagine to expose the list of subscriptions in this accessor. Moreover this feature should also be added to Kaaf IMHO

doc/2/plugins/plugin-context/accessors/realtime/index.md
```diff
@@ -0,0 +1,68 @@
...
+title: registerSubscription
+---
+
+# `realtime`
+
+Add or remove [real-time subscriptions](/core/2/guides/essentials/real-time) from the backend.
+
+## `registerSubscription`
+
+Registers a new realtime subscription. The subscription works exaclty like the one created by the [`realtime:subscribe` API endpoint](/api/controllers/realtime/subscribe/index.md). The notifications will be sent to the connection identified by the `Request` object passed to the method.
```
scottinet, line 13:
```suggestion
Registers a new realtime subscription on behalf of a client. The subscription works exactly like the one created by the [`realtime:subscribe` API endpoint](/api/controllers/realtime/subscribe/index.md). The notifications will be sent to the connection identified by the `Request` object passed to the method.
```

doc/2/plugins/plugin-context/accessors/realtime/index.md
```diff
@@ -0,0 +1,68 @@
+---
+code: true
+type: page
+title: registerSubscription
+---
+
+# `realtime`
+
+Add or remove [real-time subscriptions](/core/2/guides/essentials/real-time) from the backend.
```
scottinet, line 9:
```suggestion
Adds or removes [real-time subscriptions](/core/2/guides/essentials/real-time) from the backend.
```

doc/2/plugins/plugin-context/accessors/realtime/index.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+---
+
+### Use cases
+
+The most common use-case of this method is to implement the realtime subscriptions on the server-side. Instead of leaving the client-side with the freedom (and the responsibility) of specifying the filters for a subscription, you can open a custom API endpoint in your plugin and call `registerSubscription` inside it by passing the incoming Request. When the client hits this endpoint, they will start receiving the notifications from the newly created subscription. This way, the client will only be aware of a `subscribeToSomething` business-specific endpoint, but won't see the set of filters the endpoint encapsulates. 
+
+::: info
+Note that the same restriction can be achieved by adding a pipe on the `realtime:subscribe` API endpoint. The pipe would check the filters object and make the request fail if the filters are not allowed. This approach has two main downsides:
```
scottinet, line 65:
```suggestion
Note that the same restrictions can be achieved by adding a pipe on the `realtime:subscribe` API endpoint. The pipe would check the filters object and make the request fail if the filters are not allowed. This approach has two main downsides:
```

scottinet (changes requested):

scottinet:
@Aschen > counter-proposal :grin: 
Either have a `accessors.realtime.subscribe` method, to keep everything realtime-related under the same object or, perhaps better, make the `sdk.as(...).realtime.subscribe` work :thinking: 
(or perhaps both?)

xbill82:
@scottinet I forgot to update the docs. The code reflects my intention.

Aschen:
@scottinet We can use the `sdk.realtime.subscribe` and `sdk.realtime.unsubscribe` methods since this PR https://github.com/kuzzleio/kuzzle/pull/1645

The purpose of this PR is to allows to add subscription to another connection. The `sdk.as()` method takes an user as argument but we need the `connection` object. 
I'm not really fond of allowing to use `sdk.realtime.subscribe` to subscribe to realtime notifications for another user because this method take a callback and we just want Kuzzle to start sending realtime notification to another connection. The client can then do whatever he want with those notifications

scottinet:
@Aschen >Ah, I get it now. Thanks for the explanation

@xbill82 > The documentation page title is still wrong it seems. And now that I understand what's going on, I also wonder about the accessors.realtime.(un)subscribe name: it might be misleading, since it sounds like the usual "subscribe" in our SDKs and in our API, and yet it's a different beast.
We could either rename it to (un)register (or (un)attach?), and/or make the connection.id argument stand out independently from the request object. So we could have something like:
`accessors.realtime.register(connectionId, filters)`
or:
`accessors.realtime.attach(connectionId, filters)`

Aschen:
> So we could have something like:
> 
> `accessors.realtime.register(connectionId, filters)`
> or:
> `accessors.realtime.attach(connectionId, filters)`
> 
> WDYT?

Why don't call the accessor `subscription` instead?  So later we could imagine to expose a list of current subscriptions for example.

scottinet:
Sounds good to me.
I also insist in replacing the `request` argument with a more explicit (and probably easier to use) pair of connectionId + filters

doc/2/plugins/plugin-context/accessors/realtime/index.md
```diff
@@ -0,0 +1,117 @@
+---
+code: true
+type: page
+title: registerSubscription
```
scottinet, line 4:
```suggestion
title: realtime
```

doc/2/plugins/plugin-context/accessors/realtime/index.md
```diff
@@ -0,0 +1,117 @@
...
+---
+
+# `realtime`
+
+Adds or removes [real-time subscriptions](/core/2/guides/essentials/real-time) from the backend.
+
+## `subscribe`
+
+Registers a new realtime subscription on behalf of a client. The subscription works exactly like the one created by the [`realtime:subscribe` API endpoint](/api/controllers/realtime/subscribe/index.md). The notifications will be sent to the connection identified by the `Request` object passed to the method.
```
scottinet, line 13: This method documentation is obsolete: it doesn't accept a `Request` object anymore.

rolljee (approved):

lib/core/plugin/pluginContext.js
```diff
@@ -175,6 +175,34 @@ class PluginContext {
...
+              },
+              {
+                connectionId: connectionId,
+              });
+            return kuzzle.ask(
+              'core:realtime:subscribe',
+              request
+            );
+          },
+          unregister: (connectionId, roomId, notify) =>
```
scottinet, line 198: According to this PR documentation, shouldn't this be the following instead?

```suggestion
      Reflect.defineProperty(this.accessors, 'realtime', {
        enumerable: true,
        value: {
          subscribe: (connectionId, index, collection, filters) => {
            const request = new Request(
              {
                action: 'subscribe',
                body: filters,
                collection,
                controller: 'realtime',
                index,
              },
              {
                connectionId: connectionId,
              });
            return kuzzle.ask(
              'core:realtime:subscribe',
              request
            );
          },
          unsubscribe: (connectionId, roomId, notify) => 
```

doc/2/plugins/plugin-context/accessors/realtime/index.md
```diff
@@ -0,0 +1,117 @@
+---
+code: true
+type: page
+title: registerSubscription
+---
+
+# `realtime`
+
+Adds or removes [real-time subscriptions](/core/2/guides/essentials/real-time) from the backend.
```
Aschen, line 9:
```suggestion
Manage [real-time subscriptions](/core/2/guides/essentials/real-time) from the backend.
```
For now it's only add/remove but it may evolve

plugins/available/functional-test-plugin/functionalTestPlugin.js
```diff
@@ -118,6 +138,39 @@ class FunctionalTestPlugin {
     await this.sdk.server.info();
   }
 
+  // accessors.registerSubscription related methods ============================
+
+  async registerSubscription(request) {
+    const roomId = await this.context.accessors.realtime.subscribe(
```
scottinet, line 144: I must be missing something in the code :thinking: 
I see that you define `accessors.subscription.(un)register` methods in the plugin context, but here you seem to call `accessors.realtime.(un)subscribe` without error :thinking: 

What's going on?

doc/2/plugins/plugin-context/accessors/realtime/index.md
Aschen, line 13: Nitpicking: we try to use the term "action" and not method or endpoint which can be confusing (the first with a method in the code and the second because it's mainly related to HTTP)
```suggestion
Registers a new realtime subscription on behalf of a client. The subscription works exactly like the one created by the [realtime:subscribe](/api/controllers/realtime/subscribe/index.md) API action. The notifications will be sent to the connection identified by the `Request` object passed to the method.
```

doc/2/plugins/plugin-context/accessors/realtime/index.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+---
+
+### Use cases
+
+The most common use-case of this method is to implement the realtime subscriptions on the server-side. Instead of leaving the client-side with the freedom (and the responsibility) of specifying the filters for a subscription, you can open a custom API endpoint in your plugin and call `accessors.realtime.subscribe` inside it by passing the incoming Request. When the client hits this endpoint, they will start receiving the notifications from the newly created subscription. This way, the client will only be aware of a `subscribeToSomething` business-specific endpoint, but won't see the set of filters the endpoint encapsulates. 
+
+::: info
+Note that the same restrictions can be achieved by adding a pipe on the `realtime:subscribe` API endpoint. The pipe would check the filters object and make the request fail if the filters are not allowed. This approach has two main downsides:
```
Aschen, line 70:
```suggestion
Note that the same restrictions can be achieved by adding a pipe on the `realtime:subscribe` API action. The pipe would check the filters object and make the request fail if the filters are not allowed. This approach has two main downsides:
```

doc/2/plugins/plugin-context/accessors/realtime/index.md
```diff
@@ -0,0 +1,117 @@
...
+## `subscribe`
+
+Registers a new realtime subscription on behalf of a client. The subscription works exactly like the one created by the [`realtime:subscribe` API endpoint](/api/controllers/realtime/subscribe/index.md). The notifications will be sent to the connection identified by the `Request` object passed to the method.
+
+---
+
+### Arguments
+
+```js
+subscribe(request)
```
Aschen, line 20: This signature is outdated

scottinet (changes requested):

scottinet:
Also you need to update this PR description, as it'll be used in our changelogs.

lib/core/plugin/pluginContext.js
Aschen, line 198: I would rather use an option object for the last parameter because it's more explicit than just passing a boolean
```suggestion
          unregister: (connectionId, roomId, { notify=true } = {}) => 
```

Aschen (changes requested):

doc/2/plugins/plugin-context/accessors/realtime/index.md
scottinet: (Manages)

doc/2/plugins/plugin-context/accessors/subscription/index.md
```diff
@@ -0,0 +1,109 @@
...
+title: subscription
+---
+
+# `subscription`
+
+Adds or removes [real-time subscriptions](/core/2/guides/essentials/real-time) from the backend.
+
+## `register`
+
+Registers a new realtime subscription on behalf of a client. The subscription works exactly like the one created by the [`realtime:subscribe` API endpoint](/api/controllers/realtime/subscribe/index.md). The notifications will be sent to the connection identified by the `Request` object passed to the method.
```
scottinet, line 13:
```suggestion
Registers a new realtime subscription on behalf of a client. The subscription works exactly like the one created by the [`realtime:subscribe` API endpoint](/api/controllers/realtime/subscribe/index.md). The notifications will be sent to the connection identified by the connection identifier passed to the method.
```

scottinet (changes requested):
Request leftover

package.json
```diff
@@ -7,6 +7,7 @@
     "kuzzle": "./bin/start-kuzzle-server"
   },
   "scripts": {
+    "start:dev": "node -r ts-node/register docker/scripts/start-kuzzle-dev.ts --enable-plugins functional-test-plugin",
```
scottinet, line 10: Good idea for running kuzzle outside docker images. Though it would probably be better to start it with the debugger on, as well as the automated reloader:

```suggestion
    "start:dev": "node docker/scripts/reloader.js --inspect=0.0.0.0:9229 -r ts-node/register docker/scripts/start-kuzzle-dev.ts --enable-plugins functional-test-plugin",
```

wdyt?

scottinet (approved):

Aschen (approved):

PR #1793, opened 2020-09-25
Loosely coupling the data storage module

### LINKED ISSUE #1655: As an end user, I'd like to identify Kuzzle connection to Redis
Kuzzle could use CLIENT SETNAME in order to be easily identified.

Hardly an urgent need, but could be useful for monitoring and debug when the same redis instance is shared with multiple clients.

https://github.com/kuzzleio/kuzzle/issues/1320

depends on #1692

scottinet:
# Description

Loosely couple the storage module (i.e. everything related to Elasticsearch).

The events follow the following naming convention: `core:storage:<scope>:<object>:<action>`

With:

* `scope`: either `public` or `private`, preventing accidental access to private indexes from the API
* `object`: a storage object: `index`, `collection`, `mappings`, `document`
* `action`: the action to perform on the object

# Storage classes

Here is the new class structures to access storage. Things have slightly changed, but those familiar with the old structure shouldn't be too lost (well... is there anyone else than @Aschen familiar with this? :thinking:)

```
Storage Engine (core/storage/storageEngine.js): entry point
  |- "public" Client Adapter (core/storage/clientAdapter.js)
    |- "public" index cache (core/storage/indexCache.js)
    |- "public" Elasticsearch service (service/storage/elasticsearch.js)
  |- "private" Client Adapter
    |- "private" index cache
    |- "private" Elasticsearch service
```

How that works:

* the storage engine:
  * is tasked with instantiating a client adapter, one for public indexes, the other for private ones
  * verifies at startup that there are no duplicate indexes (we don't allow private and public indexes to have the same name)
* the client adapter:
  * registers Kuzzle events (`core:storage:...`) and wraps the ES service
  * when initialized, it populates its own index cache, and is responsible of maintaining it

# Changes

* the distinction between index scope has been renamed "private" and "public" (instead of "internal" and "public")
* add a new "Mutex" utility class, allowing to safely lock resources across a Kuzzle cluster (`util/mutex.js`)
* the `core/storage/IndexCache` class has been moved to `core/share/Store.js`
  * Its role has slightly changed: it serves as a convenience wrapper to the most used store events, and is dedicated to one index, provided to its constructor. It's now also used as a factory class to pass around to parts of the code (e.g. repository classes)
  * Contrary to its previous version (when it was called "IndexStorage"), only a handful selection of events are wrapped. This class is not meant to completely wrap the store events: only those needed by classes using it as a factory object have been wrapped
  * Its `init` method, used to bootstrap the necessary index/collections if needs be, is now protected by the new `Mutex` class, ensuring that no race condition can occur when initializing new collections
* Since the refactored `core/shared/store` safely handles its initialization with mutexes, the `safeBootstrap` class is now obsolete and has been removed
* `kuzzle.internalIndex` is now an instance of `core/kuzzle/InternalIndexHandler`
  * This class inherits `core/shared/Store` (attached to the kuzzle's internal index), and adds a few convenience methods
  * Bootstrapping the Kuzzle internal storage structure is now protected with mutexes
  * This class is now responsible of handling the secret hash. Which can now be accessed using the new `kuzzle.secret` property after kuzzle has started
* the indexes cache has been reworked: instead of a global cache for both public and private indexes, handled by the storage engine, now each client adapter has its own cache (see the storage classes summary above)

# Other changes

* remove the "fromCache" option in storage engine methods, since it seems to be obsolete
* moved the cache/store configuration of the repository classes to the constructor (prior to this, those were defined in the init function). Repository classes now take a `Store` instance as their storage option (defaults to `null`: no permanent storage)
* replace the locking mechanism of the `admin:resetSecurity` and `admin:resetDatabase` API actions: they now use the new mutex class to prevent multiple resets to occur on a kuzzle cluster
* the `core:storage:<scope>:document:create` action takes the document identifier in its option argument (instead of being a dedicated argument)
* the core modules starter has been deleted (`core/index.js`). It seemed like a good idea, until now: there are too many interconnections to be viable. Kuzzle is now responsible of starting its own modules.
* new `kuzzle.id` property: it's the cluster node's unique identifier. Currently only used for mutexes, but will have more usages with the new cluster
* mappings import is now protected with mutexes, to prevent race conditions when starting multiple kuzzle nodes at the same time with the `--mappings`  option
* use the new kuzzle node ID to identify redis connections (fixes #1655)

# Boyscout

* asyncify modified functions
* remove the `connectionId` argument from ApiKey.create (unused)
* log cluster heartbeats in a dedicated `kuzzle:cluster:heartbeat` debug namespace, and filter it out by default from the docker-compose for development


Aschen:
Why `store` for the event namespace and not `storage` ?

scottinet:
it's a bit shorter :expressionless: 

if needed this can be changed to storage, I don't have strong feelings about this

lib/api/controller/collection.js
```diff
@@ -214,26 +211,29 @@ class CollectionController extends NativeController {
...
 
-    return this.kuzzle.internalIndex.delete('validations', specificationsId)
-      .catch(error => {
-        if (error.status === 404) {
-          return {
-            acknowledged: true
-          };
-        }
+    try {
+      await this.kuzzle.internalIndex.delete('validations', specificationsId);
```
Aschen, line 219: The `internalIndex` is not loosely coupled?

scottinet: See the PR changelog: `kuzzle.internalIndex` is an instance of the `core/shared/Store` class, which is meant to be a universal factory class to access storage spaces (used in several parts of the code). Under the hood, it calls the storage module using events.

Aschen:
> it's a bit shorter expressionless
> 
> if needed this can be changed to storage, I don't have strong feelings about this

I understand you concern about the size but IMHO it's more important to have a more explicit name who match the rest of our terminology

plugins/available/kuzzle-plugin-cluster/lib/node.js
```diff
@@ -24,6 +24,7 @@ const Bluebird = require('bluebird');
 const debug = require('debug')('kuzzle:cluster');
```
Aschen, line 24: Feel so good to be able to modify the cluster directly in Kuzzle :grin:

scottinet: You can say that again :+1:

lib/util/mutex.js
```diff
@@ -0,0 +1,158 @@
...
+   *                                 if 0, locking will fail immediately if it
+   *                                 cannot lock with its 1st attempt
+   *                                 (default) if -1, will try to acquire the
+   *                                 lock indefinitely
+   * @param {Number} options.ttl - lock TTL in milliseconds (default: 5000)
+   */
+  constructor (kuzzle, id, { timeout = -1, ttl = 5000 } = {}) {
+    this.kuzzle = kuzzle;
+    this.lockId = id;
+    this.mutexId = `${kuzzle.id}/${randomBytes(16).toString('hex')}`;
```
Leodau, line 82: What if there is already a mutex with the same ID. :hand_over_mouth:

scottinet: [Alt text](https://www.youtube.com/watch?v=X4oV6rKyO5I)

lib/core/shared/store.js
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * Creates collections with the provided mappings
+   *
+   * @param {Object} collections - collections with mappings
+   *
+   * @returns {Promise}
+   */
+  createCollections (collections) {
+    return Bluebird.map(
```
Aschen, line 111: You should limit concurrency on this one because if there is a lot of collections it will put too much pressure on ES. (Omniscient crashed because of something like this)

lib/kuzzle/internalIndexHandler.js
```diff
@@ -0,0 +1,241 @@
...
+   */
+  async init () {
+    await super.init(this.config.collections);
+
+    const mutex = new Mutex(this.kuzzle, 'InternalIndexBootstrap', {
+      timeout: -1,
+      ttl: 30000,
+    });
+
+    await mutex.lock();
```
Aschen, line 108: Can't we check if the internal index is bootstrapped before trying to lock the mutex? 

Actually each node will have to wait until the mutex is free to check if the internal index exists or not.

lib/kuzzle/internalIndexHandler.js
```diff
@@ -0,0 +1,241 @@
...
+    const seed = this.kuzzle.config.security.jwt.secret
+      || crypto.randomBytes(512).toString('hex');
+
+    try {
+      await this.create('config', { seed }, {
+        id: this._JWT_SECRET_ID,
+      });
+    }
+    catch (e) {
+      if (e.id !== 'services.storage.document_already_exists') {
```
Aschen, line 210: This should not happen anymore since we have a mutex on bootstrapping.

lib/kuzzle/internalIndexHandler.js
```diff
@@ -0,0 +1,241 @@
...
+      obj.userId,
+      obj.ttl);
+
+    await ApiKey.batchExecute({ match_all: {} }, apiKeys => {
+      for (const { _source } of apiKeys) {
+        promises.push(createToken(_source));
+      }
+    });
+
+    return Bluebird.all(promises);
```
Aschen, line 237: We should put a concurrency limit here in case there is a lot of API keys

lib/util/mutex.js
```diff
@@ -0,0 +1,158 @@
...
+   * @param {Object} options - mutex options
+   * @param {Number} options.timeout - mutex lock acquisition timeout in
+   *                                 milliseconds
+   *                                 if 0, locking will fail immediately if it
+   *                                 cannot lock with its 1st attempt
+   *                                 (default) if -1, will try to acquire the
+   *                                 lock indefinitely
+   * @param {Number} options.ttl - lock TTL in milliseconds (default: 5000)
+   */
+  constructor (kuzzle, id, { timeout = -1, ttl = 5000 } = {}) {
```
Aschen, line 79: Why the `attemptDelay` is not configurable as well? We may want to use a mutex with a very small delay between each try

lib/util/mutex.js
```diff
@@ -0,0 +1,158 @@
+/*
```
Aschen, line 1: Discussion: This class should be exposed in the `kuzzle` package and the plugin context so we can use it in plugins and Kaaf.

It should be rewritten in TS and we need to find a way to get ride of the `kuzzle` constructor argument as well.

But I think we should do that later in a separate PR

Aschen (changes requested):

lib/service/storage/elasticsearch.js
```diff
@@ -750,7 +753,7 @@ class ElasticSearch extends Service {
   * @param {Object} changes - Changes wanted on documents
   * @param {Object} options - refresh (undefined), size (undefined)
   *
-  * @returns {Promise.<Object>} { successes: [_id, _source, _status], errors: [ document, status, reason ] }
+  * @returns {Promise.<{ successes: [_id, _source, _status], errors: [ document, status, reason ] }>}
```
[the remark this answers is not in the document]
scottinet: It's not the ts syntax, it's the jsdoc :thinking:

lib/kuzzle/internalIndexHandler.js
scottinet: We can't do that, because the "bootstrap done" document could be created between the check and the mutex lock.

lib/util/mutex.js
scottinet: Ah, I thought about making it available to plugins, but I didn't think that would mean writing it in typescript. This should be converted in another PR though, but yes, I agree with that.

I don't see how we could remove `kuzzle` from the constructor, since we need to emit events to access the cache module. I don't think this is a problem, though. Plugins for instance could have access to a wrapper to this class for instance. Same for kaaf I think.

lib/util/mutex.js
scottinet: I guess we could make it configurable, if we needed. This should be easy to add if the need ever arise. wdyt?

lib/kuzzle/internalIndexHandler.js
scottinet: You might be right here. We should do that in another PR though: this requires rewriting the BaseModel's batchExecute logic, and I think that this is outside the scope of this PR.

lib/service/storage/elasticsearch.js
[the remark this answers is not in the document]
Aschen: It was to be consistent with TS but I don't really care

lib/kuzzle/internalIndexHandler.js
Aschen: Oh yes you're right.

lib/kuzzle/kuzzle.js
```diff
@@ -57,6 +59,9 @@ class Kuzzle extends KuzzleEventEmitter {
       config.plugins.common.maxConcurrentPipes,
       config.plugins.common.pipesBufferSize);
 
+    // Node unique identifier
+    this.id = randomBytes(16).toString('hex');
```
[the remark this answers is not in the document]
scottinet: Ok, why not. :+1:

lib/util/mutex.js
Aschen: We may not be the one who are gonna need this but our users could need it.

lib/util/mutex.js
Aschen: We talked about putting the `kuzzle` instance in the `process` global object so the `ask` and `onAsk` method can be in an independent file and we don't need to pass the kuzzle object in every constructors.  
It will allow to expose the `Mutex` class directly as well;
What do you think?

lib/kuzzle/kuzzle.js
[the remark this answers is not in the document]
Aschen: Also, this could be used to identify Kuzzle node connection to Redis.  
If it's not too much work you could add it in this PR, otherwise we can do it in a separate PR.

See https://github.com/kuzzleio/kuzzle/issues/1655

lib/service/storage/elasticsearch.js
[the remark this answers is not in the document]
scottinet: ah, but I see what you mean, the array is badly described :+1:

lib/kuzzle/kuzzle.js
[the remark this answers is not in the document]
scottinet: already done ;)

lib/util/mutex.js
scottinet: I don't like the idea of adding properties to "process".

I would prefer if we would force the `Kuzzle` object to be a singleton, meaning that we shouldn't be able to instantiate multiple kuzzle objects. If we do so, we could simply add a static getter method returning that singleton.
This means that we could simply do this:

```js
const Kuzzle = require('kuzzle');

const kuzzle = Kuzzle.getInstance();
```

lib/kuzzle/kuzzle.js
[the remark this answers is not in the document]
scottinet: Good thinking. And done, all clients are now identified, even the cluster. Format is: "adapter name/node id"

```
127.0.0.1:6379> client list

id=105 addr=192.168.112.6:43298 fd=15 name=cluster_manager/general-sloppy-run-6762 age=253 idle=2 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0 events=r cmd=smembers
id=100 addr=192.168.112.4:57952 fd=10 name=public_adapter/military-loud-competition-5560 age=254 idle=253 flags=N db=5 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0 events=r cmd=info
id=101 addr=192.168.112.4:57954 fd=11 name=internal_adapter/military-loud-competition-5560 age=254 idle=3 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0 events=r cmd=set
id=102 addr=192.168.112.6:43254 fd=12 name=public_adapter/general-sloppy-run-6762 age=254 idle=253 flags=N db=5 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0 events=r cmd=info
id=107 addr=127.0.0.1:54748 fd=17 name= age=28 idle=0 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=26 qbuf-free=32742 obl=0 oll=0 omem=0 events=r cmd=client
id=103 addr=192.168.112.6:43256 fd=13 name=internal_adapter/general-sloppy-run-6762 age=254 idle=3 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0 events=r cmd=set
id=106 addr=192.168.112.4:58012 fd=16 name=cluster_manager/military-loud-competition-5560 age=253 idle=2 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0 events=r cmd=smembers
id=104 addr=192.168.112.5:34274 fd=14 name=cluster_manager/wry-soggy-committee-4186 age=253 idle=3 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0 events=r cmd=smembers
id=98 addr=192.168.112.5:34220 fd=8 name=internal_adapter/wry-soggy-committee-4186 age=254 idle=3 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0 events=r cmd=set
id=99 addr=192.168.112.5:34222 fd=9 name=public_adapter/wry-soggy-committee-4186 age=254 idle=253 flags=N db=5 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0 events=r cmd=info
```

lib/util/mutex.js
scottinet: ... ok. I still don't think that this is useful, but I added it anyway.

lib/kuzzle/kuzzle.js
```diff
@@ -22,11 +22,11 @@
 'use strict';
 
 const path = require('path');
-const { randomBytes } = require('crypto');
 
 const { murmurHash128: murmur } = require('murmurhash-native');
 const stringify = require('json-stable-stringify');
 const Koncorde = require('koncorde');
+const nameGenerator = require('project-name-generator');
```
Aschen, line 29: Even if I really like to have human readable name I prefer if we don't add another dependency to Kuzzle for this.

We could just take the [nouns](https://github.com/aceakash/project-name-generator/blob/master/src/nouns.js) and [adjectives](https://github.com/aceakash/project-name-generator/blob/master/src/adjectives.js) and generate the name by ourselves, what do you think?

lib/util/mutex.js
Aschen: Or we could simply expose the `kuzzle` instance directly?

```
const { kuzzle } = require('kuzzle')
```

I'm not sure we can do that thus.

We can discuss it when it comes to expose the Mutex class :+1:

scottinet: We could do what you propose. But, hmm... this would mean we could do this: 

```js
const { Kuzzle, kuzzle } = require('kuzzle');
```

bakk7

Aschen: :sweat_smile:

lib/kuzzle/kuzzle.js
scottinet: You're no fun :expressionless: 

Ok I wrote one. And while I was at it, I created a custom list of nouns and adjectives. Collision chances are about 1 to 24'359'731'200, that should be enough for our purposes :expressionless:

scottinet: This name generator is SLOW, I don't want to spend time making it quicker for now. If we want to use it for other purposes than generating a node identifier, we might need to make it better.

Aschen (approved):
LGTM :+1:

Yoann-Abbes (approved):

lib/api/controller/collection.js
```diff
@@ -174,38 +172,37 @@ class CollectionController extends NativeController {
...
-          {
-            collection,
-            index,
-            validation: specifications
-          });
-      })
-      .then(() => this.kuzzle.internalIndex.refreshCollection('validations'))
-      .then(() => this.kuzzle.validation.curateSpecification())
-      .then(() => specifications);
+    if (! isValid) {
```
Yoann-Abbes, line 185:
```suggestion
    if (!isValid) {
```

lib/core/application/backend.ts
```diff
@@ -435,7 +435,8 @@ class StorageManager extends ApplicationManager {
    */
   get client (): Client {
     if (! this._client) {
```
Yoann-Abbes, line 437:
```suggestion
    if (!this._client) {
```

lib/model/storage/baseModel.js
```diff
@@ -85,18 +84,17 @@ class BaseModel {
    */
   async save ({ userId=null, refresh } = {}) {
     if (! this.__persisted) {
```
Yoann-Abbes, line 86:
```suggestion
    if (!this.__persisted) {
```

lib/service/storage/elasticsearch.js
```diff
@@ -821,7 +824,7 @@ class ElasticSearch extends Service {
       size
     };
 
-    if (! _.isPlainObject(query)) {
+    if (! isPlainObject(query)) {
```
Yoann-Abbes, line 827:
```suggestion
    if (!isPlainObject(query)) {
```

lib/kuzzle/kuzzle.js
Leodau: :open_mouth:

scottinet: Added a lot more names: I forgot to explore greek mythology and mystical creatures. :smile:

rolljee (approved):

PR #1799, opened 2020-10-01
Add new server:openapi route

alexandrebouthinon:
## What does this PR do ?

Add new route `server:openapi` to the server controller

* Start a Kuzzle stack
* test the two URL `http://localhost:7512/_swagger&format=json` and `http://localhost:7512/_swagger&format=yaml`

### Boyscout 

* Clean JS import in serverController
* Lint using new rules


lib/api/controller/server.js
```diff
@@ -235,6 +237,28 @@ class ServerController extends NativeController {
     return apiDefinition;
   }
 
+  async swagger (request) {
+    const format = 
+      // The YAML response format is only valid with an HTTP request
+      request.context.connection.protocol === 'http'&& request.input.args.format
```
Aschen, line 243: Why can't we have YAML in WebSocket or MQTT? YAML it's just a string so we could put it in the result

package.json
```diff
@@ -96,6 +96,7 @@
     "codecov": "^3.7.2",
     "cucumber": "^6.0.5",
     "eslint": "^7.6.0",
+    "js-yaml": "^3.14.0",
```
Aschen, line 99: Did you remove the former package `jsonToYaml` ?

Aschen (changes requested):

package.json
alexandrebouthinon: Nop, `jsonToYaml` is only for JSON to YAML conversion but I need YAML to JSON conversion in unit tests to performs a few checks (this is why I add it in the `devDependencies` only)

Aschen (approved):

Aschen (changes requested):
IMHO we should change the `.kuzzlerc.sample` and also the `config.default.js` files to give the right to the anonymous user to call this action

Aschen (approved):

scottinet (approved):

PR #1801, opened 2020-10-02
Revert unintended change on `realtime:unsubscribe` returned value

Yoann-Abbes:
## What does this PR do ?

Reverse an unintended change on `realtime:unsubscribe` return value

  - Step 1 :
  - Step 2 :
  - Step 3 :  
  ...

### DIFF
--- a/lib/api/controller/realtime.js
+++ b/lib/api/controller/realtime.js
@@ -107,7 +107,7 @@ class RealtimeController extends NativeController {
 
     this.kuzzle.tokenManager.unlink(request.context.token, roomId);
 
-    return { roomId };
+    return roomId;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lib/api/controller/realtime.js
```diff
@@ -107,7 +107,7 @@ class RealtimeController extends NativeController {
 
     this.kuzzle.tokenManager.unlink(request.context.token, roomId);
 
-    return { roomId };
+    return roomId;
```
scottinet, line 110: I don't get it: the previous version seems to match the documented returned value (https://docs.kuzzle.io/core/2/api/controllers/realtime/unsubscribe/#response)

scottinet (changes requested):

lib/api/controller/realtime.js
Yoann-Abbes: with `{ roomId }` as returned value it's not the response will not have the `result` level

Yoann-Abbes: As we discussed, you did a correction, not a mistake, I'm closing this pr

Issue #1661, opened 2020-06-18
As an end user, I'd like to know when I'm using a deprecated feature


berthieresteban:
We should notify users (in development) when they are using obsolete features.

Controllers should checks for deprecation (method, parameters, etc)

Each deprecation is then added to an array returned with the response.

This array contain deprecation objects with a message and a version

SDK should then be able to print deprecation warning

https://trello.com/c/JHelN9Fa/216-adm-deprecation-module

rolljee:
- [ ] https://docs.kuzzle.io/sdk/android/3/essentials/realtime-notifications/
- [ ] https://docs.kuzzle.io/sdk/js/7/protocols/http/properties/
- [ ] https://docs.kuzzle.io/sdk/js/7/controllers/collection/update/
- [ ] https://docs.kuzzle.io/sdk/js/7/controllers/collection/create/

- [ ] https://docs.kuzzle.io/core/1/plugins/plugin-context/constructors/dsl/
- [ ] https://docs.kuzzle.io/core/2/plugins/plugin-context/errors/parseerror/
- [ ] https://docs.kuzzle.io/core/1/plugins/guides/manual-setup/prerequisites/
- [ ] https://docs.kuzzle.io/core/2/plugins/plugin-context/constructors/request/
- [ ] https://docs.kuzzle.io/core/2/plugins/guides/pipes/
- [ ] https://docs.kuzzle.io/core/2/plugins/guides/events/core-kuzzle-start/

- [ ] https://docs.kuzzle.io/core/2/guides/cookbooks/realtime-api/terms/


- [ ] https://docs.kuzzle.io/core/2/protocols/api/context/errors/
- [ ] https://docs.kuzzle.io/core/1/protocols/api/context/errors/
- [ ] https://docs.kuzzle.io/core/2/protocols/api/context/request/
- [ ] https://docs.kuzzle.io/core/1/protocols/api/context/request/ 


- [x] https://docs.kuzzle.io/core/2/api/controllers/auth/update-self/
- [x] https://docs.kuzzle.io/core/2/api/controllers/auth/credentials-exists/
- [x] https://docs.kuzzle.io/core/2/api/controllers/collection/update-mapping/
- [x] https://docs.kuzzle.io/core/2/api/controllers/auth/update-my-credentials/
- [x] https://docs.kuzzle.io/core/2/api/controllers/auth/get-my-rights/

PR #1806, opened 2020-10-08
Add deprecation warnings

rolljee:
## What does this PR do ?

Add deprecation messages

This PR resolves https://github.com/kuzzleio/kuzzle/issues/1661

TODO FILL THIS.


lib/api/controller/auth.js
```diff
@@ -286,6 +286,8 @@ class AuthController extends NativeController {
    * @returns {Promise<object>}
    */
   getMyRights (request) {
+    request.addDeprecation('2.4.0', 'http://kuzzle:7512/users/_me/_rights as been replaced by http://kuzzle:7512/_me/_rights');
```
Aschen, line 289: This deprecation warning will be printed every time this action is called, even with WebSocket and with the new HTTP route.

You should ensure to add the deprecation only when the deprecated route has been used.

Maybe it could be interesting to add a new property in the `httpRoutes.js` definition file and handle those deprecation in the funnel level instead of adding `if` in controller methods.

Maybe something like
```
{ verb: 'get', url: 'http://kuzzle:7512/users/_me/_rights', controller: 'auth', action: 'getMyRights', deprecated: { since: '2.4.0', message: 'http://kuzzle:7512/_me/_rights' } }
```

rolljee: That's sounds good 👍

lib/config/httpRoutes.js
```diff
@@ -25,322 +25,322 @@
...
+  { verb: 'put', path: '/profiles/:_id', controller: 'security', action: 'createOrReplaceProfile' },
+  { verb: 'put', path: '/roles/:_id', controller: 'security', action: 'createOrReplaceRole' },
+  { verb: 'put', path: '/credentials/:strategy/:_id/_update', controller: 'security', action: 'updateCredentials' },
+  { verb: 'put', path: '/profiles/:_id/_update', controller: 'security', action: 'updateProfile' },
+  { verb: 'put', path: '/roles/:_id/_update', controller: 'security', action: 'updateRole' },
+  { verb: 'put', path: '/users/:_id/_update', controller: 'security', action: 'updateUser' },
+  { verb: 'put', path: '/users/:_id/_replace', controller: 'security', action: 'replaceUser' },
+  { verb: 'put', path: '/profiles/_mapping', controller: 'security', action: 'updateProfileMapping' },
+  { verb: 'put', path: '/roles/_mapping', controller: 'security', action: 'updateRoleMapping' },
+  { verb: 'put', path: '/users/_mapping', controller: 'security', action: 'updateUserMapping' }
```
xbill82, line 343: Thank you Prettier ❤️

xbill82 (approved):

features/step_definitions/http.js
```diff
@@ -21,3 +21,13 @@ Then('I send the crafted HTTP multipart request', function (done) {
 Then('Kuzzle is still up', async function () {
   await this.api.serverPublicApi();
 });
+
+When('I call a deprecated method I should have good properties', async function() {
```
Aschen, line 25: This should be a unit test instead since no request are made to Kuzzle API

Aschen (changes requested):

features-sdk/step_definitions/controllers-steps.js
```diff
@@ -165,3 +166,41 @@ Then('I got an error with id {string}', function (id) {
 
   assert(this.props.error.id === id, `Expected error to have id "${id}", but got "${this.props.error.id}"`);
 });
+
+When('I call the deprecated publicApi method', async function () {
+  if (process.env.KUZZLE_PROTOCOL !== 'http') {
```
Aschen, line 172: You can use the `@realtime` annotation at the top of your feature definition to run it only with websocket and mqtt ([see here](https://github.com/kuzzleio/kuzzle/blob/master/features-sdk/Application.feature#L21))

features-sdk/step_definitions/controllers-steps.js
```diff
@@ -165,3 +166,41 @@ Then('I got an error with id {string}', function (id) {
...
+    return;
+  }
+
+  if (process.env.NODE_ENV !== 'development') {
```
Aschen, line 176: It would be better to add a `@development` annotation for this with the corresponding [hook](https://github.com/kuzzleio/kuzzle/blob/master/features-sdk/support/hooks.js)

features-sdk/step_definitions/controllers-steps.js
```diff
@@ -165,3 +166,41 @@ Then('I got an error with id {string}', function (id) {
 
   assert(this.props.error.id === id, `Expected error to have id "${id}", but got "${this.props.error.id}"`);
 });
+
+When('I call the deprecated publicApi method', async function () {
```
Aschen, line 171: This step should use the generic [API action step](https://github.com/kuzzleio/kuzzle/blob/master/features-sdk/step_definitions/controllers-steps.js#L7)
You can modify this step to also store the entire `response` object in the `this.props` object

features-sdk/step_definitions/controllers-steps.js
```diff
@@ -165,3 +166,41 @@ Then('I got an error with id {string}', function (id) {
...
+
+  const response = await this.sdk.query({
+    controller: 'server',
+    action: 'publicApi'
+  });
+
+  this.props.deprecations = response.deprecations;
+});
+
+Then('I should receive a deprecation notice', async function () {
```
Aschen, line 188: This step should be converted to a more generic step that allow to match the `response`

You could modify [this step](https://github.com/kuzzleio/kuzzle/blob/master/features-sdk/step_definitions/controllers-steps.js#L56) to allows match an array of object either in the `this.props.response` or in the `this.props.result` object.

The regex could be something like `/I should receive a ("(.*?)" )?array (of objects )?in the "(.*?)" matching:/`

```
Then I should receive a "deprecations" array of objects in the "response" matching:
```

features-sdk/step_definitions/controllers-steps.js
```diff
@@ -165,3 +166,41 @@ Then('I got an error with id {string}', function (id) {
...
+
+  if (process.env.NODE_ENV !== 'development') {
+    return;
+  }
+
+  should(this.props.deprecations).deepEqual([{ 'version': '2.5.0', 'message': 'http://kuzzle:7512/_openapi' }]);
+  this.props.deprecations = undefined;
+});
+
+Then('I shouldn\'t receive a deprecation notice', async function () {
```
Aschen, line 201: This step should also be more generic, something like
```
Then The response should contains a "deprecations" property equals to "undefined"
```

test/core/network/httpRouter/httpRouter.test.js
```diff
@@ -623,5 +624,20 @@ describe('core/network/httpRouter', () => {
         }
       });
     });
+
+    it('deprecated method should have good properties', done => {
```
Aschen, line 628: You don't need to use the `done` method I think. The test will pass if the function does not throw

Aschen, line 628: Nitpicking
```suggestion
  it('should ensure that deprecated routes have the correct properties', done => {
```

test/core/network/router/httpRequest.test.js
```diff
@@ -33,6 +35,21 @@ describe('Test: router.httpRequest', () => {
       {url: '', method: '', headers: {}});
   });
 
+  it('deprecated method should have good properties', done => {
```
Aschen, line 38: Why this test is duplicated here?

Aschen (changes requested):

features-sdk/step_definitions/controllers-steps.js
rolljee: I only want it to run when protocol is "http"

rolljee: But i will create an @http then like the @development

test/core/network/router/httpRequest.test.js
rolljee: It shouldn't 👍

features-sdk/step_definitions/controllers-steps.js
Aschen: Ok but I don't think you need to create any extra steps since it's already handled in cucumber profiles
See:
 - https://github.com/kuzzleio/kuzzle/blob/master/package.json#L23
 - https://github.com/kuzzleio/kuzzle/blob/master/cucumber.js#L28

Aschen (commented):
About the messages, I think it could be better to have human readable messages so SDK can display them directly for example.
```
deprecated: { since: '2.4.0', message: 'Use this route instead: http://kuzzle:7512/_me' } }
```

Aschen (approved):

PR #1808, opened 2020-10-09
Replace nodemon with a custom reloader

scottinet:
# Description

We currently use [nodemon](https://www.npmjs.com/package/nodemon) to force Kuzzle instances to restart whenever a file is changed, during development.

But nodemon has one infuriating caveat: it doesn't check if the spawned process is down before restarting another one. Which leads to constant errors due to the node debug port being still used when restarting a new instance, mainly because of Kuzzle's graceful shutdown (it doesn't stop immediately, it first processes remaining requests before shutting down):

Screenshot from 2020-10-14 15-23-50

I browsed NPM for another module, didn't find one that satisfied me, so... I wrote a tiny one, fit for our needs. It doesn't do much: it monitors files, and if one changes, it stops Kuzzle. If Kuzzle doesn't stop before a reasonable delay, it kills it. And once the process is really down, it restarts a new one. Simple.

# About nodemon in docker images

Don't remove nodemon from docker images just yet. I made several tests using this custom reloader, and it seems to work just fine in various situations. But there might be issues I didn't foresee on various environments, or various situations, so we should keep nodemon in the docker images for now, just in case, until we're all satisfied with it.

### DIFF
--- a/docker/scripts/reloader.js
+++ b/docker/scripts/reloader.js
@@ -35,7 +35,7 @@ const config = {
     'lib',
     'plugins',
     'node_modules',
-    'index.js',
+    'index.*s',
     'docker/scripts/*.ts',
   ],
 };
--- a/docker/scripts/reloader.js
+++ b/docker/scripts/reloader.js
@@ -79,16 +79,13 @@ function startProcess () {
 
     console.log(clc.red(`[RELOADER] Process exited with ${msg}. Waiting for a file change to restart it.`));
     childProcess = null;
+    state = stateEnum.STOPPED;
   });
 
   state = stateEnum.RUNNING;
 }
 
 async function stopProcess () {
-  if (childProcess === null) {
-    state = stateEnum.STOPPED;
-  }
-
   if (state !== stateEnum.RUNNING) {
     return;
   }
--- a/docker/scripts/reloader.js
+++ b/docker/scripts/reloader.js
@@ -1,0 +1,137 @@
+/*
+ * For development purposes only: this script is meant to be run on top
+ * of a Kuzzle starter script. It watches source files (json, ts, js) from
+ * Kuzzle core, plugins, configuration files.
+ * When a file changes, it automatically shuts the current process down, WAIT
+ * until it has correctly been stopped (or kills it after some delay), and only
+ * then restarts it.
+ *
+ * This script is meant to replace solutions like nodemon, which have a few
+ * cumbersome caveats.
+ *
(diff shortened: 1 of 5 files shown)


Aschen:
About nodemon in Docker, we have to keep it inside the images now because otherwise it would be a breaking change (like pm2)

docker/scripts/reloader.json
```diff
@@ -0,0 +1,8 @@
+{
```
Aschen, line 1: Can't we have a default config according to our needs instead of having a config file?

rolljee: I agree path may differ among projects

scottinet: The logic behind is two-fold:
* I didn't want to parse (too much) the command-line, I just a quick&forget feature to get rid of that infuriating pile of garbage that is nodemon
* this makes more apparent that a configuration exists and that this tool can be configured, think of this as a documentation of a sort

But if you think that it's far better to remove the configuration file, I can of course do it, I guess.

scottinet: I removed the configuration file. If we need this tool to be configurable, we can work on it later.

While I was at it, I added a watcher on typescript files in `docker/scripts` so changes on the functional-test script or on the starter one now trigger a restart.

docker/scripts/reloader.js
```diff
@@ -0,0 +1,129 @@
...
+const clc = require('cli-color');
+
+const config = {
+  cwd: '/var/app',
+  killDelay: 5000,
+  watch: [
+    'lib',
+    'plugins',
+    'node_modules',
+    'index.js',
```
Aschen, line 38: It should watch `index.ts` as well
```suggestion
    'index.*s',
```

Aschen (approved):

docker/scripts/reloader.js
scottinet: why should we restart kuzzle on changes made to index.ts?
it isn't used until recompiled, isn't it?

Aschen: Since with use Node with the `ts-node` package, Typescript files are compiled and interpreted on the fly in the memory so only `index.ts` is really used in development mode

scottinet: ohh, I didn't know that!

Adding this file to the watched list right away :+1:

xbill82:
Awesome job @scottinet maybe in the future we can distribute this loader as a separate NPM module and make it available to the community :)

xbill82 (approved):

PR #1818, opened 2020-10-15
Allow profile restrictions on Elasticsearch aliases

### LINKED ISSUE #1807: Cannot use aliases with uppercases in profile restrictions
# Expected Behavior

Kuzzle should be able to use ES aliases seamlessly in profile restrictions.

## Current Behavior

ES aliases don't have the same restrictions as normal indexes. Namely, uppercases are allowed in aliases, but they are not in indexes.

Kuzzle uses the same set of rules when checking whether index/collection names can be used in profile restrictions. This makes aliases containing uppercase letters unusable with Kuzzle's access control rights.

## Possible Solution

2 possible solutions:
* don't check name validity in profile restrictions. Worse that can happen: rights will be enforced on unreachable indexes or collections, which is not a big deal
* change the checks made when restricting profiles to something like this: "if the index or collection is known to the storage module cache, then allow it, then check if this is a valid index name". Since the storage module loads aliases, it'll allow known aliases in the profile restrictions, even if they contain uppercase letters

Solution 2 seems safer to me.

## Steps to Reproduce

1. create a collection

`kourou collection:create foo bar`

2. create an alias containing an uppercase letter

`echo '{"actions":[{"add": {"index": "&foo.bar", "alias": "&foo.ohNoes"}}]}' | http  POST <ES host name>:9200/_aliases`

3. restart Kuzzle so that it knows about the newly created alias

4. create a new profile restricted to the new alias:

```
kourou security:createProfile -a _id='ohnoes' --body '{
  policies: [
    {
      roleId: "default", 
      restrictedTo: [
        { index: "foo", collections: [ "ohNoes" ] }
      ]
    }
  ]
}'
```

The following error is returned by Kuzzle: 
`[X] ErrorThe collection name "ohNoes" is invalid`

Instead, the profile should be correctly updated and its rights enforced.

## Context (Environment)

Kuzzle version 2.4.5

scottinet:
# Description

Fix #1807 

ES aliases have different naming rules than indexes. When creating, replacing or updating profiles with storage restrictions, Kuzzle verifies if index or collection names are valid. Which prevents some ES aliases to be used in profiles.

This PR changes the following:
* by default, no check is done at all on index or collection names, it is the administrators responsbility to ensure that profile rights are correct (worse that can happen: a profile can be restricted to unreachable indexes/collections)
* a new `strict` option is now available to the following routes: `security:createProfile`, `security:createOrReplaceProfile`, `security:updateProfile`. This option is off by default, but if set, it makes Kuzzle verify if provided indexes or collections already **exist**, otherwise the profile modification is rejected (this is different from the previous behavior, where we only checked that the indexes/collectins **could exist**)

# Other changes

* Fix handling of functional tests with "( not)? <action>" sentences: most of them didn't fail if the action succeeded when it should not
* As a consequence, fix a failing functional test because of a misconfiguration


features-sdk/step_definitions/global.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
```
Aschen, line 1: Nitpicking: "global" steps related to the API are in the `controller-steps.js` file. Maybe it's not explicit enough and you can rename it but I would like to avoid having too many files

features-sdk/step_definitions/security-steps.js
```diff
@@ -4,49 +4,59 @@ const _ = require('lodash');
...
-
   try {
-    this.props.result = await this.sdk.security[method + 'Role'](roleId, { controllers }, options);
+    if (!strict) {
+      this.props.result = await this.sdk.security.createProfile(profileId, {
+        policies,
+      });
+    }
+    else {
+      // using sdk.query until createProfile handles the new "strict" option
```
Aschen, line 24:
```suggestion
      // @todo replace sdk.query when createProfile will handle the new "strict" option
```

Aschen (approved):
I really like the new `tryAction` for functional tests!

features-sdk/step_definitions/global.js
scottinet: oh, I didn't found them... I'll move this step in the right file right now.

Yoann-Abbes (approved):

PR #1831, opened 2020-10-20
Add a HTTP endpoint taking a Kuzzle API request in JSON format

### LINKED ISSUE #1774: Add a HTTP endpoint taking a Kuzzle API request in JSON format
## Feature description

We should have a HTTP endpoint taking a standard Kuzzle API request (as with any other protocols such as WebSocket or MQTT)

This endpoint would take a POST request containing a Kuzzle request in JSON format:

```
curl -X POST -H "Content-Type: application/json" localhost:7512/_query --data '{"controller": "server", "action": "info"}'
```

## Example usecase

HTTP is great because you can open URL in your browser or you can cache responses according to URL or verb.

But when it comes to call an HTTP API from a server, it's more convenient to have only one thing to change between API calls: the JSON payload

Also it would allow to call custom actions that did not declare HTTP routes.

Add a new route directly in the router (like the swagger one) that receive an HTTP POST request containing the JSON payload and then execute the standard Kuzzle request

Aschen:
## What does this PR do ?

Add a HTTP route `/_query` that takes a Kuzzle API request in the JSON body and then execute it.

Fix #1774

```bash
$ curl -X POST -H  "Content-Type: application/json" "localhost:7512/_query" --data '{ "controller":"server", "action":"now" }'
```

### DIFF
--- a/doc/2/api/essentials/query-syntax/index.md
+++ b/doc/2/api/essentials/query-syntax/index.md
@@ -45,7 +45,7 @@ If the form field holds a file, then the corresponding JSON key will refer to an
 
 Kuzzle also exposes an endpoint to send queries using the standard JSON request format used by other protocols.  
 
-This makes it possible to avoid with the use of HTTP routes and to send queries via the HTTP protocol in the same way as for other protocols.
+This makes it possible to avoid the use of the REST API and to send queries via the HTTP protocol the same way as for any other protocols.
 
 This endpoint is accessible on the route `POST http://<host>:<port>/_query`:
 
--- a/doc/2/api/essentials/query-syntax/index.md
+++ b/doc/2/api/essentials/query-syntax/index.md
@@ -41,6 +41,27 @@ If the form field holds a file, then the corresponding JSON key will refer to an
 - `mimetype`: MIME type
 - `file`: file content, encoded in base64
 
+### JSON query endpoint
+
+Kuzzle also exposes an endpoint to send queries using the standard JSON request format used by other protocols.  
+
+This makes it possible to avoid with the use of HTTP routes and to send queries via the HTTP protocol in the same way as for other protocols.
+
+This endpoint is accessible on the route `POST http://<host>:<port>/_query`:
+
+```bash
+$ curl -X POST -H  "Content-Type: application/json" "http://localhost:7512/_query" --data '{
+  "controller":"server", 
+  "action":"now" 
+}'
+```
+
+The body of the query will be processed by Kuzzle as a standard query.
+
+::: warning
+This endpoint does not allow to benefit from the advantages of the cache system integrated to HTTP via URLS.
+:::
+
 ---
 
 ## Other protocols
--- a/doc/2/api/essentials/query-syntax/index.md
+++ b/doc/2/api/essentials/query-syntax/index.md
@@ -59,7 +59,7 @@ $ curl -X POST -H  "Content-Type: application/json" "http://localhost:7512/_quer
 The body of the query will be processed by Kuzzle as a standard query.
 
 ::: warning
-This endpoint does not allow to benefit from the advantages of the cache system integrated to HTTP via URLS.
+This endpoint does not allow to benefit from the advantages of the cache system integrated to HTTP via URLs.
 :::
 
 ---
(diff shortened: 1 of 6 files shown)


Yoann-Abbes:
You didnt add it to `httpRoutes.js`, is that normal?

Aschen:
> You didnt add it to `httpRoutes.js`, is that normal?

Yes because it's not a route linked to a controller/action. You can call every controller/action with this route

xbill82 (approved):

features-sdk/support/world.js
```diff
@@ -40,20 +41,26 @@ class KuzzleWorld {
...
-    }
+  parseObject(dataTable) {
+    const rawContent = dataTable.rowsHash();
+    const content = {};
 
-    const content = dataTable.rowsHash();
+    for (const [path, value] of Object.entries(rawContent)) {
+      if (value.includes('_AGO_')) {
+        // format: "_5m_AGO_"
+        const timeAgo = ms(value.split('_')[1]);
```
rolljee, line 51: ms is undefined there

features-sdk/support/world.js
```diff
@@ -40,20 +41,26 @@ class KuzzleWorld {
...
 
-    for (const key of Object.keys(content)) {
-      content[key] = JSON.parse(content[key]);
+        _.set(content, path, this.props.now - timeAgo);
+      }
+      else {
+        _.set(content, path, eval(`var o = ${value}; o`));
```
rolljee, line 56: eval can be harmful

rolljee (changes requested):

rolljee (approved):

features-sdk/support/world.js
Aschen: It's only for the tests so it's not dangerous for our clients

features-sdk/Api.feature
```diff
@@ -0,0 +1,15 @@
...
+  Scenario: Send Request to the HTTP JSON endpoint
+    Given an existing collection "nyc-open-data":"yellow-taxi"
+    When I send a HTTP "post" request with:
+      | controller | "document"                      |
+      | action     | "create"                        |
+      | index      | "nyc-open-data"                 |
+      | collection | "yellow-taxi"                   |
+      | _id        | "foobar-1"                      |
+      | body       | { "name": "Aschen", "age": 27 } |
+    And The document "foobar-1" content match:
```
scottinet, line 13: (nitpicking)

```suggestion
    Then The document "foobar-1" content match:
```

scottinet (changes requested):
Missing documentation. Though I don't know exactly where to put this one :thinking: 

Maybe [here](https://docs.kuzzle.io/core/2/api/essentials/query-syntax/), in a new paragraph?

doc/2/api/essentials/query-syntax/index.md
```diff
@@ -41,6 +41,27 @@ If the form field holds a file, then the corresponding JSON key will refer to an
 - `mimetype`: MIME type
 - `file`: file content, encoded in base64
 
+### JSON query endpoint
+
+Kuzzle also exposes an endpoint to send queries using the standard JSON request format used by other protocols.  
+
+This makes it possible to avoid with the use of HTTP routes and to send queries via the HTTP protocol in the same way as for other protocols.
```
scottinet, line 48:
```suggestion
This makes it possible to avoid the use of the REST API and to send queries via the HTTP protocol the same way as for any other protocols.
```

doc/2/api/essentials/query-syntax/index.md
```diff
@@ -41,6 +41,27 @@ If the form field holds a file, then the corresponding JSON key will refer to an
...
+$ curl -X POST -H  "Content-Type: application/json" "http://localhost:7512/_query" --data '{
+  "controller":"server", 
+  "action":"now" 
+}'
+```
+
+The body of the query will be processed by Kuzzle as a standard query.
+
+::: warning
+This endpoint does not allow to benefit from the advantages of the cache system integrated to HTTP via URLS.
```
scottinet, line 62:
```suggestion
This endpoint does not allow to benefit from the advantages of the cache system integrated to HTTP via URLs.
```

scottinet (approved):

PR #1836, opened 2020-10-23
Allow kuzzle to handle HTTP forms again

scottinet:
# Description

HTTP forms (both `application/x-www-form-urlencoded` and `multipart/form-data`) weren't accepted anymore by Kuzzle, whereas all the necessary code was there. This was because of an erroneous message validation check, which was only considering `application/json` queries.

# Other changes

There were no tests to verify that HTTP forms are accepted. Functional tests have been added to prevent regressions from occuring again.


lib/core/network/protocols/http.js
```diff
@@ -78,6 +85,30 @@ class HttpMessage {
     }
   }
 
+  /**
+   * Checks that an incoming HTTP message is well-formed
+   *
+   */
+  validate () {
```
Aschen, line 92: Where this method is called? I can't find the call anywhere :thinking:

scottinet: ask my brain :expressionless:

lib/core/network/protocols/http.js
```diff
@@ -78,7 +85,33 @@ class HttpMessage {
...
+    if (this.isEmpty()) {
+      return;
+    }
+
+    const contentType = this.headers['content-type'];
```
Aschen, line 97: Should we consider the absence of `content-type` as an error?

scottinet: no, we default it to `application/json`

Aschen (approved):

rolljee (approved):

xbill82 (approved):

PR #1839, opened 2020-10-26
Forbid overriding of native controller in applications

Aschen:
## What does this PR do ?

Checks if the controller name is the same as a native one and throw an error to avoid overriding Kuzzle native controllers.

### DIFF
--- a/lib/core/plugin/pluginsManager.js
+++ b/lib/core/plugin/pluginsManager.js
@@ -636,7 +636,7 @@ class PluginsManager {
         throw assertionError.get(
           'invalid_controller_definition',
           controller,
-          'Native controllers can not be overriden');
+          'Native controllers cannot be overriden');
       }
 
       Plugin.checkControllerDefinition(controller, definition);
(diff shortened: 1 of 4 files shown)


lib/core/plugin/pluginsManager.js
```diff
@@ -625,13 +625,20 @@ class PluginsManager {
...
 
-  _initApi (plugin) {
+  async _initApi (plugin) {
     for (const [controller, definition] of Object.entries(plugin.instance.api)) {
       debug(
         '[%s][%s] starting api controller registration',
         plugin.name,
         controller);
 
+      if (await this.kuzzle.ask('kuzzle:api:funnel:controller:isNative', controller)) {
```
Aschen, line 635: IMHO for this kind of usage it could be interesting if `kuzzle.ask` does not always returns a promise

xbill82: true, but it is also interesting (more, IMHO) that the return value of `kuzzle.ask` is always the same type.

xbill82 (approved):

lib/core/plugin/pluginsManager.js
```diff
@@ -625,13 +625,20 @@ class PluginsManager {
...
+        throw assertionError.get(
+          'invalid_controller_definition',
+          controller,
+          'Native controllers can not be overriden');
```
scottinet, line 639:
```suggestion
          'Native controllers cannot be overriden');
```

scottinet (approved):

PR #1857, opened 2020-11-03
Fix the HTTP index:mDelete endpoint

scottinet:
# Description

The HTTP API endpoint for `index:mDelete` is currently `http://<host>:<port>/_mdelete` (instead of `_mDelete`), which is inconsistent with the rest of the API.

This PR keeps the old route (it's here since the beta stages of Kuzzle 4 years ago) and deprecates it in favor of a new, fixed route.

The documentation doesn't need to be changed: it already documents the correct version.

### DIFF
--- a/lib/config/httpRoutes.js
+++ b/lib/config/httpRoutes.js
@@ -46,7 +46,7 @@ const routes = [
   { verb: 'get', path: '/validations/_scroll/:scrollId', controller: 'collection', action: 'scrollSpecifications' },
   { verb: 'get', path: '/:index/_list', controller: 'collection', action: 'list' },
 
-  { verb: 'post', path: '/admin/_resetCache', controller: 'admin', action: 'resetCache' }, 
+  { verb: 'post', path: '/admin/_resetCache', controller: 'admin', action: 'resetCache' },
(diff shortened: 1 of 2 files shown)


lib/config/httpRoutes.js
```diff
@@ -287,6 +287,8 @@ const routes = [
   { verb: 'delete', path: '/:index/:collection/_mDelete', controller: 'document', action: 'mDelete' },
 
   { verb: 'delete', path: '/:index', controller: 'index', action: 'delete' },
+  { verb: 'delete', path: '/_mDelete', controller: 'index', action: 'mDelete' },
+  // @deprecated: has a typo. To be removed in the next major version
   { verb: 'delete', path: '/_mdelete', controller: 'index', action: 'mDelete' },
```
rolljee, line 292: Since this is deprecated, can we add a deprecation notice ?

rolljee (changes requested):

jenow (approved):

lib/config/httpRoutes.js
scottinet: ah, sure, I forgot about that new feature :+1:

scottinet: Done :+1:

alexandrebouthinon (approved):

rolljee (approved):

PR #1858, opened 2020-11-03
Fix index:mDelete to make it return the list of deleted indexes

scottinet:
# Description

As the title says: calling `index:mDelete` resulted in an empty response value, instead of an expected array of deleted indexes.

This bug is live since Kuzzle@2.6.0

### DIFF
--- a/test/core/storage/clientAdapter.test.js
+++ b/test/core/storage/clientAdapter.test.js
@@ -179,17 +179,22 @@ describe('#core/storage/ClientAdapter', () => {
       }
     });
 
-    describe('#index:mDelete', () => {
+    describe.only('#index:mDelete', () => {
       it('should register an "index:mDelete" event', async () => {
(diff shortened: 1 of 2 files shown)


jenow (approved):

test/core/storage/clientAdapter.test.js
```diff
@@ -179,17 +179,22 @@ describe('#core/storage/ClientAdapter', () => {
       }
     });
 
-    describe('#index:mDelete', () => {
+    describe.only('#index:mDelete', () => {
```
alexandrebouthinon, line 182:
```suggestion
    describe('#index:mDelete', () => {
```

alexandrebouthinon (changes requested):

Yoann-Abbes (approved):

alexandrebouthinon (approved):

PR #1861, opened 2020-11-05
Integrate a more resilient cluster as a core module

scottinet:
# Description

Rewrite of the cluster architecture, with 2 goals in mind:
* fully integrate the cluster plugin into the core product
* have a faster and more resilient cluster architecture. The goal here is to make it able to withstand millions of connections

A detailed description will be written later.

Fix https://github.com/kuzzleio/kuzzle/issues/1940


lib/cluster/command.js
```diff
@@ -0,0 +1,262 @@
...
+    this.node = localNode;
+    this.server = null;
+    this.protoroot = null;
+
+    this.state = stateEnum.INITIALIZING;
+  }
+
+  async init () {
+    this.protoroot = await protobuf.load(`${__dirname}/protobuf/command.proto`);
+    this.serve();
```
Aschen, line 58: This can return a rejected promise

lib/cluster/command.js
```diff
@@ -0,0 +1,262 @@
...
+   * load across all nodes)
+   *
+   * @param  {Array.<IdCard>} nodes
+   * @return {Object|null} Returns the fetched full state, or null if no node answered
+   */
+  async getFullState (nodes) {
+    let idx = Math.floor(Math.random() * Math.floor(nodes.length));
+    let fullState = null;
+
+    for (let retries = 0; fullState === null && retries < nodes.length; retries++) {
```
Aschen, line 153: `fullState` is already initialized l.151

lib/cluster/command.js
```diff
@@ -0,0 +1,262 @@
...
+      const { id, ip } = nodes[idx];
+      const req = new Request();
+
+      req.receiveTimeout = 2000;
```
Aschen, line 157: Shouldn't we use one of the cluster config like `heartbeat` or `joinTimeout` (maybe more the heartbeat value)

lib/cluster/node.js
```diff
@@ -0,0 +1,979 @@
...
+   *
+   * @param {string} roomId
+   * @return {void}
+  */
+  async countRealtimeSubscribers (roomId) {
+    return this.fullState.countRealtimeSubscriptions(roomId);
+  }
+}
+
+function checkConfig (config) {
```
Aschen, line 963: This should be handled by the config module

lib/cluster/node.js
```diff
@@ -0,0 +1,979 @@
...
+    await this.idCardHandler.addNode(id);
+
+    global.kuzzle.log.info(`[CLUSTER] Node "${id}" joined the cluster`);
+    this.trackActivity(id, ip, nodeActivityEnum.ADDED);
+
+    if ( global.kuzzle.state === kuzzleStateEnum.NOT_ENOUGH_NODES
+      && this.countActiveNodes() >= this.config.minimumNodes
+    ) {
+      global.kuzzle.state = kuzzleStateEnum.RUNNING;
+      global.kuzzle.log.warn(`[CLUSTER] Minimum number of nodes reached (${this.countActiveNodes()}). This node is now accepting requests again.`);
```
Aschen, line 159:
```suggestion
      global.kuzzle.log.info(`[CLUSTER] Minimum number of nodes reached (${this.countActiveNodes()}). This node is now accepting requests again.`);
```

lib/cluster/node.js
```diff
@@ -0,0 +1,979 @@
...
+class ClusterNode {
+  constructor () {
+    this.config = global.kuzzle.config.cluster;
+    this.heartbeatDelay = this.config.heartbeat;
+
+    checkConfig(this.config);
+
+    this.ip = getIP.address('public', this.config.ipv6 ? 'ipv6' : 'ipv4');
+
+    this.nodeId = null;
```
Aschen, line 59: Shouldn't this property named `idCard` instead ? It was confusing to me because I thought it was the node string ID

Aschen (changes requested):

lib/cluster/command.js
scottinet: Ah, yes, when binding to the network port for instance. I'll have to separate the socket creation and its listener.

lib/cluster/command.js
scottinet: It's not like this: this will loop until the number of retries has been reached, or if the fullstate is not null anymore (see l.165)

lib/cluster/subscriber.js
```diff
@@ -0,0 +1,680 @@
...
+   * @return {void}
+   */
+  async handleDocumentNotification (message) {
+    const notification = new DocumentNotification(
+      null,
+      message.scope,
+      message.action,
+      JSON.parse(message.result));
+
+    notification.node = this.remoteNodeId;
```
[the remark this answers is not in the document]
scottinet: You're right, I made a mistake in the current implementation, as it will trigger a de-opt from V8, since the instantiated class won't have the same properties whether they are created from the `core:notifier` module or from the cluster.

As discussed, I'll harmonize the way we create those notifications whatever who instantiates them.

lib/cluster/node.js
scottinet: When the quorum is not reached, a warning is logged about it, telling that Kuzzle won't accept new requests until new nodes join the cluster.
It seems preferable to me that the log line telling that the quorum is reached once again share the same level: if only warnings and above levels of logs are kept, it allows admins to have an insight of when Kuzzle stopped accepting requests, and when it started accepting them again.

lib/cluster/command.js
scottinet: Hm, good point. 

On a cluster with very large numbers of realtime filters (so with large fullstates), we might need a dedicated timeout: `heartbeat` makes no sense here, and `joinTimeout` should be greater than the time needed to receive a single fullstate, as it's meant to limit the time for a node to join a cluster (it's used for the mutex lock).

I'll create a dedicated configuration entry.

lib/cluster/node.js
scottinet: Funny how I keep telling the same thing when reviewing PRs and I still fail to apply it to myself :sweat:

lib/cluster/subscriber.js
```diff
@@ -0,0 +1,680 @@
...
+   * @param {Object} message - decoded UserNotification protobuf message
+   * @return {void}
+   */
+  async handleUserNotification (message) {
+    const notification = new UserNotification(
+      null,
+      message.user,
+      JSON.parse(message.result));
+
+    notification.status = message.status;
```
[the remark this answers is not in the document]
scottinet: Will be harmonized too.

lib/cluster/node.js
scottinet: It IS the node string ID :thinking: 
What makes you think otherwise?

scottinet:
@Aschen > requested changes applied.

lib/cluster/subscriber.js
```diff
@@ -0,0 +1,699 @@
...
+
+    if (!await this.validateMessage(message)) {
+      return;
+    }
+
+    try {
+      await this.handlers[topic].call(this, message);
+    }
+    catch (e) {
+      // @todo can't the node evict itself to save a round of heartbeat?
```
Aschen, line 220: Can't we broadcast a node eviction message to save a round of heartbeat to other nodes?

lib/cluster/subscriber.js
```diff
@@ -0,0 +1,699 @@
...
+      && this.lastMessageId.greaterThanOrEqual(message.messageId)
+    ) {
+      return false;
+    }
+
+    this.lastMessageId = this.lastMessageId.add(1);
+
+    if (this.lastMessageId.notEquals(message.messageId)) {
+      global.kuzzle.log.error(`[FATAL] Node out-of-sync: ${message.messageId - this.lastMessageId - 1} messages lost from node ${this.remoteNodeId}`);
+      global.kuzzle.shutdown();
```
Aschen, line 682: Can't we broadcast a node eviction message to save a round of heartbeat to other nodes?

lib/cluster/subscriber.js
scottinet: yes you're right for both comments

scottinet: this will also provide insight as to why a node shuts down

scottinet: Other nodes won't have to wait for a round of heartbeat: there is a shutdown handler broadcasting a `NodeShutdown` message to all other nodes.

scottinet: @Aschen > fixed. Nodes that shut down themselves because of a malfunction now first send an eviction message, with the shutdown reason in it, for activity tracking purposes.

I didn't bother doing the same when an eviction occurs due to a network split, since by definition, 1/ the cluster cannot receive the eviction message, and 2/ every other nodes will conclude on the same eviction and will log it in their activity tracking themselves.

I also adapted the unit tests accordingly, removed the custom kuzzle mock you wrote, and replaced it with our standard one. Which shed light on a bug: when handling a dump request, I incorrectly called `kuzzle.janitor.dump` instead of `kuzzle.dump`: since the standard mock is always up-to-date, this made the bug apparent :wink:

lib/cluster/subscriber.js
scottinet: Resolved (see above)

PR #1864, opened 2020-11-10
[Doc > Code-examples > IOT > JS] publish.js snippet fix

xbill82:
Fixes the `publish.js` snippet of the IOT code-example in Javascript.
Having the `client.end()` in the `finally` block prevents the `client.publish` to succeed, since `end` and `publish` get called at the same time due to the synchronicity of the MQTT library.

Putting the `client.end()` inside the callback to `publish` allows the connection to be destroyed after the message gets published.

### DIFF
--- a/doc/2/guides/code-examples/iot/javascript/snippets/publish.js
+++ b/doc/2/guides/code-examples/iot/javascript/snippets/publish.js
@@ -13,9 +13,12 @@ try {
     body: {
       command: 'battery-report'
     }
-  }));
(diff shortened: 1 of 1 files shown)


Aschen (approved):

jenow (approved):

scottinet (approved):

scottinet:
I changed the label, as this is a document fix, not an actual kuzzle bug fix.

PR #1898, opened 2020-12-04
Add new document:upsert API action

### LINKED ISSUE #1895: Add a document:createOrUpdate route
# Document

Now this one is a bit counterintuitive. Until now, there weren't use cases for an upsert API route. Because document updates apply partial content modification to documents, it's normally not something we want to create a new document on the fly with only partial content.

However, here is a use case coming from one of our clients, which to me seems to justify the addition of that new route to Kuzzle:

> Well, we have a use case where we want to maintain a list of recommendations for a user. Those reccomendations come from different sources from a kafka topic.
>
> So when a recommendation is received we want to add it to the document for that user in our recommendations collection.
> If the customer is new, the document which matches the customer id will not exist yet. So we want to update the document by changing a specific set of recommendations OR create a new document with that user id and add the recommendation type.
We don;t really want a collection per rec type, because that means millions of documents in each of those collections.

In that specific case, it is desirable to add partial content to a document, or to create a new one on the fly if needed.

scottinet:
# Description

Fix #1895 

Add a new `document:upsert` API action, able to either apply partial changes to a document, or to create a new document on the fly.

# Query syntax

```js
{
  "index": "<index>",
  "collection": "<collection>",
  "controller": "document",
  "action": "update",
  "_id": "<documentId>",
  "body": {
    "changes": {
      // document partial changes
    },
    "default": {
      // optional: document fields to add to the changes if the document
      // is created
    }
  }
}
```

The body part of the query contains 2 objects:
* `changes` (required): the partial changes to apply to existing documents. 
* `default` (optional): optional fields to add to the new document if it doesn't already exist

So, this route does this:
* if the document already exists, this API action behaves exactly as `document:update`: the content of `body.changes` is applied to the existing document (and the content of `body.default` is ignored)
* if the document doesn't exist, a new one is created, containing both the content of `body.default` and `body.changes`, IN THAT ORDER (meaning that if a field exists both in changes and in default, the one in `changes` supersedes the default one)

# API response

This route takes the same options as `document:update`: `source` (boolean), `retryOnConflict` (number), `refresh` (string)

But it behaves a bit differently with the `source` option: by default (`source = false`), it doesn't return any `_source` property in the response.

Also, compared to `document:update`, a new property is available in the response: `_created` (boolean).
If `true`, the document didn't exist and was created. It is set to `false` otherwise.

# Metadata

The content of `_kuzzle_info` varies depending on whether the document was created or updated, matching what is currently found in documents if `document:create` or `document:update` was called, respectively.

# Realtime notifications

Realtime notifications triggered match the status of the document: if it was created, a "new document" notification is fired, otherwise a "document updated" is sent to subscribers.


doc/2/api/controllers/document/upsert/index.md
```diff
@@ -0,0 +1,107 @@
...
+- `changes`: partial changes to apply to the document
+- `default`: (optional) fields to add to the document if it gets created
+
+---
+
+## Response
+
+Returns information about the updated document:
+
+- `_created`: if `true`, a new document was created, otherwise the document existed and was updated
```
Aschen, line 83: We already have a `created` boolean for the `createOrReplace` action family. IMHO we should keep the same name to ensure consistency in the actual API. (even if I prefer `_created`)

features-sdk/DocumentController.feature
```diff
@@ -648,3 +648,48 @@ Feature: Document Controller
       | _id          | _source                                                     |
       | "document-1" | { "name": "Sylvanas Windrunner", "title": "The liberator" } |
       | "document-4" | { "name": "Sylvanas Windrunner", "title": "The liberator" } |
+
+# document:upsert ===========================================================
```
Aschen, line 652: (Nitpicking) alignment
```suggestion
  # document:upsert ============================================================
  
```

lib/service/storage/elasticsearch.js
```diff
@@ -614,6 +614,70 @@ class ElasticSearch extends Service {
...
+    collection,
+    id,
+    content,
+    { defaultValues = {}, refresh, userId=null, retryOnConflict } = {})
+  {
+    const esRequest = {
+      _source: true,
+      body: {
+        doc: content,
+        upsert: Object.assign({}, defaultValues, content),
```
Aschen, line 640: Nitpicking
```suggestion
        upsert: { ...defaultValues, ...content },
```

Aschen (changes requested):
Minor changes

jenow (approved):

doc/2/api/controllers/document/upsert/index.md
```diff
@@ -0,0 +1,107 @@
...
+Method: PUT
+Body:
+```
+
+```js
+{
+  changes: {
+    // document partial changes
+  },
+  default: {
```
rolljee, line 28: i'm personnaly not found of `default` property. IMO whole document to create if needed should be inside. Is there a specific use case for that ?

rolljee (changes requested):

doc/2/api/controllers/document/upsert/index.md
scottinet: If we want to create the whole document, there is already a `document:createOrReplace` API action available (or `document:replace` if we know that the document already exists).

The use case for an upsert API action has been explained in the PR description. 

As for the `default` object, the reasoning behind this option is this: what happens if we need to update non-required fields, but the document needs to have required values because of business rules or because of strict mappings?
We might need to also provide a few required fields along with the document. And since we don't know if the document exists or not, those values will probably be default ones, meaning that we DON'T want to update them if these fields already exist.

Thus I divided the request payload into two parts: the fields we want to update and, optionally, a set of default fields values that will only be used if the document needs to be created.

Does that answer your question?

doc/2/api/controllers/document/upsert/index.md
scottinet: Dang. I didn't think about checking the response returned by createOrReplace, it's only through luck that we return the same `created` value (for once this is not intentional :sweat_smile: )

So, for the sake of homogeneity, we indeed need to rename it... It's a shame, I also prefer `_created` over `created.

doc/2/api/controllers/document/upsert/index.md
rolljee: It does indeed.

> As for the default object, the reasoning behind this option is this: what happens if we need to update non-required fields, but the document needs to have required values because of business rules or because of strict mappings?

I was mitigated because this mode of operation differs from upstert methods in general. (eg: Mongo, etc.), but it is true that the properties that are required justify the choice made.

rolljee (approved):

doc/2/api/controllers/document/upsert/index.md
scottinet: That's probably because traditional upserts work with complete documents (as with our own "createOrReplace")

I don't know about mongo, but that's the case with traditional DBs :wink: 

Our update action takes only partial changes to apply to documents, and until now, we didn't have an "updateOrReplace" action because there weren't proper use cases for it. That kind of action is a bit rare and counterintuitive at first, and it's only because one of our clients needs it that I added it to our API.

doc/2/api/controllers/document/upsert/index.md
scottinet: Response property renamed.

scottinet:
@jenow @rolljee > I have re-requested your review: I forgot to add that new route to generic document events. This is fixed (+ appropriate functional tests have been added)

doc/2/api/controllers/document/upsert/index.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+### Other protocols
+
+```js
+{
+  "index": "<index>",
+  "collection": "<collection>",
+  "controller": "document",
+  "action": "update",
```
Aschen, line 42:
```suggestion
  "action": "upsert",
```

Aschen (approved):

rolljee (approved):

jenow (approved):

PR #1907, opened 2020-12-10
Allows BackendLogger before startup

Aschen:
## What does this PR do ?

Allows to log message before application startup

### Other changes
 
  - terminate the childProcess of the reloaded.js script when the main script is terminated (e.g. CTRL+C)

### DIFF
--- a/lib/core/application/backend.ts
+++ b/lib/core/application/backend.ts
@@ -385,10 +385,11 @@ class InternalLogger extends ApplicationManager implements InternalLogger {
 
   private _log (level: string, message: any) {
     if (! this._application.started) {
-      throw runtimeError.get('unavailable_before_start', 'log');
+      console.log(util.inspect(message));
     }
-
-    this._kuzzle.log[level](util.inspect(message));
+    else {
+      this._kuzzle.log[level](util.inspect(message));
+    }
   }
 }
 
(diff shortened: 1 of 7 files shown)


xbill82 (approved):

jenow (approved):

lib/core/application/backend.ts
```diff
@@ -385,10 +385,12 @@ class InternalLogger extends ApplicationManager implements InternalLogger {
 
   private _log (level: string, message: any) {
     if (! this._application.started) {
-      throw runtimeError.get('unavailable_before_start', 'log');
+      // eslint-disable-next-line no-console
+      console.log(util.inspect(message));
+    }
+    else {
+      this._kuzzle.log[level](util.inspect(message));
     }
```
scottinet, line 393: Unnecessary: kuzzle's `log` function already does that.
See https://github.com/kuzzleio/kuzzle/blob/master/lib/kuzzle/log.js

scottinet (changes requested):

lib/core/application/backend.ts
Aschen: Yes but Kuzzle is not even instantiated before calling `Backend.start`, and I can't instantiate Kuzzle before that because Kuzzle take the configuration in the constructor so the custom configuration will not be loaded correctly:

```
// Kuzzle is instantiated when Backend is instantiated so we can use the logger
const app = new Backend('test')

app.log.info('it works')

// Kuzzle has already have been instantiated with the configuration
app.config.set('server.port', 4242)

// Kuzzle will listen on 7512
app.start()
```

scottinet: ah, yes, of course :sweat_smile:

scottinet (approved):

PR #1908, opened 2020-12-10
Expose mutex class to plugins

scottinet:
# Description

Backports the `Mutex` class from Kuzzle v2 and exposes it to plugins, allowing them to safely lock resources and prevents concurrent accesses in cluster environments.

### DIFF
--- a/doc/1/plugins/plugin-context/constructors/mutex/index.md
+++ b/doc/1/plugins/plugin-context/constructors/mutex/index.md
@@ -10,7 +10,7 @@ title: Mutex
 
 Instantiates a new mutex, that can be used to lock a resource. 
 
-Allow to have a process played on only 1 node on a Kuzzle cluster, with the other ones waiting until the lock is freed.
+Allows to have a process played on only 1 node on a Kuzzle cluster, with the other ones waiting until the lock is freed.
 
 Also works within a single node: in that case, it's better if lock attempts from different parts of the code use different Mutex instances (with the same lock ID).
 
--- a/doc/1/plugins/plugin-context/constructors/mutex/index.md
+++ b/doc/1/plugins/plugin-context/constructors/mutex/index.md
@@ -1,0 +1,87 @@
+---
+code: true
+type: page
+title: Mutex
+---
+
+# Mutex
+
+<SinceBadge version="1.12.0" />
+
+Instantiates a new mutex, that can be used to lock a resource. 
+
+Allow to have a process played on only 1 node on a Kuzzle cluster, with the other ones waiting until the lock is freed.
+
+Also works within a single node: in that case, it's better if lock attempts from different parts of the code use different Mutex instances (with the same lock ID).
+
+---
+
+## Constructor
+
+```js
+new context.constructors.Mutex(lockId, [options]);
+```
+
+<br/>
+
+| Arguments           | Type              | Description                                                                                                                     |
+| ------------------- | ----------------- | ------------------------------------------------------------------------------------------------------------------------------- |
+| `lockId`            | <pre>string</pre> | Lock unique id (must be identical across all nodes)  |
+| `options`           | <pre>object</pre> | (optional) Mutex options (see below) |
+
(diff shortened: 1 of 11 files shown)


test/services/internalEngine/bootstrap.test.js
```diff
@@ -397,7 +397,7 @@ describe('services/internalEngine/bootstrap.js', () => {
   describe('#_waitTillUnlocked', () => {
     it('should throw if locked for too long', () => {
       bootstrap.db.exists.returns(Bluebird.resolve(true));
-      mockrequire('bluebird', Object.assign(require('bluebird'), {delay: () => Bluebird.resolve()}));
+      mockrequire('bluebird', Object.assign({}, require('bluebird'), {delay: () => Bluebird.resolve()}));
```
scottinet, line 400: That one bug took me so much time to find: it prevented unit tests on the new mutex class to work properly, because Bluebird.delay was overwritten by this Object.assign command, impacting the require cache for all following tests. :expressionless:

Aschen (approved):
Great :+1:

lib/util/mutex.js
```diff
@@ -0,0 +1,165 @@
...
+
+    this._lockCoroutine = Bluebird.coroutine(function * lockfn () {
+      let duration = 0;
+
+      do {
+        this.locked = yield this.kuzzle.services.list.internalCache.set(
+          this.lockId,
+          this.mutexId,
+          'NX',
+          'PX',
```
Yoann-Abbes, line 96: What are those arguments for?

test/util/mutex.test.js
```diff
@@ -0,0 +1,186 @@
...
+            123);
+        });
+    });
+
+    it('should use random values for different mutex instances', () => {
+      const mutex1 = new Mutex(kuzzle, 'foo');
+      const mutex2 = new Mutex(kuzzle, 'foo');
+
+      should(mutex1.mutexId).not.eql(mutex2.mutexId);
+    });
```
Yoann-Abbes, line 46: Is this test relevant with only 2 mutex?

Aschen: IMHO it's enough to test with 2 different mutex since the goal is to ensure a random value is correctly added.

Maybe if we create one billion of mutex we may have collision but still I doubt it

doc/1/plugins/plugin-context/constructors/mutex/index.md
```diff
@@ -0,0 +1,87 @@
...
+title: Mutex
+---
+
+# Mutex
+
+<SinceBadge version="1.12.0" />
+
+Instantiates a new mutex, that can be used to lock a resource. 
+
+Allow to have a process played on only 1 node on a Kuzzle cluster, with the other ones waiting until the lock is freed.
```
xbill82, line 13:
```suggestion
Allow to have a process played on only 1 node on a Kuzzle cluster, with the other ones waiting until the lock is freed.
```
```suggestion
Allows to have a process played on only 1 node on a Kuzzle cluster, with the other ones waiting until the lock is freed.
```

alexandrebouthinon (approved):

lib/util/mutex.js
```diff
@@ -0,0 +1,165 @@
...
+ * Allow to have a process played on only 1 node, with the other ones waiting
+ * until the lock is freed.
+ * Also works within a single node: in that case, it's better
+ * if lock attempts from different parts of the code use different Mutex
+ * instances: each mutex will lock using a unique ID, and the lock can only
+ * be freed by that mutex instance, and it alone.
+ *
+ * /!\ Meant to be used only with 1 independant redis client (single node or
+ * cluster). Which is what Kuzzle supports today.
+ * If, in the future, Kuzzle is able to support multiple independant
```
xbill82, line 61:
```suggestion
 * If, in the future, Kuzzle is able to support multiple independent
```

xbill82 (approved):
small nitpicking

PR #1911, opened 2020-12-15
Make kuzzle instance a singleton

scottinet:
# Description

kuzzle_everywhere

Kuzzle instances are now singletons. And they are put in the global space.
This PR removes usages of `this.kuzzle` from everywhere, removing a lot of clutter.

# Why?

I need to expose the mutex class to plugins and applications. To do this properly, I need to convert the mutex object to TS.
But... the mutex object needs the Kuzzle object. And we don't want plugins/apps user to provide one, obviously.

So, I had 2 possibilities: write yet another TS wrapper that will inject the kuzzle object to the mutex object. Or get rid of that kuzzle object once and for all, simplifying the code.

Since having `this.kuzzle` everywhere was a pain point, I decided for the latter solution (mutex for plugins/apps will come in another PR)

# Other changes

Fix a bug in the graceful shutdown, where the number of remaining requests was never refreshed, thus preventing Kuzzle to shut down properly if there were pending requests.


Aschen (commented):
Just a remark about the `global.kuzzle` property visibility: should we use `Reflect.defineProperty` to hide it from enumeration and `console.log` ?

scottinet:
@Aschen > Sure. And done. :)

lib/kuzzle/kuzzle.js
```diff
@@ -55,47 +54,53 @@ const InternalIndexHandler = require('./internalIndexHandler');
...
+    }
+
     super(
       config.plugins.common.maxConcurrentPipes,
       config.plugins.common.pipesBufferSize);
 
+    Reflect.defineProperty(global, 'kuzzle', {
+      value: this,
+    });
+
```
Yoann-Abbes, line 68: I don't understand why a `global` level is needed.
Can't we get rid of it and simply write `kuzzle.something` instead of `global.kuzzle.something` ?
With the `global` It feels like there are others globals
Sorry it its a naive question

scottinet: I don't get it. Where do you want to put it?

I had another approach with a static instance getter. It also works, and it only needs a `require` from files needing it. This approach has two caveats: it's really inconvenient for our private plugins, and it's a pain in the a** to mock (I succeeded only by monkey patching mock-require :expressionless: )

But apart from this other solution, I don't see where I can store that variable.

scottinet:
The CI passed. Only sonarcloud didn't run because the token was changed and travis wasn't updated.

Aschen (approved):

Yoann-Abbes (approved):

rolljee (approved):

PR #1919, opened 2020-12-22
Add node id to response and logs

Aschen:
## What does this PR do ?

Add the node unique identifier to the response:
 - in the response payload
 - in the response headers (HTTP only)

Also add it to the log lines.


lib/api/request/requestResponse.ts
```diff
@@ -42,6 +42,8 @@ export class Headers {
       get: (target, name) => this.getHeader(name as string),
       set: (target, name, value) => this.setHeader(name as string, value),
     });
+
+    this.setHeader('X-Kuzzle-Node', (global as any).kuzzle.id);
```
rolljee, line 46: Can't this be as string ?

lib/api/request/requestResponse.ts
```diff
@@ -256,6 +258,12 @@ export class RequestResponse {
     this[_request].setResult(r);
   }
 
+  /**
+   * Node identifier
+   */
+  get node (): string {
```
rolljee, line 264: Can't this be as string ?

rolljee (changes requested):

lib/api/request/requestResponse.ts
Aschen: `kuzzle.id` is a string but I have to cast the `global` because the `kuzzle` property is not defined on it

lib/api/request/requestResponse.ts
Aschen: Same answer, the cast is for the type of `global`, not for `kuzzle.id`

rolljee (approved):

xbill82 (approved):
If you know what you're doing when monkeypatching MockRequire, I am OK with it.

Aschen:
> If you know what you're doing when monkeypatching MockRequire, I am OK with it.

Hmm actually this commit was added by @scottinet :thinking:

lib/api/request/requestResponse.ts
scottinet: This won't be necessary once #1913 gets merged, `global.kuzzle` will be accessible directly from TS scripts.

lib/kuzzle/kuzzle.js
```diff
@@ -42,11 +40,16 @@ const Logger = require('./log');
...
+const Funnel = require('../api/funnel');
+const EntryPoint = require('../core/network/entryPoint');
 const CacheEngine = require('../core/cache/cacheEngine');
 const StorageEngine = require('../core/storage/storageEngine');
 const SecurityModule = require('../core/security');
 const RealtimeModule = require('../core/realtime');
-const InternalIndexHandler = require('./internalIndexHandler');
+
+// kuzzle instance singleton
+global.kuzzle = null;
```
scottinet, line 52: @Aschen > this is now obsolete. Could you merge #1913's branch on this one, and mark this PR as dependent of it?

test/api/request/request.test.js
```diff
@@ -16,6 +16,10 @@ describe('#Request', () => {
   let nodeEnv;
 
   beforeEach(() => {
+    global.kuzzle = {
+      id: 'nasty-author-4242'
+    };
```
scottinet, line 21: With #1913, you'll have to delete `global.kuzzle` first, with `Reflect.deleteProperty`, otherwise this line will silently fail, and your tests won't behave correctly.

test/api/request/requestResponse.test.js
```diff
@@ -10,6 +10,10 @@ describe('#RequestResponse', () => {
   let req;
 
   beforeEach(() => {
+    global.kuzzle = {
+      id: 'nasty-author-4242'
+    };
```
scottinet, line 15: ditto

test/kuzzle/log.test.js
```diff
@@ -0,0 +1,40 @@
...
+    global.kuzzle = {
+      once: sinon.stub(),
+      id: 'nasty-author-4242',
+      emit: sinon.stub(),
+      asyncStore: {
+        exists: () => false,
+        has: () => true,
+        get: () => ({ id: 'request-unique-id' })
+      }
+    };
```
scottinet, line 21: (nitpicking) ...or you could simply instantiate a new KuzzleMock object, it should take care of everything (apart from the asyncStore, I don't recall having that mocked)

test/mock-kuzzle.js
```diff
@@ -0,0 +1,21 @@
...
+
+mockRequire.stopAll = () => {
+  savedStopAll();
+  mockKuzzle();
+};
+
+module.exports = mockKuzzle;
+
+mockKuzzle();
+
```
scottinet, line 21: This looks like a leftover from my first attempt to globalize the kuzzle object. This file has never been pushed to 2-dev since we now use `global.kuzzle` :thinking:

scottinet (changes requested):

scottinet (approved):

PR #1925, opened 2021-01-06
Fix OpenApi Format + Implement route testing

### LINKED ISSUE #1910: Exposed OpenApi is not valid
## Expected Behavior
The **OpenApi** exposed by Kuzzle should be correctly formatted.

## Current Behavior
Some structural errors can be found 

## Steps to Reproduce

1. Get Kuzzle's OpenApi result in yaml format. (http://localhost:7512/_openapi[?format=yaml])
2. Copy/Paste into [Swagger Editor](https://editor.swagger.io/)

image

Leodau:
## What does this PR do ?
Fixes two issues in our **OpenApi** Format:

- Seems like there was a breaking change the moment we implemented the `/_/`prefix on our plugin routes. Our generator should not add any prefixes by itself. `/_/_plugin/` & `/_plugin/_plugin/` were found.
- Some plugin **verbs** were getting translated in all caps. Make sure every verb is lowercase.

1. Get Kuzzle's OpenApi result in any format. (http://localhost:7512/_openapi[?format=yaml])
2. Copy/Paste into [Swagger Editor](https://editor.swagger.io/)

You should not have any structural issues reported by the editor & no  `/_/_plugin/` || `/_plugin/_plugin/` routes should be found.

### Other changes

- Took the opportunity to **test** our **openApi** route :rescue_worker_helmet: 
- Took the opportunity to add some unit **tests** to our **openApiGenerator** :rescue_worker_helmet: 
- :warning: Changed our **swagger.js** generator file to the more openApi related **openApiGenerator.js** If you need info about why: [SwaggerVsOpenApi](https://swagger.io/blog/api-strategy/difference-between-swagger-and-openapi/)

Closes #1910 Exposed OpenApi is not valid


lib/api/swagger.js
```diff
@@ -34,12 +34,7 @@ module.exports = function generateSwagger () {
   const routes = [];
 
   global.kuzzle.config.http.routes.forEach(_route => routes.push(_.assign({}, _route)));
-
-  global.kuzzle.pluginsManager.routes.forEach(_route => {
-    const route = _.assign({}, _route);
-    route.path = `/_plugin${route.path}`;
-    routes.push(route);
-  });
+  global.kuzzle.pluginsManager.routes.forEach(_route => routes.push(_.assign({}, _route)));
```
Aschen, line 37: Nitpicking: we should avoid lodash when we can
```suggestion
  global.kuzzle.pluginsManager.routes.forEach(_route => routes.push({ ..._route });
```

Aschen (approved):

scottinet:
I see no change in test units, and they seem to pass despite the modifications, meaning that there are some tests missing to verify that this bug doesn't appear again.

Aschen (approved):

Leodau:
> I see no change in test units, and they seem to pass despite the modifications, meaning that there are some tests missing to verify that this bug doesn't appear again.

As discussed earlier,  atm we do not perform any testing on our **openApi** generator. 
If we do so, wouldn't this mean that we would be checking/validating for an external _api/format_?

I figured, we could check the validity of our final **openApi** format with external third-party modules, but internal bugs like having a badly formatted route would mean that we have to somehow test every route generated to see if they actually correspond to the testing backend.

Is this what you had in mind? @scottinet

Shiranuit (approved):

Aschen:
@scottinet I agree with @Leodau, writing unit tests for the Open API format it's just validating the specification so IMHO it's overkill.

scottinet:
I... still disagree, sorry. The bugs solved by this PR are about how we generate the OpenAPI format, not the format itself, which should indeed be taken for granted.

It seems to me that adding a couple tests verifying that plugin and native API actions are correctly generated is not a lot to ask.
We should simply make sure that we generate the OpenAPI format correctly, to prevent such inconveniences in the future, if we have to make changes in this part of the code.

It's exactly why we test our own API routes: we take the response formats for granted (as per our own documentation), and we do not question them, we simply check that our API actions return what is expected from them.

xbill82 (changes requested):
Agree with @scottinet - missing tests... Sorry :(

Leodau:
Ok so, I'll update/change the **scope** of this PR to require as much kuzzle-side testing as possible with no direct OpenAPI format validation.

features/step_definitions/controllers-steps.js
```diff
@@ -180,6 +180,10 @@ Then('The response should contains a {string} equals to undefined', async functi
   should(this.props.response[key]).equal(undefined);
 });
 
+Then('The property {string} of the response should be equal to: {string}', async function (key, value) {
```
Aschen, line 183: This step could be adapted to be more generic like [the one matching the result](https://github.com/kuzzleio/kuzzle/blob/master/features/step_definitions/controllers-steps.js#L87)

features/step_definitions/controllers-steps.js
```diff
@@ -92,6 +92,14 @@ Then('I should receive a result matching:', function (dataTable) {
   should(this.props.result).matchObject(expectedResult);
 });
 
+Then('I should receive a response matching:', function (dataTable) {
+  const expectedResult = this.parseObject(dataTable);
+
+  should(this.props.response).not.be.undefined();
```
Aschen, line 98: Nitpicking: this will generate a better error message instead of `expected undefined to not be undefined`
```suggestion
  should(this.props).have.property('response');
```

Aschen (approved):

xbill82 (approved):

scottinet (approved):

PR #1926, opened 2021-01-06
Add cluster-wide events to applications/plugins

scottinet:
# Description

Allows applications or plugins to emit and receive cluster-wide events.

# Usage

```js
app.cluster.on('some:event', payload => { /* ... */ });

await app.cluster.broadcast('some:event', { some: 'payload' });
```

Apart from `broadcast`, a bunch of standard event emitter functions are available: `on`, `once`, `off` and `removeAllListeners`
Wildcards are not supported.

# Other changes

* Functional tests are now run on a cluster environment. This was necessary to properly test this new feature.
* The `EventHandler` type now also accepts functions that do not return promises. Useful for hooks and cluster-wide events


rolljee (approved):

lib/core/application/backendCluster.ts
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * @param  {string}        event
+   * @param  {JSONObject}    payload
+   * @return {Promise<void>}
+   */
+  async broadcast (event: string, payload: JSONObject): Promise<void> {
+    // @deprecated - to be removed with the new cluster (that event will
+    // always have an answerer)
+    if (!global.kuzzle.hasAskAnswerer('cluster:event:broadcast')) {
+      throw kerror.get('core', 'fatal', 'assertion_failed', 'Cannot use cluster methods in a single-node environment');
```
Aschen, line 38: Can't we have a simple return if Kuzzle is in single-node mode?

Otherwise developers need to check manually if Kuzzle is deployed as a cluster or not (and also there is no clean way to do this)

doc/2/framework/classes/backend-cluster/once/index.md
```diff
@@ -0,0 +1,34 @@
...
+title: once
+description: BackendCluster.once method
+---
+
+# `once()`
+
+<SinceBadge version="auto-version" />
+<CustomBadge type="error" text="Experimental: non-backward compatible changes or removal may occur in any future release."/>
+
+Listens to event emitted with [BackendCluster.broadcast](/core/2/framework/classes/backend-cluster/broadcast).
```
xbill82, line 13:
```suggestion
Listens to an event emitted with [BackendCluster.broadcast](/core/2/framework/classes/backend-cluster/broadcast).
```

lib/core/application/backendCluster.ts
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * @param  {string}        event
+   * @param  {EventHandler}      listener
+   * @return {Promise<void>}       [description]
+   */
+  async on (event: string, listener: EventHandler): Promise<void> {
+    // @deprecated - to be removed with the new cluster (that event will
+    // always have an answerer)
+    if (!global.kuzzle.hasAskAnswerer('cluster:event:on')) {
+      throw kerror.get('core', 'fatal', 'assertion_failed', 'Cannot use cluster methods in a single-node environment');
```
xbill82, line 55: Agree with @Aschen on the above and this one.

lib/core/application/backendCluster.ts
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * @param  {string}        event
+   * @param  {EventHandler}      listener
+   * @return {Promise<void>}
+   */
+  async once (event: string, listener: EventHandler): Promise<void> {
+    // @deprecated - to be removed with the new cluster (that event will
+    // always have an answerer)
+    if (!global.kuzzle.hasAskAnswerer('cluster:event:once')) {
+      throw kerror.get('core', 'fatal', 'assertion_failed', 'Cannot use cluster methods in a single-node environment');
```
xbill82, line 73: same

lib/core/application/backendCluster.ts
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * @param  {string}        event
+   * @param  {EventHandler}      listener
+   * @return {Promise<void>}
+   */
+  async off (event: string, listener: EventHandler): Promise<void> {
+    // @deprecated - to be removed with the new cluster (that event will
+    // always have an answerer)
+    if (!global.kuzzle.hasAskAnswerer('cluster:event:off')) {
+      throw kerror.get('core', 'fatal', 'assertion_failed', 'Cannot use cluster methods in a single-node environment');
```
xbill82, line 92: same

lib/core/application/backendCluster.ts
```diff
@@ -0,0 +1,113 @@
...
+   * Removes all listeners from an event.
+   *
+   * @param  {string}        event
+   * @return {Promise<void>}
+   */
+  async removeAllListeners (event: string): Promise<void> {
+    // @deprecated - to be removed with the new cluster (that event will
+    // always have an answerer)
+    if (!global.kuzzle.hasAskAnswerer('cluster:event:removeAllListeners')) {
+      throw kerror.get('core', 'fatal', 'assertion_failed', 'Cannot use cluster methods in a single-node environment');
```
xbill82, line 108: same

xbill82 (approved):
Nitpickings

lib/core/application/backendCluster.ts
scottinet: This makes sense, since this doesn't change the addon behavior in any way.

I'll make that change right away.

scottinet:
@Aschen @xbill82 > change applied, the cluster methods are now doing nothing in single node environments.

scottinet:
Putting this PR back to draft: we get random error on functional tests, probably because they are now run on a cluster env. There seems to be a race condition somewhere in our tests, probably in the hooks...

scottinet:
Ok it was only the index bootstrap timeout which was a bit too short, I've extended it from 5s to 30s

Aschen (approved):

PR #1928, opened 2021-01-07
Add document:deleteFields

### LINKED ISSUE #1773: Add a document:deleteFields action
## Feature description

Adds a `document:deleteFields` action that allows to remove field from a document without having to pass the entire document minus the field you want to be removed to `document:replace`

## Example usecase

You want to remove fields from a document but you don't want to pass the entire document

Creates a new method in the Elasticsearch service
Creates a new API action in the document controller

Shiranuit:
## What does this PR do ?

This PR add the a new method called deleteFields to the document controller, the goal is to give an array of fields to remove to avoid doing a replace with the whole document just to delete some fields.
Add Documentation and Tests for document:deleteFields

close #1773


features/DocumentController.feature
```diff
@@ -173,6 +173,25 @@ Feature: Document Controller
       | _id          |
       | "document-1" |
 
+  # document:deleteFields ===========================================================
+  @mappings
+  Scenario: Update document with and without returning updated document
```
rolljee, line 178: It's seem's to me that the name of the scenario does not reflect what done inside.
Also, they should be split into two distinct use case.

lib/service/storage/elasticsearch.js
```diff
@@ -822,6 +822,79 @@ class ElasticSearch extends Service {
...
+   */
+  deleteFields (
+    index,
+    collection,
+    id,
+    fields,
+    { refresh, userId=null } = {})
+  {
+    const
+      esIndex = this._getESIndex(index, collection),
```
rolljee, line 844: Nitpicking: one const per line

lib/service/storage/elasticsearch.js
```diff
@@ -822,6 +822,79 @@ class ElasticSearch extends Service {
...
+        }
+
+        debug('DeleteFields document: %o', esRequest);
+        return this._client.get(esRequest)
+          .then(({ body }) => {
+            for (let field of fields) {
+              _.unset(body._source, field);
+            }
+
+            _.set(body, '_source._kuzzle_info',
```
rolljee, line 870: If i'm not wrong, we want not to use lodash. This could be done anotherway without mutating the body object

rolljee (changes requested):

lib/service/storage/elasticsearch.js
Shiranuit: Is there a reason to avoid using lodash ?

doc/2/api/controllers/document/delete-fields/index.md
```diff
@@ -0,0 +1,96 @@
+---
+code: true
+type: page
+title: delete-fields
```
Aschen, line 4: Missing `description`

doc/2/api/controllers/document/delete-fields/index.md
```diff
@@ -0,0 +1,96 @@
...
+---
+
+## Query Syntax
+
+### HTTP
+
+```http
+URL: http://kuzzle:7512/<index>/<collection>/<documentId>/_fields[?refresh=wait_for]
+Method: DELETE
+Body: {}
```
Aschen, line 22: Nitpicking
```suggestion
Body: 
```

doc/2/api/controllers/document/delete-fields/index.md
```diff
@@ -0,0 +1,96 @@
...
+- `_version`: updated document version number
+
+```js
+{
+  "status": 200,
+  "error": null,
+  "index": "<index>",
+  "collection": "<collection>",
+  "controller": "document",
+  "action": "replace",
```
Aschen, line 86:
```suggestion
  "action": "deleteFields",
```

doc/2/api/controllers/document/delete-fields/index.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+### Other protocols
+
+```js
+{
+  "index": "<index>",
+  "collection": "<collection>",
+  "controller": "document",
+  "action": "replace",
```
Aschen, line 40:
```suggestion
  "action": "deleteFields",
```

lib/service/storage/elasticsearch.js
Shiranuit: After researching a bit more I saw a lot of script made to remove a field in a document and they are removing the field straight from the _source of the document

This comes from the ES doc (https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html)
```
POST test/_update/1
{
  "script" : "ctx._source.remove('new_field')"
}
```

So I was wondering, is it really a bad behaviour to mutate the body object ? If so, could you explain me why ?

lib/api/controller/document.js
```diff
@@ -559,6 +560,41 @@ class DocumentController extends NativeController {
...
+  * @returns {Promise<Object>}
+  */
+  async deleteFields (request) {
+    const id = this.getId(request);
+    const fields = this.getBodyArray(request, 'fields');
+    const userId = this.getUserId(request);
+    const refresh = this.getString(request, 'refresh', 'false');
+    const { index, collection } = this.getIndexAndCollection(request);
+
+    const modifiedRequest = await global.kuzzle.validation.validate(request, false);
```
Aschen, line 576: The validation API need the entire document content. Here the body only contains the list of fields to delete.

Since we need to rewrite the validation API, maybe we can skip it for this action for now, what do you think @scottinet ?

lib/api/controller/document.js
```diff
@@ -559,6 +560,41 @@ class DocumentController extends NativeController {
...
+      'core:storage:public:document:deleteFields',
+      index,
+      collection,
+      id,
+      fields,
+      { refresh, userId });
+
+    await this.ask(
+      'core:realtime:document:notify',
+      modifiedRequest,
```
Aschen, line 588: The notification should be done with the entire new document content

lib/service/storage/elasticsearch.js
Aschen: Yes, lodash is slow and the behavior is not explicit.

lib/service/storage/elasticsearch.js
```diff
@@ -822,6 +822,79 @@ class ElasticSearch extends Service {
...
+      .then(({ body: exists }) => {
+        if (!exists) {
+          throw kerror.get('not_found', id, index, collection);
+        }
+
+        debug('DeleteFields document: %o', esRequest);
+        return this._client.get(esRequest)
+          .then(({ body }) => {
+            for (let field of fields) {
+              _.unset(body._source, field);
```
Aschen, line 867: You should set the fields values to `undefined` instead of deleting them
```suggestion
              _.set(body._source, field, undefined);
```

lib/service/storage/elasticsearch.js
```diff
@@ -822,6 +822,79 @@ class ElasticSearch extends Service {
...
+    return this._client.exists({ id, index: esIndex })
+      .then(({ body: exists }) => {
+        if (!exists) {
+          throw kerror.get('not_found', id, index, collection);
+        }
+
+        debug('DeleteFields document: %o', esRequest);
+        return this._client.get(esRequest)
+          .then(({ body }) => {
+            for (let field of fields) {
```
Aschen, line 866:
```suggestion
            for (const field of fields) {
```

lib/service/storage/elasticsearch.js
```diff
@@ -822,6 +822,79 @@ class ElasticSearch extends Service {
...
+    };
+
+    return this._client.exists({ id, index: esIndex })
+      .then(({ body: exists }) => {
+        if (!exists) {
+          throw kerror.get('not_found', id, index, collection);
+        }
+
+        debug('DeleteFields document: %o', esRequest);
+        return this._client.get(esRequest)
```
Aschen, line 864: Avoid the usage of `then/catch` and prefer the usage of `async/await`

lib/service/storage/elasticsearch.js
```diff
@@ -822,6 +822,79 @@ class ElasticSearch extends Service {
...
+          .then(({ body }) => {
+            for (let field of fields) {
+              _.unset(body._source, field);
+            }
+
+            _.set(body, '_source._kuzzle_info',
+              _.merge(
```
Aschen, line 871: Prefer the usage of vanilla javascript when its possible

lib/service/storage/elasticsearch.js
```diff
@@ -822,6 +822,79 @@ class ElasticSearch extends Service {
...
+              body: body._source,
+              id,
+              index: esIndex,
+              refresh
+            };
+
+            assertNoRouting(newEsRequest);
+            assertWellFormedRefresh(newEsRequest);
+
+            return this._client.index(newEsRequest)
```
Aschen, line 887: Prefer `async/await`

Aschen (changes requested):

lib/service/storage/elasticsearch.js
rolljee: > Is there a reason to avoid using lodash ?

Apart from trying to stay appart from external lib as much as we can and so reduce the build size their is (from my part) not much reasons 😄 

> So I was wondering, is it really a bad behaviour to mutate the body object ? If so, could you explain why ?

From my experience of clean code mutation should be avoided.

- Mutation may lead to unexpected and hard-to-debug issues, where data becomes incorrect somewhere, and you have no idea where it happens.
- Mutation makes code harder to understand: at any time, an array or object may have a different value, so we need to be very careful when reading the code.
- Mutation of function arguments makes the behavior of a function surprising.

Shiranuit: Ok I understand now why mutations may not be desired, but in that case it seems like there is no other choice (or I might have missed it), since the only way I saw to remove a field is through the update API of ES with a script, but this adds a lot of complexity to generate a working script to remove the field that the user want to be removed, and I dont think this is the way to go, and I dont think we want the user to pass a script to update the document since we couldn't control what he could do with it.

doc/2/api/controllers/document/delete-fields/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Optional:
+
+- `refresh`: if set to `wait_for`, Kuzzle will not respond until the new document content is indexed
+
+---
+
+## Body properties
+
+- `fields`: an array of string. Each string represent a path (see lodash path) to a specific field to remove
```
xbill82, line 66:
```suggestion
- `fields`: an array of string. Each string represents a path (see [lodash path](https://lodash.com/docs/4.17.15#toPath)) to a specific field to remove
```

lib/service/storage/elasticsearch.js
```diff
@@ -822,6 +822,77 @@ class ElasticSearch extends Service {
     }
   }
 
+  /**
+   * Delete fields of a document and replace it
+   *
+   * @param {String} index - Index name
+   * @param {String} collection - Collection name
+   * @param {String} id - Document id
+   * @param {Array}  fields - Document fields
```
xbill82, line 831:
```suggestion
   * @param {Array}  fields - Document fields to be removed
```

xbill82 (approved):
Nitpicking

lib/service/storage/elasticsearch.js
```diff
@@ -822,6 +822,77 @@ class ElasticSearch extends Service {
...
+
+    try {
+      const { body: exists } = await this._client.exists({ id, index: esIndex });
+
+      if (!exists) {
+        throw kerror.get('not_found', id, index, collection);
+      }
+
+      debug('DeleteFields document: %o', esRequest);
+      const { body } = await this._client.get(esRequest);
```
scottinet, line 864: You can simply get the document, and let the error formatter do its job: it'll return a proper "document not found" error. This'll avoid doing a check request through the network for nothing:

```suggestion
      debug('DeleteFields document: %o', esRequest);
      const { body } = await this._client.get(esRequest);
```

lib/service/storage/elasticsearch.js
```diff
@@ -822,6 +822,77 @@ class ElasticSearch extends Service {
...
+      assertNoRouting(newEsRequest);
+      assertWellFormedRefresh(newEsRequest);
+
+      await this._client.index(newEsRequest);
+
+      return {
+        _id: id,
+        _source: body._source,
+        _version: body._version
+      };
```
scottinet, line 890: The version you return in your response is the one of the document you got from ES, before the fields are deleted. You need to get the new version from ES' response:

```suggestion
      const { body: updated } = await this._client.index(newEsRequest);

      return {
        _id: id,
        _source: updated._source,
        _version: updated._version
      };
```

scottinet (changes requested):

doc/2/api/controllers/document/delete-fields/index.md
```diff
@@ -0,0 +1,95 @@
+---
+code: true
+type: page
+title: delete-fields
```
Aschen, line 4: Page title must match method name
```suggestion
title: deleteFields
```

doc/2/api/controllers/document/delete-fields/index.md
```diff
@@ -0,0 +1,95 @@
...
+- `fields`: an array of string. Each string represents a path (see [lodash path](https://lodash.com/docs/4.17.15#toPath)) to a specific field to remove
+
+---
+
+## Response
+
+Returns an object containing updated document information, with the following properties:
+
+- `_id`: document unique identifier
+- `_source`: new document content
```
Aschen, line 75: Maybe we could add a `source` option to this method like for the `document:update` method so Kuzzle will return the body only if asked.

What do you think?

lib/service/storage/elasticsearch.js
```diff
@@ -822,6 +822,71 @@ class ElasticSearch extends Service {
...
+  {
+    const esIndex = this._getESIndex(index, collection);
+    const esRequest = {
+      id,
+      index: esIndex,
+    };
+
+    // metadata
+    const _kuzzle_info = {
+      author: getUserId(userId),
```
Aschen, line 851: The `author` property should not be updated since the `deleteFields` action is considered as an update.

lib/service/storage/elasticsearch.js
```diff
@@ -822,6 +822,71 @@ class ElasticSearch extends Service {
...
+      debug('DeleteFields document: %o', esRequest);
+      const { body } = await this._client.get(esRequest);
+
+      for (const field of fields) {
+        _.set(body._source, field, undefined);
+      }
+
+      body._source._kuzzle_info = Object.assign(
+        body._source._kuzzle_info || { createdAt: Date.now() },
+        _kuzzle_info);
```
Aschen, line 866: Prefer the new deconstruction syntax like:
```
const newObj = { ...oldObj, value: 42 };
```
But also here you don't need to put the current metadata in you request, ES will perform an update of the metadata you provide (`updater` and `updatedAt`) Look [here](https://github.com/kuzzleio/kuzzle/blob/master/lib/service/storage/elasticsearch.js#L601) for an example

Aschen (changes requested):
Also you should add the corresponding unit tests to your PR

lib/service/storage/elasticsearch.js
Shiranuit: Does ES do an update of the metadata even if the whole document is replaced ?

doc/2/api/controllers/document/delete-fields/index.md
Shiranuit: This could reduce the network load is the user doesn't need to get the new document, I will add an option for that

lib/service/storage/elasticsearch.js
Aschen: Oh yes you're right, I forgot that this was a replace. Then yes you need to pass the existing metadata object enriched with the `updater` and `updatedAt` properties

scottinet (approved):

Aschen:
@Shiranuit You need to add the corresponding unit tests to your PR

rolljee (approved):

doc/2/api/controllers/document/delete-fields/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Optional:
+
+- `refresh`: if set to `wait_for`, Kuzzle will not respond until the new document content is indexed
+- `source`: if set to `true`, the response will contain the new updated document
+---
+
+## Body properties
+
+- `fields`: an array of string. Each string represents a path (see [lodash path](https://lodash.com/docs/4.17.15#toPath)) to a specific field to remove
```
Leodau, line 66:
```suggestion
- `fields`: an array of strings. Each string represents a path (see [lodash path](https://lodash.com/docs/4.17.15#toPath)) to a specific field to remove
```

features/DocumentController.feature
```diff
@@ -173,6 +173,25 @@ Feature: Document Controller
       | _id          |
       | "document-1" |
 
+  # document:deleteFields ===========================================================
+  @mappings
+  Scenario: Delete fields of a document returning the document without the specified fields
```
Leodau, line 178: The scenario name has two names in one?

test/api/controller/document.test.js
```diff
@@ -1030,6 +1030,78 @@ describe('DocumentController', () => {
...
+      should(kuzzle.ask).be.calledWith(
+        'core:storage:public:document:deleteFields',
+        index,
+        collection,
+        'foobar',
+        content.fields,
+        { userId: null, refresh: 'false' });
+    });
+
+    it('should returns the entire document with source: true', async () => {
```
Leodau, line 1093:
```suggestion
    it('should return the entire document with source: true', async () => {
```

Leodau (changes requested):
Good job!

test/api/controller/document.test.js
```diff
@@ -1030,6 +1030,78 @@ describe('DocumentController', () => {
     });
   });
 
+  describe('#deleteFields', () => {
+    let content;
```
Aschen, line 1034: (Really nitpicking)
As far as I understand, this variable represents the request body so maybe the name `body` would be more appropriate

Aschen (approved):

Leodau (approved):

PR #1930, opened 2021-01-08
Allow to check user rights when using EmbeddedSDK.as impersonation

### LINKED ISSUE #1686: As an end user, I want to apply rights when using the sdk.as method
## Description

The `EmbeddedSDK.as` method allows to impersonate an user when executing an API action.

For now, this method only allows to add the user kuid in the metadata.

The method should also apply the user rights to the request.

Also, when using this method a new EmbeddedSDK is instantiated, we should find a way to avoid this instantiation.

Leodau:
## What does this PR do ?

Long story short. :fast_forward: 
Updates the `app.sdk.as` system to be less expensive and also implements a **conditional** security check _(user rights)_ before executing any action at the funnel level.

---

Current impersonation using Kuzzle's internal **EmbeddedSDK** (`app.sdk.as()`) is actually quite expensive, it would instantiate a whole new SDK with all its bells and whistles. This PR changes this by implementing a lazily dynamic meta-constructed sdk wrapper called the **ImpersonatedSDK** that is considerably smaller, grows depending on the impersonated user's usage and most importantly it works by using the already instantiated base **sdk**.

### _To Dos_
- [x] **ImpersonatedSDK** Tests
- [x] Update **EmbeddedSDK** Tests
- [x] Update **funnelProtocol** Tests
- [x] Functional **Plugin Context** Create Document As
- [x] Docs

### How should this PR be reviewed?

- :warning: Please invest particular attention into finding any breaking differences between the **EmbeddedSDK** and our new **ImpersonatedSDK**. _(Atm, the ImpersonatedSDK only proxies the **controllers** + **query** actions, would that really suffice?)._ :warning: 
- Taking a look at the modifications of the **sdk/funnelProtocol**.
- Check out the below test example to see if any other weird **case** comes to your mind.
- Nvm **sonarcloud** :man_facepalming: 

1. With kuzzle services running, `npm run dev test.ts`
```ts
import { Backend } from './lib/core/application/backend'

const app = new Backend('playground')

function assert(response, wantedId) {
    if (response._source._kuzzle_info.author !== wantedId)
        throw new Error(`Failed, ${response._source._kuzzle_info.author}!=${wantedId}`)
}

app.start()
.then(async () => {
    const proxyInstance = app.sdk.as({ _id: 'leo' }, { checkRights: true })

    assert(await app.sdk.document.create('test', 'test', { authorShouldBe: 'nobody' }), null)
    assert(await proxyInstance.document.create('test', 'test', { authorShouldBe: 'leo' }), 'leo')
    assert(await app.sdk.document.create('test', 'test', { authorShouldBe: 'nobody' }), null)
    assert(await app.sdk.as({ _id: 'alyx' }).document.create('test', 'test', { authorShouldBe: 'alyx' }), 'alyx')
    assert(await app.sdk.document.create('test', 'test', { authorShouldBe: 'nobody' }), null)
    assert(await app.sdk.as({ _id: 'leo' }).document.create('test', 'test', { authorShouldBe: 'leo' }), 'leo')
    assert(await app.sdk.document.create('test', 'test', { authorShouldBe: 'nobody' }), null)

    // This doesn't work, you can't use '.as' with an already impersonated sdk instance.
    // await proxyInstance.as({ _id: 'alyx' }).document.create('test', 'test', { authorShouldBe: 'alyx' });

    // Should these be also made to work?
    // console.log(proxyInstance.protocol, proxyInstance.sdkVersion, proxyInstance.sdkName);
    // console.log(app.sdk.protocol, app.sdk.sdkVersion, app.sdk.sdkName);

})
.catch(console.error)
```

### Other changes

- Updated the **FunnelProtocol** to be able to handle performant user Impersonation on demand. _(With backwards compatibility)_
- Added **app** _(which is a Backend instance)_ to the **global object**. :flags: 
- Moved and renamed native Controller files. _(api/controllers/)_ and their tests!
- Had to update the UTs for the newly dissected **BackendPlugin**, Backend is a singleton now?

### Boyscout

- Removes a UT step duplicate (authController.test)
- Adds a new assert **assertionError** when trying to re-define the **global.app** instance (:warning: This makes the Backend class behave like a single instance only! (hard singleton)

Closes #1686


lib/api/funnel.js
```diff
@@ -25,16 +25,7 @@ const Bluebird = require('bluebird');
...
-const AuthController = require('./controller/auth');
-const BulkController = require('./controller/bulk');
-const CollectionController = require('./controller/collection');
-const DocumentController = require('./controller/document');
-const IndexController = require('./controller/index');
-const MemoryStorageController = require('./controller/memoryStorage');
-const RealtimeController = require('./controller/realtime');
-const SecurityController = require('./controller/security');
-const ServerController = require('./controller/server');
+const nativeControllers = require('./controllers');
```
Aschen, line 28: I would rather keep importing the original class name so it's easier to identify them later
```suggestion
const {
  AdminController,
  AuthController,
  BulkController,
  CollectionController,
  DocumentController,
  IndexController,
  MemoryStorageController,
  RealtimeController,
  SecurityController,
  ServerController,
} = require('./controllers');
```

lib/api/controllers.js
```diff
@@ -0,0 +1,35 @@
...
+  Admin: require('./controller/admin'),
+  Auth: require('./controller/auth'),
+  Bulk: require('./controller/bulk'),
+  Collection: require('./controller/collection'),
+  Document: require('./controller/document'),
+  Index: require('./controller/index'),
+  Ms: require('./controller/memoryStorage'),
+  Realtime: require('./controller/realtime'),
+  Security: require('./controller/security'),
+  Server: require('./controller/server'),
```
Aschen, line 34: I would rather match the export with the real class name. Also the file should be renamed to match the class declared inside.

Then you will be able to rename the `api/controllers.js` file to `api/controllers/index.js` to follow the Node.js module naming convention.

```suggestion
  AdminController: require('./controllers/adminController'),
  AuthController: require('./controllers/authController'),
  BulkController: require('./controllers/bulkController'),
  CollectionController: require('./controllers/collectionController'),
  DocumentController: require('./controllers/documentController'),
  IndexController: require('./controllers/indexController'),
  MemoryStorageController: require('./controllers/memoryStorageController'),
  RealtimeController: require('./controllers/realtimeController'),
  SecurityController: require('./controllers/securityController'),
  ServerController: require('./controllers/serverController'),
```

lib/api/funnel.js
```diff
@@ -96,17 +87,17 @@ class Funnel {
...
+    this.controllers.set('bulk', new nativeControllers.Bulk());
+    this.controllers.set('collection', new nativeControllers.Collection());
+    this.controllers.set('document', new nativeControllers.Document());
+    this.controllers.set('index', new nativeControllers.Index());
+    this.controllers.set('realtime', new nativeControllers.Realtime());
+    this.controllers.set('security', new nativeControllers.Security());
+    this.controllers.set('server', new nativeControllers.Server());
+    this.controllers.set('admin', new nativeControllers.Admin());
+
+    const msController = new nativeControllers.Ms();
```
Aschen, line 100:
```suggestion
  AdminController: require('./controllers/adminController'),
  AuthController: require('./controllers/authController'),
  BulkController: require('./controllers/bulkController'),
  CollectionController: require('./controllers/collectionController'),
  DocumentController: require('./controllers/documentController'),
  IndexController: require('./controllers/indexController'),
  MemoryStorageController: require('./controllers/memoryStorageController'),
  RealtimeController: require('./controllers/realtimeController'),
  SecurityController: require('./controllers/securityController'),
  ServerController: require('./controllers/serverController'),
  
  const msController = new MemoryStorageController();
```

lib/core/shared/sdk/impersonatedSdk.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+'use strict';
+
+const nativeControllers = require('../../../api/controllers');
+
+class ImpersonatedSDK {
+  constructor (sdk, userId) {
+    this.sdk = sdk;
+    this.userId = userId;
```
Aschen, line 29: Nitpicking: This could be renamed to `kuid` to follow Kuzzle domain

lib/core/shared/sdk/impersonatedSdk.js
```diff
@@ -0,0 +1,70 @@
...
+
+    for (const controller of Object.keys(nativeControllers)) {
```
Aschen, line 31: Nitpicking: This should be renamed to something more explicit representing the content of the variable (since it's a class it should start with a capital letter): `ControllerClass`

lib/core/shared/sdk/impersonatedSdk.js
```diff
@@ -0,0 +1,70 @@
...
+      const controllerName = controller.toLowerCase();
+      const subProxyName = `${controllerName}ProxyMethod`;
+
+      if (! Object.prototype.hasOwnProperty.call(
```
Aschen, line 35: Why not simply use `this.sdk[controllerName]` ?

lib/core/shared/sdk/impersonatedSdk.js
```diff
@@ -0,0 +1,70 @@
...
+        this.sdk,
+        controllerName)) {
+        continue;
+      }
+
+      Reflect.defineProperty(this, controllerName, {
+        get: () => {
+          if (! Object.prototype.hasOwnProperty.call(this, subProxyName)) {
```
Aschen, line 43: Ditto, why using `hasOwnProperty` ?

lib/core/shared/sdk/impersonatedSdk.js
```diff
@@ -0,0 +1,70 @@
...
+          }
+
+          return this[subProxyName];
+        }
+      });
+    }
+  }
+
+  query (request, options = {}) {
+    request.__userId__ = this.userId;
```
Aschen, line 64: Nitpicking: the "hidden" variable in the request payload could be named `__kuid__`

Aschen (changes requested):
I know it's still in draft but I suggested some changes :+1:

lib/api/funnel.js
Leodau: This was part of my thinking process also, BUT, the idea of importing the 'whole' controllers module instead of one by one, was so when we update our controller list, there would be only one place to do so.

lib/api/controllers.js
Leodau: Ofc, having a **controllers/index.js** is my main intention.

Leodau: Also, there is an inconsistency between our controllers naming and its **sdk.** counterparts.

Like with our  'alias' for MemoryStorage here:
https://github.com/kuzzleio/kuzzle/blob/master/lib/api/funnel.js#L111
And the whole DocumentController -> _document_.

So I thought this turnaround would spare me from doing name parsing in the **proxyNaming** at the **impersonatedSdk level**.

lib/core/application/backend.ts
```diff
@@ -571,6 +571,10 @@ export class Backend {
       writable: true
     });
 
+    Reflect.defineProperty(global, 'app', {
```
Aschen, line 574: An error should be raised (or even better, an assertion) if the `app` property is already defined

lib/core/shared/sdk/funnelProtocol.js
```diff
@@ -53,6 +55,17 @@ class FunnelProtocol extends KuzzleEventEmitter {
...
+  async _impersonate(requestedUserId) {
+    if (typeof requestedUserId !== 'string') {
+      throw kerror.get('plugin', 'context', 'invalid_user');
+    }
+
+    // Make sure to request the user only if needed
+    if (! this.requestOptions.user || this.requestOptions.user._id !== requestedUserId) {
+      this.requestOptions.user = await global.kuzzle.ask('core:security:user:get', requestedUserId);
+    }
+  }
```
Aschen, line 67: Nitpicking
```suggestion
  async _impersonate (requestedKuid) {
    if (typeof requestedKuid !== 'string') {
      throw kerror.get('plugin', 'context', 'invalid_user');
    }

    // Make sure to request the user only if needed
    if (! this.requestOptions.user || this.requestOptions.user._id !== requestedKuid) {
      this.requestOptions.user = await global.kuzzle.ask('core:security:user:get', requestedKuid);
    }
  }
```

lib/core/shared/sdk/funnelProtocol.js
```diff
@@ -63,25 +76,47 @@ class FunnelProtocol extends KuzzleEventEmitter {
...
           id: await global.kuzzle.ask('core:network:internal:connectionId:get'),
           protocol: this.id
         },
-        user: this.user
+        user: null
       };
     }
 
+    // Use the provided request.UserId, otherwise use the impersonatedId previously given at funnel construction
+    if (Object.prototype.hasOwnProperty.call(request, '__kuid__')) {
```
Aschen, line 91: Why using `hasOwnProperty` and not `request.__kuid__` directly?

lib/core/shared/sdk/impersonatedSdk.js
```diff
@@ -0,0 +1,71 @@
...
+
+const nativeControllers = require('../../../api/controllers');
+
+class ImpersonatedSDK {
+  constructor (kuid) {
+    this.kuid = kuid;
+
+    for (const ControllerClass of Object.keys(nativeControllers)) {
+      // @todo in the future, native controller names should be normalized to sdk naming
+      const controllerName = ControllerClass
```
Aschen, line 32: I didn't realize that since you only iterate on the keys then `ControllerClass` is not a constructor but just a name.. Sorry for that
```suggestion
    for (const controller of Object.keys(nativeControllers)) {
      // @todo in the future, native controller names should be normalized to sdk naming
      const controllerName = controller
```

Aschen (changes requested):

Aschen (changes requested):
Missing functional tests, otherwise everything is ok for me :+1:

xbill82 (commented):
I find the whole architecture of this system a little weird. Let me know if I get this correctly:
* `EmbeddedSDK` gets instantiated once by `Backend`, when the global `app` object is created.
* `app` (and thus `app.sdk`) is meant to be a singleton, even if the implementation of this pattern is quite exotic (`global.app` with an assertion). The problem here is that the "globality" of the SDK is defined in the `Backend` class, rather in the `EmbeddedSDK` class
* The static `EmbeddedSDK.as()` method returns a new instance of `ImpersonatedSDK` (which is pretty elegant), wrapping the global instance of `EmbeddedSDK` (`global.app.sdk`).

It seems to me we could avoid this roundabout and just let `EmbeddedSDK` behave like the real Singleton pattern and be able to pass its unique instance to the constructor of `ImpersonatedSDK`. Or the `as()` method could be a non-static one, to call directly on the `global.app.sdk` object. I feel like we oscillate between the class and object (at global level) nature of the symbols, which I find misleading.
I don't know. I smell spaghetti (not strictly inherent to this PR, tho... It comes from the deep levels of the arch).

But I do understand that the real Singleton here is `Backend`, and that `sdk` is just one of its members and that the use of `global` is just a shortcut to avoid importing `Backend` in all the files. And I don't like it.

And I do understand that **changing this is costly and probably not worth the fuss**

xbill82:
Ok, spoke with @Aschen nevermind my comment!

xbill82 (approved):

plugins/available/functional-test-plugin/functionalTestPlugin.js
```diff
@@ -119,6 +119,16 @@ class FunctionalTestPlugin {
         });
     };
 
+    // Embedded SDK.as() Impersonation =========================================
+    this.controllers.impersonate = { createDocumentAs: 'createDocumentAs' };
+
+    this.routes.push({
+      action: 'createDocumentAs',
+      controller: 'impersonate',
+      path: '/impersonate/createDocumentAs/:userId',
```
Aschen, line 128: Nitpicking
```suggestion
      path: '/impersonate/createDocumentAs/:kuid',
```

plugins/available/functional-test-plugin/functionalTestPlugin.js
```diff
@@ -260,6 +270,20 @@ class FunctionalTestPlugin {
 
     return { result: helloName };
   }
+
+  /**
+   * Tests the EmbeddedSDK's impersonating feature using the new
+   * ImpersonatedSDK wrapper
+   */
+  async createDocumentAs (request) {
+    const sdkInstance = await this.sdk.as({ _id: request.input.args.userId });
```
Aschen, line 279:
```suggestion
    const sdkInstance = await this.sdk.as({ _id: request.input.args.kuid });
```

plugins/available/functional-test-plugin/functionalTestPlugin.js
```diff
@@ -260,6 +270,20 @@ class FunctionalTestPlugin {
...
+
+    return sdkInstance.document.create(
+      'nyc-open-data',
+      'yellow-taxi',
+      { shouldBeCreatedBy: request.input.args.userId },
```
Aschen, line 284:
```suggestion
      { shouldBeCreatedBy: request.input.args.kuid },
```

Aschen (approved):

jenow (approved):

Aschen (changes requested):
Documentation must be updated:
 - https://docs.kuzzle.io/core/2/guides/develop-on-kuzzle/embedded-sdk/#user-impersonation
 - https://docs.kuzzle.io/core/2/framework/classes/embedded-sdk/as/

Aschen (changes requested):
:warning: We need a workshop because there is an usecase where a developer may want to add Kuzzle metadata to the document but does not want to apply the user rights

rolljee (approved):

lib/types/Global.ts
```diff
@@ -13,4 +14,4 @@ global.NODE_ENV = process.env.NODE_ENV;
 
 export {};
 
-/* eslint-enable @typescript-eslint/no-namespace */
+/* eslint-enabl2e @typescript-eslint/no-namespace */
```
Aschen:
```suggestion
/* eslint-enable @typescript-eslint/no-namespace */
```

alexandrebouthinon (approved):

lib/core/shared/sdk/funnelProtocol.js
```diff
@@ -32,16 +32,18 @@ class FunnelProtocol extends KuzzleEventEmitter {
...
 
     this.id = 'funnel';
-    this.user = user;
-
     this.requestOptions = null;
+    this._impersonateKuid = null;
 
-    if (this.user && typeof this.user._id !== 'string') {
-      throw kerror.get('plugin', 'context', 'invalid_user');
+    if (user) {
```
Aschen, line 38: The `FunnelProtocol` should not be instantiated with a user object anymore. It's a leftover from the previous implementation of the `as` method where the `EmbeddedSDK` could be instantiated with an user.

Now the `EmbeddedSDK` should never be instantiated with an user since we have the `ImpersonatedSDK` for this.

You should also remove the `user` argument from the `EmbeddedSDK` constructor

lib/core/shared/sdk/impersonatedSdk.js
```diff
@@ -0,0 +1,73 @@
...
+        .replace('Controller', '')
+        .replace('MemoryStorage', 'ms')
+        .toLowerCase();
+      const subProxyName = `${controllerName}ProxyMethod`;
+
+      if (! global.app.sdk[controllerName]) {
+        continue;
+      }
+
+      Reflect.defineProperty(this, controllerName, {
```
Aschen, line 43: Can we have some explicit comments for this section ? Metaprogramming spell are powerful but not easy to understand and master for newcomers ;-)

lib/types/Global.ts
```diff
@@ -4,6 +4,7 @@ declare global {
   namespace NodeJS {
     interface Global {
       kuzzle: any;
+      app: any;
```
Aschen, line 7: We should specify the real type here since it's already written in TS

Aschen (changes requested):

lib/core/shared/sdk/funnelProtocol.js
Leodau: Oh ofc, getting that user in the funnel's constructor actually induced some underlying logic complexities. I thought we couldn't remove it due to 'trying to maintain backwards compatibility'.

Aschen (approved):

plugins/available/functional-test-plugin/functionalTestPlugin.js
```diff
@@ -104,10 +104,10 @@ class FunctionalTestPlugin {
     this.pipes['plugin-functional-test-plugin:testPipesReturn'] =
       async name => `Hello, ${name}`;
 
-    // Pipe declared with a function name
+    // Pipe declared with a function name ======================================
```
scottinet, line 107: (nitpicking / for another PR) when I see that kind of thing, this triggers a reflex telling me that the file is too large, and that the code should be split to make it more readable

scottinet (changes requested):

lib/core/shared/sdk/funnelProtocol.js
```diff
@@ -64,13 +69,34 @@ class FunnelProtocol extends KuzzleEventEmitter {
...
       };
     }
 
+    // Use the provided request.kuid, otherwise use the impersonatedId previously given at funnel construction
+    if (request.__kuid__) {
+      await this._impersonate(request.__kuid__);
+    }
+    else { // Reset user if impersonation is not required
+      this.requestOptions.user = null;
+    }
```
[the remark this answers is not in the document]
Aschen: Since we already have the `kuid` I'm not sure we need to get the entire user object the the `kuid` is enough for the rights and metadata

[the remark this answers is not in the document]
scottinet: It seems to be insufficient, since the FunnelProtocol needs it. And IIRC there is only 1 FunnelProtocol for all impersonated SDKs: loading the User object in an async task means that race conditions can (and probably will) occur when processing concurrent requests.
Or did I miss something?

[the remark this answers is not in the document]
Leodau: Seems like the only place where the user is required, it actually uses only the **_id** in it. (KuzzleContext)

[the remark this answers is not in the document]
Leodau: Ok, so. Now we store the user and the required request options in the local context instead, this should solve our race condition issue. Can you guys confirm this? @scottinet @Aschen

lib/core/shared/sdk/impersonatedSdk.js
```diff
@@ -0,0 +1,83 @@
...
+        continue;
+      }
+
+      // Lazily wrap in runtime; native 'sdk' controllers and induce custom 'query' beaviour
+      Reflect.defineProperty(this, controllerName, {
+        get: () => {
+          // If the requested controller has not yet been proxied
+          if (! this[controllerProxy]) {
+
+            // Create a proxy object to injects our local 'query' behaviour
```
Leodau, line 49:
```suggestion
            // Create a proxy object to inject our local 'query' behaviour
```

lib/core/shared/sdk/impersonatedSdk.js
```diff
@@ -0,0 +1,83 @@
...
+            // in the underlying base 'sdk' action methods so we can send
+            // impersonation configuration (kuid & checkRights) to the funnel
+            this[controllerProxy] = new Proxy(global.app.sdk[controllerName], {
+              get: (controllerInstance, actionName) => {
+
+                // Returns the original SDK action method but bind a context
```
Leodau, line 55:
```suggestion
                // Return the original SDK action method BUT bind to this local context
```

lib/core/shared/sdk/impersonatedSdk.js
```diff
@@ -0,0 +1,83 @@
...
+                // with our custom 'query method
+                return controllerInstance[actionName].bind({
```
Aschen, line 57: We need to merge the original context (`this`) with our new one who contains the query method because some controller method are using other references of the original context (e.g [auth.checkToken](https://github.com/kuzzleio/sdk-javascript/blob/master/src/controllers/Auth.ts#L193) or [realtime.subscribe](https://github.com/kuzzleio/sdk-javascript/blob/master/src/controllers/Realtime.ts#L165) for example)

Aschen (changes requested):

Aschen (approved):

scottinet (approved):

alexandrebouthinon (approved):

PR #1941, opened 2021-01-20
Hotfix: Restore API keys when the cache is wiped clean

scottinet:
# Description

This PR solves 2 problems.

First, API keys are stored in Elasticsearch, and synchronized in Redis. They are supposed to be loaded into Redis during the bootstrap of the internal indexes, which are run only once during the lifetime of Kuzzle: when it's started for the first time. 
If the Redis cache is wiped, or when Kuzzle is started for the first time with an ES storage restored from a snapshot, Kuzzle skips the internal index bootstrap, and API keys aren't synchronized in Redis, making them unusable.

Second, loading API keys in the internal index bootstrapper doesn't work. This wasn't visible because the bootstrap only occured when Kuzzle is started for the first time, on a blank ES storage. But the API key loader uses the Token Repository, which isn't initialized until after the internal indexes are bootstrapped.

# Solution

The API keys synchronization is now the responsibility of the Token Repository, and a different bootstrap key is used. Contrary to the one used by the internal index bootstrap, this bootstrap key is stored in Redis.
This way, if the cache is wiped, or if Kuzzle is started from a restored ES storage, the Token Repository correctly restores the API keys in the cache, and this is done only once, whatever the number of Kuzzle nodes in a cluster environment.

# Why is this a hotfix?

Because one of our clients is currently experiencing this issue, and it's a blocking one.

### DIFF
--- a/test/core/security/tokenRepository.test.js
+++ b/test/core/security/tokenRepository.test.js
@@ -13,23 +13,35 @@ const {
   UnauthorizedError
 } = require('../../../index');
 const KuzzleMock = require('../../mocks/kuzzle.mock');
+const MutexMock = require('../../mocks/mutex.mock');
 
 const Token = require('../../../lib/model/security/token');
 const User = require('../../../lib/model/security/user');
-const TokenRepository = require('../../../lib/core/security/tokenRepository');
 const Repository = require('../../../lib/core/shared/repository');
+const ApiKey = require('../../../lib/model/storage/apiKey');
 
-describe('Test: security/tokenRepository', () => {
+describe.only('Test: security/tokenRepository', () => {
   let kuzzle;
+  let TokenRepository;
   let tokenRepository;
 
   beforeEach(() => {
     kuzzle = new KuzzleMock();
     kuzzle.secret = 'test-secret';
 
+    mockrequire('../../../lib/util/mutex', { Mutex: MutexMock });
+    TokenRepository = mockrequire.reRequire('../../../lib/core/security/tokenRepository');
+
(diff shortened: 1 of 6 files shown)


test/core/security/tokenRepository.test.js
```diff
@@ -13,25 +13,37 @@ const {
...
+const MutexMock = require('../../mocks/mutex.mock');
 
 const Token = require('../../../lib/model/security/token');
 const User = require('../../../lib/model/security/user');
-const TokenRepository = require('../../../lib/core/security/tokenRepository');
 const Repository = require('../../../lib/core/shared/repository');
+const ApiKey = require('../../../lib/model/storage/apiKey');
 
-describe('Test: security/tokenRepository', () => {
+describe.only('Test: security/tokenRepository', () => {
```
Leodau, line 23:
```suggestion
describe('Test: security/tokenRepository', () => {
```

Leodau (approved):

test/core/security/tokenRepository.test.js
alexandrebouthinon, line 23:
```suggestion
describe('Test: security/tokenRepository', () => {
```

alexandrebouthinon (changes requested):

alexandrebouthinon (approved):

test/core/security/tokenRepository.test.js
scottinet: dang it >_<

Yoann-Abbes (approved):

PR #1949, opened 2021-01-28
Allow to generate human readable kuid for users

Shiranuit:
This PR changes the way kuids are generated when creating a new user:
-  Now we are generating human readable kuid by default
- It is possible to choose between human readable kuid or standard uuid thanks to the request optional parameter `kuid` that can be either set to `human` or `uuid`

### DIFF
--- a/doc/2/api/controllers/security/create-first-admin/index.md
+++ b/doc/2/api/controllers/security/create-first-admin/index.md
@@ -56,7 +56,8 @@ Body:
   },
   // optional
   "reset": <boolean>,
-  "_id": "<kuid>"
+  "_id": "<kuid>",
+  "kuid": "human"
 }
 ```
 
@@ -68,7 +69,7 @@ Body:
 
 - `_id`: specify the administror [kuid](/core/2/guides/main-concepts/authentication#kuzzle-user-identifier-kuid), instead of letting Kuzzle generate a random identifier.
 - `reset` (boolean): if true, restricted rights are applied to the `anonymous` and `default` roles (by default, these roles don't have any restriction).
-
+- `kuid`: if set to `human`, Kuzzle will generate a human readable id, otherwise if set to `uuid` Kuzzle will generate a standard uuid (default: `"human"`)
 ---
 
 ## Body properties
--- a/lib/api/controller/security.js
+++ b/lib/api/controller/security.js
@@ -31,6 +31,9 @@ const formatProcessing = require('../../core/auth/formatProcessing');
 const ApiKey = require('../../model/storage/apiKey');
 const kerror = require('../../kerror');
 const { has, get } = require('../../util/safeObject');
+const nameGenerator = require('../../util/name-generator');
+const { v4: uuidv4 } = require('uuid');
+const { until } = require('async');
 
 /**
  * @class SecurityController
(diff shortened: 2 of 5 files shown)


lib/api/controller/security.js
```diff
@@ -1160,31 +1162,51 @@ class SecurityController extends NativeController {
    * @private
    */
   async _persistUser(request, profileIds, content) {
-    const id = this.getId(request, ifMissingEnum.GENERATE);
+    const humanReadable = this.getBodyString(request, 'kuid', 'human') !== 'uuid';
```
Aschen, line 1165: This should be an argument and not part of the body

Aschen (changes requested):

Aschen:
@Shiranuit You need to add the proper labels to your PR

doc/2/api/controllers/security/create-first-admin/index.md
```diff
@@ -68,7 +69,7 @@ Body:
 
 - `_id`: specify the administror [kuid](/core/2/guides/main-concepts/authentication#kuzzle-user-identifier-kuid), instead of letting Kuzzle generate a random identifier.
 - `reset` (boolean): if true, restricted rights are applied to the `anonymous` and `default` roles (by default, these roles don't have any restriction).
-
+- `kuid`: if set to `human`, Kuzzle will generate a human readable id, otherwise if set to `uuid` Kuzzle will generate a standard uuid (default: `"human"`)
```
xbill82, line 72: Actually, it would be cool that the auto-generated KUID for the first admin contain the word `admin` or something similar.

xbill82 (approved):

doc/2/api/controllers/security/create-first-admin/index.md
scottinet: Since kuid cannot be easily changed, what if admin priviledges get removed from that user?

alexandrebouthinon: I agree with @scottinet . Moreover in the Admin Console, it's easy to differentiate an admin via his profiles so it's not embarrassing IMO
image

lib/api/controller/security.js
```diff
@@ -1159,32 +1165,51 @@ class SecurityController extends NativeController {
...
+            throw kerror.get(
+              'security',
+              'credentials',
+              'database_inconsistency',
+              id);
+          }
         }
       }
-    }
+    } while (alreadExists);
```
alexandrebouthinon, line 1212:
```suggestion
    } while (alreadyExists);
```

lib/api/controller/security.js
```diff
@@ -1159,32 +1165,51 @@ class SecurityController extends NativeController {
...
-  async _persistUser(request, profileIds, content) {
-    const id = this.getId(request, ifMissingEnum.GENERATE);
+  async _persistUser(request, profileIds, content, humanReadableId=true) {
+    const generator = humanReadableId ? nameGenerator : uuidv4;
     const credentials = this.getBodyObject(request, 'credentials', {});
     const strategies = Object.keys(credentials);
-
+    
+    let id = null;
+    let alreadExists = false;
```
alexandrebouthinon, line 1174:
```suggestion
    let alreadyExists = false;
```

lib/api/controller/security.js
```diff
@@ -1159,32 +1165,51 @@ class SecurityController extends NativeController {
...
-        const exists = this.getStrategyMethod(strategy, 'exists');
-        if (await exists(request, id, strategy)) {
-          throw kerror.get(
-            'security',
-            'credentials',
-            'database_inconsistency',
-            id);
+      );
+
+      if (id) {
```
alexandrebouthinon, line 1187: I don't understand why it should be checked that id is not undefined. It will never be the case according to:
```js
      id = this.getId(
        request,
        ifMissingEnum.GENERATE,
        () => {
          generated = true;
          return `kuid-${generator()}`;
        }
      );
```

lib/api/controller/security.js
```diff
@@ -1159,32 +1165,51 @@ class SecurityController extends NativeController {
...
+            throw kerror.get(
+              'security',
+              'credentials',
+              'unknown_strategy',
+              strategy);
+          }
+
+          const exists = this.getStrategyMethod(strategy, 'exists');
+          alreadExists = await exists(request, id, strategy);
+          if (alreadExists) {
```
alexandrebouthinon, line 1199:
```suggestion
          if (alreadyExists) {
```

lib/api/controller/security.js
```diff
@@ -1159,32 +1165,51 @@ class SecurityController extends NativeController {
...
+          if (!global.kuzzle.pluginsManager.listStrategies().includes(strategy)) {
+            throw kerror.get(
+              'security',
+              'credentials',
+              'unknown_strategy',
+              strategy);
+          }
+
+          const exists = this.getStrategyMethod(strategy, 'exists');
+          alreadExists = await exists(request, id, strategy);
```
alexandrebouthinon, line 1198:
```suggestion
          alreadyExists = await exists(request, id, strategy);
```

alexandrebouthinon (changes requested):

lib/api/controller/security.js
Shiranuit: You are right, this check is not useful, removed

lib/api/controller/security.js
```diff
@@ -1159,13 +1165,25 @@ class SecurityController extends NativeController {
    * @returns {Promise}
    * @private
    */
-  async _persistUser(request, profileIds, content) {
-    const id = this.getId(request, ifMissingEnum.GENERATE);
+  async _persistUser(request, profileIds, content, humanReadableId=true) {
```
Aschen, line 1168:
```suggestion
  async _persistUser(request, profileIds, content, { humanReadableId=true } = {}) {
```

Options should be passed as an object. This improve readability:
```
function foobar (index, collection, skipRights=false) { ... }
function barfoo (index, collection, { skipRights=false } = {}) { ... }

foobar('idx', 'col', true); // what does true mean ??
barfoo('idx', 'col', { skipRights: true }); // explicite
```

lib/api/controller/security.js
```diff
@@ -1159,13 +1165,25 @@ class SecurityController extends NativeController {
...
+    const generator = humanReadableId ? nameGenerator : uuidv4;
     const credentials = this.getBodyObject(request, 'credentials', {});
     const strategies = Object.keys(credentials);
-
+    
+    let id = null;
```
Aschen, line 1173: Nitpicking: if you want to initialize this variable then you could use an empty string to keep the same type after assignment

Aschen (approved):
Last nitpicking

Aschen (approved):

alexandrebouthinon (approved):

PR #1957, opened 2021-02-04
Allow "inner_hits" key in ES search body whitelist

### LINKED ISSUE #1956: Allow "inner_hits" key in Elasticsearch search body whitelist
Add the [inner_hits](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/inner-hits.html) ES key for nested query inside our [whitelist](https://github.com/kuzzleio/kuzzle/blob/master/lib/service/storage/elasticsearch.js#L105)

MathieuVeber:
Closes #1956

## What does this PR do ?
Adding compatibility with inner_hits [see here](https://discord.com/channels/306054601436561410/694124702053957720/806623592246804531)

### DIFF
--- a/lib/config/default.config.js
+++ b/lib/config/default.config.js
@@ -250,7 +250,7 @@ module.exports = {
       }
     },
     internalIndex: {
-      bootstrapLockTimeout: 5000
+      bootstrapLockTimeout: 60000
     },
     storageEngine: {
       aliases: ['storageEngine'],
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -468,7 +468,7 @@
     //       Maximum amount of time (in milliseconds)
     //       to wait for a concurrent database bootstrap
     "internalIndex": {
-      "bootstrapLockTimeout": 5000
+      "bootstrapLockTimeout": 60000
     },
 
     // [internalCache]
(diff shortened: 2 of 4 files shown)


lib/config/default.config.js
```diff
@@ -250,7 +250,7 @@ module.exports = {
       }
     },
     internalIndex: {
-      bootstrapLockTimeout: 5000
+      bootstrapLockTimeout: 60000
```
Aschen, line 253: Why this change?

Aschen (changes requested):
Documentation must be updated accordingly https://github.com/kuzzleio/kuzzle/blob/master/doc/2/guides/main-concepts/querying/index.md

lib/config/default.config.js
MathieuVeber: I don't know this is coming from @scottinet's commit

Aschen (approved):

lib/config/default.config.js
scottinet: Yes it does.
I cherry-picked a commit from the 2-dev branch to master to fix our CI a few days ago. I don't get why it appears here. :thinking: 
But it doesn't really matter.

scottinet (approved):

PR #1958, opened 2021-02-04
Add an application-level PING for WebSocket

Yoann-Abbes:
## What does this PR do ?

We need a keep-alive system on client side.
That means that the client has to ping Kuzzle and Kuzzle response to it.
With the node library, it's automatic, the response of Kuzzle is automatic since the ping frame sent is a _real_ one.

However, through the browser that is not possible since there is no API implemented to send pings.
We needed something higher level, like a custom message to interpret.
See the related PR on the SDK JS https://github.com/kuzzleio/sdk-javascript/pull/586

:small_blue_diamond:  This tiny code came from the conclusion that's it's the only way to go after several tries. 

Check out this branch of the sdk js https://github.com/kuzzleio/sdk-javascript/pull/586
Create an `index.html` and import the SDK JS
```html
<!DOCTYPE html>
<html>

<head>
  <meta charset="utf-8" />
  <meta http-equiv="X-UA-Compatible" content="IE=edge" />
  <title>Kuzzle SDK Playground</title>
  <script src="../sdks/sdk-javascript/dist/kuzzle.js"></script> //replace with your path
</head>

<body>
  <script>
    var kuzzle = new KuzzleSDK.Kuzzle(
      new KuzzleSDK.WebSocket('localhost')
    );
    kuzzle.on('networkError', error => {
      console.error('Network Error: ', error);
    });
    kuzzle.on('connected', () => {
      console.log('Successfully connected to Kuzzle');
    });
    kuzzle.connect();
  </script>
</body>

</html>
```
 and a `index.js` 
```js
const {
  Kuzzle,
  WebSocket
} = require('../sdks/sdk-javascript/');

const kuzzle = new Kuzzle(
  new WebSocket('localhost')
);

kuzzle.on('networkError', error => {
  console.error('Network Error:', error);
});
const run = async () => {
  try {
    // Connects to the Kuzzle server
    await kuzzle.connect();
  } catch (error) {
    console.error(error.message);
  }
};
run();
```
Add console logs after sending/receiving messages
Launch the node app and your html page
See logs and Kuzzle still alive


xbill82 (approved):
Looks fine to me but, why the Netlify build is broken?

lib/core/network/protocols/websocket.js
```diff
@@ -208,6 +208,12 @@ class WebSocketProtocol extends Protocol {
 
     try {
       parsed = JSON.parse(data);
+      if (parsed.p && parsed.p === '1') {
+        debug('[%s] sending a `pong` message', connection.id);
```
Aschen, line 212: Nitpicking
```suggestion
        debug('[%s] sending back a "pong" message', connection.id);
```

Aschen (changes requested):
We need to document this behavior somewhere, here for example https://docs.kuzzle.io/core/2/api/protocols/websocket/

features/Websocket.feature
```diff
@@ -0,0 +1,10 @@
+Feature: WebSocket
+
+  # WebSocket ===========================================================
+
+  Scenario: Check reponse message from Kuzzle with a specific payload
+    Given I open a new local websocket connection
+    When I send the message '{"p":"1"}' to Kuzzle through websocket
+    Then I should receive a response message from Kuzzle through websocket matching:
+      |  p  |  '1'  |
+    Then I terminate the websocket connection
```
Aschen, line 10: This should be a hook on a tag (like the one on `@securities`) because if the step line 8 fail then the step line 10 will not be executed and the websocket connexion will not be terminated

features/step_definitions/websocket-steps.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+const should = require('should');
+const { Given, Then, When } = require('cucumber');
+const ws = require('ws');
+
+Given('I open a new local websocket connection', async function () {
+  this.props.client = new ws('ws://localhost:7512');
+  const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
```
Aschen: Why are you waiting 1 sec here? 1 second is huge, it will the slow down all the tests

features/step_definitions/websocket-steps.js
```diff
@@ -0,0 +1,28 @@
...
+  this.props.client.on('message', data => {
+    this.props.result = JSON.parse(data);
+  });
+});
+
+When('I send the message {string} to Kuzzle through websocket', function (message) {
+  this.props.client.send(message);
+});
+
+Then('I should receive a response message from Kuzzle through websocket matching:', function (dataTable) {
```
Aschen, line 23: We may not have received the answer here

Aschen (changes requested):

features/step_definitions/websocket-steps.js
Yoann-Abbes: I removed it, that was just a test

features/step_definitions/websocket-steps.js
Yoann-Abbes: Kuzzle sends automatically a message when it received one, I'm just checking if it's the good one

Aschen: Yes but when this step is executed, maybe we still didn't receive a response from Kuzzle

doc/2/api/protocols/websocket/index.md
```diff
@@ -63,3 +63,11 @@ The listening port can be modified under the `server.port` section of the [confi
 HTTP and WebSocket protocols share the same underlying server instance.  
 Modifying the listening port will impact these two protocols.
 :::
+
+<SinceBadge version="2.10.0"/>
```
scottinet, line 67:
```suggestion
<SinceBadge version="auto-version">
```

doc/2/api/protocols/websocket/index.md
```diff
@@ -63,3 +63,11 @@ The listening port can be modified under the `server.port` section of the [confi
...
+:::info
+We needed the client to be able to ping Kuzzle to have a keep-alive system.
+Through the browser, that was not possible to send real formatted pings websocket frames since there is no API implemented for that unlike the one of Node.
+We needed something higher level.
+That is the reason why when you are using the SDK Javascript, the ping/pong between Kuzzle and your client is a custom message interpreter.
+:::
```
scottinet, line 73: You need to fully document the feature, including the expected application-level PING format, and the corresponding PONG response.

Also, I would change how this documentation is worded (to be put in a chapter of its own, and not as a side-note):

```suggestion
:::info
Though Kuzzle's WebSocket server is fully compliant with the RFC6455, meaning (among other things) that Kuzzle will respond to PING packets with standard PONG ones, an additional PING request has been added in the protocol's application layer.

This application-level PING has been especially added for web browsers, which don't allow sending PING packets. This can be troublesome if a web application needs to know if a connection has been severed, or if Kuzzle is configured to be in passive mode (i.e. it won't send PING requests by itself, and will close sockets if they are idle for too long).

When run in a browser, our Javascript SDK uses that feature for its keep-alive mechanism.
:::
```

lib/core/network/protocols/websocket.js
```diff
@@ -208,6 +208,12 @@ class WebSocketProtocol extends Protocol {
 
     try {
       parsed = JSON.parse(data);
+      if (parsed.p && parsed.p === 1) {
```
scottinet, line 211: You also need to verify that this is the only property of the received JSON payload.

For all we know, there might be a custom API action out there having a numeral `p` parameter...

lib/core/network/protocols/websocket.js
```diff
@@ -208,6 +208,12 @@ class WebSocketProtocol extends Protocol {
...
+        debug('[%s] sending back a "pong" message', connection.id);
+        wsConnection.alive = true;
+        this._send(connection.id, '{"p":1}');
```
scottinet, line 214: The PONG response needs to be different than the PING query:

```suggestion
        this._send(connection.id, '{"p":2}');
```

scottinet (changes requested):

Aschen (approved):

features/step_definitions/websocket-steps.js
```diff
@@ -0,0 +1,32 @@
...
+    });
+  });
+});
+
+When('I send the message {string} to Kuzzle through websocket', function (message) {
+  this.props.client.send(message);
+});
+
+Then('I wait to receive a websocket response from Kuzzle', function() {
+  return new Promise((resolve) => {
```
Aschen, line 21: Very good :+1:

alexandrebouthinon (approved):

scottinet (approved):

PR #1959, opened 2021-02-05
Enable usage of HttpOnly Cookies to authenticate browsers

Shiranuit:
## What does this PR do ?

The goal of this PR is to add an option that enable the support of Browser Cookie as a Jwt
# To Do / Done
- [x] Add a Kuzzle config option to enable support for Cookie Authentication
- [x] Add config check when `supportCookieAuthentication` is `enabled`, to avoid incompatibilities in config
- [x] `auth:login` should send token as cookie when requested cookieOnly if Kuzzle supportCookieAuthentication is enabled
- [x] `auth:refreshToken` should send token as cookie when requested cookieOnly if Kuzzle supportCookieAuthentication is enabled
- [x] `auth:checkToken` should check use token in the cookie when requested cookieOnly if Kuzzle supportCookieAuthentication is enabled
- [x] The funnel should not use the token in the cookie when a request to `auth:login` or `auth:checkToken` is made with cookieOnly and Kuzzle supportCookieAuthentication is enabled
- [x] Fix documentation on error `configuration.incompatible_config`
- [x] Add unit tests for `auth:login`
- [x] Add unit tests for `auth:logout`
- [x] Add unit tests for `auth:checkToken`
- [x] Add unit tests for `auth:refreshToken`
- [x] Add unit tests for `funnel:checkRights`
- [x] Add functional tests for `auth:login`
- [x] Add functional tests for `auth:logout`
- [x] Add functional tests for `auth:checkToken`
- [x] Add functional tests for `auth:refreshToken`
- [x] Add functional tests for `funnel:checkRights`
- [x] Add documentation for the `cookieOnly` request's option
- [x] Add documentation for Kuzzle `http.supportCookieAuthentication` config


Shiranuit:
## Add Configuration [security.supportBrowserCookieAsJwt]
### When enabled:
Check if the [http] config is compatible with the option [security.supportBrowserCookieAsJwt]
#### In development:
- If incompatible raise an Exception that cancel the boot of Kuzzle
#### In production
- If incompatible disable the option and then prints a warning that explains why the option has been disabled

Set Access-Control-Allow-Credentials to true

## auth:login and auth:refreshToken
### When `cookieOnly` is set to `true` and `[security.supportBrowserCookieAsJwt]` is set to `true`

Kuzzle will send an object with the following properties (e.g without the jwt property)
```js
{ 
          _id: token.userId,
          expiresAt: token.expiresAt,
          ttl: token.ttl,  
}
```
and set the `Set-Cookie` as the `jwt` in the response header

### Otherwise both controller methods will work as usual

Why ? This will allow the SDK to detect if Kuzzle supportBrowserCookieAsJwt is disabled or not supported in the current version,
with this information we take action directly in the SDK to print a warning or raise an Exception

## Funnel behaviour when [security.supportBrowserCookieAsJwt] is true
When supportBrowserCookieAsJwt is enabled and a request has been made with the param `cookieOnly` the funnel is going to check if the request can be processed by checking the right as usual but this time, it will use the information contained in the cookie to replace the expected token in the request

### Troubles
For now if we make a request to `/_login/local?cookieOnly` with the credentials for the first time, everything works perfectly
But when we want to login with other credentials or if the jwt contained in the cookie as expired, because the param `cookieOnly` is set to true, the funnel is currently checking for the validity of the jwt sent in the cookie causing the request to will fail since the cookie is not valid anymore and preventing us to login.

This needs to be fixed to avoid checking the validity of the old jwt contained in the cookie when an `auth:login` request if made with `cookieOnly` since we want the SDK to login again and not check token validity

.kuzzlerc.sample
```diff
@@ -212,6 +212,10 @@
     // The profileIds applied to a user created with the API action
     // "security:createRestrictedUser"
     "restrictedProfileIds": ["default"],
+    // [supportBrowserCookieAsJwt]
+    // Enable Kuzzle secure support of Jwt in Cookie for the browser
+    // This option however is not compatible with [http.accessControlAllowOrigin] set to '*'
+    "supportBrowserCookieAsJwt": false,
```
Aschen, line 218: Nitpicking: I don't really like that name, also because we try to avoid the term `jwt` and use authentication token instead.

Proposals:
  - `authenticateWithCookie`
  - `supportCookieAuthentication` 
  - ...

lib/api/controller/auth.js
```diff
@@ -326,7 +365,14 @@ class AuthController extends NativeController {
...
   async checkToken (request) {
-    const token = this.getBodyString(request, 'token');
+    let token = null;
+
+    const isCookieOnly = request.input.args.cookieOnly;
+    if (!global.kuzzle.config.security.supportBrowserCookieAsJwt || !isCookieOnly) {
+      token = this.getBodyString(request, 'token');
+    } else {
+      token = request.context.token.jwt;
+    }
```
Aschen, line 375: Nitpicking: coding standards
```suggestion
    } 
    else {
      token = request.context.token.jwt;
    }
```

lib/api/controller/auth.js
```diff
@@ -326,7 +365,14 @@ class AuthController extends NativeController {
    * @returns {Promise<object>}
    */
   async checkToken (request) {
-    const token = this.getBodyString(request, 'token');
+    let token = null;
```
Aschen, line 368: If you really want to initialize this variable then you can use an empty string to keep same type after assignment

lib/api/controller/auth.js
```diff
@@ -259,6 +260,44 @@ class AuthController extends NativeController {
...
     }
 
+    // Only if the support of Browser Cookie as Jwt is enabled
+    // otherwise we should send a normal response because
+    // even if the SDK / Browser can handle the cookie,
+    // Kuzzle would not be capable of doing anything with it
+    if (global.kuzzle.config.security.supportBrowserCookieAsJwt) {
+      const isCookieOnly = request.input.args.cookieOnly;
+
+      if (isCookieOnly) {
```
Aschen, line 270: Redondant if, you should have both conditions in the same if. Also, since you use this at 3 different places you could DRY this code

lib/api/controller/auth.js
```diff
@@ -536,6 +582,40 @@ class AuthController extends NativeController {
...
+          _id: token.userId,
+          expiresAt: token.expiresAt,
+          ttl: token.ttl,  
+        };
+
+        request.setResult(resultContent,
+          {
+            headers: {
+              'Set-Cookie': Cookie.serialize(
+                'jwt',
```
Aschen, line 603: We try to not use `jwt` anymore. You should use `authToken` instead for example. (We want to introduce `refreshToken` as well in the futur)

lib/api/controller/auth.js
```diff
@@ -536,6 +582,40 @@ class AuthController extends NativeController {
...
+      const isCookieOnly = request.input.args.cookieOnly;
+
+      if (isCookieOnly) {
+        const resultContent = { 
+          _id: token.userId,
+          expiresAt: token.expiresAt,
+          ttl: token.ttl,  
+        };
+
+        request.setResult(resultContent,
```
Aschen, line 599:
```suggestion
        request.setResult(null,
```

Since the first argument is not used by Kuzzle to construct the response (but rather the return of the controller action method), we tend to put `null` as the first argument here to avoid confusion. 

Also, this makes `resultContent` useless

lib/api/funnel.js
```diff
@@ -368,6 +369,24 @@ class Funnel {
...
+    if (global.kuzzle.config.security.supportBrowserCookieAsJwt) {
+      
+      const isCookieOnly = request.input.args.cookieOnly;
+      // When a request id made with CookieOnly set to true
+      // We try to use the jwt cookie if present as jwt token
+      // otherwise check for jwt as input
+      if (isCookieOnly) {
+
+        if (request.input.headers && request.input.headers.cookie) {
+          const cookie = Cookie.parse(request.input.headers.cookie);
```
Aschen, line 383: Can this lead to a parsing error? Also since the client want a `cookieOnly` authentification, if the `cookie.jwt` property is not set (l 385) then it should be considered as an error

lib/core/backend/backend.ts
```diff
@@ -567,6 +567,28 @@ export class Backend {
       throw runtimeError.get('already_started', 'start');
     }
 
+    // When the support of cookie as jwt is enabled
+    // we need to perform some checks before starting Kuzzle
+    if (this.config.content.security.supportBrowserCookieAsJwt) {
+      if (this.config.content.http.accessControlAllowOrigin === '*') {
+        if (process.env.NODE_ENV === 'development') {
```
Aschen, line 574: `process.env` calls should be avoided in the code since each call will lead to an expensive syscall to get the env.

You can use our custom `global.NODE_ENV` variable which is set at startup.
```suggestion
        if (global.NODE_ENV === 'development') {
```

Also I think we should stop Kuzzle startup sequence even in production environment instead of (almost) silently start Kuzzle without a key security feature

lib/core/backend/backend.ts
```diff
@@ -567,6 +567,28 @@ export class Backend {
       throw runtimeError.get('already_started', 'start');
     }
 
+    // When the support of cookie as jwt is enabled
+    // we need to perform some checks before starting Kuzzle
+    if (this.config.content.security.supportBrowserCookieAsJwt) {
```
Aschen, line 572: 3 levels of `if` is too much IMHO

Aschen (changes requested):
Good job :+1: This feature is very much appreciated.

Some few remarks on the code (I saw it's still a draft but we need to move fast on this one so we could ship this in Kuzzle 2.10.0)

Shiranuit:
Thanks for the feedback, I will apply your suggestion ASAP

lib/api/funnel.js
Shiranuit: Cookie.parse can only fail when the string given is empty, wich cannot be the case here

Also I dont think if the client want a `cookieOnly` authentication that the this should be considered as an error, since it may be the first time the user makes a request to Kuzzle to authenticate himself and the cookie might not be set at this time

I will made a slight modification to make sure that if  the request is made with `cookieOnly` that only the cookie should be used even if it's not defined, this way I let the funnel handle the case when the token is missing

Shiranuit:
The previous trouble with login when there already was a Cookie with a potential expired token as been fixed by not using the token present in the Cookie when a request to `auth:login` or `auth:checkToken` has been made with cookieOnly and Kuzzle supportCookieAuthentication is enabled

Shiranuit:
# To Do

- [ ] Fix documentation on error `configuration.incompatible_config`
- [ ] Add unit tests for `auth:login`
- [ ] Add unit tests for `auth:checkToken`
- [ ] Add unit tests for `auth:refreshToken`
- [ ] Add documentation for the `cookieOnly` request's option
- [ ] Add documentation for Kuzzle `security.supportCookieAuthentication` config

lib/api/controller/auth.js
```diff
@@ -188,9 +189,74 @@ class AuthController extends NativeController {
         request.context.token);
     }
 
+    if (global.kuzzle.config.security.supportCookieAuthentication
```
Aschen, line 192: Nitpicking: coding standard
```suggestion
    if ( global.kuzzle.config.security.supportCookieAuthentication
```

lib/api/controller/auth.js
```diff
@@ -188,9 +189,74 @@ class AuthController extends NativeController {
...
+          }
+        });
+    }
+
     return {acknowledged: true};
   }
 
+  // Used to send the Token using different ways when in cookieOnly mode. (DRY)
+  async _sendToken(token, request) {
+    // Only if the support of Browser Cookie as Jwt is enabled
```
Aschen, line 217:
```suggestion
    // Only if the support of Browser Cookie as Authentication Token is enabled
```

lib/api/controller/auth.js
```diff
@@ -188,9 +189,74 @@ class AuthController extends NativeController {
...
+    // otherwise we should send a normal response because
+    // even if the SDK / Browser can handle the cookie,
+    // Kuzzle would not be capable of doing anything with it
+    if (global.kuzzle.config.security.supportCookieAuthentication
```
Aschen, line 221: Nitpicking: coding standards
```suggestion
    if ( global.kuzzle.config.security.supportCookieAuthentication
```

lib/api/controller/auth.js
```diff
@@ -188,9 +189,74 @@ class AuthController extends NativeController {
...
+      && request.input.args.cookieOnly
+    ) {
+      // Here we are not sending jwt when cookieOnly is set to true
+      // This allow us to detect if kuzzle does support cookie as jwt directly from the SDK
+      // if the response contains the jwt this means that the configuration [security.supportCookieAuthentication] is set to false
+      // or that the version of kuzzle doesn't support the feature Browser Cookie as Jwt
```
Aschen, line 227: Nitpicking: domain wording
```suggestion
      // Here we are not sending auth token when cookieOnly is set to true
      // This allow us to detect if kuzzle does support cookie as auth token directly from the SDK
      // if the response contains the auth token this means that the configuration [security.supportCookieAuthentication] is set to false
      // or that the version of kuzzle doesn't support the feature Browser Cookie as Authentication Token
```

lib/api/controller/auth.js
```diff
@@ -326,7 +387,14 @@ class AuthController extends NativeController {
...
   async checkToken (request) {
-    const token = this.getBodyString(request, 'token');
+    let token = '';
+
+    if (!global.kuzzle.config.security.supportCookieAuthentication || !request.input.args.cookieOnly) {
+      token = this.getBodyString(request, 'token');
+    }
+    else {
+      token = request.input.jwt;
+    }
```
Aschen, line 397: Nitpicking: (code complexity) it's more readable to have a logical AND rather than NOR ... OR NOR ...

```suggestion
    if ( global.kuzzle.config.security.supportCookieAuthentication 
      && request.input.args.cookieOnly
    ) {
      token = request.input.jwt;
    }
    else {
      token = this.getBodyString(request, 'token');
    }
```

lib/api/funnel.js
```diff
@@ -368,9 +369,29 @@ class Funnel {
    * @returns {Promise<Request>}
    */
   async checkRights (request) {
+   
+    let skipTokenVerification = false;
+    // When the Support of Cookie as Jwt is enabled we check if a jwt cookie is present
+    // When a request is made with CookieOnly set to true
+    // We try to use the jwt cookie if present as jwt token
+    // otherwise check for jwt as input
```
Aschen, line 377: Nitpicking: domain wording
```suggestion
    // When the Support of Cookie as Authentication Token is enabled we check if an auth token cookie is present
    // When a request is made with CookieOnly set to true
    // We try to use the auth token cookie if present as auth token
    // otherwise check for auth token as input
```

lib/api/funnel.js
```diff
@@ -368,9 +369,29 @@ class Funnel {
...
+    if (global.kuzzle.config.security.supportCookieAuthentication
```
Aschen, line 378: Nitpicking: coding standard
```suggestion
    if ( global.kuzzle.config.security.supportCookieAuthentication
```

lib/config/default.config.js
```diff
@@ -105,6 +105,7 @@ module.exports = {
 
   security: {
     restrictedProfileIds: ['default'],
+    supportCookieAuthentication: false,
```
Aschen, line 108: This should be true by default IMHO
```suggestion
    supportCookieAuthentication: true,
```

Aschen (approved):
Last nitpicking 

Also I think we should enable this option by default

lib/api/controller/auth.js
```diff
@@ -188,9 +189,74 @@ class AuthController extends NativeController {
...
+          headers: {
+            'Set-Cookie': Cookie.serialize(
+              '__Host-authToken',
+              token.jwt,
+              {
+                expires: new Date(token.expiresAt),
+                httpOnly: true,
+                path: '/',
+                sameSite:'none',
+                secure: true,
```
Leodau, line 240: As already demanded/discussed, can you explain why you chose to use the `secure: true` property?
What about when in dev mode where **https** is not enabled?

[Forbid non-https cookies with Secure flag](https://blog.dareboost.com/en/2019/03/secure-cookies-secure-httponly-flags/#forbid-to-use-a-cookie-without-https-thanks-to-the-secure-flag)
[Other issues when using secure + httpOnly](https://webmasters.stackexchange.com/questions/53178/what-are-the-problems-of-httponly-and-secure-cookies#:~:text=The%20problem%20with%20secure%20cookies,can%20save%20your%20server%20resources.)

lib/api/funnel.js
Leodau, line 377: Please drop the `jwt` naming, prefer `authToken` (also, avoid repeating words in the same comment :exploding_head: )

lib/config/default.config.js
Shiranuit: IDK the problem with this option is that the CORS Header cannot be defined as a wildcard, so if we defined it to true by default, we can't set the `http.accessControlAllowOrigin` to `*`

Leodau (changes requested):
:+1: Could you provide a more practical/explicit approach in how we could test this PR?

lib/api/controller/auth.js
Shiranuit: Secure cookie cannot be delivered when requests are not made with the https protocol, but they can be delivered to localhost even if it's not https.

I think it's a bigger concern to send token through http than not being able to use cookies for your website when in http

Shiranuit: You right, I just saw that I didn't push something, because I stashed it, now `sameSite` is set to `'strict'` and `secure` has been removed allowing us to use the cookie even in an http context, which was not the case before.

`sameSite: 'strict'` is a better choice since the browser will not send cookies when the request is not made directly from the same domain as the cookie was obtained, this helps us to avoid CSRF attacks

Aschen: I agree with @Leodau , this feature should work in both HTTP and HTTPS environments

Aschen (changes requested):
Missing functional tests

lib/api/controller/auth.js
Shiranuit: This is the case, as I mentionned above, cookies works both in HTTP and HTTPS

.kuzzlerc.sample
```diff
@@ -212,6 +212,10 @@
     // The profileIds applied to a user created with the API action
     // "security:createRestrictedUser"
     "restrictedProfileIds": ["default"],
+    // [supportCookieAuthentication]
+    // Enable Kuzzle secure support of Jwt in Cookie for the browser
+    // This option however is not compatible with [http.accessControlAllowOrigin] set to '*'
+    "supportCookieAuthentication": false,
```
scottinet, line 218:
```suggestion
    // [supportCookieAuthentication]
    // Enables Kuzzle support for authentication tokens in cookies, for browsers
    // If activated, [http.accessControlAllowOrigin] must be set to a value different than '*'
    "supportCookieAuthentication": false,
```

scottinet: Also, I think this would be better to put that option in the HTTP section of the configuration, especially since there are interactions with other options there.

doc/2/api/errors/error-codes/core/index.md
```diff
@@ -58,6 +58,7 @@ description: Error codes definitions
 | core.configuration.invalid_type<br/><pre>0x00030001</pre>  | [InternalError](/core/2/api/errors/error-codes#internalerror) <pre>(500)</pre> | Invalid type for the configuration parameter "%s" (expected: %s) | Wrong configuration parameter type |
 | core.configuration.out_of_range<br/><pre>0x00030002</pre>  | [InternalError](/core/2/api/errors/error-codes#internalerror) <pre>(500)</pre> | The configuration value set for "%s" is outside the allowed range (allowed: %s) | A configuration value exceeds the allowed range |
 | core.configuration.cannot_parse<br/><pre>0x00030003</pre>  | [InternalError](/core/2/api/errors/error-codes#internalerror) <pre>(500)</pre> | Unable to read kuzzlerc configuration file: %s | The kuzzlerc configuration file is badly formatted. |
+| core.configuration.incompatible_config<br/><pre>0x00030004</pre>  | [InternalError](/core/2/api/errors/error-codes#internalerror) <pre>(500)</pre> | Incompatible configuration: %s | The kuzzlerc configuration file has incompatible configurations |
```
scottinet, line 61: Keep it short: you don't need to remind that the incompatibility comes from the configuration, since the error pertains to the `core.configuration` domain.

```suggestion
| core.configuration.incompatible<br/><pre>0x00030004</pre>  | [InternalError](/core/2/api/errors/error-codes#internalerror) <pre>(500)</pre> | Incompatible configuration: %s | The kuzzlerc configuration file has incompatible configurations |
```

lib/api/funnel.js
```diff
@@ -368,9 +369,29 @@ class Funnel {
...
+   
+    let skipTokenVerification = false;
+    // When the Support of Cookie as Authentication Token is enabled we check if an auth token cookie is present
+    // When a request is made with CookieOnly set to true
+    // We try to use the auth token cookie if present as auth token
+    // otherwise check for auth token as input
+    if ( global.kuzzle.config.security.supportCookieAuthentication
+      && request.input.args.cookieOnly
+      && request.input.headers
+      && request.input.headers.cookie
```
scottinet, line 381: You should use `has(object, property)` from `lib/util/safeObject.js`, it's safer.

```suggestion
      && has(request.input.headers, 'cookie')
```

lib/api/funnel.js
```diff
@@ -23,6 +23,7 @@
 
 const Bluebird = require('bluebird');
 const Deque = require('denque');
+const Cookie = require('cookie');
```
scottinet, line 26: Missing dependency in `package.json`: this only works because this dependency is imported by `winston-elasticsearch`

lib/api/funnel.js
```diff
@@ -368,9 +369,29 @@ class Funnel {
...
+    // When the Support of Cookie as Authentication Token is enabled we check if an auth token cookie is present
+    // When a request is made with CookieOnly set to true
+    // We try to use the auth token cookie if present as auth token
+    // otherwise check for auth token as input
+    if ( global.kuzzle.config.security.supportCookieAuthentication
+      && request.input.args.cookieOnly
+      && request.input.headers
+      && request.input.headers.cookie
+    ) {
+      const cookie = Cookie.parse(request.input.headers.cookie);
```
scottinet, line 383: `Cookie.parse` may throw if an invalid cookie is provided. This should be caught and a proper `invalid_cookie` error should be returned.

lib/core/backend/backend.ts
```diff
@@ -567,6 +567,16 @@ export class Backend {
...
 
+    // When the support of cookie as jwt is enabled
+    // we need to perform some checks before starting Kuzzle
+    if (this.config.content.security.supportCookieAuthentication) {
+      if (this.config.content.http.accessControlAllowOrigin === '*') {
+        throw kerror.get('core', 'configuration', 'incompatible_config', 'Option [security.supportCookieAuthentication] is incompatible with [http.accessControlAllowOrigin] set to \'*\'');
+      }
+      // Mandatory to enable cookie transmission
+      this.config.content.http.accessControlAllowCredentials = true;
+    }
```
scottinet, line 578: It's the responsibility of the configuration loader to run that kind of checks (in `lib/config`).

scottinet (changes requested):

lib/api/funnel.js
```diff
@@ -368,9 +370,37 @@ class Funnel {
...
+        request.input.jwt = cookie.authToken || null;
+        
+        skipTokenVerification = request.input.controller === 'auth'
+          && (request.input.action === 'login' || request.input.action === 'checkToken');
+      } catch {
+        throw kerror.get(
+          'security',
+          'cookie',
+          'invalid'
+        );
```
scottinet, line 396: try..catch should be specialised as much as possible, for better maintenability: this helps maintainers knowing what exactly is caught and why

```suggestion
      let cookie;
      
      try {
        cookie = Cookie.parse(request.input.headers.cookie);
      } 
      catch {
        throw kerror.get('security', 'cookie', 'invalid');
      }
        
      request.input.jwt = cookie.authToken || null;
        
      skipTokenVerification = request.input.controller === 'auth'
        && (request.input.action === 'login' || request.input.action === 'checkToken');
```

scottinet (changes requested):

features/AuthController.feature
```diff
@@ -56,6 +56,20 @@ Feature: Auth Controller
...
+  @security
+  Scenario: Login with cookieOnly
+    When I successfully execute the action "auth":"login" with args:
+      | strategy   | "local"                                              |
+      | body       | { "username": "test-admin", "password": "password" } |
+      | cookieOnly | true                                                 |
+    Then I should receive a result matching:
+      | _id       | "_STRING_"  |
+      | expiresAt | "_NUMBER_"  |
+      | ttl       | "_NUMBER_"  |
```
Aschen, line 70: You can use only one step to check.
Available special placeholders are here: https://github.com/kuzzleio/kuzzle/blob/master/features/support/assertions.js
```suggestion
      | ttl       | "_NUMBER_"      |
      | jwt     | "_UNDEFINED_" | 
```

features/AuthController.feature
```diff
@@ -56,6 +56,20 @@ Feature: Auth Controller
       | _id        | _source.userId | _source.ttl | _source.expiresAt | _source.description | _source.fingerprint |
       | "_STRING_" | "test-admin"   | -1          | -1                | "Sigfox API key"    | "_STRING_"          |
 
+  # auth:login
+
+  @security
+  Scenario: Login with cookieOnly
+    When I successfully execute the action "auth":"login" with args:
+      | strategy   | "local"                                              |
+      | body       | { "username": "test-admin", "password": "password" } |
```
Aschen, line 65: Nitpicking: you can use lodash syntax to construct objects
```suggestion
      | body.username       | "test-admin" |
      | body.password        | "password"  |
```

features/step_definitions/controllers-steps.js
```diff
@@ -121,6 +121,12 @@ Then('The result should contain a property {string} of type {string}', function
   should(typeof property).be.eql(type);
 });
 
+Then('The result should not contain a property {string}', function (path) {
```
Aschen, line 124: You should remove this step

Aschen (changes requested):

features/AuthController.feature
Shiranuit: This seems not to be working

scottinet (approved):

Leodau:
Could you please **resolve** the conversations **on** which you have already applied any suggestions?

Leodau (approved):

doc/2/api/protocols/http/index.md
```diff
@@ -64,3 +64,22 @@ The listening port can be modified under the `server.port` section of the [confi
 HTTP and WebSocket protocols share the same underlying server instance.  
 Modifying the listening port will impact these two protocols.
 :::
+
+### HTTP-only options
+
+Only the HTTP protocol does support the following options in a [KuzzleRequest](../../payloads/request/index.md)
```
scottinet, line 70: We usually use absolute paths in URLs. Also, `index.md` won't exist once the documentation have been generated, since markdowns are used to build HTML files.

When unsure, don't hesitate to test your documentation by running `npm run doc-dev` and by navigating the local copy of the docs on your browser.

```suggestion
Only the HTTP protocol does support the following options in a [KuzzleRequest](/core/2/api/payloads/request)
```

doc/2/api/protocols/http/index.md
```diff
@@ -64,3 +64,22 @@ The listening port can be modified under the `server.port` section of the [confi
...
+
+| Property         | Type        | Description                                                                 |
+| ---------------- | ----------- | --------------------------------------------------------------------------- |
+| **`cookieOnly`** | **boolean** | **Instruct Kuzzle to store / use authentication token within HTTP Cookies** |
```
scottinet, line 74: Our documentation style is as follows:
* property/argument names are to be put between quotes
* types are to be put between `pre` tags 

Not really practical for us, but the result is clear and nice :)

```suggestion
| `cookieOnly` | <pre>boolean</pre> | Instructs Kuzzle to store / use authentication tokens within secured HTTP cookies |
```

doc/2/api/protocols/http/index.md
```diff
@@ -64,3 +64,22 @@ The listening port can be modified under the `server.port` section of the [confi
...
+
+#### cookieOnly
+
+When the property `cookieOnly` is set to `true` in an HTTP request,
+Kuzzle will `store` or `load` the authentication token directly from an [HTTP Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies),
+and will not use the `jwt` property.
+
+::: warning
+The property `cookieOnly` is only supported when the option `supportCookieAuthentication` is enabled in the [Kuzzle Configuration](../../../guides/advanced/configuration/index.md).
```
scottinet, line 83: Invalid URL:

```suggestion
The property `cookieOnly` is only supported when the option `supportCookieAuthentication` is enabled in the [Kuzzle Configuration](/core/2/guides/advanced/configuration).
```

doc/2/api/protocols/http/index.md
scottinet: Also, since you're at it, could you add a documentation for the `pretty` option?

E.g.

```
| `pretty` | <pre>boolean</pre> | Prettify the JSON response, making it more readable for humans |

File: doc/2/api/protocols/http/index.md [hunk 32]
```diff
@@ -64,3 +64,22 @@ The listening port can be modified under the `server.port` section of the [confi
...
+Otherwise Kuzzle will treat the request as if `cookieOnly` wasn't set.
```
scottinet, line 84:
```suggestion
Otherwise Kuzzle will ignore the `cookieOnly` option.
```

doc/2/guides/main-concepts/authentication/index.md
```diff
@@ -169,6 +169,17 @@ kourou auth:login -a strategy=local --body '{
 }'
 ```
 
+### Authentication Token in the Browser
+
+When you're making `HTTP request` from a Browser you can instruct Kuzzle
```
scottinet, line 174:
```suggestion
When you're sending HTTP requests from a browser you can instruct Kuzzle
```

doc/2/guides/main-concepts/authentication/index.md
```diff
@@ -169,6 +169,17 @@ kourou auth:login -a strategy=local --body '{
...
+to `load` and `store` authentication tokens within an [HTTP Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies).
+This is possible thanks to the [HTTP-only](../../../api/protocols/http/index.md#http-only-options) option [cookieOnly](../../../api/protocols/http/index.md#cookieonly)
```
scottinet, line 176:
```suggestion
This is possible thanks to the [HTTP-only](/core/2/api/protocols/http#http-only-options) option [cookieOnly](/core/2/api/protocols/http#cookieonly)
```

doc/2/guides/main-concepts/authentication/index.md
```diff
@@ -169,6 +169,17 @@ kourou auth:login -a strategy=local --body '{
...
+
+::: warning
+For this to work you should enable the option `supportCookieAuthentication` in the [Kuzzle Configuration](../../advanced/configuration/index.md)
```
scottinet, line 179:
```suggestion
This only works if the option `supportCookieAuthentication` is enabled in the [Kuzzle Configuration](/core/2/guides/advanced/configuration)
```

scottinet (changes requested):

doc/2/api/protocols/http/index.md
```diff
@@ -64,3 +64,23 @@ The listening port can be modified under the `server.port` section of the [confi
...
 Modifying the listening port will impact these two protocols.
 :::
+
+### HTTP-only options
+
+Only the HTTP protocol does support the following options in a [KuzzleRequest](/core/2/api/payloads/request)
+
+| Property     | Type               | Description                                                                           |
+| ------------ | ------------------ | ------------------------------------------------------------------------------------- |
+| `cookieOnly` | <pre>boolean</pre> | Instructs Kuzzle to store / use authentication tokens within secured HTTP Cookies     |
```
scottinet, line 74: Sorry for mentioning this only now, but after thinking about it, I don't like the name `cookieOnly`, I don't find it self-explaining.

What about somethiing like `cookieAuth` or something like that? (open to debate obviously)

Leodau: Yep, I like this idea !
Something like `cookieAuth` makes sense, `cookieOnly` makes me think that _sometimes_ cookie **and** sent tokens could be allowed at the same time?

lib/config/default.config.js
```diff
@@ -47,9 +47,9 @@ module.exports = {
    */
   http: {
     routes: require('./httpRoutes'),
-    accessControlAllowOrigin: '*',
+    accessControlAllowOrigin: 'localhost',
```
Yoann-Abbes, line 50: Should not be localhost

Yoann-Abbes (dismissed):

lib/config/default.config.js
Shiranuit: outdated comment

doc/2/api/protocols/http/index.md
```diff
@@ -64,3 +64,23 @@ The listening port can be modified under the `server.port` section of the [confi
 HTTP and WebSocket protocols share the same underlying server instance.  
 Modifying the listening port will impact these two protocols.
 :::
+
```
Yoann-Abbes, line 67: Missinge badge `SinceVersion`

doc/2/guides/main-concepts/authentication/index.md
```diff
@@ -169,6 +169,17 @@ kourou auth:login -a strategy=local --body '{
 }'
 ```
 
+### Authentication Token in the Browser
```
Yoann-Abbes, line 172: Same as above

Yoann-Abbes (changes requested):

Yoann-Abbes (approved):

doc/2/api/protocols/http/index.md
```diff
@@ -64,3 +64,24 @@ The listening port can be modified under the `server.port` section of the [confi
...
+<SinceBadge version="auto-version"/>
+### HTTP-only options
+
+Only the HTTP protocol does support the following options in a [KuzzleRequest](/core/2/api/payloads/request)
+
+| Property     | Type               | Description                                                                           |
+| ------------ | ------------------ | ------------------------------------------------------------------------------------- |
+| `cookieAuth` | <pre>boolean</pre> | Instructs Kuzzle to store / use authentication tokens within secured HTTP Cookies     |
```
scottinet, line 75: well... now that option is only available for a few actions in the `auth` controller (login, checkToken, refreshToken).

So this piece of documentation should be removed from global HTTP options, and instead be added to each one of those routes.

doc/2/api/protocols/http/index.md
```diff
@@ -64,3 +64,24 @@ The listening port can be modified under the `server.port` section of the [confi
...
+
+#### cookieAuth
+
+When the property `cookieAuth` is set to `true` in an HTTP request,
+Kuzzle will `store` or `load` the authentication token directly from an [HTTP Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies),
+and will not use the `jwt` property.
+
+::: warning
+The property `cookieAuth` is only supported when the option `supportCookieAuthentication` is enabled in the [Kuzzle Configuration](/core/2/guides/advanced/configuration).
+Otherwise Kuzzle will ignore the `cookieAuth` option.
```
scottinet, line 86: Obsolete: this option is now only available if CORS is configured.

scottinet (changes requested):

doc/2/api/controllers/auth/logout/index.md
```diff
@@ -45,6 +45,9 @@ Headers: Authorization: "Bearer <authentication token>"
 
 * `global`: if `true`, also revokes all other active sessions instead of just the current one (default: `false`)
 
+* `cookieAuth`: Enable the erase of the token in the [HTTP Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies)
```
scottinet, line 48:
```suggestion
* `cookieAuth`: Erase of the token in the [HTTP Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies)
```

doc/2/api/protocols/http/index.md
```diff
@@ -64,3 +64,12 @@ The listening port can be modified under the `server.port` section of the [confi
 HTTP and WebSocket protocols share the same underlying server instance.  
 Modifying the listening port will impact these two protocols.
 :::
+
+<SinceBadge version="auto-version"/>
```
scottinet, line 68: Can be removed: the `pretty` option exists since before kuzzle 1.0

```suggestion
```

doc/2/guides/main-concepts/authentication/index.md
```diff
@@ -169,6 +169,19 @@ kourou auth:login -a strategy=local --body '{
 }'
 ```
 
+### Authentication Token in the Browser
+
+When you're sending HTTP requests from a browser you can instruct Kuzzle
+to `load` and `store` authentication tokens within an [HTTP Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies).
+If a cookie is stored in the browser
```
scottinet, line 177: dangling sentence?

scottinet (changes requested):
a few minor documentation fixes and this is good to go for me

features/AuthController.feature
```diff
@@ -19,6 +19,32 @@ Feature: Auth Controller
...
+      | controller | "auth"                                               |
+      | action     | "login"                                              |
+      | strategy   | "local"                                              |
+      | body       | { "username": "test-admin", "password": "password" } |
+      | cookieAuth | true                                                 |
+    When I send a HTTP "POST" request with:
+      | controller | "auth"                                                      |
+      | action     | "checkToken"                                                |
+      | cookieAuth | true                                                        |
+      | headers    | { cookie: this.props.rawResponse.headers['set-cookie'][0] } |
```
Aschen, line 34: Whoa good one! You find this yourself :heart_eyes:

features/AuthController.feature
```diff
@@ -19,6 +19,32 @@ Feature: Auth Controller
...
+    Then I should receive a result matching:
+      | valid     | true         |
+      | kuid      | "test-admin" |
+      | expiresAt | "_NUMBER_"   |
+    When I send a HTTP "POST" request with:
+      | controller | "auth"                                                      |
+      | action     | "checkToken"                                                |
+      | cookieAuth | true                                                        |
+      | headers    | { cookie: 'authToken=wrongtoken' }                                              |
```
Aschen, line 43: Nitpicking: you can construct object "on the fly"
```suggestion
      | headers.cookie    | 'authToken=wrongtoken'                                              |
```

features/AuthController.feature
```diff
@@ -19,6 +19,32 @@ Feature: Auth Controller
       | valid | false      |
       | state | "_STRING_" |
 
+  @security @http
+  Scenario: Check if a token in the cookie is valid
+    Given I send a HTTP "POST" request with:
+      | controller | "auth"                                               |
+      | action     | "login"                                              |
+      | strategy   | "local"                                              |
+      | body       | { "username": "test-admin", "password": "password" } |
```
Aschen, line 28: Nitpicking: you can construct object "on the fly"
```suggestion
      | body.username       | "test-admin"  |
      | body.password        | "password"  |
```

features/AuthController.feature
```diff
@@ -56,6 +82,66 @@ Feature: Auth Controller
...
 
+  # auth:login =================================================================
+
+  @security @http
+  Scenario: Auth login with cookies
+    When I send a HTTP "POST" request with:
+      | controller | "auth"                                               |
+      | action     | "login"                                              |
+      | strategy   | "local"                                              |
+      | body       | { "username": "test-admin", "password": "password" } |
```
Aschen, line 93:
```suggestion
      | body.username       | "test-admin"  |
      | body.password        | "password"  |
```

features/AuthController.feature
```diff
@@ -56,6 +82,66 @@ Feature: Auth Controller
...
+  
+  # auth:logout ================================================================
+
+  @security @http
+  Scenario: Auth logout with cookies
+    When I send a HTTP "POST" request with:
+      | controller | "auth"                                               |
+      | action     | "login"                                              |
+      | strategy   | "local"                                              |
+      | body       | { "username": "test-admin", "password": "password" } |
```
Aschen, line 111:
```suggestion
      | body.username       | "test-admin"  |
      | body.password        | "password"  |
```

features/AuthController.feature
```diff
@@ -56,6 +82,66 @@ Feature: Auth Controller
...
+      | action     | "logout"                                                    |
+      | cookieAuth | true                                                        |
+      | headers    | { cookie: this.props.rawResponse.headers['set-cookie'][0] } |
+    Then The raw response should match:
+      | headers.set-cookie | [ /authToken=null;.*/ ] |
+    And I send a HTTP "POST" request with:
+      | controller | "auth"                                               |
+      | action     | "login"                                              |
+      | strategy   | "local"                                              |
+      | body       | { "username": "test-admin", "password": "password" } |
```
Aschen, line 124:
```suggestion
      | body.username       | "test-admin"  |
      | body.password        | "password"  |
```

features/AuthController.feature
```diff
@@ -56,6 +82,66 @@ Feature: Auth Controller
...
+
+  # auth:refreshToken ==========================================================
+
+  @security @http
+  Scenario: Auth refreshToken with cookies
+    Given I send a HTTP "POST" request with:
+      | controller | "auth"                                               |
+      | action     | "login"                                              |
+      | strategy   | "local"                                              |
+      | body       | { "username": "test-admin", "password": "password" } |
```
Aschen, line 135:
```suggestion
      | body.username       | "test-admin"  |
      | body.password        | "password"  |
```

Aschen (approved):
LGTM :clap: :+1:

Leodau (approved):
:+1:

alexandrebouthinon (approved):
👏

features/AuthController.feature
Shiranuit: Has I said in your previous comment  I tried, but this doesn't seem to work

features/AuthController.feature
Shiranuit: Same as above, It seems that I cannot construct "object on the fly"

features/AuthController.feature
Shiranuit: Same as above, It seems that I cannot construct "object on the fly"

features/AuthController.feature
Shiranuit: Same as above, It seems that I cannot construct "object on the fly"

features/AuthController.feature
Shiranuit: Same as above, It seems that I cannot construct "object on the fly"

scottinet (approved):

Issue #1850, opened 2020-10-30
Node.js 14 support


Aschen:
## Feature description

Since Node.js 14 is going to be in LTS the 27 october, we should add official support for it and start using it by default.

We need to run the unit tests and also the functional tests with Node.js 14

PR #1971, opened 2021-02-09
Official support for Node.js 14.17.0

Yoann-Abbes:
## What does this PR do ?

Since Node.js 14 is in LTS since the 27 october, we should add official support for it and start using it by default.

We need to run the unit tests and also the functional tests with Node.js 14

https://github.com/kuzzleio/kuzzle/issues/1850


docker-compose.yml
```diff
@@ -36,6 +36,7 @@ x-kuzzle-config: &kuzzle-config
     - KUZZLE_SECRETS_FILE=${KUZZLE_SECRETS_FILE:-/var/app/features/fixtures/secrets.enc.json}
     # Variables used by the development scripts
     - NODE_12_VERSION=12.13.0
```
Aschen, line 38: Well, I know it's not yours but this version of node is not even present in our matrix `node-version: ['14.15.4', '12.16.3', '12.20.0']`

Aschen:
I think we should update the docker images as well (`docker/images/*`) but I'm not sure if it will be a breaking or not :thinking: 

What do you think @alexandrebouthinon ?

alexandrebouthinon (approved):

Aschen (changes requested):
This need to be updated with the latest Node.js 14 version

Shiranuit:
Updated to use lastest NodeJS version (14.17.0)

PR #1975, opened 2021-02-11
Upgrade plugin logger version to 3.0.2

alexandrebouthinon:
Upgrade Kuzzle plugin logger version to 3.0.2. The 3.0.2 use `winston-syslog=2.4.4` which fix compilation issues

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "jsonwebtoken": "^8.5.1",
     "koncorde": "^3.0.0",
     "kuzzle-plugin-auth-passport-local": "^6.2.0",
-    "kuzzle-plugin-logger": "^2.0.7",
+    "kuzzle-plugin-logger": "^3.0.2",
     "kuzzle-sdk": "^7.5.3",
     "kuzzle-vault": "^2.0.4",
     "lodash": "4.17.20",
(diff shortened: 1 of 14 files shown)


Yoann-Abbes (approved):

Shiranuit (approved):

scottinet (approved):
This PR does quite a lot more than just upgrading our dependencies :sweat_smile:

Aschen:
> This PR does quite a lot more than just upgrading our dependencies sweat_smile

Because we had to fix the cluster tests (they weren't in the CI)

package.json
```diff
@@ -11,7 +11,7 @@
     "start:dev": "node docker/scripts/reloader.js --inspect=0.0.0.0:9229 -r ts-node/register docker/scripts/start-kuzzle-dev.ts --enable-plugins functional-test-plugin",
     "prepublishOnly": "npm run build",
     "build": "npm run build-ts",
-    "clean": "npm run build | grep TSFILE | cut -d' ' -f 2 | xargs rm",
+    "clean": "touch index.ts && npm run build | grep TSFILE | cut -d' ' -f 2 | xargs rm",
```
alexandrebouthinon, line 14: A better and cleaner way to do that:
```suggestion
    "clean-ts": "git clean -fdX -e node_modules",
    "clean": "git clean -fdX",
```

Aschen: I don't like this command because it removes EVERY untracked files from the working dir :expressionless:  
I just loss some temp working files

Aschen: When it comes to removing files I want to know exactly what I am removing

PR #1977, opened 2021-02-17
Add full path in notification updated fields

Aschen:
## What does this PR do ?

The `_updatedFields` inside notifications only returns names of first level keys.   
Now it will return complete path to every leaf values that changed.  

 - Create an index and a collection: `kourou collection:create test test`
 - Create a document: `kourou document:create test test doc-1 '{ foo: "bar" }'`
 - Subscribe to notifications: `kourou sdk:execute  'return sdk.realtime.subscribe("test", "test", {}, console.log)'`
 - Update it: `kourou document:update test test doc-1 '{ foo: "bar", nested: { obj: { value: 42 } } }'`

```js
{
  status: 200,
  requestId: '6348f29d-0d5f-4c37-86e4-e59df2e089e7',
  timestamp: 1613553208086,
  volatile: {
    client: 'kourou@0.18.0',
    sdkInstanceId: 'c669c180-254d-4bdf-9c50-59a7b1d37dcd',
    sdkName: 'js@7.5.3'
  },
  index: 'test',
  collection: 'test',
  controller: 'document',
  action: 'update',
  protocol: 'websocket',
  scope: 'in',
  result: {
    _id: 'doc-1',
    _source: { foo: 'bar', _kuzzle_info: [Object], nested: [Object] },
    _updatedFields: [ 'foo', 'nested.obj.value' ]
  },
  type: 'document',
  node: 'wide-historian-1693',
  event: 'write',
  room: 'b0c0382588bc32c6c352d5a75484c8b3-cf3d9b363d487dd10bf35c786e06879b'
}
```

### Other changes

 - fix translation of `in` Koncorde clause to ES `terms` clause


lib/service/storage/queryTranslator.js
```diff
@@ -68,7 +68,7 @@ const KONCORDE_CLAUSES_TO_ES = {
     }
   }),
   in: content => ({
-    term: {
+    terms: {
```
Leodau, line 71: Wouldn't this be a breaking change? (given that for a time, **term** was the usable 'term')

lib/api/controller/document.js
```diff
@@ -744,6 +741,25 @@ class DocumentController extends NativeController {
       successes: response.items
     };
   }
+
+  _extractUpdatedFields (document, path = null, paths = []) {
```
Leodau, line 745: Why would this function define so many other possibilities and usages (given those extra params) if we don't use them :thinking: ?

scottinet:
Could be a bug fix instead of an enhancement.

lib/service/storage/queryTranslator.js
scottinet: To me it's not a breaking change, since the external API doesn't change. Rather, it's a bug fix since a `in` term in Koncorde allows to match multiple values, which needs the `terms` key in ES queries (`term` only allows matching a single value)

scottinet (approved):

alexandrebouthinon (approved):
Nit: Since it modify response, maybe we should add a check in document:update functional test

Aschen:
> Nit: Since it modify response, maybe we should add a check in document:update functional test

I would if the realtime functional tests weren't done with the legacy framework :sweat_smile:

PR #1990, opened 2021-02-26
Expose vault before startup

Aschen:
## What does this PR do ?

Allow to use the secrets vault before application startup.

This allows to use a secret inside the configuration for example.

I couldn't find a proper solution since the Vault is initialized when Kuzzle object is instantiated and we cannot instantiate it before application startup because then the custom configuration will not be taken into account. 

```
app.config.set('services.storageEngine.client', {
  node: 'http://elasticsearch:9200',
  auth: {
    username: 'elastic',
    password: app.vault.secrets.xpack.password,
  }
});
```

### Boyscout

  - add missing license header


lib/core/backend/backend.ts
```diff
@@ -293,11 +297,18 @@ class BackendVault extends ApplicationManager {
    * Decrypted secrets
    */
   get secrets () : JSONObject {
-    if (! this._application.started) {
-      throw runtimeError.get('unavailable_before_start', 'vault.secrets');
+    // We need to load the secrets before Kuzzle start so when can use them
```
alexandrebouthinon, line 300:
```suggestion
    // We need to load the secrets before Kuzzle start so we can use them
```

alexandrebouthinon (approved):
👍

scottinet (approved):

PR #1991, opened 2021-02-28
Prefix ApiKey tokens and node name

Aschen:
## What does this PR do ?

We are prefixing generated kuid for the sake of readability, so users know instantly to what kind of object correspond to the ID (`kuid-xxx-xxx-42`).

This PR adds a prefix for:
 - Node name `knode-` (e.g. `knode-responsive-pankhurst-54260`)
 - ApiKey authentication token `kapikey-` (e.g. `kapikey-eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJfaWQiOiJrdWlkLWZsYXQtZXBpY3VydXMtMjYzMDgiLCJpYXQiOjE2MTQ0OTMwOTF9.BBdNKG4IJa7nWEDZQaeAU-QMD9f8Kpj2t1HIkn4OPcg`)

The goal here is to allow users to manipulate understandable strings and not just obscure IDs.

I was also thinking about prefixing the authentication token but I'm not sure about this one.  

Also, if you have any other ideas I'm in :+1: 

(I will add unit tests if we agreed on this)

### Other changes
 
  - The name generator method now takes a `prefix` argument


scottinet (commented):
Proposal: `kauth` instead of `kapikey`, and use the same prefix for regular authentication tokens.

That way, you don't need to add/remove the prefix for API keys.

Added advantage: this don't leak more information about the nature of the auth token. I don't think this is a security problem by any mean, but I tend to be super-paranoid when it comes to that kind of subject.

Speaking of which... is it really a good idea to add a "kuzzle" prefix to auth tokens, which gives information about what kind of backend is installed for a given app?

wdyt?

Aschen:
My goal was to be able to differentiate API keys and simple authentication token :sweat_smile: 

About the security concern, to me the biggest disadvantage is that you can search for kuzzle token on internet, github, with regex etc.
It's already very easy to know that the backend is Kuzzle (API handshake, errors code, etc)

I know that Stripe have prefix on their API keys for example and they handle billions of payments: https://stripe.com/docs/api/authentication

scottinet:
hm... good point about searchable tokens. But this can be an advantage too: it allows for easy security checks forbidding the release of private information on public platforms, with a mere regex :thinking:

With your explanation I better see the goal of that PR. It can be convenient for users to be able to easily distinguish between regular auth token and api keys.

This would also allow us to lazily load API keys, instead of being forced to upload them in redis during startup, reducing the risks for desyncs.

I'll recheck your PR, there are a few things that bugs me in your implementation.

lib/core/security/tokenRepository.js
```diff
@@ -393,13 +394,16 @@ class TokenRepository extends Repository {
       debug('Loading API keys into Redis');
 
       const promises = [];
-      const createToken = async ({token, ttl, userId }) => {
-        return await this.persistForUser(token, userId, ttl);
+      const createToken = async apiKey => {
+        return await this.persistForUser(
+          apiKey.authToken(),
+          apiKey._source.userId,
+          apiKey._source.ttl);
```
scottinet, line 401: Why not include the token prefixes when you store it?
And in all cases, in fact.
You could isolate the prefix removal in the token verification code (with a `@deprecated` check to skip prefix removal if there isn't one, to ensure backward-compatibility).

I'm asking because I worry about potential inconsistencies, where in some parts of the code the tokens don't have prefixes, and where they do in some other parts. It can be hard to follow and to debug.

Especially with the fingerprint that seems to be computed w/ the prefix, whereas the token is stored without one.

scottinet (changes requested):

lib/core/security/tokenRepository.js
Aschen: Very good point :+1:

scottinet (approved):

alexandrebouthinon (approved):
Nice 👍 Could we prefix user and document auto generated ID like `kuser-xxx-xxx` for users and `kdoc-xxx-xxx`?

scottinet:
Technically we could: ES only generates an ID if we don't provide one. But I don't think that we should: first this means that we must check for id collisions (even w/ uuid, collisions might occur on some edge cases), and ES probably does that in a very efficient way. Moreover, I don't think that a prefix on document ids will help much, documents are already clearly identified in payloads.

lib/core/security/tokenRepository.js
```diff
@@ -301,6 +309,12 @@ class TokenRepository extends Repository {
     return userToken;
   }
 
+  removeTokenPrefix (token) {
+    return token
+      .replace(TokenRepository.AUTH_PREFIX, '')
+      .replace(TokenRepository.APIKEY_PREFIX);
```
Leodau, line 315: Won't the second **.replace** insert a `'undefined'` if no **APIKEY** is found?

Leodau (changes requested):

Leodau (approved):

PR #1996, opened 2021-03-03
Allows dynamic pipe registration

Aschen:
## What does this PR do ?

Allows to dynamically register pipes.

When the application is already started, `app.pipe.register` can be used with the `dynamic: true` option to dynamically register a pipe.  
The method will returns a `pipeId` which can be then used later to unregistere the pipe with `app.pipe.unregister`.

Defining dynamic pipes at return need to maintain a state of the registered pipes so new nodes joining the cluster can also register the same pipes.  
This dynamic state management is not handled by Kuzzle so developers should implements their own state management routines.

This feature is meant to be left undocumented since it's quite and advanced usage and we didn't had other use cases than our Workflows plugin.

  - Step 1 : Run functional tests

### Other

 - Expose an usable version of Koncorde

### DIFF
--- a/features/support/hooks.js
+++ b/features/support/hooks.js
@@ -91,6 +91,12 @@ Before({ tags: '@http' }, async function () {
   }
 });
 
+Before({ tags: '@not-http' }, async function () {
+  if (process.env.KUZZLE_PROTOCOL !== 'websocket') {
+    return 'skipped';
+  }
+});
+
 // firstAdmin hooks ============================================================
 
 Before({ tags: '@firstAdmin' }, async function () {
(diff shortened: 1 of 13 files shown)


scottinet (approved):

alexandrebouthinon (approved):

features/support/hooks.js
```diff
@@ -91,6 +91,12 @@ Before({ tags: '@http' }, async function () {
   }
 });
 
+Before({ tags: '@not-http' }, async function () {
+  if (process.env.KUZZLE_PROTOCOL !== 'websocket') {
```
alexandrebouthinon, line 95:
```suggestion
  if (process.env.KUZZLE_PROTOCOL === 'http') {
```

alexandrebouthinon (changes requested):

alexandrebouthinon (approved):

Leodau (approved):

PR #2003, opened 2021-03-04
Fix import of ACLs w/ credentials at startup

scottinet:
# Description

When starting Kuzzle with ACL fixtures, if users to import contain credentials, then the import fails with an error telling that whatever authentication strategy is used doesn't exist.
And for good reason: plugins are initialized only after the ACLs are imported, making it impossible to import users with credentials that way.

This PR fixes the issue by reordering the starting sequence a bit: plugins are now started before ACLs are imported.
This should not cause any issue (and even more so if the application framework is used), as plugins usually don't need rights to execute requests.


package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "kuzzle",
   "author": "The Kuzzle Team <support@kuzzle.io>",
-  "version": "2.10.0",
+  "version": "2.10.1",
   "description": "Kuzzle is an open-source solution that handles all the data management through a secured API, with a large choice of protocols.",
   "bin": {
-    "kuzzle": "./bin/start-kuzzle-server"
+    "kuzzle": "bin/start-kuzzle-server"
```
Leodau, line 7: :eyes:

Leodau (approved):

package.json
scottinet: wat

scottinet: I just ran some tests: it looks like `npm version patch` also sanitizes the package.json file :thinking: 

I don't think that this changes anything. Do you want me to revert that change (made by NPM)?

scottinet: I upgraded to NPM7 a couple days ago, so it might be a new feature.

Leodau: Oh this was a NPM7 doing? :eyes: 
Go for it then, it would be modified by sanitization anyway.
:+1:

alexandrebouthinon (approved):

PR #2005, opened 2021-03-04
Deprecate listing of realtime collections

### LINKED ISSUE #1822: Deprecate listing of realtime collections
## Feature description

We should deprecate the listing of realtime collections:
 - when the action is called with default params: do nothing
 - when the action is called with `type: 'realtime'`: add deprecation

Also deprecate in the documentation

### Possible implementation

Check the `type` parameter of the `collection:list` action and add a [deprecation](https://github.com/kuzzleio/kuzzle-common-objects#addDeprecationversion-message) on the request

MathieuVeber:
Closes #1822

## What does this PR do ?

Add deprecation to realtime listed collections route

Try to access http://localhost:7512/index/_list?type=realtime of your instance of kuzzle in local and you see in the result a deprecated field

### Other changes
Doc updated

### Boyscout

- Fix a command given in this [section](https://docs.kuzzle.io/core/2/guides/getting-started/customize-api-behavior/#api-events) of the tutorial in order to be consistent with docker.
-  Adding deprecated for some actions in collectionController in order to match the doc

### DIFF
--- a/lib/api/controllers/collectionController.js
+++ b/lib/api/controllers/collectionController.js
@@ -300,7 +300,7 @@ class CollectionController extends NativeController {
     }
 
     if (request.input.args.type) {
-      request.addDeprecation('auto-version',
+      request.addDeprecation('2.10.2',
         'The "type" argument and this route returning a list of realtime collections have both been deprecated since Kuzzle version 2.10.1. This feature might be removed in a future major version. To get a list of realtime collections, use this API action instead: "realtime:list".');
     }
 
--- a/lib/api/controllers/collectionController.js
+++ b/lib/api/controllers/collectionController.js
@@ -297,7 +297,7 @@ class CollectionController extends NativeController {
     }
 
     if (request.input.args.type) {
-      request.addDeprecation('2.10.1',
+      request.addDeprecation('auto-version',
         'The "type" argument and this route returning a list of realtime collections have both been deprecated since Kuzzle version 2.10.1. This feature might be removed in a future major version. To get a list of realtime collections, use this API action instead: "realtime:list".');
     }
 
--- a/lib/api/controllers/collectionController.js
+++ b/lib/api/controllers/collectionController.js
@@ -58,6 +58,9 @@ class CollectionController extends NativeController {
    * @returns {Promise.<Object>}
    */
   async updateMapping (request) {
+    request.addDeprecation('2.1.0',
+      'This action has been deprecated since Kuzzle version 2.1.0. This feature might be removed in a future major version. To update a collection, use this API action instead: "collection:update".');
+
     const
       { index, collection } = this.getIndexAndCollection(request),
       mapping = this.getBody(request);
@@ -353,6 +356,8 @@ class CollectionController extends NativeController {
     // @deprecated sending directly the mappings is deprecated since 2.1.0
     if (body.properties || body.dynamic || body._meta) {
       config.mappings = body;
+      request.addDeprecation('2.1.0',
+        'Using "properties", "dynamic" or "_meta" fields in the body has been deprecated since Kuzzle version 2.1.0. This feature might be removed in a future major version. Use "settings" and "mappings" fields instead');
     }
     else {
       config = body;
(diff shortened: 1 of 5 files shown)


lib/config/httpRoutes.js
```diff
@@ -44,6 +44,7 @@ const routes = [
   { verb: 'get', path: '/:index/:collection/_search', controller: 'document', action: 'search' },
   { verb: 'get', path: '/:index/:collection/_specifications', controller: 'collection', action: 'getSpecifications' },
   { verb: 'get', path: '/validations/_scroll/:scrollId', controller: 'collection', action: 'scrollSpecifications' },
+  { verb: 'get', path: '/:index/_list&type=realtime', controller: 'collection', action: 'list', deprecated: { since: '2.1.4', message: 'Use this route instead: http://kuzzle:7512/_listSubscriptions' } }, // @deprecated
```
scottinet, line 47: Doesn't work if the invoked URL isn't exactly that one.

For instance, this URL won't return the deprecation notice because of the `pretty` argument:

`http://localhost:7512/foo/_list?pretty&type=realtime`

Also doesn't work if the API action is invoked using another protocol than HTTP.

The way I see the linked issue, the whole `type` argument should be deprecated, meaning that the default output (which uses `type=all`) should also print a meaningful message.

About the message: it should be more explicit: what is deprecated exactly?
Something along these lines could do: `The list of realtime collections has been deprecated since Kuzzle version <version>, and might be removed in a future major version. To get a list of realtime collections, use this API action instead: "realtime:list".`

Also, why is it deprecated since Kuzzle 2.1.4? :thinking: 
We should never deprecate a feature retroactively, it's confusing for users.

Lastly, the documentation should also be updated, explaining what is deprecated, why, since which version, and what to use instead.

scottinet (changes requested):

lib/config/httpRoutes.js
MathieuVeber: About why 2.1.4 and updating the doc, it's because it's already deprecated see [here](https://docs.kuzzle.io/core/2/api/controllers/collection/list/#optional), I'll correct the other points

Aschen: I think we have a design flaw here because it was only the pagination options that were deprecated with the `2.1.4`. The badges should be put on the corresponding options lines

Aschen:
IMHO the change-log tag should not be `new-feature` but rather `enhancement`

lib/api/controllers/collectionController.js
```diff
@@ -296,6 +296,11 @@ class CollectionController extends NativeController {
       throw kerror.get('api', 'assert', 'invalid_argument', '"all", "stored", "realtime"');
     }
 
+    if (request.input.args.type) {
+      request.addDeprecation('2.10.1',
```
Aschen, line 300: You should use the `auto-version` string instead of the version number. (Your feature will probably be released in the 2.10.2)

Aschen (changes requested):

lib/api/controllers/collectionController.js
MathieuVeber: Done! (Didn't know it also worked outside DeprecatedBadge)

lib/api/controllers/collectionController.js
```diff
@@ -296,6 +299,11 @@ class CollectionController extends NativeController {
       throw kerror.get('api', 'assert', 'invalid_argument', '"all", "stored", "realtime"');
     }
 
+    if (request.input.args.type) {
+      request.addDeprecation('auto-version',
```
scottinet, line 303: Unfortunately, the `auto-version` replacement currently only works for the documentation. And I don't know if we want to automatically make the replacement in the code itself. :thinking: 

I think that you'll need to put a hard-coded version number there. For now the next version will be `2.10.2`.

Aschen: IMHO we could look into the code to replace those strings as well

Aschen: Maybe we could have something more explicit like `{auto-version}`

Aschen (approved):

lib/api/controllers/collectionController.js
scottinet: Could be a lead. In that case, I would go full explicit, with something like: `release-tools:auto-version` to make sure that we can't mistake this with any other string.

Leodau: I'm not fund of modifying strings within the code, how can we know that we would **never** write `release:auto-version` in a comment or a log?

Issue #1791, opened 2020-09-24
Adds a "strict" option to the m* actions to return an error if some action fails


Aschen:
## Feature description

When using the `document:m*` or the `bulk:import/mWrite` action  (`document:mCreate`, etc), the action returns an object with 2 array `successes` and `errors`.

The API always returns a 200 ok status and if there is any error-ed actions, they will be in the `errors` array.

We should add a `strict` option to those actions to return an error if there is at least one error-ed action. This option will be `false` by default to not introduce breaking changes.

Also if there is no elements in the `successes` array, the API should return an error.

## Possible solution (optional)

Adds the `strict` option to the following routes:
 - `document:m*`
 - `bulk:import`
 - `bulk:mWrite`

Creates a new error code that indicate that at least one of the action failed or none succeed.

PR #2007, opened 2021-03-08
Add a "strict" option to the m* actions to return an error if some action fails

MathieuVeber:
Closes [#1791](https://github.com/kuzzleio/kuzzle/issues/1791)

## What does this PR do ?
Now you can add a strict option to the multiple actions so that it throw a Partial Error (206) if some of the actions fails (instead of just listing errors in the result)

### Other changes
New error, updated doc, tests added


Aschen (approved):

scottinet (changes requested):
LGTM

But you have 2 small changes to perform before this PR gets accepted, I don't believe we briefed you on those ones:

* when adding new error codes, you need to run the `doc/build-error-codes.js` script and push the changes: it updates the error codes documentation. That's why you have 1 check failing in the CI (we verify that the documentation is synchronized with our error codes)
* you should never push changes to the `package-lock.json` file, unless you made some changes in `package.json`
*

lib/kerror/codes/2-api.json
```diff
@@ -140,6 +140,12 @@
           "code": 9,
           "message": "Admin user is already set.",
           "class": "PreconditionError"
+        },
+        "incomplete_multiple_request": {
+          "description": "Failed to execute some or all actions requested",
+          "code": 10,
+          "message": "At least one of the %s actions failed.",
+          "class": "PartialError"
```
scottinet, line 148: You shouldn't throw a PartialError, because that error type has a 206 status code. This, in fact, should be a Partial Success return type, and not an error class (the PartialError class is a bit of an anomaly, we might remove it in the future)

scottinet (changes requested):

doc/2/api/controllers/document/m-get/index.md
```diff
@@ -58,6 +58,10 @@ Method: GET
 - `collection`: collection name
 - `index`: index name
 
+### Optional:
+
+- `strict`: if set, an error will occur if at least one document could not been retrieved <SinceBadge version="auto-version" />
```
Leodau, line 63:
```suggestion
- `strict`: if set, an error will occur if any of the documents could not be retrieved <SinceBadge version="auto-version" />
```

doc/2/api/controllers/document/m-update/index.md
```diff
@@ -82,6 +82,7 @@ Body:
 - `refresh`: if set to `wait_for`, Kuzzle will not respond until the updates are indexed
 - `retryOnConflict`: conflicts may occur if the same document gets updated multiple times within a short timespan in a database cluster. You can set the `retryOnConflict` optional argument (with a retry count), to tell Kuzzle to retry the failing updates the specified amount of times before rejecting the request with an error.
 - `silent`: if set, then Kuzzle will not generate notifications <SinceBadge version="2.9.2" />
+- `strict`: if set, an error will occur if at least one document has not been updated <SinceBadge version="auto-version" />
```
Leodau, line 85:
```suggestion
- `strict`: if set, an error will occur if a document was not updated <SinceBadge version="auto-version" />
```

Leodau (changes requested):
I think I found some _frenglish_ wording in the documentation,

Could you please update any of the **"if at least + could not been"** wording, by something more like the 2 examples given below?

- By the use of **'any of the'**
- Or, bringing the main subject to a singular form. _(documents-> a document)_

lib/api/controllers/documentController.js
```diff
@@ -170,16 +170,22 @@ class DocumentController extends NativeController {
...
     this.assertNotExceedMaxFetch(ids.length);
 
-    return this.ask('core:storage:public:document:mGet', index, collection, ids)
-      .then(({ items, errors }) => {
-        return {
-          errors,
-          successes: items
-        };
-      });
+    const {items, errors} = await this.ask('core:storage:public:document:mGet', index, collection, ids);
```
Leodau, line 177:
```suggestion
    const { items, errors } = await this.ask('core:storage:public:document:mGet', index, collection, ids);
```

lib/kerror/codes/2-api.json
MathieuVeber: Oh this makes sense but it questions the purpose of the issue, @Aschen may had another thing in mind while creating this issue... Should we discuss about this?

Aschen: It should be a BadRequestError to me, even if some documents succeed to be written

scottinet: Hm, bad requests are more for erroneous requests (invalid syntax/payload).

I don't have a ready-to-use solution, so here are a few thoughts:
* we do have individual error, for each failed document. If there is 1 error, or if all errors are of the same type, we could throw that error instead. Problem is, our current error system does not handle polymorphic errors
* we could have some kind of error wrapper, used to aggregate multiple errors. That could be useful for the validation module (when we rework it). Let's call it `MultipleError` for the sake of example (name to be debated). Since this is not something that exists in HTTP error codes, and since we don't know what kind of underlying error it contains, we could use the generic `400` error code. I know it's the BadRequest one, but I lack better idea (least we use a currently unused HTTP code for our own needs)

Aschen: I like the second option :+1:

MathieuVeber: Let's go for it then, I'll opt for 400 for this MultipleErrorsError

Aschen: `MultipleError` ?

scottinet (approved):

Aschen (changes requested):

Aschen (approved):

Leodau (approved):
GJ :clap:

Leodau (changes requested):
Could you please explore the idea of adding (or modifying) a functional test?
Would it be interesting/required for this 'enhancement'?

lib/kerror/errors/multipleErrorsError.ts
```diff
@@ -38,11 +38,16 @@ export class MultipleErrorsError extends KuzzleError {
...
   toJSON () {
-    const serialized = super.toJSON();
+    const serialized = super.toJSON();    
 
-    serialized.errors = this.errors.map(error => error.toJSON());
     serialized.count = this.count;
 
+    try {
+      serialized.errors = this.errors.map(error => error.toJSON());
+    } catch (error) {
```
Aschen, line 47: Wrong indentation rules

lib/kerror/errors/multipleErrorsError.ts
```diff
@@ -38,11 +38,16 @@ export class MultipleErrorsError extends KuzzleError {
   }
 
   toJSON () {
-    const serialized = super.toJSON();
+    const serialized = super.toJSON();
```
Aschen, line 41:
```suggestion
    const serialized = super.toJSON();
```

Aschen (changes requested):

Aschen (approved):
LGTM :+1:

Leodau (approved):

PR #2011, opened 2021-03-10
Allow fine-tuning Redis connections using Kuzzle configuration

### LINKED ISSUE #2010: Allow fine-tuning Redis connections using Kuzzle configuration
# Main task

At this moment only the Elasticsearch can be configured using options given in the Kuzzle configuration file. Problem is there are edge cases when dialling with PaaS Redis services:
```js
  _buildClient () {
    return new IORedis(this._config.node);
  }

  _buildClusterClient () {
    return new IORedis.Cluster(this._config.nodes, { enableReadyCheck: true });
  }
```
We need to ensure users could pass the IORedis constructor options object using the Kuzzle configuration file.

# Other tasks

## Be able to override `dnsLookup` callback

This is required when connecting to an AWS ElastiCache Cluster with In-transit encryption enabled (see [IORedis documentation](https://github.com/luin/ioredis/blob/master/README.md#special-note-aws-elasticache-clusters-with-tls)).
I suggest `overrideDnsValidation` boolean option in Redis config (`memoryStorage` and `internalCache`) in kuzzlerc.

```json

{
  "services": {
    "internalCache": {
      "backend": "redis",
      "database": 0,
      "overrideDnsValidation": true,
      "nodes": [{
        "host": "clustercfg.prod-skybet-kuzzle.o6xc43.use1.cache.amazonaws.com",
        "port": 6379
      }]
    },
    "memoryStorage": {
      "backend": "redis",
      "database": 0,
      "overrideDnsValidation": true,
      "nodes": [{
        "host": "clustercfg.prod-skybet-kuzzle.o6xc43.use1.cache.amazonaws.com",
        "port": 6379
      }]
    }
  }
}
```

## Be able to use other databases IDs with Clusters
For now, we can only use Redis database index `0`. Due to those lines in the [init method](https://github.com/kuzzleio/kuzzle/blob/8668a12df05315686520d9043d1fdd5b784d22de/lib/service/cache/redis.js#L53):
```js
        if (this._config.database > 0) {
          this.client.select(
            this._config.database,
            error => error ? reject(error) : resolve());
        }
        else {
          resolve();
        }
```
`select` action is not allowed in Redis Cluster context.
I suggest to use the `options.db` in the [IORedis client](https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options)

alexandrebouthinon:
## What does this PR do ?

fix #2010 

Launch Kuzzle stack and play with IORedis configuration from Kuzzle configuration

### Boyscout

Improve Redis mocks and add RedisClusterClient mock


alexandrebouthinon:
Unit tests freeze but you can still review the code

.kuzzlerc.sample
```diff
@@ -476,19 +476,29 @@
...
+    //       https://github.com/luin/ioredis/blob/master/API.md#redis--eventemitter
     "internalCache": {
       "backend": "redis",
+      "database": 0,
       "node": {
         "host": "localhost",
         "port": 6379
+      },
+      "options": {
+        "db": 0
```
scottinet, line 501: When dealing with deprecated options, you should make sure that behaviors are clearly defined, otherwise this makes things confusing for users and this can be detrimental when migrating Kuzzle to the next major version.

Here is what I propose:

* remove the `database` option from the `.kuzzlerc.sample` configuration file, keep it in the documentation, and explain that `options.db` takes precedence over it
* remove `options.db` from the default configuration file (`default.config.js`), and use the old `database` property instead: that way, older client configurations will still override it, and newer ones will make it ignored
* add a `@deprecated` tag in the redis service code, when using that `database` variable, to make sure we'll spot it and remove it when working on the next major version

.kuzzlerc.sample
```diff
@@ -476,19 +476,29 @@
...
     // 1. using a single Redis database:
-    //   node:
+    //   * node:
     //     * host:
     //         The host on which Redis can be reached.
     //         Can take an IP address, an URI or a hostname
     //     * port:
     //         The port on which Redis is running its database:
-    //     * (optional) database:
+    //     * (optional) database (DEPRECATED):
```
scottinet, line 485:
```suggestion
    //     * (optional, deprecated) database:
```

.kuzzlerc.sample
```diff
@@ -476,19 +476,29 @@
...
     //         ID of the redis database (default: 0)
+    //         NOTE: this option is deprecated and will be removed in Kuzzle v3.
+    //               Considere using of options.db as a replacement.
```
scottinet, line 488:
```suggestion
    //               Consider using "options.db" as a replacement. If both options are
    //               set, then "options.db" will take precedence.
```

.kuzzlerc.sample
```diff
@@ -476,19 +476,29 @@
...
+    //   * (optional) options: 
+    //       Redis specific options compatible with IORedis.
+    //       See Redis client constructor available options:
+    //       https://github.com/luin/ioredis/blob/master/API.md#redis--eventemitter
```
scottinet, line 492:
```suggestion
    //       https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options
```

.kuzzlerc.sample
```diff
@@ -504,17 +514,27 @@
...
     //         or more slaves
-    //     * (optional) database:
+    //     * (optional) database (DEPRECATED):
     //         ID of the redis database (default: 0)
+    //         NOTE: this option is deprecated and will be removed in Kuzzle v3.
+    //               Considere using of options.db as a replacement.
+    //   * (optional) options: 
+    //       Redis specific options compatible with IORedis.
+    //       See Redis client constructor available options:
+    //       https://github.com/luin/ioredis/blob/master/API.md#redis--eventemitter
```
scottinet, line 524: apply changes from above comments

.kuzzlerc.sample
```diff
@@ -524,28 +544,138 @@
     //         Can be an IP address, an URI or a hostname
     //     * port:
     //        Network port opened by the redis cluster node
+    //   * (optional) options: 
+    //       Redis specific options compatible with IORedis.
+    //       See Redis client constructor available options:
+    //       https://github.com/luin/ioredis/blob/master/API.md#redis--eventemitter
```
scottinet, line 550:
```suggestion
    //       https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options
```

.kuzzlerc.sample
```diff
@@ -524,28 +544,138 @@
...
+    //   * (optional) clusterOptions: 
+    //       Redis Cluster specific options compatible with IORedis.
+    //       See Redis Cluster client constructor available options:
+    //       https://github.com/luin/ioredis/blob/master/API.md#cluster--eventemitter
```
scottinet, line 554:
```suggestion
    //       https://github.com/luin/ioredis/blob/master/API.md#new-clusterstartupnodes-options
```

.kuzzlerc.sample
```diff
@@ -524,28 +544,138 @@
...
+    //   * (optional) overrideDnsLookup: 
+    //       Only available when using a Redis Cluster config.
+    //       Use with caution, if set to true override DNS validation for TLS certificates
```
scottinet, line 557:
```suggestion
    //       Use with caution: if set to true, overrides the DNS validation for TLS certificates
```

.kuzzlerc.sample
```diff
@@ -524,28 +544,138 @@
...
+    //     * port:
+    //         The port on which Redis is running its database:
+    //     * (optional) database (DEPRECATED):
+    //         ID of the redis database (default: 0)
+    //         NOTE: this option is deprecated and will be removed in Kuzzle v3.
+    //               Considere using of options.db as a replacement.
+    //   * (optional) options: 
+    //       Redis specific options compatible with IORedis.
+    //       See Redis client constructor available options:
+    //       https://github.com/luin/ioredis/blob/master/API.md#redis--eventemitter
```
scottinet, line 595:
```suggestion
    //       https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options
```

.kuzzlerc.sample
```diff
@@ -524,28 +544,138 @@
...
+    //   * node:
+    //     * host:
+    //         The host on which Redis can be reached.
+    //         Can take an IP address, an URI or a hostname
+    //     * port:
+    //         The port on which Redis is running its database:
+    //     * (optional) database (DEPRECATED):
+    //         ID of the redis database (default: 0)
+    //         NOTE: this option is deprecated and will be removed in Kuzzle v3.
+    //               Considere using of options.db as a replacement.
```
scottinet, line 591: apply above requested changes

.kuzzlerc.sample
```diff
@@ -524,28 +544,138 @@
...
+    //         Group of Redis instances composed of a master and one
+    //         or more slaves
+    //     * (optional) database (DEPRECATED):
+    //         ID of the redis database (default: 5)
+    //         NOTE: this option is deprecated and will be removed in Kuzzle v3.
+    //               Considere using of options.db as a replacement.
+    //   * (optional) options: 
+    //       Redis specific options compatible with IORedis.
+    //       See Redis client constructor available options:
+    //       https://github.com/luin/ioredis/blob/master/API.md#redis--eventemitter
```
scottinet, line 627:
```suggestion
    //       https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options
```

.kuzzlerc.sample
```diff
@@ -524,28 +544,138 @@
...
+    //   * nodes: array of master nodes of the cluster
+    //     * host:
+    //         Host name/address of a redis cluster node
+    //         Can be an IP address, an URI or a hostname
+    //     * port:
+    //        Network port opened by the redis cluster node
+    //   * (optional) options: 
+    //       Redis specific options compatible with IORedis.
+    //       See Redis client constructor available options:
+    //       https://github.com/luin/ioredis/blob/master/API.md#redis--eventemitter
```
scottinet, line 653:
```suggestion
    //       https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options
```

.kuzzlerc.sample
```diff
@@ -524,28 +544,138 @@
...
+    //   * (optional) clusterOptions: 
+    //       Redis Cluster specific options compatible with IORedis.
+    //       See Redis Cluster client constructor available options:
+    //       https://github.com/luin/ioredis/blob/master/API.md#cluster--eventemitter
```
scottinet, line 657:
```suggestion
    //       https://github.com/luin/ioredis/blob/master/API.md#new-clusterstartupnodes-options
```

.kuzzlerc.sample
```diff
@@ -524,28 +544,138 @@
...
+    //   * (optional) overrideDnsLookup: 
+    //       Only available when using a Redis Cluster config.
+    //       Use with caution, if set to true override DNS validation for TLS certificates
```
scottinet, line 660:
```suggestion
    //       Use with caution: if set to true, it overrides the DNS validation for TLS certificates
```

lib/service/cache/redis.js
```diff
@@ -51,11 +51,22 @@ class Redis extends Service {
    * @returns {Promise}
    */
   _initSequence() {
+    // @deprecated @todo Remove those lines for Kuzzle v3 then 
+    // remove 'database' from .kuzzlerc.sample and default.config.js
+    if (this._config.database) {
+      this._config.options = { ...this._config.options, db: this._config.database };
+    }
```
scottinet, line 58: This makes the deprecated option take precedence over the newer one, this should be the other way around (see my previous comment)

.kuzzlerc.sample
```diff
@@ -476,19 +476,29 @@
...
+    //   * node:
     //     * host:
     //         The host on which Redis can be reached.
     //         Can take an IP address, an URI or a hostname
     //     * port:
     //         The port on which Redis is running its database:
-    //     * (optional) database:
+    //     * (optional) database (DEPRECATED):
     //         ID of the redis database (default: 0)
+    //         NOTE: this option is deprecated and will be removed in Kuzzle v3.
```
Leodau, line 487: Are we sure we want to say it will be removed in v3?
We just pushed it from v2 to v3... :lollipop:

scottinet (changes requested):

.kuzzlerc.sample
Leodau, line 488:
```suggestion
    //               Please consider to use options.db as a replacement.
```
https://textranch.com/201155/we-consider-using/or/we-consider-to-use/

.kuzzlerc.sample
```diff
@@ -504,17 +514,27 @@
     //     * name:
     //         Group of Redis instances composed of a master and one
     //         or more slaves
-    //     * (optional) database:
+    //     * (optional) database (DEPRECATED):
     //         ID of the redis database (default: 0)
+    //         NOTE: this option is deprecated and will be removed in Kuzzle v3.
+    //               Considere using of options.db as a replacement.
```
Leodau, line 520: Ditto

.kuzzlerc.sample
Leodau, line 591: Ditto

.kuzzlerc.sample
```diff
@@ -524,28 +544,138 @@
...
+    //               Can be an IP address, an URI or a hostname
+    //           * port:
+    //               Network port opened by Redis on the sentinel server
+    //     * name:
+    //         Group of Redis instances composed of a master and one
+    //         or more slaves
+    //     * (optional) database (DEPRECATED):
+    //         ID of the redis database (default: 5)
+    //         NOTE: this option is deprecated and will be removed in Kuzzle v3.
+    //               Considere using of options.db as a replacement.
```
Leodau, line 623: Ditto

lib/service/cache/redis.js
```diff
@@ -51,11 +51,22 @@ class Redis extends Service {
    * @returns {Promise}
    */
   _initSequence() {
+    // @deprecated @todo Remove those lines for Kuzzle v3 then
```
Leodau, line 54:
```suggestion
    // @deprecated Remove those lines for Kuzzle v3 then 
```
We usually only add one of those tags, in this case it would be` @deprecated`

lib/service/cache/redis.js
```diff
@@ -51,11 +51,22 @@ class Redis extends Service {
...
+    // remove 'database' from .kuzzlerc.sample and default.config.js
+    if (this._config.database) {
+      this._config.options = { ...this._config.options, db: this._config.database };
+    }
+
     if (this._config.nodes) {
-      this.client = this._buildClusterClient();
+      this._config.clusterOptions = this._config.overrideDnsLookup
+        ? { ...this._config.clusterOptions, dnsLookup: (address, callback) => callback(null, address) }
+        : this._config.clusterOptions;
```
Leodau, line 63: I think this condition should be more verbose and less long :exploding_head: ?
```suggestion
      if (this._config.overrideDnsLookup) {
        this._config.clusterOptions.dnsLookup = (address, callback) => callback(null, address);
      }
```

Leodau (changes requested):

.kuzzlerc.sample
alexandrebouthinon: I just copy deprecation message for pcre support https://github.com/kuzzleio/kuzzle/blob/8668a12df05315686520d9043d1fdd5b784d22de/.kuzzlerc.sample#L739

lib/service/cache/redis.js
alexandrebouthinon: Nice catch but I'll need to keep object destructuring because `this._config.clusterOptions` can be undefined

.kuzzlerc.sample
alexandrebouthinon: I'll apply your first and second point, thanks 😄

.kuzzlerc.sample
Leodau: Yep, I figured... Just saying that how are we sure we will remove it v3? :headphones:

lib/service/cache/redis.js
Leodau: I figured, it is not in our style to construct properties on the fly like this. Can't we initialize this option elsewhere?
Why would the **init_sequence** modify directly the received config?

alexandrebouthinon: Ok found the problem. The undefined error on `clusterOptions` can only append on in unit tests (since object is initialised in the default config). I just need to update my tests to do so.

.kuzzlerc.sample
alexandrebouthinon: Because we always remove deprecated feature when release a new major version 😉

scottinet: Nope. Some deprecated features might stay in the next major version, for instance if the deprecation occurs not too long before the new major release, to allow our users to adjust.

scottinet (approved):

.kuzzlerc.sample
```diff
@@ -476,19 +476,29 @@
...
     //     * host:
     //         The host on which Redis can be reached.
     //         Can take an IP address, an URI or a hostname
     //     * port:
     //         The port on which Redis is running its database:
-    //     * (optional) database:
+    //     * (optional, deprecated) database: 
     //         ID of the redis database (default: 0)
+    //         NOTE: this option is deprecated and will be removed in Kuzzle v3.
+    //               Please consider to use 'options.db' as a replacement. If both
```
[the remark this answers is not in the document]
alexandrebouthinon: I prefer `Use 'options.db' instead` 👍

Leodau (approved):

PR #2012, opened 2021-03-10
Allow overriding configuration properties with JSON strings

scottinet:
# Description

Properties of the `kuzzlerc` configuration file can be overridden using environment variables. This is a native property of the `rc` module that we use to load configuration values.

Problem is: `rc` does only handle scalar values. Meaning that users cannot update arrays using environment variables.

This PR proposes to solve this issue by allowing users to pass JSON strings in their environment variable: with this PR, Kuzzle will parse those strings as a post-processing to the `rc` configuration loading.

This is done by passing the `*json:` prefix to a JSON string (see the updated documentation for examples).

# How to test this PR

Run Kuzzle with, for instance, this environment variable:

`kuzzle_foo__bar__baz='*json:{"hey": { "this": [ "is", "pretty", "cool" ] } }'`

And then run a `server:getConfig` action:

```bash
kourou server:getConfig
```

The configuration should show something like this: 

```
...
"foo": {
  "bar": {
    "baz": {
      "hey": {
        "this": [ "is", "pretty", "cool" ]
      }
    }
  }
}
...
```

### DIFF
--- a/doc/2/guides/advanced/configuration/index.md
+++ b/doc/2/guides/advanced/configuration/index.md
@@ -44,6 +44,16 @@ For example, the `.kuzzlerc` parameter `services.storageEngine.host` in example 
 export kuzzle_services__storageEngine__host="http://localhost:9200"
 ```
 
+You can also pass stringified JSON values this way to override non-scalar values such as objects or arrays.  
+To do so, prefix a valid stringified JSON with `*json:` to instruct Kuzzle to parse the content of the value as JSON.
+
+Examples:
+
+```bash
+export kuzzle_security__restrictedProfileIds='*json:["default","foo","bar"]'
+export kuzzle_services__common='*json:{"defaultInitTimeout":120000, "retryInterval":1000}'
+```
+
 ### Docker Compose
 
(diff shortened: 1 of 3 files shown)


doc/2/guides/advanced/configuration/index.md
```diff
@@ -44,6 +44,16 @@ For example, the `.kuzzlerc` parameter `services.storageEngine.host` in example
 export kuzzle_services__storageEngine__host="http://localhost:9200"
 ```
 
+You can also pass stringified JSON values this way to override non-scalar values such as objects or arrays.  
+To do so, prefix a valid stringified JSON with `*json:` to instruct Kuzzle to parse the content of the value as JSON.
```
Aschen, line 48: Why `*json` as placeholder ? (the `*`)

alexandrebouthinon: I think the is to avoid conflict with parameters name `jsonSomething`

alexandrebouthinon (approved):

Aschen (approved):

PR #2013, opened 2021-03-11
Throw when cookieAuth is present and CORS headers not properly configured

Shiranuit:
## What does this PR do ?

This PR ensure that Kuzzle will throw an error if the option cookieAuth is present when the CORS headers are not properly configured to support the cookie authentication  feature
- Correct funnel behaviour
- Tests to ensure everything works properly

### DIFF
--- a/lib/api/funnel.js
+++ b/lib/api/funnel.js
@@ -373,6 +373,16 @@ class Funnel {
    */
   async checkRights (request) {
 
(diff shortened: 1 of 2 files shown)


Leodau (approved):
Can you please update your description and tell us why we are suddenly removing 2 entire unit tests?

Aschen (approved):

Issue #1973, opened 2021-02-10
Implement a dedicated Role Controller


Leodau:
## Feature description
As the title suggests, create a new stand-alone **Role Controller** and update the **Security Controller** to alias/call the newly added controller actions. Make use of this opportunity to clean/update actions if required.

> :information_source:  To prevent breaking changes, the security controller would still be present, but deprecated. 

#### Also:

- Add a new `role:refresh` action to force [**es** refresh](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/docs-refresh.html).
- Update **docs** accordingly.
- :warning: Make sure to update our default user rights.

## Actions

- `role:create`
- `role:createOrReplace`
- `role:delete`
- `role:get`
- `role:mget`
- `role:mapping`
- `role:search`
- `role:update`
- `role:updateMapping`  :thinking:
- `role:refresh` :new:

PR #2016, opened 2021-03-13
Implement a dedicated User Controller

### LINKED ISSUE #1972: Implement a dedicated User Controller
## Feature description
As the title suggests, create a new stand-alone **User Controller** and update the **Security Controller** to alias/call the newly added controller actions. Make use of this opportunity to clean/update actions if required.

> :information_source:  To prevent breaking changes, the security controller would still be present, but deprecated. 

#### Also:

- Add a new `user:refresh` action to force [**es** refresh](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/docs-refresh.html).
- Update **docs** accordingly.
- :warning: Make sure to update our default user rights.

## Actions

- `user:create`
- `user:createRestricted`
- `user:delete`
- `user:get`
- `user:rights`
- `user:mapping`
- `user:strategies`
- `user:mDelete`
- `user:mGet`
- `user:replace`
- `user:revoke` :thinking: 
- `user:scroll`
- `user:search`
- `user:update`

Leodau:
## What does this PR do?

Towards the goal of splitting our huge `SecurityController`, creates a new standalone `UserController`, adds an abstraction layer `NativeSecurityController` and updates the `Security Controller`.

### Standalone UserController

1. Update documentation, deprecate old actions and migrate doc for new actions
2. Migrate and rename actions from SecurityController to UserController, deprecate the old ones
3. Update HTTP routes for each actions
4. Migrate unit tests for the new controller and check that new actions are called by old controller
5. Migrate or write functional tests for migrated actions

#### Summary of migrations

| SecurityController | UserController | Special Case |
| ------------------------ | ------------------- | ------------------ |
| `createUser` | `create` |
| `createRestrictedUser` | | Migrated in the AuthController as `auth:signin` |
| `getUser`| `get`|
| `mGetUsers`| `mGet`|
| `searchUsers`| `search`|
| `scrollUsers`| `scroll`|
| `updateUser`| `update`|
| `replaceUser`| `replace`|
| `deleteUser`| `delete`|
| `mDeleteUsers`| `mDelete`|
| `getUserMapping`| `getMappings`| Result: `{ mapping: {...} }` => `{ mappings, dynamic, _meta }` |
| `updateUserMapping`| `updateMappings`| Result: `{ properties: {...} }` => `{ properties, dynamic, _meta }` |
| `getUserRights`| `getRights`| Result: `{ hits: [...] }` => `{ rights: [...] }` |
| `checkRights`| `isAllowed`|
| `getUserStrategies`| `getStrategies`| Result: `{ strategies: [...], total: x }` => `{ strategies: [...] }` |
| `revokeTokens`| `revokeTokens`|
| `refresh` | `refresh`| `security:refresh` must still be used for `role` or `profile` collection |

#### About permissions

It would be costly, fastidious and useless after Kuzzle v3 to keep the same permissions between `security:oldAction` and `user:newAction`. Since Kuzzle determines permissions on a whitelist basis, every newly created actions will not be allowed unless it is explicitely stated. In that case, the developer is obviously up to date and **should be aware of this duplication of actions**. It is now his job to **watch out for incoherent permissions** during transition.

### AbstractSecurityController

From now, a security-related controller (aka `UserController`, `SecurityController`...) will extend `AbstractSecurityController` to take advantage of private generic functions since they share common logic.

#### Inheritance

- `UsualController` > `NativeController` > `BaseController`
- `SecurityRelatedController` > `AbstractSecurityController` > `NativeController` > `BaseController`

#### Generic methods

| Generic name | Used by |
| ------------------- | ------------- |
| `_get` | `security:getProfile`, `security:getRole`, `user:get` |
| `_mGet` | `security:mGetProfiles`, `security:mGetRoles`, `user:mGet` |
| `_delete` | `security:deleteProfile`, `security:deleteRole`, `user:delete` |
| `_mDelete` | `security:mDeleteProfiles`, `security:mDeleteRoles`, `user:mDelete` |
| `_refresh` | `security:refresh`, `user:refresh` |

### Other changes

- References to user related routes in the documentation updated
- Merge SecurityController method `_getSearchPageSize` into `request.getSearchParams` allowing to easily set a maximum default size as well as a useful 'NotExceedMaxFetch' check
- `security:createFirstAdmin` has been deprecated in favor of `user:create` used with admin profile
- Private method `securityController._persist` used by `create*` actions has been merged with `'core:security:user:create'` which was only asked by the first one anyway
- Deprecation warnings during functional tests disabled

### How this should be manually tested?

1. Clone and run Kuzzle
2. Try an old action (`security:getUser` for example), you should see some deprecration message in the response
3. Try a new action (`user:get` for example), you can now use any user actions, go check the doc!

### What's next?
The next step to successfully split the `SecurityController` is to implement a [RoleController](https://github.com/kuzzleio/kuzzle/issues/1973).
While doing this, you may add new methods to the abstraction layer since 'roles' and 'profiles' share even more common logic (`_update` and `_search` might be some good candidates)

Closes #1972


Aschen:
The mappings related methods should be put in plural to respect our domain:
 - `mapping` => `mappings`
 - `updateMapping` => `updateMappings`

Aschen:
1) We should just check that the correct method is called (and stub the method).

2) IMHO we need an abstraction layer because operations on users, roles and profiles are quite similar (e.g. `getRole`, `getUser` and `getProfile` had almost similar code). Then we will be able to benefits from this layer when extracting role and profile controllers.

3) Ideally we should but I start to think that all this "restful" compliance is kinda useless (and boring :sweat_smile: ) since we use SDKs

4) To me the  `checkRights` action belongs to the user controller since we check a user rights

5) I'm not sure we want to keep this action actually

lib/config/httpRoutes.js
```diff
@@ -217,7 +217,7 @@ const routes = [
   { verb: 'post', path: '/roles/_search', controller: 'security', action: 'searchRoles' },
   { verb: 'post', path: '/users/_search', controller: 'user', action: 'search', deprecated: { since: '2.10.3', message: 'Use this route instead: http://kuzzle:7512/user/_search' } },
   { verb: 'post', path: '/credentials/:strategy/:_id/_validate', controller: 'security', action: 'validateCredentials' },
-  { verb: 'post', path: '/_checkRights', controller: 'auth', action: 'checkRights' },
+  { verb: 'post', path: '/_checkRights', controller: 'user', action: 'checkRights' },
```
Aschen: I'm not sure about this change, to me we have 2 `checkRights` action because the first one is meant to be used by the current logged user to check his own rights and another one that allows admins to check a user rights.

So to me it's more the route l221 that need to be moved in the `user` controller

Leodau: Yep, that was an error. I never meant to change the `auth:checkRights`. I did update this accordingly tho, seems like you saw an outdated file.

lib/config/httpRoutes.js
```diff
@@ -71,21 +71,28 @@ const routes = [
   { verb: 'get', path: '/profiles/:_id', controller: 'security', action: 'getProfile' },
   { verb: 'get', path: '/profiles/:_id/_rights', controller: 'security', action: 'getProfileRights' },
   { verb: 'get', path: '/roles/:_id', controller: 'security', action: 'getRole' },
-  { verb: 'get', path: '/users/:_id', controller: 'security', action: 'getUser' },
-  { verb: 'get', path: '/users/:_id/_strategies', controller: 'security', action: 'getUserStrategies' },
-  { verb: 'get', path: '/users/_mGet', controller: 'security', action: 'mGetUsers' },
-  { verb: 'get', path: '/users/:_id/_rights', controller: 'security', action: 'getUserRights' },
+  { verb: 'get', path: '/users/:_id', controller: 'security', action: 'getUser', deprecated: { since: '2.13.0', message: 'Use this route instead: http://kuzzle:7512/user/:_id' } }, // @deprecated
```
Aschen, line 74: Why not using `auto-version` placeholder?

Aschen (changes requested):

doc/2/api/controllers/user/rights/index.md
```diff
@@ -0,0 +1,80 @@
...
+- `index`: index name
+- `collection`: collection name
+- `value`: tell if access is `allowed` or `denied`.
+
+```js
+{
+  "status": 200,
+  "error": null,
+  "result": {
+    "hits": [
```
Aschen, line 57: Should be `rights` instead

doc/2/api/controllers/user/strategies/index.md
```diff
@@ -0,0 +1,58 @@
...
+- `strategies`: array containing all the available authentication strategies on the requested user.
+- `total`: total number of strategies found for that user.
+
+```js
+{
+  "status": 200,
+  "error": null,
+  "result": {
+    "strategies": ["local"],
+    "total": 1
```
Aschen, line 55: Useless property

Aschen:
Can we have this PR on a dedicated branch like `split-security-controller` ? So we can split the 3 controllers on this branch and then release everything at the same time

lib/api/controllers/userController.js
```diff
@@ -0,0 +1,579 @@
...
+    this.getStrategyMethod = global.kuzzle.pluginsManager.getStrategyMethod
+      .bind(global.kuzzle.pluginsManager);
+  }
+
+  /**
+   * @param {Request} request
+   * @returns {Promise}
+   * @private
+   */
+  async _persist (request, profileIds, content, { humanReadableId = true } = {}) {
```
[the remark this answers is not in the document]
MathieuVeber: Done, merge with create method which was only called by _persist

doc/2/api/controllers/auth/signin/index.md
```diff
@@ -21,8 +21,8 @@ This method allows users with limited rights to create other accounts, but block
 ### HTTP
 
 ```http
-URL: http://kuzzle:7512/user/<_id>/_createRestricted[?refresh=wait_for]
-URL(2): http://kuzzle:7512/user/_createRestricted[?refresh=wait_for]
+URL: http://kuzzle:7512/<_id>/_signin[?refresh=wait_for]
```
Aschen, line 24: For the url `sign-in` would be nicer

lib/api/controllers/userController.js
```diff
@@ -23,25 +23,20 @@
...
+const AbstractSecurityController = require('./base/abstractSecurityController');
 
 /**
  * @class UserController
  */
-class UserController extends NativeSecurityController {
+class UserController extends AbstractSecurityController {
   constructor() {
     super([
       'create',
```
Aschen, line 39: I know it's really nitpicking but can you sort the action names alphabetically ?

doc/2/api/controllers/user/update-mappings/index.md
```diff
@@ -0,0 +1,71 @@
...
+
+```http
+URL: http://kuzzle:7512/user/_mappings
+Method: PUT
+Body:
+```
+
+```js
+{
+  "properties": {
```
Aschen, line 27: The example should illustrate the fact that developers can also change the `metadata` and `dynamic` properties like the `collection:updateMappings` API action

lib/api/controllers/userController.js
```diff
@@ -0,0 +1,367 @@
...
+
+  /**
+   * Get the user mapping
+   *
+   * @returns {Promise}
+   */
+  async getMappings () {
+    const { properties } = await global.kuzzle.internalIndex.getMapping('users');
+
+    return { mappings: properties };
```
Aschen, line 249: This API action should also returns the `dynamic` and `metadata`

Aschen (changes requested):

Aschen (changes requested):

lib/api/controllers/userController.js
```diff
@@ -244,9 +244,9 @@ class UserController extends AbstractSecurityController {
    * @returns {Promise}
    */
   async getMappings () {
-    const { properties } = await global.kuzzle.internalIndex.getMapping('users');
+    const { _meta, dynamic, properties } = await global.kuzzle.internalIndex.getMapping('users');
 
-    return { mappings: properties };
+    return { _meta, dynamic, mappings: properties };
```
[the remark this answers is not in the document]
MathieuVeber: Sure I didn't know mappings means the whole thing

Aschen (approved):

Shiranuit (approved):

PR #2017, opened 2021-03-14
Implement requestResponse.configure()

### LINKED ISSUE #1658: As a plugin developer, I'd like to set raw response more easily
Add a `request.response.makeRaw()` method allowing to make a request result a raw one, preventing the need of having a "setResponse(response, opts)" triggered before... returning the response itself.

Leodau:
## What does this PR do ?
Adds a **configure** function in our **requestResponse** class in order to ease **response.raw** access.

`request.response.configure()`

> Nvm _codecov_ 🤦 is taking the wrong branch/commit

Closes #1658

### DIFF
--- a/doc/2/framework/classes/request-response/configure/index.md
+++ b/doc/2/framework/classes/request-response/configure/index.md
@@ -9,7 +9,8 @@ description: RequestResponse class configure() method
 
 <SinceBadge version="auto-version" />
 
-Allows to directly set internal configuration options.
+Allows to configure how the API response should be sent to the requesting client.
+
 
 ### Arguments
 
--- a/doc/2/framework/classes/request-response/configure/index.md
+++ b/doc/2/framework/classes/request-response/configure/index.md
@@ -26,11 +26,11 @@ configure(
 
 The `options` object may contain the following properties:
 
-| Name | Type | Description                      | Default |
-|------|------|----------------------------------|---------|
-| `headers` | <pre>JSONObject</pre> | Additional response protocol headers | `null` |
-| `status` | <pre>integer</pre> | KuzzleRequest status code, following the HTTP standard | `200` |
-| `format` | <pre>string</pre> | The response format, as a `standard` Kuzzle response or in a unwrapped `raw` format instead | `null` |
+| Name | Type | Description<br/>(default)        |
+|------|------|----------------------------------|
+| `headers` | <pre>JSONObject</pre>(`null`) | Additional response protocol headers |
+| `status` | <pre>integer</pre>(`200`) | KuzzleRequest status code, following the HTTP standard |
+| `format` | <pre>string</pre>(`null`) | The response format, as a `standard` Kuzzle response or in a unwrapped `raw` format instead |
 
 ### Example
 
--- a/doc/2/framework/classes/request-response/configure/index.md
+++ b/doc/2/framework/classes/request-response/configure/index.md
@@ -1,0 +1,45 @@
+---
+code: true
+type: page
+title: configure
+description: RequestResponse class configure() method
+---
+
+# configure
+
+<SinceBadge version="auto-version" />
+
+Allows to directly set internal configuration options.
+
+### Arguments
+
+```ts
+configure(
+  options: {
+    headers?: JSONObject;
+    status?: number;
+    raw?: boolean;
+  }): void;
+```
+
+</br>
+
+The `options` object may contain the following properties:
+
+| Name | Type | Description                      | Default |
+|------|------|----------------------------------|---------|
+| `headers` | <pre>JSONObject</pre> | Additional response protocol headers | `null` |
+| `status` | <pre>integer</pre> | KuzzleRequest status code, following the HTTP standard | `200` |
+| `raw` | <pre>boolean</pre> | Instead of a Kuzzle response, forward the result directly to the client, without being converted to an API response payload (can be used to answer in a different format than JSON) | `false` |
+
+### Example
+
+```js
+request.response.configure(
+  headers: {
+    'Location': 'http://kuzzle.io'
+  },
+  status: 302,
+  raw: true,
+);
+```
--- a/doc/2/framework/classes/request-response/configure/index.md
+++ b/doc/2/framework/classes/request-response/configure/index.md
@@ -16,9 +16,9 @@ Allows to directly set internal configuration options.
 ```ts
 configure(
   options: {
-    headers?: JSONObject;
-    status?: number;
-    raw?: boolean;
+    headers?: JSONObject,
+    status?: number,
+    format?: 'standard' | 'raw'
   }): void;
 ```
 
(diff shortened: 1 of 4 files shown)


Aschen (commented):
IMHO this method should allows to set the `headers` and the `status` like the [setResult](https://docs.kuzzle.io/core/2/framework/classes/kuzzle-request/set-result/#arguments) method.

Leodau:
> IMHO this method should allows to set the `headers` and the `status` like the [setResult](https://docs.kuzzle.io/core/2/framework/classes/kuzzle-request/set-result/#arguments) method.
> 
> Also, the documentation should be updated accordingly the framework reference but also this page https://docs.kuzzle.io/core/2/guides/develop-on-kuzzle/api-controllers/#return-a-custom-response

I agree, but then.. **makeRaw()** would not be a good name for this then. 🤔

Aschen:
> I agree, but then.. **makeRaw()** would not be a good name for this then. thinking

I agree as well :sweat_smile: 
The purpose of this method is to send a custom response different from the standardized Kuzzle API response format.

For this we need to use the `raw` attribute so Kuzzle does not put the handler function return into the `result` property but we can also customize the HTTP headers and status.

(Again the HTTP protocol make everything more difficult :sweat_smile: )

Leodau:
> IMHO this method should allows to set the `headers` and the `status` like the [setResult](https://docs.kuzzle.io/core/2/framework/classes/kuzzle-request/set-result/#arguments) method.
> 
> Also, the documentation should be updated accordingly the framework reference but also this page https://docs.kuzzle.io/core/2/guides/develop-on-kuzzle/api-controllers/#return-a-custom-response

But why would we add a `request.response.f(header, status, isRaw)` to the class, IF what we are looking is to effectively remove the need of doing **request.setResult**???

Aschen:
> But why would we add a `request.response.f(header, status, isRaw)` to the class, IF what we are looking is to effectively remove the need of doing **request.setResult**???

Actually the `setResult` method is what we need. The only problem is that the first argument is supposed to be the action result but we set it to null because Kuzzle always use the handler function return

scottinet:
What about this?

```js
request.response.configure({
  headers?: JSONObject,
  status?: number,
  standardize?: boolean,
});
```

This is self-documented: this method configures the response.

The `standardize` option (name to be debated) replaces the `raw` one and instructs Kuzzle to standardize the response into a Kuzzle response one (default: `true`).

wdyt?

Leodau:
> What about this?
> 
> ```js
> request.response.configure({
>   headers?: JSONObject,
>   status?: number,
>   standardize?: boolean,
> });
> ```
> 
> This is self-documented: this method configures the response.
> 
> The `standardize` option (name to be debated) replaces the `raw` one and instructs Kuzzle to standardize the response into a Kuzzle response one (default: `true`).
> 
> Lastly: an option object instead of arguments, because I feel we might add more options in the future. But this is just a feeling, I might be mistaken on that one.
> 
> wdyt?

Awesome, I really like the `configure` idea.
Btw for the 'standardize' option, should we stay around the `raw: true` instead, so we don't have too many differences between **setResult** and **configure**?. 

> We could also, add the **standardize** option to **setResult** and deprecate 'raw'?

Aschen:
> Awesome, I really like the `configure` idea.
> Btw for the 'standardize' option, should we stay around the `raw: true` instead, so we don't have too many differences between **setResult** and **configure**?.
> 
> Or we could add the **standardize** option to **setResult** and deprecate 'raw'?

I agree with your proposal but I would go even further and deprecate the `setResult` method which was first meant to an internal usage I think.

scottinet:
> Awesome, I really like the `configure` idea.
> Btw for the 'standardize' option, should we stay around the `raw: true` instead, so we don't have too many differences between **setResult** and **configure**?.

Like @Aschen, I don't really care about `setResult`. We cannot deprecate it internally (we kinda need it), but its usage by users could, as setting a request's result should be handled solely by Kuzzle.

As for that option name:

* `raw`: I don't like it. And I did come up with that name :expressionless: 
* `standardize` is a tad better, but I'm not entirely convinced about it either

If you have a better name, either to mean that the response will be embedded in a standard Kuzzle response, or that it will be sent raw, don't hesitate to submit it :)

Aschen:
1) `standardize: true`
2) `standardResponse: true`
3) `format: "standard|raw"`

scottinet:
```js
request.response.configure({ asKuzzleResponse: false });
```

Leodau:
Ok, `format` it is then :partying_face: 

`request.response.configure({ format: 'standard|raw' });`

Aschen:
About the changelog, it's more an enhancement to me since the feature already exists

alexandrebouthinon (approved):

doc/2/framework/classes/request-response/configure/index.md
```diff
@@ -0,0 +1,45 @@
...
+    status?: number,
+    format?: 'standard' | 'raw'
+  }): void;
+```
+
+</br>
+
+The `options` object may contain the following properties:
+
+| Name | Type | Description                      | Default |
```
Aschen, line 29: This is not how we document default option, you can look here for example https://github.com/kuzzleio/sdk-javascript/blob/master/doc/7/controllers/document/search/index.md#options

lib/api/request/requestResponse.ts
```diff
@@ -264,6 +264,33 @@ export class RequestResponse {
   get node (): string {
     return (global as any).kuzzle.id;
   }
+
+  /**
+   * Configure the response
```
Aschen, line 269: Missing jsdoc about the options object

Aschen (changes requested):

lib/api/request/requestResponse.ts
```diff
@@ -267,6 +267,12 @@ export class RequestResponse {
 
   /**
    * Configure the response
+   *
+   * @param {Object} [options]
+   * @param {JSONObject} [options.headers] - Additional protocol headers
+   * @param {Number} [options.status=200] - HTTP status code
+   * @param {String} [options.format] - Response format, standard or raw
```
Aschen, line 274:
```suggestion
   * @param [options]
   * @param [options.headers] - Additional protocol headers
   * @param [options.status=200] - HTTP status code
   * @param [options.format] - Response format, standard or raw
```

Aschen (approved):

scottinet (changes requested):

doc/2/framework/classes/request-response/configure/index.md
```diff
@@ -0,0 +1,45 @@
...
+type: page
+title: configure
+description: RequestResponse class configure() method
+---
+
+# configure
+
+<SinceBadge version="auto-version" />
+
+Allows to directly set internal configuration options.
```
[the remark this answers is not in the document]
Leodau: @scottinet 
Yep, makes sense.
How about 
```suggestion
Allows to configure how the API response should be sent to the requesting client

```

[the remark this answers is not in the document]
scottinet: It's better :+1: 
Suggestion applied.

scottinet (approved):

PR #2019, opened 2021-03-15
Add “suggest” to the whitelisted ES search body keys

pierreminik:
## What does this PR do ?

Adds the [`suggest`](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-suggesters.html) keyword to whitelisted ES search body.

### DIFF
--- a/doc/2/guides/main-concepts/querying/index.md
+++ b/doc/2/guides/main-concepts/querying/index.md
@@ -27,6 +27,7 @@ Elasticsearch supports many keywords in a search query root level. For security 
   - `search_timeout`
   - `size`
   - `sort`
+  - `suggest`
   - `_name`
   - `_source`
   - `_source_excludes`
--- a/lib/api/controllers/documentController.js
+++ b/lib/api/controllers/documentController.js
@@ -95,6 +95,7 @@ class DocumentController extends NativeController {
       hits: result.hits,
       remaining: result.remaining,
       scrollId: result.scrollId,
+      suggest: result.suggest,
       total: result.total,
     };
   }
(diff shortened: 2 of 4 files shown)


Aschen (changes requested):
Thanks for your contribution, it's very much appreciated ;-)

Can you also update the documentation to reflect this change ? https://github.com/kuzzleio/kuzzle/blob/master/doc/2/guides/main-concepts/querying/index.md

Also since the result of the `suggest` keyword are not including in the hits array, you will need to add it in the return of the controller action here: https://github.com/kuzzleio/kuzzle/blob/master/lib/api/controllers/documentController.js#L93

pierreminik:
I hope I've done it correctly. 😊

Aschen:
If you could just add/modify an unit test in the document controller, verifying that the `suggest` property from the Elasticsearch result is correctly returned in the document controller action it would be perfect :+1:

pierreminik:
I haven't quite catched how the mocking works so I'm not sure if it's the implementation failing or test not properly test up. I'll figure it out.

pierreminik:
It seems to go through now. I'm not sure if the `suggest` keyword should be available when doing a `scroll` besides the `search`?

Aschen (approved):

Aschen:
> It seems to go through now. I'm not sure if the `suggest` keyword should be available when doing a `scroll` besides the `search`?

Since the `_formatSearchResult` is used by both `search` and `scroll` it's already available in scroll as well :+1:

scottinet (approved):

PR #2029, opened 2021-03-22
WebSocket: enable support for per-message compression

### LINKED ISSUE #1734: Allow per message compression for websocket
## Feature description

Per message compression is currently [explicitly deactivated server-side](https://github.com/kuzzleio/kuzzle/blob/master/lib/core/network/protocols/websocket.js#L104) due to catastrophic memory fragmentation problems encountered in Node.js when heavily using zlib (https://github.com/nodejs/node/issues/8871).

A solution seems to have been found and merged a few days ago (https://github.com/nodejs/node/pull/34048). We need to keep an eye on it: if this lands in Node.js 12, we could then allow per-message compression in websockets.

If we do though, this would need to be deactivated by default. A new flag in Kuzzle's configuration files would allow to activate that feature, with a proper documentation explaining from which version of node.js that option can be safely used.

scottinet:
# Description

Closes #1734 

This PR enables per-message compression for websocket connections.

To do this, this PR replaces the [ws](https://www.npmjs.com/package/ws) module with the [uWebSockets.js](https://github.com/uNetworking/uWebSockets.js) one: the latter is faster, more stable, and not prone to the catastrophic memory fragmentation issue with Node.js (mitigated in later Node.js 12.x  versions).

Since µwebsockets uses C sockets, we cannot replace the WebSocket layer without replacing the HTTP one too: we cannot reuse the Node.js HTTP socket for WebSocket, and we don't want Kuzzle to have 2 different network ports for HTTP and WebSocket.
So this PR also lets µws handle HTTP connections, hence the sizeable refactor.

# Enhancements

* The `allow-encoding` HTTP header is now properly interpreted: before this PR, the priority argument `<algorithm>;q=<priority>` was ignored
* Content parsing for multipart form requests is now done by µws, letting us remove the `busboy` dependency
* Due to how µws handles idle client connections and heartbeat, I had to rework the websocket options in our RC file:
  * the `heartbeat` option is now deprecated and ignored
  * the `idleTimeout` now cannot be deactivated, and its value cannot be set lower than 1000ms. To prevent breaking changes with existing installs, this option is defaulted to 60000ms if set to a value lower than 1000.
  * a new `compression` option has been added. It takes a boolean (`false` by default): setting this option to `true` enables per-message compression
* Add a configurable rate limiter for websocket connections (disabled by default)

# Other changes

* Debug domain for the network layer has been renamed `kuzzle:network:*` (from `kuzzle:entry-point:*`)
* The `server:getConfig` removed the entire `http` configuration from its output, instead of just the HTTP routes one (`http.routes`). This has been fixed, as not returning the http config is a bit confusing (I know I was perplexed when trying to figure why I didn't have HTTP config in that route output)
* Default timeout of cucumber.js has been increased from 5000ms to 30000ms, to prevent functional tests to fail in github actions

# Bug fixes

* Numbers as raw request results are now correctly handled (trying to send a number as a raw response results in an error without this PR)
* Make a realtime functional test (using the functional tests plugin) reentrant, allowing it to be run multiple times on the same node


lib/core/network/protocols/http+websocket.js
```diff
@@ -0,0 +1,920 @@
...
+  disconnect (connectionId, message = null) {
+    debugWS('[%s] forced disconnect', connectionId);
+
+    const socket = this.socketByConnectionId.get(connectionId);
+
+    if (!socket) {
+      return;
+    }
+
+    socket.end(1011, message ? Buffer.from(message) : WS_GENERIC_CLOSE_MESSAGE);
```
Aschen, line 241: What is this number `1011` ?

lib/core/network/protocols/http+websocket.js
```diff
@@ -0,0 +1,920 @@
...
+    this.entryPoint.removeConnection(connection.id);
+    this.connectionBySocket.delete(socket);
+    this.backpressureBuffer.delete(socket);
+    this.socketByConnectionId.delete(connection.id);
+  }
+
+  wsOnMessageHandler (socket, data) {
+    const connection = this.connectionBySocket.get(socket);
+
+    if (!data || data.byteLength === 0 || !connection) {
```
Aschen, line 277: How can we receive a message from an orphan socket ?

lib/core/network/protocols/http+websocket.js
```diff
@@ -0,0 +1,920 @@
...
+        if (result.content && typeof result.content === 'object') {
+          result.content.room = result.requestId;
+        }
+        this.wsSend(socket, Buffer.from(JSON.stringify(result.content)));
+      });
+    }
+    catch (e) {
+      const errobj = {
+        error: {
+          message: e.message
```
Aschen, line 337: Don't we send our other error properties? Like `id`, `code`, `stack` etc?

Also, why not using the `wsSendError` method?

lib/core/network/protocols/http+websocket.js
```diff
@@ -0,0 +1,920 @@
...
+    if (socket.getBufferedAmount() < WS_MAX_BACKPRESSURE) {
+      socket.cork(() => socket.send(payload));
+    }
+    else {
+      const buffer = this.backpressureBuffer.get(socket);
+      buffer.push(payload);
+
+      // Client socket too slow: we need to close it
+      if (buffer.length > WS_BACKPRESSURE_BUFFER_MAX_LENGTH) {
+        socket.end(1011, WS_BACKPRESSURE_MESSAGE);
```
Aschen, line 399: What is this number `1011` ?

lib/core/network/protocols/http+websocket.js
```diff
@@ -0,0 +1,920 @@
...
+
+    if (encoding !== null && encoding[1].toLowerCase() !== 'utf-8') {
+      this.httpSendError(
+        message,
+        response,
+        kerrorHTTP.get('unsupported_charset', encoding[1].toLowerCase()));
+      return;
+    }
+
+    this.httpReadData(message, response, (err) => {
```
Aschen, line 444: Nitpicking
```suggestion
    this.httpReadData(message, response, err => {
```

lib/core/network/protocols/http+websocket.js
```diff
@@ -0,0 +1,920 @@
...
+
+    if (response.aborted) {
+      return;
+    }
+
+    this.entryPoint.newConnection(message.connection);
+
+    global.kuzzle.router.http.route(message, request => {
+      this.entryPoint.logAccess(request, message);
+      this.entryPoint.removeConnection(message.connection.id);
```
Aschen, line 559: Why removing the connection at the beginning and not at the end?

lib/core/network/protocols/http+websocket.js
```diff
@@ -0,0 +1,920 @@
...
+
+        this.httpUncompressStep(encodings, inflated, cb, index + 1);
+      });
+    }
+    else {
+      cb(kerrorHTTP.get('unsupported_compression', encoding));
+    }
+  }
+
+  parseWebSocketOptions () {
```
Aschen, line 815: This shouldn't be the responsability of the config module?

package.json
```diff
@@ -78,14 +76,14 @@
...
-    "utf-8-validate": "^5.0.4",
     "uuid": "^8.3.2",
+    "uWebSockets.js": "github:uNetworking/uWebSockets.js#v18.14.0",
     "validator": "^13.5.2",
     "winston": "^3.3.3",
     "winston-elasticsearch": "0.15.2",
     "winston-syslog": "^2.4.4",
     "winston-transport": "^4.4.0",
-    "ws": "^7.4.4",
+    "ws": "^7.4.3",
```
Aschen, line 86: Why do we still need the `ws` library?

Aschen (changes requested):
Welcome back uWS :tada:

lib/core/network/protocols/http+websocket.js
scottinet: https://tools.ietf.org/html/rfc6455#page-46

lib/core/network/protocols/http+websocket.js
scottinet: Websocket termination code, see the rfc: https://tools.ietf.org/html/rfc6455#page-46

lib/core/network/protocols/http+websocket.js
scottinet: Because there are multiple exit routes: the connection might have already been aborted, or it can aborts before we finish compressing the response, or during the time we send the response (see tryEnd/onWritable)

We might as well remove the ClientConnection object now: it's only the Kuzzle representation of a network connection, and at that point of time, it doesn't exist anymore (its associated request has been processed already)

package.json
scottinet: Argh. Merge mishap surely, I DID remove it. :expressionless:

scottinet: fixed

lib/core/network/protocols/http+websocket.js
Aschen: Can we have a named constant instead of a number? It's more readable than having to read the RFC :sweat_smile:

lib/core/network/protocols/http+websocket.js
Aschen: Can we have a named constant instead of a number? It's more readable than having to read the RFC :sweat_smile:

lib/core/network/protocols/http+websocket.js
scottinet: Yes, it should :+1:

lib/core/network/protocols/http+websocket.js
scottinet: Good point :+1:

lib/core/network/protocols/http+websocket.js
```diff
@@ -0,0 +1,920 @@
...
+    const payload = Buffer.allocUnsafe(payloadByteSize + 255);
+
+    let offset = payloadByteSize - 1;
+
+    payload.write(stringified, 0);
+    payload.write(JSON_ROOM_PROPERTY, offset);
+
+    offset += JSON_ROOM_PROPERTY.length;
+
+    for (const channel of data.channels) {
```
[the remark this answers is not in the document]
scottinet: welp

it makes hardly a difference:

```
for(;;) x 1,554,697 ops/sec ±0.93% (92 runs sampled)
for(const ... of ...) x 1,512,130 ops/sec ±0.66% (94 runs sampled)
Fastest is for(;;)
```

lib/core/network/protocols/http+websocket.js
scottinet: Indeed.

lib/core/network/protocols/http+websocket.js
scottinet: Hm... This bit of code comes from our holder websocket message handler, I didn't take a good look at it. After some analysis, I don't see the point of this try..catch at all: the entire process is safe, and the callback is always invoked with an error result if something wrong happens.

I think I'll remove this bit of code altogether.

lib/core/network/protocols/http+websocket.js
```diff
@@ -0,0 +1,920 @@
...
+      this.initWebSocket();
+    }
+
+    if (this.httpConfig.enabled) {
+      this.initHttp();
+    }
+
+    this.server.listen(entrypoint.config.port, socket => {
+      if (!socket) {
+        throw new Error(`[http/websocket] fatal: unable to listen to port ${entrypoint.config.port}`);
```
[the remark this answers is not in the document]
scottinet: I don't see the point, as this error will stop Kuzzle from starting, and will be print in the logs. It's like our assertion errors: we don't need to standardize them since they won't be returned by our API.

lib/core/network/protocols/http+websocket.js
scottinet: Ah, yes, it IS useful: it's there to catch errors when promoting the payload to a KuzzleRequest object. I've properly separated the code to make it clearer, and the code now uses wsSendError, as it should have since the beginning.

scottinet:
@Aschen > I applied your requested changes.

lib/core/network/protocols/http+websocket.js
scottinet: I've moved the option checks to the config module. However the "parseXxxOptions()" are still needed to prepare the actual protocols options according to what has been configured.

lib/core/network/protocols/http+websocket.js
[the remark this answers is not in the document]
Aschen: Looks like v8 is getting better and better :sweat_smile:

features/PluginContext.feature
```diff
@@ -32,14 +32,15 @@ Feature: Plugin context
 
   @realtime
   Scenario: Subscribe and unsubscribe to realtime notifications
+    Given a plugin realtime subscription
```
Aschen, line 35: This could be replaced by 
```suggestion
    Given I successfully execute the action "functional-test-plugin/realtime":"subscribeOnce"
```

And the associated step can be removed

Aschen (approved):

features/PluginContext.feature
scottinet: Less is more :+1: 

Changes applied.

lib/core/network/protocols/http+websocket.js
```diff
@@ -0,0 +1,910 @@
+/*
```
Leodau, line 1: Wot, I don't like the use of characters like `+` in file naming, kindda weird.
Also, I believe this goes against node export naming convention, shouldn't the file name reflect the name of the class defined within?

lib/core/network/protocols/http+websocket.js
```diff
@@ -0,0 +1,910 @@
...
+const kerrorHTTP = kerror.wrap('network', 'http');
+const debugWS = debug('kuzzle:network:protocols:websocket');
+const debugHTTP = debug('kuzzle:network:protocols:http');
+
+// The idleTimeout option should never be deactivated, so instead we use
+// a default value for backward-compatibility
+const DEFAULT_IDLE_TIMEOUT = 60000;
+
+// Size of backpressure an individual socket can handle before needing to drain
+const WS_MAX_BACKPRESSURE = 4096;
```
Leodau, line 49: What does this 4096 mean?
`Back-pressure` as = `size`?

lib/core/network/protocols/http+websocket.js
```diff
@@ -0,0 +1,910 @@
...
+
+// Applicative WebSocket PONG message for browsers
+const WS_APP_PONG_RESPONSE = Buffer.from('{"p":2}');
+
+// Used by the broadcast method to build JSON payloads while limiting the
+// number of JSON serializations
+const JSON_ROOM_PROPERTY = ',"room":"';
+const JSON_ENDER = '"}';
+
+// pre-computed messages & errors
```
Leodau, line 64:
```suggestion
// Pre-computed messages & errors
```

Leodau (changes requested):

lib/core/network/protocols/http+websocket.js
scottinet: It's the size in bytes, I'll amend the comment.

Backpressure is the right term for it though, but it should be the backpressure buffer size: since websocket is on top of TCP, the server expects an ACK packet each time it sends something to a client. If the client fails to send an acknowledgement, or if it is too slow, or if there are network problems, then the server HAS to keep the payload it needs to send in a buffer. This is backpressure: a load imposed to something, because of something else (here, the server has to keep things in memory because of client sockets).

If we want to be exact about the name of the constant, then it should be `WS_MAX_BACKPRESSURE_BUFFER_SIZE`, but it's a bit long to my taste.

lib/core/network/protocols/http+websocket.js
scottinet: You're no fun: I do like esoteric characters in file names, messing up with regexes when looking for them. :expressionless:

scottinet:
@Leodau > changes applied.

Leodau (approved):

PR #2045, opened 2021-04-13
Hotfix duplicated cors headers on HTTP Request

Shiranuit:
## What does this PR do ?
This PR is fixes the unwanted behaviour of CORS headers being duplicated when kuzzle is responding to an HTTP Request.
Duplicating CORS headers on a Response causes the browsers that sent the request to throw a CORS error.

I also added more test cases to avoid this issue in the future.

### DIFF
--- a/lib/api/request/requestResponse.ts
+++ b/lib/api/request/requestResponse.ts
@@ -322,11 +322,16 @@ export class RequestResponse {
   /**
    * Adds new multiple headers.
    */
-  setHeaders (headers: JSONObject) {
+  setHeaders (headers: JSONObject, ifNotPresent:boolean = false) {
     assert.assertObject('headers', headers);
 
     if (headers) {
-      Object.keys(headers).forEach(name => this.setHeader(name, headers[name]));
+      Object.keys(headers).forEach(name => {
+        // When ifNotPresent is set to true, only set the header if no value has been defined before
+        if (!ifNotPresent || this.getHeader(name) === undefined) {
+          this.setHeader(name, headers[name]);
+        }
+      });
     }
   }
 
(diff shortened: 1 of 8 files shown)


Leodau (approved):

Aschen (approved):

Aschen (changes requested):
Missing version bump

Shiranuit:
**Update**: Found some code somewhere else that could duplicate headers, this is now fixed

**Added**:
- `RequestResponse.setHeaders` method now has a new parameter named `ifNotPreset`, if set to `true` only set the header if the header is not already present

lib/api/request/requestResponse.ts
```diff
@@ -322,11 +322,16 @@ export class RequestResponse {
   /**
    * Adds new multiple headers.
    */
-  setHeaders (headers: JSONObject) {
+  setHeaders (headers: JSONObject, ifNotPresent:boolean = false) {
     assert.assertObject('headers', headers);
 
     if (headers) {
-      Object.keys(headers).forEach(name => this.setHeader(name, headers[name]));
+      Object.keys(headers).forEach(name => {
```
Aschen, line 329: Avoid the usage of `forEach`, prefer vanilla `for` loop

Aschen (changes requested):
You must bump the package version with `npm version patch`

xbill82 (approved):

lib/api/request/requestResponse.ts
```diff
@@ -322,11 +322,16 @@ export class RequestResponse {
   /**
    * Adds new multiple headers.
    */
-  setHeaders (headers: JSONObject) {
+  setHeaders (headers: JSONObject, ifNotPresent = false) {
     assert.assertObject('headers', headers);
 
     if (headers) {
-      Object.keys(headers).forEach(name => this.setHeader(name, headers[name]));
+      for (const name in headers) {
```
Aschen, line 329:
```suggestion
      for (const name of headers) {
```

Aschen (approved):

PR #2048, opened 2021-04-19
Framework: allow the execution of code only once

MathieuVeber:
## What does this PR do ?

Add an install method available for Kuzzle as a Framework allowing developers to run a piece of code only once during the lifetime of the app. This can be useful when considering db migrations or setup with external services and many more use cases...

- Initialize a kuzzle app `kourou app:scaffold playground`
- Open the file app.ts
- Between `const app = new Backend('playground')` and `app.start()` write `app.install('id', () => { console.log('I will be print only once') });`
- Run kuzzle `docker-compose up`
- Watch the logs and see your message 

### Other changes

- Classes inside lib/core/backend/backend.ts and their tests (test/core/backend/Backend.test.js) extracted in several classes (Review tip: without changing code inside these classes)
  - backendConfig
  - backendController
  - backendHook
  - backendPipe
  - backendStorage
  - backendVault
  - internalLogger
- Deleted type export internalLogger since the class is now exported directly
- Doc for install method
- Unit tests added (Review tip: in Backend.test.js and kuzzle.test.js)
- Added a collection %kuzzle.installations in default.config.js
- Fix versioning in the RequestInput doc


lib/config/default.config.js
```diff
@@ -335,6 +335,13 @@ module.exports = {
               ttl: { type: 'keyword' },
               token: { type: 'keyword' }
             }
+          },
+          installations: {
+            dynamic: 'false',
```
Aschen, line 340: Should be `strict`

lib/config/default.config.js
```diff
@@ -335,6 +335,13 @@ module.exports = {
...
+            properties: {
+              handler: { type: 'text' },
```
Aschen, line 342: Why storing the code of the function?

lib/kuzzle/kuzzle.js
```diff
@@ -197,6 +199,24 @@ class Kuzzle extends KuzzleEventEmitter {
 
       this.log.info('[✔] Core components loaded');
 
+      if ( options.installations && options.installations.length) {
+        const promises = [];
+        for (const installation of options.installations) {
+          promises.push(this.install(installation.id, installation.handler));
```
Aschen, line 205: I think installations should be done sequentially, in the order they are registered.

Otherwise we won't be able to have a mapping migration system that execute each migration one after each other.

Also this code should be moved into a function and this function should be called in the init sequence. We try to keep only function calls in the init sequence

test/kuzzle/kuzzle.test.js
```diff
@@ -233,4 +237,48 @@ describe('/lib/kuzzle/kuzzle.js', () => {
...
+    });
+
+    afterEach(() => {
+      Date.now.restore();
+    });
+
+    it('should call the handler and work properly', async () => {
+      const result = await kuzzle.install('id', handler);
+
+      should(kuzzle.ask).be.calledTwice();
```
Aschen, line 259: Can you check that `kuzzle.ask` has been called with the correct arguments please ? This function could literally be used to do anything :sweat_smile:

test/kuzzle/kuzzle.test.js
```diff
@@ -233,4 +237,48 @@ describe('/lib/kuzzle/kuzzle.js', () => {
...
+      should(handler).be.calledOnce();
+      should(result).match(true);
```
Aschen, line 261: Nitpicking
```suggestion
      should(result).be.true();
```

test/kuzzle/kuzzle.test.js
```diff
@@ -233,4 +237,48 @@ describe('/lib/kuzzle/kuzzle.js', () => {
...
+        'installations',
+        'id');
+      should(kuzzle.ask).be.neverCalledWith(
+        'core:storage:private:document:create',
+        'kuzzle',
+        'installations',
+        { handler: handler.toString(), installedAt: Date.now() },
+        { id: 'id' });
+      should(handler).not.be.called();
+      should(result).match(false);
```
Aschen, line 281:
```suggestion
      should(result).be.false();
```

test/kuzzle/kuzzle.test.js
```diff
@@ -7,6 +7,7 @@ const rewire = require('rewire');
 const Bluebird = require('bluebird');
```
Aschen, line 7: Missing a test for the function that plays the installations

Aschen (changes requested):

lib/config/default.config.js
MathieuVeber: I tought it would be useful to keep a record of what have been done but it may not be necessary

lib/core/backend/internalLogger.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import util from 'util';
+
+import _ from 'lodash';
```
Leodau, line 24: Unused module **lodash**, check other files too.
```suggestion
```

MathieuVeber: Yep I was working on it :wink:

lib/kuzzle/kuzzle.js
MathieuVeber: I'm not sure I've precisely understood what you meant but I moved the code outside the start function, tell me if it's ok

lib/kuzzle/kuzzle.js
```diff
@@ -247,6 +253,54 @@ class Kuzzle extends KuzzleEventEmitter {
...
+            'installations',
+            { handler: installation.handler.toString(), installedAt: Date.now() },
+            { id: installation.id });
+
+          executedCount += 1;
+        }
+      }
+    }
+    catch (error) {
+      throw kerror.get( 'plugin', 'runtime', 'unexpected_error', error);
```
Yoann-Abbes, line 295: I think it may be better to add a more precise error for this task

Yoann-Abbes (changes requested):

lib/config/default.config.js
Aschen: It that case maybe a description of what this setup does is better

Aschen (approved):

Yoann-Abbes (approved):

lib/kuzzle/kuzzle.js
```diff
@@ -247,6 +253,56 @@ class Kuzzle extends KuzzleEventEmitter {
...
+
+          await this.ask(
+            'core:storage:private:document:create',
+            'kuzzle',
+            'installations',
+            { handler: installation.handler.toString(), installedAt: Date.now() },
+            { id: installation.id });
+
+          executedCount += 1;
+        }
```
xbill82, line 296: As a user, I'd like to have a warning here, saying the install code has not been re-executed.

xbill82 (dismissed):

lib/kuzzle/kuzzle.js
Aschen: IMHO it's not a warning because that's the expected behavior to not re-execute already executed installations

MathieuVeber: There is already a short fedback of what have been done [here](https://github.com/kuzzleio/kuzzle/pull/2048/files#diff-ee0c987c21f421e4392b8cc04b7737eb94ff6c2b903c02c2bbfb509af36acb43R204), is that sufficient?

doc/2/framework/classes/backend/install/index.md
```diff
@@ -0,0 +1,29 @@
...
+
+<SinceBadge version="auto-version" />
+<CustomBadge type="error" text="Experimental: non-backward compatible changes or removal may occur in any future release."/>
+
+Register code which will be executed only once during the lifetime of the app.
+
+::: info
+This method can only be used before the application is started because the handler is called when the app is launching. If the handler throws, the app will be unable to start.
+:::
+
```
Leodau, line 18: Wot? I don't see this as accurate as you may think. 
Why talk about the "lifetime" of the app if this is executed _pre-start_?

Instead of using a phrase and then making an _info_ block to explain it further, we could just simply explain it directly right?

```suggestion
Register setup code that runs only once and before the application starts.

::: info
If this method throws, the application won't start.
:::
```

lib/kuzzle/kuzzle.js
```diff
@@ -197,6 +199,10 @@ class Kuzzle extends KuzzleEventEmitter {
 
       this.log.info('[✔] Core components loaded');
 
+      const executedCount = await this.install(options.installations);
+
+      this.log.info(`[✔] ${executedCount} new installations executed`);
+
```
Leodau, line 205: To me, having a log saying `"N new installations executed"` makes no sense :facepalm:

Shiranuit (approved):

Leodau (changes requested):
I'm really not keen on the "_installations_" word, we decided on using the `install()` name for this functionality, but we never talked about having the 'code blocks' registered/executed getting called "_installations_"

Does this really feel right to you @xbill82 ?
The more I read it, the more it makes sense. But at first glance, I did not actually understand it. :thinking: :thought_balloon:

xbill82:
My 2 cents, this is not an `installation` but rather a `beforeBoot` lifecycle hook.

lib/kuzzle/kuzzle.js
Leodau: What he is trying to say, is that the execution order should always be preserved!

MathieuVeber: Yep, I've changed the promise part since, to keep the installation order same as written by the developer

doc/2/framework/classes/backend/install/index.md
MathieuVeber: Your suggestion is quiet clear and it's far better but I feel like we still miss the "ever" part that I was trying to say by "lifetime of the app". When I read "that runs only once and before the application starts" I understand "each time I start the app my function will be executed" which is not the case... As long as you don't clear ES, the function will never run twice.

MathieuVeber: Maybe we should add in the info block something like:
```suggestion
Register setup code that runs only once (as long as the database is preserved) and before the application starts.

::: info
If this method throws, the application won't start.
:::

```

lib/kuzzle/kuzzle.js
MathieuVeber: Since we don't really know how to name those installations / setups / deployements(:trollface:), we may let the developer with whatever names he has chosen and print a feedback for each of those (which should make @xbill82 happy), I see something like that:
```javascript
this.log.info('[✔] ${installation.id}: Successfully executed');
this.log.info('[✔] ${installation.id}: Already executed on ${retreivedInstallation.installedAt}');
```
```sh-session
[✔] Mappings 1.8.0: Already executed on Tue Aug 19 2020 23:15:30 GMT+0100 (Central European Standard Time)
[✔] Mappings 2.4.3: Successfully executed
```

xbill82: I see your point, Mathieu. Bootstrap is not the word: I understand the "only once" character of the thing. I' out of ideas for this feature: but installation looks ugly. Did you already consider "Setup"? Even if it's not the word either...

MathieuVeber: "Setup" may be prettier but since we've chosen "install" for the name of the function I feel like it's less consistent... Do you think they'll understand the link between install and setup when working on the backend?

xbill82: I have to admit I have no more ideas on this. Install feels ugly, but seems to be the only word that fits.

Aschen:
> My 2 cents, this is not an `installation` but rather a `beforeBoot` lifecycle hook.

It's not exactly a lifecycle hook, those kind of hooks are expected to be executed when a node is starting. (e.g. `kuzzle:state:*`events, internal index init, etc)

Here we want a hook who is triggered the first time the code is deployed and executed by the application

doc/2/framework/classes/backend/install/index.md
scottinet: It's not during the lifetime of an application. Rather, it's once during the lifetime of an app _environment_.
Also, I don't quite like the "as long as the database is preserved", as it introduces a technical detail into a feature description.

What about this?

```suggestion
Register code executed when the application starts, and only once on any given environment. Once successfully executed, the code associated to an install identifier will never be run again on that environment.

::: info
If this method throws, the application won't start.
:::
```

lib/kuzzle/kuzzle.js
scottinet: I agree with @Leodau , I don't see the point of that log message.

I would rather have a log explaining what new installs have been run (with their ids), and nothing otherwise.

Also, since we went for  `app.install`, we might as well go all the way with that name, should make things easier when users are trying to understand log messages.

lib/kuzzle/kuzzle.js
scottinet: I don't see why we should show something that has not been executed in the logs.

About your need @xbill82 , perhaps an API route returning all the installed code (id, timestamp) could do the trick: that way admins can see what installs have been performed, and when. When debugging their app, this should be sufficient (and more to the point) to be reassured about their install code, without having to scan lengthy log messages, and without cluttering those.

WDYT?

lib/kuzzle/kuzzle.js
MathieuVeber: Go for it, I agree with that :+1:

lib/kuzzle/kuzzle.js
MathieuVeber: That seems to me like a good idea. If @xbill82 agrees, should I open an issue on that or implement it directly in this PR?

lib/kuzzle/kuzzle.js
```diff
@@ -247,6 +251,60 @@ class Kuzzle extends KuzzleEventEmitter {
...
+   */
+  async install (installations) {
+    if ( !installations || !installations.length) {
+      return;
+    }
+    
+    const mutex = new Mutex('backend:installations');
+    await mutex.lock();
+
+    try{
```
Leodau, line 269:
```suggestion
    try {
```

lib/kuzzle/kuzzle.js
```diff
@@ -247,6 +251,60 @@ class Kuzzle extends KuzzleEventEmitter {
...
+      for (const installation of installations) {
+        const alreadyInstalled = await this.ask(
+          'core:storage:private:document:exist',
+          'kuzzle',
+          'installations',
+          installation.id );
+
+        if (!alreadyInstalled) {
```
Leodau, line 277:
```suggestion
        if (! isAlreadyInstalled) {
```

lib/kuzzle/kuzzle.js
```diff
@@ -247,6 +251,60 @@ class Kuzzle extends KuzzleEventEmitter {
...
 
+  /**
+   * Execute multiple handlers only once on any given environment
+   * 
+   * @param {Array<{ id: string, handler: () => void, description?: string }>} installations - Array of unique methods to execute
+   * 
+   * @returns {Promise<void>}
+   */
+  async install (installations) {
+    if ( !installations || !installations.length) {
```
Leodau, line 262:
```suggestion
    if (! installations || ! installations.length) {
```

lib/kuzzle/kuzzle.js
```diff
@@ -247,6 +251,60 @@ class Kuzzle extends KuzzleEventEmitter {
...
+      return;
+    }
+    
+    const mutex = new Mutex('backend:installations');
+    await mutex.lock();
+
+    try{
+      for (const installation of installations) {
+        const alreadyInstalled = await this.ask(
```
Leodau, line 271:
```suggestion
        const isAlreadyInstalled = await this.ask(
```

Leodau (changes requested):

lib/kuzzle/kuzzle.js
Leodau: I know better would be `wasAlreadyInstalled` but let's say I like **is** better. :wink:

Leodau (approved):

PR #2051, opened 2021-04-27
Cluster: die if unable to refresh the ID card in time

scottinet:
# Description

Each cluster node maintains its own ID card in redis, allowing other nodes to discover the node, and connect to it.

ID cards are temporary redis key, meaning that they disappear after a time. It's intended: if a node crashes, its ID card disappears and future nodes won't try to connect to it.

But if a node is too slow to refresh its own ID card, then the key disappears: future nodes don't know that this node exists, they never form a cluster, and this makes the cluster inconsistent.

With this PR, each time a node refreshes its ID card, it checks that the refresh was successful: Redis returns 1 if the key is refreshed, and 0 if the key doesn't exist. In the latter case, the node now detects it, and shuts itself down.

### DIFF
--- a/lib/cluster/idCardHandler.js
+++ b/lib/cluster/idCardHandler.js
@@ -61,10 +61,11 @@ class IdCard {
  * Handles the ID Key stored in Redis, holding node information
  */
 class ClusterIdCardHandler {
-  constructor (ip, refreshDelay) {
+  constructor (node) {
+    this.node = node;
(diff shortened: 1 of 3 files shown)


xbill82 (approved):
Been following the story. Hope this fix works! 💪

Shiranuit (approved):

Leodau (approved):

PR #2054, opened 2021-04-30
Improve API performances

scottinet:
# Description

After switching our network layer to µWebSockets, we expected an improvement in Kuzzle's performances: we identified Kuzzle's network layer as its main performances bottleneck, and µws is by far quicker than Node.js' HTTP and than the `ws` module we used before.

Benchmarks demonstrated that there was no net gain, and this bothered me. So I ran a few profiling analysis, and found other bottlenecks: access logs and UUID generation.

## Access Logs

Access logs use the Winston logging module, which is... CPU-hungry. Even with all transports disabled (`silent` set to `true` in their configurations), winston still consumes a lot of resources, virtually doing nothing.

So, this PR does the following:
* If no winston transport is activated (no transport configured, or all transports silenced), then access logs are completely deactivated: Winston is not even instantiated anymore.
* If access logs are active, they are isolated in a separate worker thread, to allow winston to have its own event loop it can slow down all it wants.

## UUID generation

Another main resource consumption identified by profiling analysis comes from the `KuzzleRequest` object: 8% of Kuzzle's CPU consumption was dedicated to UUIDs generation.
These UUID are used to generate unique request identifiers.

To reduce resources consumption, UUID was replaced by [NanoID](https://www.npmjs.com/package/nanoid), which serves the same purposes, with the same risks of collision, but in a more efficient way.

# Other changes

Access logs attempted to infer the request's user ID by decoding the public part of a token (authentication bearer for HTTP, and `jwt` for other protocols). 
Since now access logs are performed once requests are submitted to the API, the token is already decoded, so this token decoding part was obsolete: it has been removed from the code.

# Benchmark results

When access logs are disabled (which is the most common use case), this PR brings an improvement of roughly 60% of raw performances:

```
===============================================================================
=== Real-time control step, measuring raw Kuzzle response time
100% | Requests completed: 300000 | Elapsed: 21s | 14465 requests/s | 0 notifications received
===============================================================================
=== Geopolygon subscriptions listening to geopoints in France
Starting subscriptions...
100 subscriptions done (took: 194ms)
100% | Requests completed: 300000 | Elapsed: 25s | 11937 requests/s | 54678 notifications received
Unsubscribing...
Unsubscriptions done (took: 172ms)
===============================================================================
=== Geopolygon subscriptions listening to geopoints in the US
Starting subscriptions...
100 subscriptions done (took: 70ms)
100% | Requests completed: 300000 | Elapsed: 33s | 9123 requests/s | 549669 notifications received
Unsubscribing...
Unsubscriptions done (took: 155ms)
```

### DIFF
--- a/lib/api/request/kuzzleRequest.ts
+++ b/lib/api/request/kuzzleRequest.ts
@@ -20,6 +20,7 @@
  */
 
 import * as uuid from 'uuid';
+import { nanoid } from 'nanoid';
 import { JSONObject } from 'kuzzle-sdk';
 
 import { RequestInput } from './requestInput';
@@ -59,7 +60,7 @@ export class KuzzleRequest {
   public id: string;
 
   constructor (data: any, options: any) {
-    this[_internalId] = uuid.v4();
+    this[_internalId] = nanoid();
     this[_status] = 102;
     this[_input] = new RequestInput(data);
     this[_context] = new RequestContext(options);
@@ -74,7 +75,7 @@ export class KuzzleRequest {
 
     this.id = data.requestId
       ? assert.assertString('requestId', data.requestId)
-      : uuid.v4();
+      : nanoid();
 
     this[_timestamp] = data.timestamp || Date.now();
 
--- a/lib/core/network/accessLogger.js
+++ b/lib/core/network/accessLogger.js
@@ -1,0 +1,330 @@
+/*
+ * Kuzzle, a backend software, self-hostable and ready to use
+ * to power modern apps
+ *
+ * Copyright 2015-2020 Kuzzle
+ * mailto: support AT kuzzle.io
+ * website: http://kuzzle.io
+ *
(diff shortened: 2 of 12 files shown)


Shiranuit (approved):

lib/api/request/kuzzleRequest.ts
```diff
@@ -20,6 +20,7 @@
  */
 
 import * as uuid from 'uuid';
```
rolljee, line 22: Can this whole file make the use of nanoId instead of uuid.v4 ?

scottinet:
@rolljee > UUIDs are still used as the default document ID generator. As NanoIds have a different form, I prefer to avoid changing the way our default IDs look.

With this new version, UUID is rarely used, so this shouldn't have an impact on performances.

rolljee (approved):

PR #2055, opened 2021-04-30
Support for multiple domains in Access Control Allow Origin

Shiranuit:
## What does this PR do ?

This PR adds the support of multiple values in the config `http.accessControlAllowOrigin`
By using coma separated value in a string like this:
```json
{
  "http": {
    "accessControlAllowOrigin": "foo, bar, localhost"
  }
}
```
Or by using an  array directly
```json
{
  "http": {
    "accessControlAllowOrigin": [
      "foo",
      "bar",
      "localhost"
    ]
  }
}
```

When a request containing the header `Origin` is received, the funnel will first check if the config `http.accessControlAllowOrigin` contains `*`,  or if the Origin value is contained in `http.accessControlAllowOrigin` in this case the request continue to follow the validation process (`checkRights`, etc...), otherwise the request will be rejected with an `unauthorized_origin` error.

Request not containing the header `Origin` are not affected by the modification described above and follow the usual validation process (`checkRights`, etc...).


lib/cluster/node.js
```diff
@@ -376,6 +376,15 @@ class ClusterNode {
           return;
         }
 
+        // Verify that no other node share the same IP address as this one
+        const duplicate = nodes.filter(node => node.ip === this.ip);
+
+        if (duplicate.length > 0) {
+          global.kuzzle.log.error(`[FATAL] Another node share the same IP address as this one (${this.ip}): ${duplicate[0].id}. Shutting down.`);
```
xbill82, line 383: This looks like @scottinet 's hotfix... Wasn't this living on a separate PR?

scottinet (approved):

xbill82 (approved):

lib/cluster/node.js
Shiranuit: IDK, I just based this branch on 2-dev, it seems pretty weird

xbill82: Ok, I thought you merged his branch while still in PR... BTW yes it's really weird git sees these lines as diff... Did you merge 2-dev on this branch lately? Even if it should make no difference, sinc diffs on GH are performed on the latest commit of the destination branch... OH! Wait! I see your destination branch is `master`!

Shiranuit: Ahh, you're right, I did put the wrong branch, I will changed it to 2-dev 👍

scottinet: Oh, right. It's fortunate you raised an issue about this one @xbill82: I saw this too, but just thought that it was some rebase mishap or something like that, and paid it no mind :sweat_smile:

Shiranuit:
I did some minor corrections:
- Updated docker-compose (added admin console to Access-Control-Allow-Origin)
- Funnel: Changed the way the unauthorized_origin error is returned because it was causing a malformed HTTP Response
- Funnel & AuthController -> replace `request.input.args.cookieAuth` with `request.getBoolean('cookieAuth')`

scottinet (approved):

scottinet:
@Shiranuit > the CI fails because you need to update the errors documentation: run `node doc/build-error-codes.js` and then push the changes

alexandrebouthinon (approved):

lib/config/index.js
```diff
@@ -212,4 +214,22 @@ function checkClusterOptions (config) {
...
+  if (typeof cfg.accessControlAllowOrigin === 'string') {
+    cfg.accessControlAllowOrigin =
+      cfg.accessControlAllowOrigin
+        .split(',')
+        .map(value => value.trim());
+  }
+
+  config.internal.allowAllOrigins = cfg.accessControlAllowOrigin.includes('*'); // Stored to avoid doing includes multiple times later
+  config.internal.cookieAuthentication = !cfg.accessControlAllowOrigin.includes('*'); // If there is no wild card, cookieAuthentication is enabled
+}
```
Leodau, line 233:
```suggestion
function preprocessHttpOptions (config) {
  const httpConfig = config.http;

  if (httpConfig === undefined) {
    return;
  }

  if (typeof httpConfig.accessControlAllowOrigin === 'string') {
    httpConfig.accessControlAllowOrigin =
      httpConfig.accessControlAllowOrigin
        .split(',')
        .map(value => value.trim());
  }

  // Stored to avoid doing includes multiple times later
  config.internal.allowAllOrigins = httpConfig.accessControlAllowOrigin.includes('*');

  // If there is no wild card, cookieAuthentication is enabled
  config.internal.cookieAuthentication = !httpConfig.accessControlAllowOrigin.includes('*');
}
```

lib/config/index.js
```diff
@@ -50,13 +50,14 @@ function load () {
   checkHttpOptions(config);
   checkWebSocketOptions(config);
   checkClusterOptions(config);
-
+
```
Leodau, line 53:
```suggestion

```

lib/api/funnel.js
```diff
@@ -263,6 +264,21 @@ class Funnel {
       return 1;
     }
 
+    const httpCfg = global.kuzzle.config.http;
+    if (request.input.headers
+      && has(request.input.headers, 'origin')
+      && !global.kuzzle.config.internal.allowAllOrigins
+      && !httpCfg.accessControlAllowOrigin.includes(request.input.headers.origin) // If the domain is not present in the list
```
Leodau, line 271: Optional Nit
```suggestion
    const httpConfig = global.kuzzle.config.http;
    if (request.input.headers
      && has(request.input.headers, 'origin')
      && !global.kuzzle.config.internal.allowAllOrigins
      && !httpConfig.accessControlAllowOrigin.includes(request.input.headers.origin) // If the domain is not present in the list
```

lib/core/network/httpRouter/index.js
```diff
@@ -156,6 +155,21 @@ class Router {
       // Set Headers if not present
       routeHandler.request.response.setHeaders(this.defaultHeaders, true);
 
+      
+      if (message.headers && message.headers.origin) {
```
Leodau, line 159: Seems like duplicated code in **L158** & **L186** -
Is the intention to have explicit duplicated code or could this be abstracted?

lib/api/funnel.js
Leodau: I saw that we are using the `httpCfg` style in other places (protocol), you can ignore this comment if you think httpCfg > httpConfig.

Leodau (changes requested):

Leodau (approved):

PR #2056, opened 2021-05-03
Update cluster topology during handshake

scottinet:
# Description

Each cluster node maintains its own cluster topology, meaning it keeps track of all nodes it is connected to.
This is used when a node leaves the cluster (normal exit, or due to an eviction): each node checks its own topology against every other node topologies, and if an inconsistency is found, the algorithm is made in such a way that nodes isolated from the cluster shut down to prevent cluster splits.

Now, there is a bug in the current implementation: the topology is only updated when a new node is added to the cluster. Meaning that when a new node is added, other nodes are notified about it and they update their topology.
But... due to an oversight, the topology wasn't updated during a handshake: this means that new nodes joining the cluster have an empty topology, whereas existing nodes will have that new node added to theirs.

This messes up the cluster consistency check obviously.

# How to reproduce the bug

Spawn a cluster of 3 nodes using the dev docker-compose. Kill the 1st node that spawned: the other two nodes will see that their topologies don't match:
* the second node that spawned has the third one in its topology (node added after the handshake)
* the third one has an empty topology (topology not updated during the handshake)

The cluster decides that the third one is isolated, and it is killed.

# Expected behavior

Once a handshake finishes, the joining node must add existing ones to its topology.
Killing the first node should not impact the other two ones (unless there is a real network split, obviously, but that shouldn't happen in test environments)

### DIFF
--- a/lib/cluster/node.js
+++ b/lib/cluster/node.js
@@ -437,6 +437,7 @@ class ClusterNode {
           this.remoteNodes.delete(nodeId);
         }
         else {
+          await this.idCardHandler.addNode(nodeId);
           subscriber.sync(handshakeData.lastMessageId);
           global.kuzzle.log.info(`Successfully completed the handshake with node ${nodeId}`);
         }
--- a/test/cluster/node.test.js
+++ b/test/cluster/node.test.js
@@ -640,6 +640,7 @@ describe('#Cluster Node', () => {
       should(node.command.getFullState).not.called();
       should(node.fullState.loadFullState).not.called();
       should(node.heartbeatTimer).not.be.null();
+      should(node.idCardHandler.addNode).not.called();
     });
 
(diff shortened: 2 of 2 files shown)


Shiranuit (commented):
Can't put it directly on the lines so I ask in this message

In lib/cluster/node.js:374
```javascript
if (nodes.length === 0) {
  this.trackActivity(this.nodeId, this.ip, nodeActivityEnum.ADDED);
  return;
}
```

I was wondering, shouldn't the mutex be released before returning ? can't the mutex prevent future node handshake operation if he has not be released before since this mutex he done through Redis if I am not wrong ?

alexandrebouthinon (approved):

scottinet:
@Shiranuit > the `finally` clause is always executed, even in the case of an return

Shiranuit (approved):

PR #2057, opened 2021-05-04
Prevent conflicting names/characters in new indexes/collections

scottinet:
# Description

When creating indexes or collections, it is possible to include a `+` character in them, or to name them `_all`.
This makes those indexes/collections unusable: Kuzzle impose a strict control to prevent multi-indexes queries, which suppose the use of either a `+` character or the `_all` ES keyword.

Since ES index names are not escapable to prevent multi-indexes queries, we have to forbid those characters/keywords from being used in index/collection names.

# How to reproduce

Create a new collection with a `+` character in its name, or name it `_all`:

* `kourou collection:create foo 'bar+qux'`
* `kourou collection:create foo _all`

Then try to use these collections to create documents or try to search in those: Kuzzle will complain and will make the query fail.

With this PR, it is no longer possible to create such indexes or collections.

### DIFF
--- a/test/service/storage/elasticsearch.test.js
+++ b/test/service/storage/elasticsearch.test.js
@@ -1396,7 +1396,7 @@ describe('Test: ElasticSearch service', () => {
     let query;
     let changes;
     let request;
-  
+
     beforeEach(() => {
       query = {
         match: { foo: 'bar' }
@@ -3703,7 +3703,7 @@ describe('Test: ElasticSearch service', () => {
       toImport[1]._source.default.country = 'Vietnam';
 
       const result = await elasticsearch.mUpsert(index, collection, documents);
-    
+
       should(elasticsearch._mExecute).be.calledWithMatch(
         esRequest,
         toImport,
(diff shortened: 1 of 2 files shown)


test/service/storage/elasticsearch.test.js
```diff
@@ -4463,13 +4463,32 @@ describe('Test: ElasticSearch service', () => {
...
-          internalESIndex = internalES._getESIndex('nepali', 'mehry');
+        const publicESIndex = publicES._getESIndex('nepali', 'liia');
+        const internalESIndex = internalES._getESIndex('nepali', 'mehry');
 
         should(publicESIndex).be.eql('&nepali.liia');
         should(internalESIndex).be.eql('%nepali.mehry');
       });
+
+      it('should throw if the index is invalid', () => {
+        for (const invalid of [ null, '', 123, true, 'foo+bar', '_all', 'HELP']) {
```
alexandrebouthinon, line 4474: 👌

alexandrebouthinon (approved):

Shiranuit (approved):

PR #2059, opened 2021-05-05
Websocket cookie propagation after Upgrade

Shiranuit:
## What does this PR do ?

This PR finalize the websocket cookie support, by adding an handler when an http request needs to be upgraded to a websocket connection.
This upgrade handler retrieve the cookie if present in the http request and store it in the socket UserData before upgrading the request, this will then trigger the websocket connection and set the cookie header for every subsequent request made with the established connection.

### DIFF
--- a/test/core/network/protocols/websocket.test.js
+++ b/test/core/network/protocols/websocket.test.js
@@ -11,6 +11,7 @@ const ClientConnection = require('../../../../lib/core/network/clientConnection'
 const KuzzleMock = require('../../../mocks/kuzzle.mock');
 const uWSMock = require('../../../mocks/uWS.mock');
 const EntryPointMock = require('../../../mocks/entrypoint.mock');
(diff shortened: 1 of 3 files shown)


rolljee (approved):

xbill82 (approved):

test/core/network/protocols/websocket.test.js
```diff
@@ -132,6 +134,37 @@ describe('core/network/protocols/websocket', () => {
     });
   });
 
+  describe('upgrade connection', () => {
+    beforeEach(() => httpWs.init(entryPoint));
+
+    it('should upgrade the connection and forward and store the cookie if present when upgrading', () => {
```
Leodau, line 140: I do not understand exactly the description of this scenario,
Can you update this to be more explicit?

Leodau (approved):

Issue #2035, opened 2021-03-30
Getting started, General Purpose Backend: some links are not working


jsatge:
## Expected Behavior
The links should redirect on an existing page.

## Current Behavior
Some links redirect on a non-existent page ("C# .NET Core" link on the Backend part, all links in Frontend Web part, "React Native", "Flutter" and "C# .NET Core" links on the Frontend Mobile part).
![Capture d’écran du 2021-03-30 17-05-35](https://user-images.githubusercontent.com/62239716/113012149-ca3d8880-917a-11eb-989b-3cb11e419e9d.png)

## Possible Solution
Delete the "/core/2" on the url to redirect on the right page.

Issue #2034, opened 2021-03-30
Getting started, Customize API Behaviour: import for KuzzleRequest is not specified


jsatge:
## Current Behavior
If we simply follow the documentation the code doesn't compile because "KuzzleRequest" is not found.

## Possible Solution
Precise that you have to add the import for KuzzleRequest in your code.

Issue #2033, opened 2021-03-30
Getting started, Authenticate Users: in an exemple the command and the results are in the same frame


jsatge:
## Current Behavior
 
![screenshot](https://user-images.githubusercontent.com/62239716/113008838-b2b0d080-9177-11eb-98f7-dc92ec038f79.png)

## Possible Solution

The command to enter and the result could be separated so the user can simply copy the command.

PR #2062, opened 2021-05-14
Fix documentation (dead links, troubleshooting, missing import)

### LINKED ISSUE #1997: Troubleshooting about permissions when running Kuzzle with docker
## Expected Behavior
When running `npm run dev:docker`, ES, Redis and Kuzzle services are supposed to be up and ready

## Current Behavior
When running `npm run dev:docker`, Kuzzle container fail because of missing permissions to access docments created by `kourou app:scaffold my-project`

## Possible Solution
Complete the [installation](https://docs.docker.com/engine/install/linux-postinstall/) of Docker to manage docker as a non root user

## Steps to Reproduce
Follow step by step the [documentation](https://docs.kuzzle.io/core/2/guides/getting-started/run-kuzzle/#let-s-go), especially the let's go section.

## Context (Environment)
Kuzzle version: 2.x
Node.js version: 12.18.1

MathieuVeber:
## What does this PR do ?

- Correct dead links
- Add precision for Kuzzle installation prerequisite
- Add missing import in code snippets

Closes #1997, #2033, #2034, #2035

### DIFF
--- a/doc/2/guides/getting-started/run-kuzzle/index.md
+++ b/doc/2/guides/getting-started/run-kuzzle/index.md
@@ -32,7 +32,7 @@ You can install Kourou globally by using NPM: `npm install -g kourou`
 
 ::: warning
 Kuzzle uses compiled C++ dependencies so a compile toolchain (a C++ compiler like g++ or clang, make and python) is necessary to run `npm install kuzzle`.  
-For the sake of simplicity we will use a Docker and Docker Compose throughout this guide.  
+For the sake of simplicity we will use a Docker and Docker Compose throughout this guide. Make sure to have them correctly installed before going further, especially the [Docker as a non root user section](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user) which is a recurrent trouble.
 ::: 
 
 ### Let's go!
--- a/doc/2/guides/getting-started/run-kuzzle/index.md
+++ b/doc/2/guides/getting-started/run-kuzzle/index.md
@@ -32,7 +32,11 @@ You can install Kourou globally by using NPM: `npm install -g kourou`
 
 ::: warning
 Kuzzle uses compiled C++ dependencies so a compile toolchain (a C++ compiler like g++ or clang, make and python) is necessary to run `npm install kuzzle`.  
-For the sake of simplicity we will use a Docker and Docker Compose throughout this guide. Make sure to have them correctly installed before going further, especially the [Docker as a non root user section](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user) which is a recurrent trouble.
+For the sake of simplicity we will use a Docker and Docker Compose throughout this guide. 
+
+::: info
+If you encounter issues when trying to use Docker with a non-root user, please follow the [Docker as a non root user guide](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user).
+:::
 ::: 
 
 ### Let's go!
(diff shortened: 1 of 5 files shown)


doc/2/guides/getting-started/run-kuzzle/index.md
```diff
@@ -32,7 +32,7 @@ You can install Kourou globally by using NPM: `npm install -g kourou`
 
 ::: warning
 Kuzzle uses compiled C++ dependencies so a compile toolchain (a C++ compiler like g++ or clang, make and python) is necessary to run `npm install kuzzle`.  
-For the sake of simplicity we will use a Docker and Docker Compose throughout this guide.  
+For the sake of simplicity we will use a Docker and Docker Compose throughout this guide. Make sure to have them correctly installed before going further, especially the [Docker as a non root user section](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user) which is a recurrent trouble.
```
Leodau, line 35:
```suggestion
For the sake of simplicity we will use Docker and Docker Compose throughout this guide. Make sure to have them correctly installed before going further, especially the [Docker as a non root user section](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user) which is a potential vulnerability.
```

Leodau (approved):

doc/2/guides/getting-started/run-kuzzle/index.md
MathieuVeber: I'm not sure "vulnerability" is the right word for that, it's more like people often can't run a Kuzzle for the first time because they try to use is as a non root user... Maybe something more like:
```suggestion
For the sake of simplicity we will use Docker and Docker Compose throughout this guide. Make sure to have them correctly installed before going further, especially the [Docker as a non root user section](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user) which is a potential issue.
```

doc/2/guides/introduction/general-purpose-backend/index.md
```diff
@@ -190,18 +190,18 @@ Or start to develop a client application by using one of our [SDKs](/sdk):
  - [Node.js](/sdk/js/7/getting-started/node-js)
  - [Java](/sdk/jvm/1/getting-started/java)
  - [Kotlin](/sdk/jvm/1/getting-started/kotlin)
- - [C# .NET Core](/sdk/csharp/2/getting-started/standalone/)
+ - [C# .NET Core](https://docs.kuzzle.io/sdk/csharp/2/getting-started/standalone/)
```
scottinet, line 193: That's not the way to solve this deadlink issue: our documentation system is capable of inferring URLs as long as they point to our documentation itself.

But there is a trick you need to know: for it to work, you cannot put a trailing slash at the end of an URL. I don't remember why, I think that's a known issue with how vuesomething processes URLs.

That's why the Node, java or kotlink links work, and why the C#, react, vue and others don't.

So you should do this instead, and apply that change to other links below:

```suggestion
 - [C# .NET Core](/sdk/csharp/2/getting-started/standalone)
```

scottinet (changes requested):

doc/2/guides/getting-started/run-kuzzle/index.md
rolljee: I personnaly like the way @scottinet proposed it 👍

rolljee (approved):

doc/2/guides/introduction/general-purpose-backend/index.md
MathieuVeber: Ok thanks for the trick, I wouldn't have guessed it otherwise :+1:

xbill82 (approved):

scottinet (approved):

PR #2064, opened 2021-05-18
Add methods to list and get loaded plugins

Aschen:
## What does this PR do ?

Adds two methods:
 - `BackendPlugin.list`: list loaded plugins
 - `BackendPlugin.get`: get a loaded plugin instance
 
 ### Boyscout

  - remove obsolete `CHANGELOG.md` file
  - use the new `request.response.configure` method instead of `request.setResult`


scottinet (commented):
It would be nice to have an API route to list plugins too, in the `server` controller.
Something like `server:plugins` for instance.

This would be helpful when trying to get that information, as this would prevent having to scroll through the very lengthy `server:info` response.

lib/kerror/codes/4-plugin.json
```diff
@@ -87,6 +87,12 @@
           "code": 14,
           "message": "Cannot define new controllers in the \"api\" and the \"controllers\" objects at the same time",
           "class": "PluginImplementationError"
+        },
+        "plugin_not_found": {
+          "description": "The request plugin does not exists or have not been loaded yet.",
+          "code": 15,
+          "message": "Plugin \"%s\" not found.%s",
+          "class": "PluginImplementationError"
```
scottinet, line 95: I don't think this is a plugin implementation error. Wouldn't a `NotFoundError` be more appropriate?

scottinet (changes requested):

Aschen:
> It would be handy to have an API route to list plugins too, in the `server` controller.
> Something like `server:plugins` for instance.
> 
> This would be helpful when trying to get that information, as this would prevent having to scroll through the very lengthy `server:info` response, or resorting to study Kuzzle's logs.

Even if I agree with you that the `server:info` response is quite huge and it's an hassle to scroll it to the plugins part (specially because the API part is at the beginning and very long), I'm not sure that we need another API action just to returns an array of plugin names.

scottinet:
> Even if I agree with you that the `server:info` response is quite huge and it's an hassle to scroll it to the plugins part (specially because the API part is at the beginning and very long), I'm not sure that we need another API action (with tests, doc, etc) just to returns an array of plugin names.

Well... I had a use case just a few days ago, with a client having trouble to know if a plugin was there or not. 
Again, this is only a "nice to have" feature, but one that will surely prove useful when doing support.

scottinet (approved):

Aschen:
@scottinet But why didn't you asked for the response of the `server:info` route ?

Maybe an alternative could be to provide a Kourou command and parse the `server:info` response

lib/api/controllers/serverController.js
```diff
@@ -222,17 +223,19 @@ class ServerController extends NativeController {
     };
   }
 
+  async plugins () {
+
```
xbill82, line 227: What is this for?

xbill82 (approved):

lib/api/controllers/serverController.js
Aschen: Leftover

PR #2068, opened 2021-05-21
Cluster: better IP inference and configuration

scottinet:
# Description

The current cluster configuration has problems with IP detection:
* when setting private addresses, the cluster regularly chooses the loopback IP address, which causes problems obviously as nodes will try to connect to themselves (partially mitigated with #2050)
* not enough configuration available: it's impossible to choose the right IP address if multiple network interfaces are available with the same IP family and the same type of IPs (public or private)
* the kuzzlerc documentation about the `ip` setting is vague, and it's easy to get it wrong

This PR solves these problem by:
* rewriting the `ip` configuration documentation in the kuzzlerc documentation
* introducing a new `interface` configuration, allowing admins to choose the network interface, IP or MAC address they want the node to use

To do that, I removed the obsolete and clunky `ip` NPM module. Instead, I rely on native Node.js features, and  sift through its results according to the provided configuration.
And I automatically remove any and all internal addresses, unfit for cluster communications (e.g. loopback addresses).

Also, the default `ip` target will now be "private", as the most common use case is to use a kuzzle cluster over a LAN.
It was switched to "public" because the "ip" module regularly returned the loopback address when asked for a private one.

### DIFF
--- a/lib/cluster/node.js
+++ b/lib/cluster/node.js
@@ -72,10 +72,11 @@ function isPrivateIP(ip) {
 
 /**
  * Return the first IP address matching the provided configuration
- * @param  {String} options.family     [description]
- * @param  {[type]} options.interface: netInterface  [description]
- * @param  {[type]} options.ip         [description]
- * @return {[type]}                    [description]
+ * @param  {Object} [options]
+ * @param  {String} [options.family] IP family (IPv4 or IPv6)
+ * @param  {String} [options.interface] Network interface/IP/MAC to use
+ * @param  {String} [options.ip] Used to target public or private addresses
+ * @return {String|null}
  */
 function getIP ({ family = 'IPv4', interface: netInterface, ip } = {}) {
   const mustBePrivate = ip === 'private';
--- a/lib/cluster/node.js
+++ b/lib/cluster/node.js
@@ -59,9 +59,10 @@ function isPrivateIP(ip) {
   const exploded = ip.split('.').map(s => Number.parseInt(s));
 
   return exploded[0] === 10
-    || exploded[0] === 127
     || (exploded[0] === 172 && exploded[1] >= 16 && exploded[1] <= 31)
     || (exploded[0] === 192 && exploded[1] === 168)
+    // Should never be considered: this range is reserved for internal uses only
+    || exploded[0] === 127
     // 169.254.x.x addresses are... particular. I *think* they are usable if
     // multiple machines share the same DHCP rules, but I don't really know.
     // This might be an edge case.
--- a/lib/cluster/node.js
+++ b/lib/cluster/node.js
@@ -529,11 +529,11 @@ class ClusterNode {
         else {
           await this.idCardHandler.addNode(nodeId);
           subscriber.sync(handshakeData.lastMessageId);
-          global.kuzzle.log.info(`Successfully completed the handshake with node ${nodeId}`);
+          global.kuzzle.log.info(`[CLUSTER] Successfully completed the handshake with node ${nodeId}`);
         }
       }
 
-      global.kuzzle.log.info('Successfully joined the cluster.');
+      global.kuzzle.log.info('[CLUSTER] Successfully joined the cluster.');
       this.trackActivity(this.nodeId, this.ip, nodeActivityEnum.ADDED);
     }
     finally {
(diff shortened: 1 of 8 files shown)


lib/cluster/node.js
```diff
@@ -50,7 +133,17 @@ class ClusterNode {
...
+    const family = this.config.ipv6 ? 'IPv6' : 'IPv4';
+
+    this.ip = getIP({
+      family,
+      interface: this.config.interface,
+      ip: this.config.ip,
+    });
+
+    if (this.ip === null) {
+      throw new Error(`[FATAL] Cluster: no suitable IP address found with the provided configuration (family: ${family}, interface: ${this.config.interface}, ip: ${this.config.ip})`);
```
Aschen, line 145: Same comment about cluster errors that prevent Kuzzle to start

lib/config/index.js
```diff
@@ -211,7 +211,8 @@ function checkClusterOptions (config) {
 
   assert(typeof cfg.ipv6 === 'boolean', '[FATAL] kuzzlerc.cluster.ipv6: boolean expected');
 
-  assert(typeof cfg.ip === 'string', '[FATAL] kuzzlerc.cluster.ip: value must be a string');
+  assert(!cfg.ip || ['private', 'public'].includes(cfg.ip), '[FATAL] kuzzlerc.cluster.ip: invalid value (accepted values: public, private)');
```
Aschen, line 214: Same comment here about the error message prefix but maybe we could have a `[CONFIG]` prefix?

I know it's not quite related to the scope of this PR so we can first talk about it and then do it in a different PR.
I started to think about this when I was trying to debug one of our client with a huge log file and I was trying to find only cluster related messages.

package.json
```diff
@@ -58,7 +58,6 @@
     "eventemitter3": "^4.0.7",
     "inquirer": "^8.0.0",
     "ioredis": "^4.27.2",
-    "ip": "^1.1.5",
```
Aschen, line 61: Less is more <3

Aschen (changes requested):

lib/cluster/node.js
```diff
@@ -35,6 +37,87 @@ const ClusterState = require('./state');
...
+
+/**
+ * Test an IP address and determine if it's in the public or private range.
+ *
+ * @param  {String}  ip
+ * @return {Boolean}
+ */
+function isPrivateIP(ip) {
+  if (net.isIP(ip) === 0) {
+    throw new Error(`[FATAL] "${ip}" is not a valid IP address`);
```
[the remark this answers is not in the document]
scottinet: Ok for both remark.

Aschen (approved):

lib/cluster/node.js
```diff
@@ -35,6 +38,85 @@ const ClusterState = require('./state');
...
+    || (exploded[0] === 172 && exploded[1] >= 16 && exploded[1] <= 31)
+    || (exploded[0] === 192 && exploded[1] === 168)
+    // Should never be considered: this range is reserved for internal uses only
+    || exploded[0] === 127
+    // 169.254.x.x addresses are... particular. I *think* they are usable if
+    // multiple machines share the same DHCP rules, but I don't really know.
+    // This might be an edge case.
+    // For now I'll consider them as private addresses, but we might want to
+    // consider them as internal ones if that poses problems in the future.
+    || (exploded[0] === 169 && exploded[1] === 254);
```
alexandrebouthinon, line 70: I was thinking about APIPA addresses this weekend. If I remember correctly they do not allow network communication with other addresses of the same type. They are more of a stopgap to allow more "classical" services that need an IP address to start without any problems. To me we should return an error if the only available addresses are APIPAs. Otherwise we may find ourselves in error cases that are very annoying to debug.

What do you think?

alexandrebouthinon (approved):

lib/cluster/node.js
scottinet: I thought I read something like that, but I was not entirely sure.

If APIPA addresses are not reachable from other machines, then we must consider them as internal addresses, exactly like loopback ones.

I'll amend my PR to reflect that.

rolljee (approved):

alexandrebouthinon (approved):

Issue #2067, opened 2021-05-20
Kuzzle action with ES queries should not tolerate script keyword


MathieuVeber:
## Example

URL: `http://kuzzle:7512/<index>/<collection>/_search`
Method: POST
Body:
```json
{
  "query": {
		"script": {
			"script": {
				"inline": "doc['name'].value.length() < params.length",
				"params": {
					"length": 5
				}
			}
		}
	}
}
```

## Expected Behavior

An error should occur and the script should not be executed

Response:
```json
{
  "action": "search",
  "collection": "test",
  "controller": "document",
  "error": {
    "code": 16842792,
    "id": "services.storage.invalid_search_query",
    "message": "The argument \"script\" is not allowed in a search query.",
    "stack": "      BadRequestError: The argument \"script\" is not allowed in a search query.\n      [...Kuzzle internal calls deleted...]\n>    at ElasticSearch._sanitizeSearchBody (/var/app/lib/service/storage/elasticsearch.js:2701:22)\n>    at ElasticSearch.search (/var/app/lib/service/storage/elasticsearch.js:344:18)\n>    at /var/app/lib/core/storage/clientAdapter.js:644:28\n>    at Kuzzle.ask (/var/app/lib/kuzzle/event/kuzzleEventEmitter.js:192:28)\n>    at DocumentController.search (/var/app/lib/api/controllers/documentController.js:87:31)\n>    at doAction (/var/app/lib/api/funnel.js:846:47)\n>    at Funnel.processRequest (/var/app/lib/api/funnel.js:510:34)",
    "status": 400
  },
  "index": "index",
  "node": "knode-educated-droopy-59177",
  "requestId": "8cd4e516-9eae-44b4-a0b7-99432dbf389d",
  "result": null,
  "status": 200,
  "volatile": null
}
```

## Current Behavior

The script is executed successfuly

Response:
```json
{
  "action": "search",
  "collection": "test",
  "controller": "document",
  "error": null,
  "index": "index",
  "node": "knode-educated-droopy-59177",
  "requestId": "8cd4e516-9eae-44b4-a0b7-99432dbf389d",
  "result": {
    "hits": [
      {
        "_id": "1",
        "_score": 1,
        "_source": {
          "age": 3,
          "name": "test",
          "_kuzzle_info": {
            "author": "-1",
            "createdAt": 1621418041525,
            "updatedAt": null,
            "updater": null
          }
        }
      }
    ],
    "total": 1
  },
  "status": 200,
[...]

PR #2070, opened 2021-05-26
Kuzzle action with ES queries should not tolerate script keyword

MathieuVeber:
## What does this PR do ?

Ensure that scripts are forbidden in an ES query.
A check has been added in _sanitizeSearchBody which should be fine for almost all requests (current and future) and bulk import has been specifically handled.
A new related error has been created. Tests have been added for both new _scriptCheck and old _sanitizeSearchBody

Follow the instructions written in the linked issue #2067

### Boyscout

Safer isPlainObject helper

### DIFF
--- a/lib/service/storage/elasticsearch.js
+++ b/lib/service/storage/elasticsearch.js
@@ -2729,7 +2729,7 @@ class ElasticSearch extends Service {
   _scriptCheck(object) {
     for (const [key, value] of Object.entries(object)) {
       if (this.scriptKeys.includes(key)) {
-        throw kerror.get('invalid_search_query', key);
+        throw kerror.get('invalid_query_keyword', key);
       }
       else if (typeof value === 'object' && value !== null) {
         this._scriptCheck(value);
--- a/lib/service/storage/elasticsearch.js
+++ b/lib/service/storage/elasticsearch.js
@@ -124,6 +124,12 @@ class ElasticSearch extends Service {
       '_source_includes'
     ];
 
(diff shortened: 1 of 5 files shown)


lib/service/storage/elasticsearch.js
```diff
@@ -2711,6 +2721,22 @@ class ElasticSearch extends Service {
...
+   * Throw if any script keyword is contained in the object  
+   * 
+   * @param {Object} object
+   */
+  _scriptCheck(object) {
+    for (const [key, value] of Object.entries(object)) {
+      if (this.scriptKeys.includes(key)) {
+        throw kerror.get('invalid_query_keyword', key);
+      }
+      else if (typeof value === 'object' && value !== null) {
```
scottinet, line 2734: `isPlainObject` should be used instead: as it is, this function will also scan arrays. This won't have any impact on the feature (array indexes are numerals), but this makes Kuzzle do useless work.

MathieuVeber: Indeed we should avoid useless work but we also need to go deeper in the array (in a boolean query for example), I'll correct that consequently.

scottinet: oh, you're right, I didn't consider arrays containing subqueries

don't change it then, it's fine like this (maybe add a quick comment to prevent others to make the same mistake than me?)

Yoann-Abbes (approved):

Leodau (approved):

PR #2081, opened 2021-06-08
Allows to use aliased collection with forbidden characters

Aschen:
## What does this PR do ?

Collection cannot contain some characters (like uppercase for example).

It's possible to create an alias with some of the characters forbidden by Kuzzle and making it pointing to another indice representing an actual Kuzzle collection.

Kuzzle was checking collection name validity even when performing a search or creating a document so it was not possible to use those aliased collections.

This check is now only performed when an index or a collection is created

### DIFF
--- a/lib/service/storage/elasticsearch.js
+++ b/lib/service/storage/elasticsearch.js
@@ -1959,7 +1959,7 @@ class ElasticSearch extends Service {
    *
    * @returns {Promise.<{ items, errors }>
    */
-  async mCreateOrReplace(
+  async mCreateOrReplace (
     index,
     collection,
     documents,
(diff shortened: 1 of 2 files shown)


lib/service/storage/elasticsearch.js
```diff
@@ -2590,8 +2590,6 @@ class ElasticSearch extends Service {
    * @returns {String} esIndex name (eg: '&nepali#liia')
    */
   _getESIndex (index, collection) {
-    this._assertValidIndexAndCollection(index, collection || '');
```
Aschen, line 2593: Actual fix

Yoann-Abbes (approved):

Shiranuit (approved):

Issue #2075, opened 2021-05-27
Add a max TTL config when generating API key in the auth controller


Aschen:
## Feature Description

A max TTL value should be configurable when users want to generate an API key with the `auth` controller.

This limit should not apply to API key generated with the `security` controller

PR #2083, opened 2021-06-09
Add a configurable maximum TTL for API keys

berthieresteban:
## What does this PR do ?

Add an `apiKey` section and a `maxTTL` property to default config
Force to `bypassMaxTTL` when api key creation requested from `security` controller
Use apiKey `maxTTL` value for api-keys
#2075 

  - Step 1 : Set an `apiKey.maxTTL` in config
  - Step 2 : Try api key creation using both `auth` and `security` controllers and with higher/lower ttl

### DIFF
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -240,6 +240,16 @@
       "maxTTL": -1,
       "secret": null
     },
+    // [apiKey]
+    // configuration for the api keys
+    //
+    //   * maxTTL:
+    //      Maximum duration in milliseconds a token can be requested
+    //      to be valid.
+    //      If set to -1 (default), no maximum duration is set.
+    "apiKey": {
+      "maxTTL": -1
+    },
     // [default]
     // The default role defines permissions for all users,
     // until an administrator configures the backend rights
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -241,12 +241,11 @@
       "secret": null
     },
     // [apiKey]
-    // configuration for the api keys
+    // API Keys configuration
     //
     //   * maxTTL:
-    //      Maximum duration in milliseconds a token can be requested
-    //      to be valid.
-    //      If set to -1 (default), no maximum duration is set.
+    //      Maximum time to live of an API Key, in milliseconds
+    //      If set to -1 (default), API Keys can be created with an infinite duration.
     "apiKey": {
       "maxTTL": -1
     },
--- a/doc/2/guides/advanced/api-keys/index.md
+++ b/doc/2/guides/advanced/api-keys/index.md
@@ -25,6 +25,9 @@ By default, **API keys do not expire**. But it is possible to specify the durati
 ::: info
 It is also possible to set a maximum validity period for an API key under the key `security.apiKey.maxTTL` in the Kuzzle configuration.
 This limit will only apply to API key created with the `auth` controller.
+Possible values: 
+  - `<= -1`: disable the use of maxTTL
+  - `>= 0`: enable maxTTL with setted value (`0` will invalid all your API keys at their creation)
 :::
 
 It is also necessary to **provide a description** of the API key.
--- a/doc/2/guides/advanced/api-keys/index.md
+++ b/doc/2/guides/advanced/api-keys/index.md
@@ -24,6 +24,7 @@ By default, **API keys do not expire**. But it is possible to specify the durati
 
 ::: info
 It is also possible to set a maximum validity period for an API key under the key `security.apiKey.maxTTL` in the Kuzzle configuration.
+This limit will only apply to API key created with the `auth` controller.
 :::
 
 It is also necessary to **provide a description** of the API key.
--- a/doc/2/guides/advanced/api-keys/index.md
+++ b/doc/2/guides/advanced/api-keys/index.md
@@ -22,6 +22,10 @@ Administrators can create API keys for other users using the [security:createApi
 
 By default, **API keys do not expire**. But it is possible to specify the duration of an API key using the argument `expiresIn`.
 
+::: info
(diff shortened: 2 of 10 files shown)


.kuzzlerc.sample
```diff
@@ -240,6 +240,16 @@
       "maxTTL": -1,
       "secret": null
     },
+    // [apiKey]
+    // configuration for the api keys
```
scottinet, line 244:
```suggestion
    // API Keys configuration
```

.kuzzlerc.sample
```diff
@@ -240,6 +240,16 @@
...
+    //
+    //   * maxTTL:
+    //      Maximum duration in milliseconds a token can be requested
+    //      to be valid.
+    //      If set to -1 (default), no maximum duration is set.
```
scottinet, line 249:
```suggestion
    //      Maximum time to live of an API Key, in milliseconds
    //      If set to -1 (default), API Keys can be created with an infinite duration.
```

lib/core/security/tokenRepository.js
```diff
@@ -189,9 +189,16 @@ class TokenRepository extends Repository {
 
     const parsedExpiresIn = parseTimespan(expiresIn);
 
+    let maxTTL;
+    if (type === 'apiKey') {
+      maxTTL = global.kuzzle.config.security.apiKey.maxTTL;
+    } else {
+      maxTTL = global.kuzzle.config.security.jwt.maxTTL;
+    }
```
scottinet, line 197: (nitpicking) More concise way of writing this:

```suggestion
    const maxTTL = type === 'apiKey'
      ? global.kuzzle.config.security.apiKey.maxTTL
      : global.kuzzle.config.security.jwt.maxTTL;
```

scottinet (approved):
A few nitpickings, but otherwise LGTM

lib/core/security/tokenRepository.js
```diff
@@ -189,9 +189,16 @@ class TokenRepository extends Repository {
...
+
     if ( ! bypassMaxTTL
-      && global.kuzzle.config.security.jwt.maxTTL > 0
-      && ( parsedExpiresIn > global.kuzzle.config.security.jwt.maxTTL
+      && maxTTL > 0
```
scottinet, line 200: Hm, edge case: setting this configuration `0` will act as if it was set to `-1`. Is this intended?
What behavior should we define if maxTTL is 0?

Proposal: it makes no sense having a TTL of a few milliseconds, so a lower bound value might be necessary. Something like 30s or 60s for instance, since anything below that is a bit too short to be really usable. So if maxTTL is set to anything between 0 and that value, it would be defaulted to that lower bound. 

wdyt?

scottinet (changes requested):

scottinet:
Also, I don't think this is a new feature: you're adding configurability to an existing feature, so it's more an enhancement to me.

doc/2/guides/advanced/api-keys/index.md
```diff
@@ -22,6 +22,10 @@ Administrators can create API keys for other users using the [security:createApi
 
 By default, **API keys do not expire**. But it is possible to specify the duration of an API key using the argument `expiresIn`.
 
+::: info
+It is also possible to set a maximum validity period for an API key under the key `security.apiKey.maxTTL` in the Kuzzle configuration.
```
Aschen, line 26: You should add a sentence saying that this limit only apply to API key created with the `auth` controller

Aschen, line 26: You should add a sentence saying that this limit only apply to API key created with the `auth` controller

lib/api/controllers/authController.js
```diff
@@ -117,6 +117,7 @@ class AuthController extends NativeController {
 
     const apiKey = await ApiKey.create(user, expiresIn, description, {
       apiKeyId,
+      bypassMaxTTL: false,
```
Aschen, line 120: Useless since that's the default value

lib/model/storage/apiKey.js
```diff
@@ -90,10 +90,10 @@ class ApiKey extends BaseModel {
     user,
     expiresIn,
     description,
-    { creatorId=null, apiKeyId=null, refresh } = {}
+    { creatorId=null, apiKeyId=null, refresh, bypassMaxTTL=false } = {}
```
Aschen, line 93: You should have this option documentation in the JSdoc

Aschen (changes requested):

Shiranuit (approved):

lib/core/security/tokenRepository.js
berthieresteban: Your solution sounds good to me but I don't really know if we should decide about a minimum value since I don't know the level of rigidity we decided for other configuration properties
I think we need to handle that case anyway because -1 should be the only value that could disable the use of maxTTL
We also could just handle 0 as any other maxTTL value with `maxTTL >  -1` and let this to the user

scottinet: well, that's another way to see it: no restriction, let kuzzle act exactly as told, and if users want to put unusable configuration, then let them

I'm ok with that.

Aschen (approved):

lib/core/security/tokenRepository.js
berthieresteban: Any other opinion ? @Shiranuit @Aschen

Leodau: No restriction makes sense, well documented is mandatory tho. :+1:

Shiranuit: I'm ok with both of your solutions, but maybe we should try to keep the same principle across Kuzzle, I mean, if we choose to not babysit the users and let them configure Kuzzle the way they intended too, even if this make no sense, then we should apply this principle all the time if possible to be coherent.

scottinet (approved):

Issue #1882, opened 2020-11-24
Use the authentication token fingerprint as API key ID


Aschen:
## Feature description

We should use the authentication token fingerprint as the API key IDs

PR #2084, opened 2021-06-09
Use API keys fingerprint as their default ID

berthieresteban:
## What does this PR do ?

Use the authentication token fingerprint as default api key ID.
#1882 

  - Step 1 : create an api-key without apiKeyId
  - Step 2 : check the match of apiKeyId and fingerprint

### DIFF
--- a/lib/model/storage/apiKey.js
+++ b/lib/model/storage/apiKey.js
@@ -102,7 +102,7 @@ class ApiKey extends BaseModel {
     const apiKey = new ApiKey({
       description,
       expiresAt: token.expiresAt,
-      fingerprint: fingerprint,
+      fingerprint,
       token: token.jwt,
       ttl: token.ttl,
       userId: user._id,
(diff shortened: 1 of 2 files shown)


lib/model/storage/apiKey.js
```diff
@@ -98,15 +98,16 @@ class ApiKey extends BaseModel {
       type: 'apiKey',
     });
 
+    const fingerprint = sha256(token.jwt);
     const apiKey = new ApiKey({
       description,
       expiresAt: token.expiresAt,
-      fingerprint: sha256(token.jwt),
+      fingerprint: fingerprint,
```
scottinet, line 105: (nitpicking)

```suggestion
      fingerprint,
```

scottinet (approved):

scottinet:
I have changed the title: those are used to create Kuzzle's changelog, so we need to have explicit and understandable titles.

Aschen (approved):

Shiranuit (approved):

PR #2085, opened 2021-06-10
Expose Kuzzle API limits

### LINKED ISSUE #1963: Expose Kuzzle API capacities
## Feature description

Expose API capacities to users. Then they can know how many documents they can write or how many login per second they can make.

## Example usecase

SDKs could then automatically split `document:m*` actions to comply with the API limits. 

`server:capacity`
`server:limits`

berthieresteban:
## What does this PR do ?

fix #1963 
Expose `server:limits` returning the following properties from kuzzle configuration:
  - concurrentRequests
  - documentsFetchCount
  - documentsWriteCount
  - loginsPerSecond

  - Step 1 : kourou server:limits

### DIFF
--- a/doc/2/api/controllers/server/limits/index.md
+++ b/doc/2/api/controllers/server/limits/index.md
@@ -1,0 +1,55 @@
+---
+code: true
+type: page
+title: limits
+---
+
+# limits
+
+
+
+Returns the kuzzle configuration limits.
+
+---
+
+## Query Syntax
+
+### HTTP
+
+```http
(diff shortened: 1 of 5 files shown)


Aschen (changes requested):
Rather than having a new API action that should be exposed to the SDK we could use the `publicApi` action which is already supposed to be open to allows SDK to works with HTTP protocol.

We also talked about something lower than an API action, like during the HTTP or WS handshake.

IMHO we should talk about the implementation

PR #2093, opened 2021-06-21
Fix after subscribe event

Aschen:
## Description

The event `core:realtime:user:subscribe:after` was not triggered after a successful new subscription
The event `core:realtime:user:unsubscribe:after` was not triggered after unsubscribing


Shiranuit (commented):
While we are here fixing the emit of `core:realtime:user:*` events, I dont see the event `core:realtime:user:unsubscribe:after` anywhere is that normal ?

Shiranuit (approved):

PR #2094, opened 2021-06-22
Fix auth:logout expiring apikey

Shiranuit:
## What does this PR do ?

This PR fixes an unwanted behaviour where ApiKeys used to login were expired when `auth:logout` was called
We now check if the token we are trying to expires is an ApiKey before deleting it.
`TokenRepository.deleteByKuid` has now an option named `keepApiKeys` that specifies that ApiKeys should be preserved, this is used when calling `auth:logout` with the `global=true` option, to avoid revoking sessions that are logged in with an ApiKeys, thus expiring the ApiKeys.

Added documentation and tests

### DIFF
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -175,7 +175,7 @@ class AuthController extends NativeController {
    * @param {KuzzleRequest} request
    * @returns {Promise<object>}
    */
-  async logout(request) {
+  async logout (request) {
     if (!global.kuzzle.config.internal.cookieAuthentication
       || !request.getBoolean('cookieAuth')
     ) {
@@ -191,10 +191,7 @@ class AuthController extends NativeController {
           {keepApiKeys: true});
       }
       else if ( request.context.token 
-        && !await global.kuzzle.ask(
-          'core:security:token:isApiKey',
-          request.context.token.jwt
-        )
+        && request.context.token.type !== 'apiKey'
       ) {
         await global.kuzzle.ask(
           'core:security:token:delete',
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -175,8 +175,8 @@ class AuthController extends NativeController {
    * @param {KuzzleRequest} request
    * @returns {Promise<object>}
    */
-  async logout (request) {
-    if ( !global.kuzzle.config.internal.cookieAuthentication
+  async logout(request) {
+    if (!global.kuzzle.config.internal.cookieAuthentication
       || !request.getBoolean('cookieAuth')
     ) {
       this.assertIsAuthenticated(request);
@@ -187,9 +187,13 @@ class AuthController extends NativeController {
       if (request.getBoolean('global')) {
         await global.kuzzle.ask(
           'core:security:token:deleteByKuid',
-          request.getKuid());
+          request.getKuid(),
+          {keepApiKeys: true});
       }
-      else {
+      else if (! await global.kuzzle.ask(
+        'core:security:token:isApiKey',
+        request.context.token
+      )) {
         await global.kuzzle.ask(
           'core:security:token:delete',
           request.context.token);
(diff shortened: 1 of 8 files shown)


lib/core/security/tokenRepository.js
```diff
@@ -319,6 +329,19 @@ class TokenRepository extends Repository {
       .replace(TokenRepository.APIKEY_PREFIX, '');
   }
 
+  /** Verifies if a token is an ApiKey
+   * 
+   * @param {String} encodedToken
+   * @returns {Boolean}
+   */
+  isApiKey(token) {
```
Aschen, line 337: Instead of having a new method I would rather have an additional property in the `Token` model so anyone can easily know if a token is an authentication token or an API key

lib/core/security/tokenRepository.js
```diff
@@ -347,7 +370,7 @@ class TokenRepository extends Repository {
    * @param {string} kuid
    * @returns {Promise}
    */
-  async deleteByKuid (kuid) {
+  async deleteByKuid (kuid, {keepApiKeys = false} = {}) {
```
Aschen, line 373: IMHO it should keep the API keys by default because API keys are not supposed to be deleted massively like this

Aschen (changes requested):

lib/api/controllers/authController.js
```diff
@@ -187,12 +187,19 @@ class AuthController extends NativeController {
...
+          'core:security:token:isApiKey',
+          request.context.token.jwt
+        )
+      ) {
         await global.kuzzle.ask(
           'core:security:token:delete',
-          request.context.token);
+          request.context.token
+        );
       }
```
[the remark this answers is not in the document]
Shiranuit: Doing that will be a breaking change no ?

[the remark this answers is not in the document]
Aschen: It's not a breaking change to add property to the object returned by an existing API action

Aschen (approved):

lib/model/security/token.js
```diff
@@ -34,6 +34,16 @@ class Token {
     this.jwt = data.jwt || null;
     this.refreshed = Boolean(data.refreshed);
   }
+
+  get type() {
+    if (this.jwt && this.jwt.startsWith(Token.APIKEY_PREFIX)) {
```
xbill82, line 39: 🤔 is this system going to work when upgrading Kuzzle on a system that already contains tokens and API keys?
Maybe not, but in the meanwhile, such a Kuzzle would not tell the difference between API keys and tokens, so it's not bad.

test/api/controllers/authController.test.js
```diff
@@ -511,6 +511,16 @@ describe('Test the auth controller', () => {
...
     });
+
+    it('should not expires the token if this is an API Key', async () => {
+      Object.defineProperty(request.context.token, 'type', { get: () => 'apiKey'});
+      const response = await authController.logout(request);
+
+      should(kuzzle.ask)
+        .not.be.calledWith('core:security:token:delete', request.context.token);
+      
+      should(response.responseObject).be.instanceof(Object);
```
xbill82, line 522: What is this for? 😄

test/api/controllers/authController.test.js
```diff
@@ -573,6 +583,16 @@ describe('Test the auth controller', () => {
 
       return should(authController.logout(request)).be.rejectedWith(error);
     });
+
+    it('should not expires the token if this is an apikey', async () => {
```
xbill82, line 587:
```suggestion
    it('should not expire the token if this is an apikey', async () => {
```

xbill82 (approved):

lib/model/security/token.js
Shiranuit: Yes this will work after an upgrade even if there are tokens and API Keys contained, because tokens and api keys are already stored with a prefix (`kauth-` for tokens, and `apikey-` for api keys) this means that even if Kuzzle get's an upgrade and is restarted, tokens and apikeys will still be stored with their prefix.

test/api/controllers/authController.test.js
Shiranuit: Leftover from when I deleted unused test cases after changing the way we detect API Keys

PR #2099, opened 2021-06-24
[Hotfix] Fix race condition when creating/deleting hidden collection

Aschen:
## What does this PR do ?

Fix a race condition occurring when deleting/creating the hidden collection in index.

This error occur if multiple collection are created/deleted concurrently on the same index.

### Other changes
 -  create the hidden collection with only one shard and replica;


Shiranuit (approved):

Shiranuit:
We recently got this issue with our node_modules where `request-promise` doesn't install the `request` library automatically anymore, could you add it to the package.json so it will be installed when doing `npm install` ?
```shell
###
### The "request" library is not installed automatically anymore.
### But is a dependency of "request-promise".
### Please install it with:
### npm install request --save
###
```

PR #2106, opened 2021-07-13
Framework: BackendImport

### LINKED ISSUE #2077: Add methods to configure database and rights before startup
## Description

Allows developers to load default mappings and rights before application startup.

```
app.database.configure({
  indexA: {
    collectionZ: { mappings, settings },
    collectionW: { mappings, settings },
  }
});

app.role.configure({
  roleA: { /* role definition */ },
  roleB: { /* role definition */ },
})

app.profile.configure({
  profileA: { /* profile definition */ },
  profileB: { /* profile definition */ },
})

```

All method call should be idempotent (multiple call allowed)

MathieuVeber:
Closes #2077

## What does this PR do ?

Add a new `BackendImport` class accessible from `backend.import`.
This class offers 5 new methods:

- `mappings` Import indexes and their collections mappings (and settings)
- `profiles` Import profiles
- `roles` Import roles
- `userMappings` Import user mappings
- `users` Import users. If they already exist in DB at starting phase, we can `skip` their importation or `overwrite` DB to prevent any change on these users. 

All of these methods are __idempotent__ (can be call multiples times). Also, the __call order doesn't matter__ because it's only when the app starts that import will be made (in the right order) in the database. Finally, all of these methods except `userMappings` come with __merge ability__ when called multiple times (old and new profiles, roles, users will be combined as long as collections contained in same index).

### Changes related to functional tests 

- From now, tests order matters. BackendImport tests need to be executed before resetting database (hence **#** in `#BackendImport.feature`)
- `docker/script/start-kuzzle-dev`: `backend._support` has been replaced by multiple `backend.import` which means dev environment is slightly different.
- Deprecation warnings are removed for all tests

- Clone Kuzzle
- Import some mappings
- Start Kuzzle and see the result on the admin console


doc/2/framework/classes/backend-import/mappings/index.md
```diff
@@ -0,0 +1,52 @@
...
+|------------|-----------------------|-------------------------------|
+| `mappings` | <pre>JSONObject</pre> | Object containing index and their collection [mappings](/core/2/guides/main-concepts/data-storage#mappings-properties). |
+
+## Usage
+
+```js
+app.import.mappings({
+  index1: {
+    collection1: [
+      {/* mappings, settings */},
```
Aschen, line 37: At least one real example could be useful for developpeurs

doc/2/framework/classes/backend-import/profiles/index.md
```diff
@@ -0,0 +1,38 @@
...
+
+| Argument   | Type                  | Description                   |
+|------------|-----------------------|-------------------------------|
+| `profiles` | <pre>JSONObject</pre> | Object containing profiles and their [definitions](/core/2/guides/main-concepts/permissions#profiles). |
+
+## Usage
+
+```js
+app.import.profiles({
+  profileA: { /* profile definition */ },
```
Aschen, line 35: Same comment here, a real example could be usefull

doc/2/framework/classes/backend-import/roles/index.md
```diff
@@ -0,0 +1,38 @@
...
+
+| Argument | Type                  | Description                   |
+|----------|-----------------------|-------------------------------|
+| `roles`  | <pre>JSONObject</pre> | Object containing roles and their [definitions](/core/2/guides/main-concepts/permissions#roles). |
+
+## Usage
+
+```js
+app.import.roles({
+  roleA: { /* role definition */ },
```
Aschen, line 35: Same comment here, a real example could be usefull

doc/2/framework/classes/backend-import/users/index.md
```diff
@@ -0,0 +1,39 @@
...
+Loads users into the app.
+
+This method is idempotent. If a user is defined multiple times before the app starts with the same kuid, only the last one will be retained.
+
+::: info
+This method can only be used before the application is started.
+:::
+
+```ts
+users(users: JSONObject): void
```
Aschen, line 22: Missing `onExistingUsers` option, also it should be an object like `users(users, { onExistingUsers: 'fail' })`

lib/core/backend/backendImport.ts
```diff
@@ -0,0 +1,174 @@
...
+
+const assertionError = kerror.wrap('validation', 'assert');
+const runtimeError = kerror.wrap('plugin', 'runtime');
+
+export class BackendImport extends ApplicationManager {
+
+  /**
+   * Import mappings.
+   *
+   * Example:
```
Aschen, line 35: You should use the `@example` tag instead

lib/core/backend/backendImport.ts
```diff
@@ -0,0 +1,174 @@
...
+      // If some collections have already been defined, only their last mappings will be retained
+      this._application._import.mappings[index] =
+        Object.assign({}, this._application._import.mappings[index], mappings[index]);
+    }
+  }
+
+  /**
+   * Import profiles.
+   *
+   * Example:
```
Aschen, line 69: You should use the `@example` tag instead

lib/core/backend/backendImport.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    // If some profiles have already been defined, only their last definition will be retained
+    this._application._import.profiles =
+      Object.assign({}, this._application._import.profiles, profiles);
+  }
+
+  /**
+   * Import roles
+   *
+   * Example:
```
Aschen, line 94: You should use the `@example` tag instead

lib/core/backend/backendImport.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    // If some roles have already been defined, only their last definition will be retained
+    this._application._import.roles =
+      Object.assign({}, this._application._import.roles, roles);
+  }
+
+  /**
+   * Import user mappings.
+   *
+   * Example:
```
Aschen, line 119: You should use the `@example` tag instead

lib/core/backend/backendImport.ts
```diff
@@ -0,0 +1,174 @@
...
+      throw assertionError.get('invalid_type', 'mappings', 'object');
+    }
+
+    this._application._import.userMappings = mappings;
+  }
+
+  /**
+   * Import users.
+   *
+   * Example:
```
Aschen, line 144: You should use the `@example` tag instead

lib/core/backend/backendImport.ts
```diff
@@ -0,0 +1,174 @@
...
+   * {
+   *   kuidA: { user content },
+   *   kuidB: { user content },
+   * }
+   *
+   * @param users Object containing users and their content
+   * @param onExistingUsers Default to `skip`. Strategy to adopt when trying to create an already existing user.
+   */
+  users (users: JSONObject, onExistingUsers?: 'fail' | 'overwrite' | 'skip') : void {
```
Aschen, line 153: As said earlier, open should be passed in object for readability and extensibility

lib/kuzzle/kuzzle.js
```diff
@@ -173,8 +173,12 @@ class Kuzzle extends KuzzleEventEmitter {
 
       await this.validation.curateSpecification();
 
+      await this.internalIndex.updateMapping('users', options.import.userMappings);
```
Aschen, line 176: What if the mappings are null?

lib/kuzzle/kuzzle.js
```diff
@@ -187,11 +191,17 @@ class Kuzzle extends KuzzleEventEmitter {
...
       // ACLs must be loaded after plugins has started, otherwise importing
       // users with credentials will fail and will prevent Kuzzle from starting
-      if (options.securities) {
-        this.log.info('[ℹ] Loading default rights... This can take some time.');
-        await this.ask('core:security:load', options.securities, { force: true });
-        this.log.info('[✔] Default rights loaded');
-      }
+      this.log.info('[ℹ] Loading default rights... This can take some time.');
+
+      await this.ask('core:security:load', options.securities, { force: true });
```
Aschen, line 196: Users, profiles and rights should be loaded either with the deprecated `securities` object or with the new `import.*` objects.

If someone try to use both then we should stop the start sequence

lib/kuzzle/kuzzle.js
```diff
@@ -173,8 +173,12 @@ class Kuzzle extends KuzzleEventEmitter {
 
       await this.validation.curateSpecification();
 
+      await this.internalIndex.updateMapping('users', options.import.userMappings);
+
       await this.ask('core:storage:public:mappings:import', options.mappings);
 
+      await this.ask('core:storage:public:mappings:import', options.import.mappings);
```
Aschen, line 180: We should ensure that users are not trying to use both `import.*` and the deprecated way with `mappings`

Aschen (changes requested):

lib/kuzzle/kuzzle.js
MathieuVeber: They can't, there is an [object type check](https://github.com/kuzzleio/kuzzle/pull/2106/files#diff-aa04b68e2a0c83aa37ee408b049c460ca78e2bda7e4da0b439c13cfaf0500311R52) when the user imports mappings in first place

Aschen: Yes but if none mappings are set by the developer ?

Aschen (changes requested):
Can you add some functional tests for those methods please?

lib/core/backend/backendImport.ts
```diff
@@ -0,0 +1,177 @@
...
+    users: JSONObject,
+    options: { onExistingUsers?: 'fail' | 'overwrite' | 'skip' } = {}
+  ) : void {
+    if (this._application.started) {
+      throw runtimeError.get('already_started', 'import');
+    }
+    else if (! isPlainObject(users)) {
+      throw assertionError.get('invalid_type', 'users', 'object');
+    }
+    else if (options.onExistingUsers) {
```
Yoann-Abbes, line 163: Shoulnt we log a warning if we actually skip?

MathieuVeber: IMHO I don't think so because it's a wanted behavior but it's up for debate :wink: 
The way I see this:

1. First time the dev start Kuzzle, an admin user is added with `skip` option
2. The dev stop Kuzzle for X motive
3. Then he restart Kuzzle and expects to have its admin user but doesn't care anymore of what import is written as long as it has worked in the first place
4. The server will be stopped and started multiple times and the admin won't be mentionned but will be there durably

However, by focusing on this option for this comment, I realized that `fail` option makes no sense so I will delete it. Also I think it's not clear that `onExistingUsers` is global and applies for all `import.users` (previous and future unless a more recent option has been given), so I will add a precision in the documentation.

MathieuVeber:
> Can you add some functional tests for those methods please?

Sure I will add them, I didn't know we were writing functional tests for Backend features or when starting kuzzle

Aschen:
> Sure I will add them, I didn't know we were writing functional tests for Backend features or when starting kuzzle

Actually I remember having so many problems with the former mappings and rights loading system that I think it will be great to test this.  It just can be some objects loaded at start in the functional test applications and having their existence checked in a scenario

lib/core/backend/backendImport.ts
xbill82: As a fervent supporter of explicit behavior, I add my +1 to skip warnings.

lib/core/security/userRepository.js
```diff
@@ -290,8 +290,20 @@ class UserRepository extends Repository {
       throw kerror.get('security', 'user', 'anonymous_profile_required');
     }
 
-    await this.persistToDatabase(user, databaseOptions);
-    await this.persistToCache(user, cacheOptions);
+    try {
+      await this.persistToDatabase(user, databaseOptions);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.log('DB');
```
xbill82, line 297: Are these intentionals?

lib/kuzzle/kuzzle.js
```diff
@@ -305,6 +298,107 @@ class Kuzzle extends KuzzleEventEmitter {
...
+          throw kerror.get(
+            'plugin',
+            'runtime',
+            'incompatible',
+            '_support.securities',
+            'import profiles roles or users');
+        }
+        else if (isSecuritiesToSupport) {
+          await this.ask('core:security:load', toSupport.securities, { force: true });
+          this.log.info('[✔] Securities import successful');
```
xbill82, line 383: When will we stop using the "securities" word?

https://capital.com/fr/titres-definition

xbill82 (approved):

lib/core/security/userRepository.js
MathieuVeber: No it's just for debugging, btw I should get back to this bug!

lib/kuzzle/kuzzle.js
MathieuVeber: Sure I was foolishly repeating the pattern, `permissions` will be more accurate :wink:

lib/core/backend/backendImport.ts
MathieuVeber: Ok! The majority has spoken, I will add an explicit feedback for these actions

Aschen (approved):

PR #2107, opened 2021-07-14
Faster collection truncating with deleteByQuery

Aschen:
## What does this PR do ?

When truncating a collection, Kuzzle delete and then re-create the underlying Elasticsearch indice.   
However indice creation is not parallelizable inside an ES cluster so if we truncate many collections it will be quite slow when the deleteByQuery action is completely parallel. 

This PR change `collection:truncate` to use `deleteByQuery` which is almost always faster. 

```
// 2 000 000 documents
Big collection: delete + create collection: 192.851ms
Big collection: delete documents: 347.221ms

// 10 000 documents
Medium collection: delete + create collection: 186.285ms
Medium collection: delete documents: 90.816ms

// 1000 documents
Small collection: delete + create collection: 139.613ms
Small collection: delete documents: 79.563ms

Truncate 10 collections in parallel: documents: 64.560ms
Truncate 10 collections in parallel: collection: 1263.654ms
```

### DIFF
--- a/doc/2/api/controllers/collection/truncate/index.md
+++ b/doc/2/api/controllers/collection/truncate/index.md
@@ -19,7 +19,7 @@ Documents removed that way do not trigger real-time notifications.
 ### HTTP
 
 ```http
-URL: http://kuzzle:7512/<index>/<collection>/_truncate
+URL: http://kuzzle:7512/<index>/<collection>/_truncate[?strategy=<documents|collection>]
 Method: DELETE
 ```
 
@@ -30,7 +30,10 @@ Method: DELETE
   "index": "<index>",
   "collection": "<collection>",
   "controller": "collection",
-  "action": "truncate"
+  "action": "truncate",
+
+  // optional
+  "strategy": "<documents|collection>"
 }
 ```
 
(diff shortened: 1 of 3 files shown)


xbill82 (approved):

alexandrebouthinon (commented):
Ok for me. But have you tested the solution of deleting and recreating the index? Maybe it's even faster

Yoann-Abbes (approved):

Aschen:
> Ok for me. But have you tested the solution of deleting and recreating the index? Maybe it's even faster

That's the whole purpose of this PR :sweat_smile: Before we were deleting and then creating the ES index but it was slower than using `deleteByQuery`

alexandrebouthinon:
> > Ok for me. But have you tested the solution of deleting and recreating the index? Maybe it's even faster
> 
> That's the whole purpose of this PR :sweat_smile: Before we were deleting and then creating the ES index but it was slower than using `deleteByQuery`

Oh! Was thinking there was a dedicated Elasticsearch API route to truncate index 😅

alexandrebouthinon (approved):

PR #2108, opened 2021-07-19
Allows to manually enable or disable cookie authentication

Aschen:
## What does this PR do ?

Cookie authentication was enabled by default but automatically disabled if `accessControlAllowOrigin` contained a wildcard.

This PR introduce a new configuration option that allows to manually enable or disable cookie authentication.

PR #2111, opened 2021-07-23
Enhance CORS origin verification

Shiranuit:
:warning: Depends on
* #2108 

## What does this PR do ?

The goal of this PR is to enhance the verification of a request's origin by adding the possibility of using Regular Expressions instead of just plain string comparison.

To do that a new option has been added to the kuzzlerc `http.accessControlAllowOriginUseRegExp`, this option when set to true converts every strings in `http.accessControlAllowOrigin` into Regular Expressions used to verifiy if a request origin should be allowed based on it's origin.

This PR also adds a new pipe `request:beforeExecution` that can be used to modify request / prevent it's execution.
This pipe can be used to verify CORS for example by throwing in the pipe.

### RegExp Performance is not a big deal here

It is true that Regular Expressions can slowdown applications when not used properly or when used often in a hot / critical code path like it's the case in the funnel, however in our case RegExp are compiled at the start of Kuzzle, and only used to verify if a request containing the `origin` header can be authorized.

The `origin` header is sent by Browsers in requests to prevent Cross Origin Resource Sharing (CORS), which means that every requests sent by our SDK don't go through this code path and are not verified by a RegExp (except the Javascript SDK & Dart SDK when used in a browser).

On my machine I tested `10 000 urls` against some RegExp in about 10ms,
which means that I can verify up to `1 000 000 urls / sec`.

### Regular Expression Dangers

If bad Regular Expression are used this could increase the time complexity exponentially leading to a Catastrophic Backtracking slowing down Kuzzle and causing what we call a REDOS (Regular Expression Denial Of Service) when used improperly.
That's why Regular Expression must be used with care and by people that know that they are doing.

REDOS can be defeated using a non backtracking regex engine like [RE2](https://github.com/google/re2) (see [node-re2](https://www.npmjs.com/package/re2)).
RE2 goal is to guarantee a linear time performance for RegExp evaluation, and protect applications from untrusted RegExp.
But I dont think it is necessary since people who are allowed to modify the config and write RegExp in the config are trusted people that know what they are doing in principle.

### Test
This script generate a bunch of random urls and tests them against multiple regex.
```js
const crypto = require('crypto');

const patterns = [
    '(https?://)?localhost/?',
    '(https?://)?next\-console\.kuzzle\.io/?',
    '(https?://)?console\.kuzzle\.io/?',
    '(https?://)?app\.somename\.kuzzle\.io/?',
    '(https?://)?api\.somename\.kuzzle\.io/?',
    '(https?://)?api\.kuzzle\.io/?',
    '(https?://)?api\.io/?',
    '(https?://)?api\.some\.name\.kuzzle\.io/?',
];

const regexPatterns = patterns.map(pattern => new RegExp(pattern));

function genRandomUrls(count) {
    let urls = [];

    const urlPlaceholders = [
        'http://$NAME',
        'https://$NAME',
        '$NAME',
    ];

    const namePlaceholders = [
        'next-console.kuzzle.io',
        'app.somename.kuzzle.io',
        'api.somename.kuzzle.io',
        'localhost',
        'console.kuzzle.io',
    ];

    for (let i = 0; i < count; i++) {
        const randomName = namePlaceholders[crypto.randomInt(namePlaceholders.length)];

        const splitName = randomName.split('.');

        if (Math.random() < 0.5) {
            const name = crypto.randomBytes(crypto.randomInt(15) + 1).toString('hex');
            splitName[crypto.randomInt(splitName.length)] = name;            
        }

        const randomUrl = urlPlaceholders[crypto.randomInt(urlPlaceholders.length)];
        const url = randomUrl
            .replace('$NAME', splitName.join('.'));

        urls.push(url);
    }

    return urls;
}

function verifyUrl(url) {
    for (const re of regexPatterns) {
        if (re.test(url)) {
            return true;
        }
    }
    return false;
}

function test(urls) {
    let count = 0;
    for (const url of urls) {
        count += verifyUrl(url);
    }
    return count;
}

const urls = genRandomUrls(10000);

const start = Date.now();
const validUrlCount = test(urls);
const stop = Date.now();
const elapsedTime = stop-start;

console.log(`Total valid urls: ${validUrlCount}`);
console.log(`Tested urls: ${urls.length}`);
console.log(`Elapsed time: ${elapsedTime}ms`);
console.log(`${Math.floor(urls.length / (elapsedTime / 1000))} url / s`);
```


xbill82 (approved):

lib/api/funnel.js
```diff
@@ -827,6 +825,30 @@ class Funnel {
 
     return error;
   }
+
+  /**
+   * Verifies if requests sent from a specific origin are allowed
+   * @param {string} origin 
+   * @returns 
+   */
+  _isOriginAuthorized(origin) {
```
Aschen, line 834:
```suggestion
  _isOriginAuthorized (origin) {
```

lib/core/network/entryPoint.js
```diff
@@ -245,19 +245,30 @@ class EntryPoint {
...
-        _res.setError(error);
-      }
-
-      this.logAccess(connection, _res);
-
-      const response = _res.response.toJSON();
-
-      cb(removeErrorStack(response));
-    });
+    global.kuzzle.pipe('request:beforeExecution', request)
```
Aschen, line 248: This should be after the funnel throttle mechanism

Aschen (changes requested):

lib/api/funnel.js
```diff
@@ -332,6 +326,47 @@ class Funnel {
...
+  /**
+   * 1/ checking if the requesting user has the right credentials
+   * 2/ send the request itself to the corresponding controller+action
+   * 
+   * @param {Request} request
+   * @param {Function} callback
+   * @returns 
+   */
+  async _preprocessRequest(request, callback) {
+    global.kuzzle.asyncStore.set('REQUEST', request);
```
Aschen, line 338: This is set twice, here and line 292 before the triggering of the `request:beforeExecution` event

lib/api/funnel.js
```diff
@@ -332,6 +326,47 @@ class Funnel {
...
 
+  /**
+   * 1/ checking if the requesting user has the right credentials
+   * 2/ send the request itself to the corresponding controller+action
+   * 
+   * @param {Request} request
+   * @param {Function} callback
+   * @returns 
+   */
+  async _preprocessRequest(request, callback) {
```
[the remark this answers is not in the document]
Shiranuit: I did that to split a bit the logic to improve the readability because the promise chaining was big

PR #2116, opened 2021-07-27
Fix createOrReplace realtime notification not working

Shiranuit:
## What does this PR do ?

This PR fixes value `created` return by elasticsearch createOrReplace / _mExecute that was using an undefined value instead of doing a string comparison on `body.result === 'created'`


Aschen (approved):
Can you add an unit test to this to ensure it's not going to happen again ?

Yoann-Abbes (approved):

PR #2117, opened 2021-07-30
Use aliases instead of indices by default in ElasticSearch

### LINKED ISSUE #2065: Use ES alias by default
## Feature Description

By default we should always use indices through an alias to facilitate maintenance

MathieuVeber:
Closes #2065

## What does this PR do ?

Change default behavior with ElasticSearch indice. An alias is associated with each indice following the same pattern index + collection. From now, Kuzzle only use aliases instead of indices for each request allowing multiple use cases on indices (such as shrinking or renaming) and facilitating overall maintenance.

- Remove reference to aliases anywhere else than in elasticsearch file (indices are implicitely aliases and that must be invisible)
- Refacto `ES index` reference to `indice` for a better understanding and to be consistent with the doc
- ElasticSearch file uses almost always aliases and has new helpers (`_getAlias`, `_getIndice`, `_getAvailableIndice`, `_getAliasFromIndice`) to facilitate their usage in different use cases well described.
- New error to preserve alias/index unique relationship
- Main concepts doc updated
- New name-generator mock
- Boyscout: Refacto some functions of `lib/service/storage/elasticsearch` with `async`/`await` instead of `.then`/`.catch`

### ElasticSearch indice/alias concerns

Using only aliases instead of indices which can evolve independently require some precautions.

| Concern | Why? | Solution |
| --------------------------- | ------------ | ------------ |
| Using a name as an alias that also refers to a different indice name. | For each ES request, referring to an alias or an indice is equivalent. We have to use the same `index` field. | Add an `ALIAS_PREFIX` (currently @) to always differentiate an alias name from an indice name. |
| The `ALIAS_PREFIX` can be confused with an index or collection char. | The `ALIAS_PREFIX` can not be added to the `FORBIDDEN_CHARS` without making a breaking change | `INDEX_PREFIX` comes between `ALIAS_PREFIX` and `index.collection`. Since the `INDEX_PREFIX` is already forbidden and always there, the `ALIAS_PREFIX` can be distinguished from a char contained in an index or collection thanks to its position. |
| Unable to create an index because the indice name is already taken even though the alias name is not. | Aliases and indices should evolve independently. If an alias has changed but the indice did not, we could theoretically create a new independant indice with the "old" alias name but when removing the prefix to have the "raw" index name, we would be falling onto the existing index.| Instead of just removing the alias prefix, we use a new method `_getAvailableIndice` which is in charge of checking for this situation and generating a unique suffix if necessary.  |
| Actual Kuzzle users do not have aliases. | This breaking change would lead to multiple errors because Kuzzle would not consider their actual indexes/collections | Add a new method `_ensureAliasConsistency` which runs when Kuzzle starts. This function is in charge of checking for this situation and adding, if necessary, missing aliases based on their existing indices. |

### DIFF
--- a/lib/service/storage/elasticsearch.js
+++ b/lib/service/storage/elasticsearch.js
@@ -2783,7 +2783,7 @@ class ElasticSearch extends Service {
         index[INDEX_PREFIX_POSITON_IN_INDEX] === this._indexPrefix
         && ! aliases.some((alias) => alias.name === index));
 
-      let esRequest = { actions : [] };
+      const esRequest = { actions : [] };
       for (const index of indexesWithoutAlias) {
         esRequest.actions.push({ add : { alias: `@${index}`, index } });
       }
@@ -2819,7 +2819,7 @@ class ElasticSearch extends Service {
    * @returns {String} index name
    */
   _extractIndex (esIndex) {
-    let prefixPosition = esIndex.search(ALIAS_PREFIX) === -1
+    const prefixPosition = esIndex.search(ALIAS_PREFIX) === -1
       ? INDEX_PREFIX_POSITON_IN_INDEX
       : INDEX_PREFIX_POSITON_IN_ALIAS;
 
--- a/lib/service/storage/elasticsearch.js
+++ b/lib/service/storage/elasticsearch.js
@@ -1235,10 +1235,10 @@ class ElasticSearch extends Service {
     }
 
     const esRequest = {
-      aliases: {
-        [this._getESIndex(index, collection)]: {}
-      },
       body: {
+        aliases: {
+          [this._getESIndex(index, collection)]: {}
+        },
         mappings: {},
         settings
       },
@@ -1289,7 +1289,7 @@ class ElasticSearch extends Service {
    */
   async getMapping (index, collection, { includeKuzzleMeta=false } = {}) {
     const
-      esIndex = this._getESIndex(index, collection),
+      esIndex = await this._getESIndexFromAlias(index, collection),
       esRequest = {
         index: esIndex
       };
@@ -1325,7 +1325,7 @@ class ElasticSearch extends Service {
    */
   async updateCollection (index, collection, { mappings={}, settings={} } = {}) {
     const esRequest = {
-      index: this._getESIndex(index, collection)
+      index: await this._getESIndexFromAlias(index, collection)
     };
 
     // If either the putMappings or the putSettings operation fail, we need to
@@ -1501,10 +1501,12 @@ class ElasticSearch extends Service {
 
       await this._client.indices.create({
         ...esRequest,
-        aliases: {
-          [this._getESIndex(index, collection)]: {}
-        },
-        body: { mappings },
+        body: {
+          aliases: {
+            [this._getESIndex(index, collection)]: {}
+          },
+          mappings
+        }
       });
 
       return null;
@@ -2739,9 +2741,7 @@ class ElasticSearch extends Service {
    * @returns {boolean} `true` if the raw ES index name is available, `false` otherwise
    */
   async _isESIndexAvailable (esIndex) {
-    const { body } = await this._client.indices.get({ index: esIndex });
-
-    return body.length === 0;
+    return ! await this._client.indices.exists({ index: esIndex }).body;
   }
 
   /**
@@ -2783,12 +2783,14 @@ class ElasticSearch extends Service {
         index[INDEX_PREFIX_POSITON_IN_INDEX] === this._indexPrefix
         && ! aliases.some((alias) => alias.name === index));
 
-      const esRequest = { actions : [] };
+      const esRequest = { body:{ actions : [] } };
       for (const index of indexesWithoutAlias) {
-        esRequest.actions.push({ add : { alias: `@${index}`, index } });
+        esRequest.body.actions.push({ add : { alias: `@${index}`, index } });
       }
 
-      await this._client.indices.updateAliases(esRequest);
+      if (esRequest.body.actions.length > 0) {
+        await this._client.indices.updateAliases(esRequest);
+      }
     }
     catch (error) {
       throw this._esWrapper.formatESError(error);
@@ -2883,10 +2885,10 @@ class ElasticSearch extends Service {
 
     try {
       await this._client.indices.create({
-        aliases: {
-          [this._getESIndex(index, HIDDEN_COLLECTION)]: {}
-        },
         body: {
+          aliases: {
+            [this._getESIndex(index, HIDDEN_COLLECTION)]: {}
+          },
           settings: {
             number_of_replicas: 1,
             number_of_shards: 1,
--- a/lib/service/storage/elasticsearch.js
+++ b/lib/service/storage/elasticsearch.js
@@ -39,6 +39,7 @@ const { isPlainObject } = require('../../util/safeObject');
 const scopeEnum = require('../../core/storage/storeScopeEnum');
 const extractFields = require('../../util/extractFields');
 const { Mutex } = require('../../util/mutex');
+const { generateRandomCustom } = require('../../util/name-generator');
 
 const SCROLL_CACHE_PREFIX = '_docscroll_';
 
@@ -49,6 +50,8 @@ const CHILD_MAPPING_PROPERTIES = ['type'];
 const HIDDEN_COLLECTION = '_kuzzle_keep';
 const PRIVATE_PREFIX = '%';
 const PUBLIC_PREFIX = '&';
+const INDEX_PREFIX_POS = 1; // For index alias name (O for raw index name)
+const ALIAS_PREFIX = '@'; // @todo next major release: Add ALIAS_PREFIX in FORBIDDEN_CHARS
 const NAME_SEPARATOR = '.';
 const FORBIDDEN_CHARS = `\\/*?"<>| \t\r\n,+#:${NAME_SEPARATOR}${PUBLIC_PREFIX}${PRIVATE_PREFIX}`;
 const DYNAMIC_PROPERTY_VALUES = ['true', 'false', 'strict'];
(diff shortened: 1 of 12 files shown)


lib/service/storage/elasticsearch.js
```diff
@@ -49,6 +50,9 @@ const CHILD_MAPPING_PROPERTIES = ['type'];
 const HIDDEN_COLLECTION = '_kuzzle_keep';
 const PRIVATE_PREFIX = '%';
 const PUBLIC_PREFIX = '&';
+const INDEX_PREFIX_POSITON_IN_INDEX = 0;
+const INDEX_PREFIX_POSITON_IN_ALIAS = 1;
+const ALIAS_PREFIX = '@'; // @todo next major release: Add ALIAS_PREFIX in FORBIDDEN_CHARS
```
xbill82, line 55: This deserves an official announcement, doesn't it?

lib/service/storage/elasticsearch.js
```diff
@@ -2620,15 +2677,124 @@ class ElasticSearch extends Service {
   }
 
   /**
-   * Gets the name of an esIndex for an index + collection
+   * Given index + collection, returns the well formated esIndex alias name
```
xbill82, line 2680:
```suggestion
   * Given index + collection, returns the well formatted esIndex alias name
```

xbill82 (approved):

lib/service/storage/elasticsearch.js
```diff
@@ -431,13 +435,18 @@ class ElasticSearch extends Service {
...
+      return {
         _id: body._id,
         _source: body._source,
         _version: body._version
-      }))
-      .catch(error => this._esWrapper.reject(error));
+      };
+    }
+    catch (error) {
+      return this._esWrapper.reject(error);
```
Aschen, line 448: You should rather use `throw this._esWrapper.formatESError(error);` to stay consistent with the rest of the codebase.

Also since we are using `æsync/await`, it's better to use `try/catch` rather than resolved or rejected promises.

lib/service/storage/elasticsearch.js
```diff
@@ -1627,13 +1676,13 @@ class ElasticSearch extends Service {
...
     try {
-      ({ body } = await this._client.cat.indices({ format: 'json' }));
+      ({ body } = await this._client.cat.aliases({ format: 'json' }));
     }
     catch (error) {
       throw this._esWrapper.formatESError(error);
     }
 
-    const esIndexes = body.map(({ index: esIndex }) => esIndex);
+    const esIndexes = body.map(({ alias: esIndex }) => esIndex);
```
Aschen, line 1685: Nit - naming: Are we listing indices (esIndexes) or aliases ?

lib/service/storage/elasticsearch.js
```diff
@@ -1649,13 +1698,13 @@ class ElasticSearch extends Service {
...
     try {
-      ({ body } = await this._client.cat.indices({ format: 'json' }));
+      ({ body } = await this._client.cat.aliases({ format: 'json' }));
     }
     catch (error) {
       throw this._esWrapper.formatESError(error);
     }
 
-    const esIndexes = body.map(({ index }) => index);
+    const esIndexes = body.map(({ alias: index }) => index);
```
Aschen, line 1707: Nit - naming: Are we listing indices (esIndexes) or aliases ?

lib/service/storage/elasticsearch.js
```diff
@@ -1668,16 +1717,19 @@ class ElasticSearch extends Service {
...
```
Aschen, line 1732: Same comment about naming

lib/service/storage/elasticsearch.js
```diff
@@ -235,11 +239,11 @@ class ElasticSearch extends Service {
...
     for (const [esIndex, esIndexInfo] of Object.entries(body.indices)) {
-      const [ indexName, collectionName ] = esIndex
-        .substr(1, esIndex.length)
-        .split(NAME_SEPARATOR);
+      const alias = await this._getAliasFromESIndex(esIndex);
+      const indexName = this._extractIndex(alias);
+      const collectionName = this._extractCollection(alias);
 
-      if ( esIndex[0] !== this._indexPrefix || collectionName === HIDDEN_COLLECTION ) {
+      if ( alias[INDEX_PREFIX_POSITON_IN_ALIAS] !== this._indexPrefix || collectionName === HIDDEN_COLLECTION ) {
```
Aschen, line 246: Nit: This line should be folded

lib/service/storage/elasticsearch.js
```diff
@@ -2620,15 +2677,124 @@ class ElasticSearch extends Service {
...
+
+  /**
+   * Gets the concrete esIndex name from an alias (index + collection)
+   *
+   * @param {String} index
+   * @param {String} collection Default to `HIDDEN_COLLECTION`
+   *
+   * @returns {String} Available esIndex name (eg: '&nepali.liia')
+   */
+  async _getSafeESIndex (index, collection = HIDDEN_COLLECTION) {
```
Aschen, line 2755: I don't get it why you put a default param here. IMHO it's more easier to understand if this method is always called with the same argument, a user collection or the hidden collection

lib/service/storage/elasticsearch.js
```diff
@@ -2620,15 +2677,124 @@ class ElasticSearch extends Service {
...
+    let esIndex = this._getESIndex(index, collection).substr(INDEX_PREFIX_POSITON_IN_ALIAS);
+    let available = await this._isESIndexAvailable(esIndex);
+
+    while (! available) {
+      esIndex = generateRandomCustom(
+        (adjective, name) => `${this._indexPrefix}${adjective}${NAME_SEPARATOR}${name}`);
```
Aschen, line 2761: IMHO it's better to try to keep the indice name as close as possible as the one asked by suffixing a random number or word

lib/service/storage/elasticsearch.js
```diff
@@ -2620,15 +2677,124 @@ class ElasticSearch extends Service {
...
+   * @param {String} index
+   * @param {String} collection Default to `HIDDEN_COLLECTION`
+   *
+   * @returns {String} Available esIndex name (eg: '&nepali.liia')
+   */
+  async _getSafeESIndex (index, collection = HIDDEN_COLLECTION) {
+    let esIndex = this._getESIndex(index, collection).substr(INDEX_PREFIX_POSITON_IN_ALIAS);
+    let available = await this._isESIndexAvailable(esIndex);
+
+    while (! available) {
```
Aschen, line 2759: You could refactor this with a `do...while` loop

lib/service/storage/elasticsearch.js
```diff
@@ -2620,15 +2677,124 @@ class ElasticSearch extends Service {
...
+   */
+  async _ensureAliasConsistency () {
+    try {
+      const { body } = await this._client.cat.indices({ format: 'json' });
+      const esIndexes = body.map(({ index }) => index);
+      const aliases = await this.listAliases();
+
+      const indexesWithoutAlias = esIndexes.filter((index) =>
+        index[INDEX_PREFIX_POSITON_IN_INDEX] === this._indexPrefix
+        && ! aliases.some((alias) => alias.name === index));
```
Aschen, line 2784: Nit: coding standard
```suggestion
      const indexesWithoutAlias = esIndexes.filter(index =>
              index[INDEX_PREFIX_POSITON_IN_INDEX] === this._indexPrefix
        && ! aliases.some(alias => alias.name === index));
```

lib/service/storage/elasticsearch.js
```diff
@@ -2620,15 +2677,124 @@ class ElasticSearch extends Service {
...
+
+      const esRequest = { body:{ actions : [] } };
+      for (const index of indexesWithoutAlias) {
+        esRequest.body.actions.push({ add : { alias: `@${index}`, index } });
```
Aschen, line 2788:
```suggestion
        esRequest.body.actions.push({ add : { alias: `${ALIAS_PREFIX}${index}`, index } });
```

Aschen (changes requested):
Can we rename all references to `ESIndex` to `Indice` ? So the codebase use the same domain as the documentation:
 - index: Kuzzle index
 - collection: Kuzzle collection
 - indice: ES index
 
 Also, this page should be updated accordingly https://docs.kuzzle.io/core/2/guides/main-concepts/data-storage/#internal-representation

lib/service/storage/elasticsearch.js
MathieuVeber: No, I thought about it and that's why I've chosen to have the `index_prefix` coming between the `alias_prefix` and `index.collection`. Since the `index_prefix` is already forbidden and always there, we should not confused an `@` contained in and index or collection and the proper `alias_prefix` thanks to its position.
But still, I will check if I missed something and add it to the concerns array above

lib/service/storage/elasticsearch.js
MathieuVeber: I thought about that but I was afraid to exceed the maximum size which would have led to another case to handle.
What you're suggesting seems better to me but what would you do in the case I've mentioned in order to keep this logic?

MathieuVeber:
> Can we rename all references to `ESIndex` to `Indice` ? So the codebase use the same domain as the documentation:
> 
>     * index: Kuzzle index
> 
>     * collection: Kuzzle collection
> 
>     * indice: ES index
> 
> 
> Also, this page should be updated accordingly https://docs.kuzzle.io/core/2/guides/main-concepts/data-storage/#internal-representation

Yes the actual naming is really messy, I will clean that up following this naming system, it should be easier for everyone to understand what is going on when looking at elastic file and others.

lib/service/storage/elasticsearch.js
Aschen: You could try to keep the original name and add a random number. If it exceed the size you can truncate the end of the collection name for example

Aschen (approved):

PR #2124, opened 2021-08-13
Send tokens to every user with persistent connection

Shiranuit:
## What does this PR do ?

This PR changes the way we handle tokens used with persistent connection.
Connections can only be associated to one token, but tokens can now be associated to multiple connections, this way when a Token expires every connection using the token will be notified that the token as expired.

Prior to this feature token and connection were as linked or unlinked when connection were subscribing / unsubscribing to a room using the realtimeController.
Now connection and tokens are directly linked in the funnel when processing request that contains a Token, or when calling `auth:login`, and connection and tokens are unlinked on `auth:logout`

I added the option `notifyTokenExpired` that work per protocol and enable the sending of token expired notifications.

### DIFF
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -182,6 +182,13 @@ class AuthController extends NativeController {
       this.assertIsAuthenticated(request);
     }
 
+    if (request.context.connection.protocol === 'websocket') {
+      // Unlink connection so the connection will not be notified when the token expires.
+      global.kuzzle.tokenManager.unlink(
+        request.context.token,
+        request.context.connection.id);
+    }
+
     if (request.context.user._id !== this.anonymousId) {
 
       if (request.getBoolean('global')) {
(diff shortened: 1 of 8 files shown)


lib/api/controllers/authController.js
```diff
@@ -182,6 +182,13 @@ class AuthController extends NativeController {
       this.assertIsAuthenticated(request);
     }
 
+    if (global.kuzzle.config.internal.notifiableProtocols.includes(request.context.connection.protocol)) {
+      // Unlink connection so the connection will not be notified when the token expires.
+      global.kuzzle.tokenManager.unlink(
```
xbill82, line 187: Why do we need to prevent this notification if the token is revoked on logout?

xbill82 (approved):

alexandrebouthinon (approved):

jenow (approved):

PR #2130, opened 2021-08-26
Support ES inner_hits keyword

MathieuVeber:
## What does this PR do ?

In this [PR](https://github.com/kuzzleio/kuzzle/pull/1957), Kuzzle allowed `inner_hits` keyword for search queries. However, this wasn't the good thing to do since `inner_hits` is not a root keyword and the issue was with the response format. This PR fix, once and for all, this issue allowing durably the use of [inner_hits API](https://www.elastic.co/guide/en/elasticsearch/reference/current/inner-hits.html) with Kuzzle.

 - Clone and run kuzzle
 - Create index + collection with these mappings `{ "comments": { "type": "nested" } }` and `dynamic = true`
 - Add some nested documents:
```
{
  "title": "Test title",
  "comments": [
    {
      "author": "kimchy",
      "number": 1
    },
    {
      "author": "nik9000",
      "number": 2
    }
  ]
}
```
 - Use `document:search` with `inner_hits` keyword:
```
{
  "query": {
    "nested": {
      "path": "comments",
      "query": {
        "match": { "comments.number": 2 }
      },
      "inner_hits": {} 
    }
  }
}
```
 - You should see the result containing `inner_hits` as requested

### DIFF
--- a/lib/service/storage/elasticsearch.js
+++ b/lib/service/storage/elasticsearch.js
@@ -116,7 +116,6 @@ class ElasticSearch extends Service {
       'explain',
       'from',
       'highlight',
-      'inner_hits',
       'query',
       'search_after',
       'search_timeout',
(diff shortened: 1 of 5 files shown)


lib/service/storage/elasticsearch.js
```diff
@@ -395,11 +394,30 @@ class ElasticSearch extends Service {
...
+        _id: hit._id,
+        _score: hit._score,
+        _source: hit._source,
+        highlight: hit.highlight,
+      };
+    }
+
+    function formatInnerHits (innerHits) {
+      if (! innerHits) {
+        return undefined;
```
Aschen, line 408:
```suggestion
        return;
```

Aschen (approved):

Shiranuit (approved):

PR #2141, opened 2021-09-17
Refresh ID Card inside Worker Thread to not be impacted by the load on the main Event Loop 

Shiranuit:
⚠️ Depends on #2145 

## What does this PR do ?

This PR exports the ID Card renewal mechanism of the IDCardHandler outside to be used inside a Worker Thread instead of the same Event Loop where Kuzzle is running.

### Why ?
Since a setInterval was used prior to this PR to refresh the ID Card periodically inside Redis, when the Kuzzle event loop is saturated because of the load, the refresh of the ID Cards takes too much time and the node will kill itself because it could not refresh his ID Card in time.

Now this mechanism is done in his own Worker Thread that has his own Event Loop.

`npm run  test`


lib/cluster/workers/IDCardRenewer.js
```diff
@@ -0,0 +1,90 @@
...
+    await this.initRedis(config.redis);
+
+    /**
+     * Early renew, otherwise we need to wait the refreshDelay before the first ID Card renewal
+     * which could be enough to make the ID Card expire.
+     */
+    try {
+      await this.renewIDCard(); 
+    } catch (error) {
+      // Do nothing
```
alexandrebouthinon, line 34: At least log something about the Redis error that could happend in the `renewCard` function

alexandrebouthinon (changes requested):

xbill82 (approved):
👏

alexandrebouthinon (approved):

lib/cluster/idCardHandler.js
```diff
@@ -99,28 +100,42 @@ class ClusterIdCardHandler {
...
-          // Unable to refresh the key in time before it expires
-          // => this node is too slow, we need to remove it from the cluster
-          if (refreshed === 0) {
-            await this.node.evictSelf('Node too slow: ID card expired');
-          }
-        }
+    this.refreshWorker = await this._constructWorker(`${__dirname}/workers/IDCardRenewer.js`);
+    this.refreshWorker.unref();
+
+    this.refreshWorker.on('message', async (message) => {
```
Aschen, line 106: Nit
```suggestion
    this.refreshWorker.on('message', async message => {
```

lib/cluster/idCardHandler.js
```diff
@@ -99,28 +100,42 @@ class ClusterIdCardHandler {
...
+        config: global.kuzzle.config.services.internalCache,
+        name: 'internal_adapter',
       },
-      this.refreshDelay);
+      refreshDelay: this.refreshDelay,
+    });
+  }
+
+  // Used to Mock the creation of a worker for the tests
+  async _constructWorker(path) {
```
Aschen, line 130: Why this is async?

lib/cluster/workers/IDCardRenewer.js
```diff
@@ -0,0 +1,105 @@
...
+    this.refreshDelay = config.refreshDelay;
+    
+    /**
+    * Since we do not have access to the Kuzzle Context we can't use kuzzle.ask('core:cache:internal:*',...),
+    * so we need to have an instance of Redis similar to the one used in the Cache Engine
+    */
+    try {
+      await this.initRedis(config.redis);
+    } catch (error) {
+      this.parentPort.postMessage({
```
Aschen, line 28: nit
```suggestion
    } 
    catch (error) {
      this.parentPort.postMessage({
```

lib/cluster/workers/IDCardRenewer.js
```diff
@@ -0,0 +1,105 @@
...
+          this.nodeIdKey,
+          this.refreshDelay * 1.5);
+        // Unable to refresh the key in time before it expires
+        // => this node is too slow, we need to remove it from the cluster
+        if (refreshed === 0) {
+          await this.dispose();
+          this.parentPort.postMessage({ error: 'Node too slow: ID card expired' });
+        }
+      }
+    } catch (error) {
```
Aschen, line 62: Wrong catch indentation

lib/cluster/workers/IDCardRenewer.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  async dispose() {
+    if (!this.disposed) {
+      this.disposed = true;
+      clearInterval(this.refreshTimer);
+      this.refreshTimer = null;
+      try {
+        await this.redis.commands.del(this.nodeIdKey);
+      } catch (error) {
```
Aschen, line 77: Wrong catch indentation

lib/cluster/workers/IDCardRenewer.js
```diff
@@ -0,0 +1,105 @@
...
+      global.kuzzle.log = {
+        debug: console.debug, // eslint-disable-line no-console
+        error: console.error, // eslint-disable-line no-console
+        info: console.info, // eslint-disable-line no-console
+        warn: console.warn, // eslint-disable-line no-console
+      };
+      
+      idCardRenewer.init(message);
+    } else if (message.action === 'dispose') {
+      idCardRenewer.dispose();
```
Aschen, line 100: I know it's not very useful but you should still attach a `catch` handler to the returned promise

Aschen (changes requested):

lib/cluster/idCardHandler.js
Shiranuit: No reason 😅

lib/cluster/workers/IDCardRenewer.js
```diff
@@ -0,0 +1,111 @@
...
+      });
+      return;
+    }
+
+    /**
+     * Early renew, otherwise we need to wait the refreshDelay before the first ID Card renewal
+     * which could be enough to make the ID Card expire.
+     */
+    await this.renewIDCard();
+    if (!this.disposeed) {
```
Aschen, line 40: Typo
```suggestion
    if (!this.disposed) {
```

Aschen (changes requested):

lib/cluster/workers/IDCardRenewer.js
Shiranuit: good catch 😅

lib/cluster/workers/IDCardRenewer.ts
```diff
@@ -1,19 +1,24 @@
-'use strict';
-const { isMainThread, parentPort } = require('worker_threads');
-const Redis = require('../../service/cache/redis');
+import { isMainThread, parentPort } from 'worker_threads';
+import Redis from '../../service/cache/redis';
+import { JSONObject } from 'kuzzle-sdk';
```
Aschen, line 3: Nit: import order
```suggestion
import { isMainThread, parentPort } from 'worker_threads';

import { JSONObject } from 'kuzzle-sdk';

import Redis from '../../service/cache/redis';
```

Aschen (approved):

Shiranuit:
Waiting for [Wait index cache propagation before allowing node to process requests](#2145) to be ready before merging this PR

lib/cluster/workers/IDCardRenewer.ts
Aschen: (It's still not the good order)

lib/cluster/workers/IDCardRenewer.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export class IDCardRenewer {
+  private disposed = false;
+  private redis : Redis = null;
+  private refreshTimer : ReturnType<typeof setInterval>;
+  private nodeIdKey : string = null;
+  private refreshDelay = 2000;
+  private parentPort : any = null;
```
Aschen, line 11: Why do you need to keep an instance reference on the global `parentPort` variable?

Aschen (changes requested):

lib/cluster/workers/IDCardRenewer.ts
Shiranuit: Just to be able to mock it in the tests

Aschen (approved):

PR #2145, opened 2021-09-23
Wait index cache propagation before allowing nodes to process requests

Shiranuit:
## What does this PR do ?

This PR forces the Kuzzle Nodes to wait for the index cache propagation to be finished before accepting requests.

### Why ?

Before the different nodes were checking if specific mutexes could be acquired and the node that acquired the mutex will do the import of either `user mappings`, `mappings`, `fixtures` or `permissions`.

If all mutexes were acquired during the boot time of one node, the node will then skip the imports since another node was doing the work to import the data inside the index cache, but the issue was that the node were not waiting for other nodes to have finished importing the data inside the index cache and were accepting requests before everything was initialized.

### Other changes

Convert Kuzzle class to TypeScript

### DIFF
--- a/lib/kuzzle/kuzzle.ts
+++ b/lib/kuzzle/kuzzle.ts
@@ -23,35 +23,37 @@
 
 const path = require('path');
 
-const { murmurHash128: murmur } = require('murmurhash-native');
-const stringify = require('json-stable-stringify');
-const { Koncorde } = require('koncorde');
-const Bluebird = require('bluebird');
-const segfaultHandler = require('segfault-handler');
-const _ = require('lodash');
+import { murmurHash128 as murmur } from 'murmurhash-native';
+import stringify from 'json-stable-stringify';
+import { Koncorde } from 'koncorde';
+import Bluebird from 'bluebird';
+import segfaultHandler from 'segfault-handler';
+import _ from 'lodash';
 
-const kuzzleStateEnum = require('./kuzzleStateEnum');
-const KuzzleEventEmitter = require('./event/kuzzleEventEmitter');
-const EntryPoint = require('../core/network/entryPoint');
-const Funnel = require('../api/funnel');
-const PassportWrapper = require('../core/auth/passportWrapper');
-const PluginsManager = require('../core/plugin/pluginsManager');
-const Router = require('../core/network/router');
-const Statistics = require('../core/statistics');
-const { TokenManager } = require('../core/auth/tokenManager');
-const Validation = require('../core/validation');
-const Logger = require('./log');
-const vault = require('./vault');
-const DumpGenerator = require('./dumpGenerator');
-const AsyncStore = require('../util/asyncStore');
-const { Mutex } = require('../util/mutex');
-const kerror = require('../kerror');
-const InternalIndexHandler = require('./internalIndexHandler');
-const CacheEngine = require('../core/cache/cacheEngine');
-const StorageEngine = require('../core/storage/storageEngine');
-const SecurityModule = require('../core/security');
-const RealtimeModule = require('../core/realtime');
-const Cluster = require('../cluster');
+import kuzzleStateEnum from './kuzzleStateEnum';
+import KuzzleEventEmitter from './event/kuzzleEventEmitter';
+import EntryPoint from '../core/network/entryPoint';
+import Funnel from '../api/funnel';
+import PassportWrapper from '../core/auth/passportWrapper';
+import PluginsManager from '../core/plugin/pluginsManager';
+import Router from '../core/network/router';
+import Statistics from '../core/statistics';
+import { TokenManager } from '../core/auth/tokenManager';
+import Validation from '../core/validation';
+import Logger from './log';
+import vault from './vault';
+import DumpGenerator from './dumpGenerator';
+import AsyncStore from '../util/asyncStore';
+import { Mutex } from '../util/mutex';
+import kerror from '../kerror';
+import InternalIndexHandler from './internalIndexHandler';
+import CacheEngine from '../core/cache/cacheEngine';
+import StorageEngine from '../core/storage/storageEngine';
+import SecurityModule from '../core/security';
+import RealtimeModule from '../core/realtime';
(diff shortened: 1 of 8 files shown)


lib/kuzzle/kuzzle.ts
```diff
@@ -79,7 +82,56 @@ Reflect.defineProperty(global, 'kuzzle', {
...
+  // Network entry point
+  public entryPoint: EntryPoint;
+
+  // Validation core component
+  public validation: Validation;
+  
+  // Dump generator
+  public dumpGenerator: DumpGenerator;
+
+  // Vault component (will be initialized after bootstrap)
```
Aschen, line 114: It's really a nitpicking but are the comments displayed by the IDE when they are with `//` and not  `/* */` ?

lib/kuzzle/kuzzle.ts
```diff
@@ -79,7 +82,56 @@ Reflect.defineProperty(global, 'kuzzle', {
  * @extends EventEmitter
  */
 class Kuzzle extends KuzzleEventEmitter {
-  constructor (config) {
+  public config: JSONObject;
```
Aschen, line 85: Why everything is public? IMHO most of those properties should be private instead

lib/kuzzle/kuzzle.ts
```diff
@@ -252,7 +292,7 @@ class Kuzzle extends KuzzleEventEmitter {
    *
    * @returns {Promise<void>}
    */
-  async install (installations) {
+  async install (installations: Array<InstallationConfig>): Promise<void> {
```
Aschen, line 295: Prefer this notation (more simple)
```suggestion
  async install (installations: InstallationConfig[]): Promise<void> {
```

lib/kuzzle/kuzzle.ts
```diff
@@ -299,6 +339,145 @@ class Kuzzle extends KuzzleEventEmitter {
...
+  async emit (...args: any[]) {
+    return super.emit(...args);
+  }
+
+  // For testing purpose
+  async pipe (...args: any[]) {
+    return super.pipe(...args);
+  }
+
+  private async _importUserMappings(config: {
```
Aschen, line 357: The first `_` is useless with typescript (we used it to indicate a private method)
Apply to other private methods

Aschen (changes requested):

lib/kuzzle/kuzzle.ts
Shiranuit: Ah didn't know that

lib/kuzzle/kuzzle.ts
```diff
@@ -299,6 +371,166 @@ class Kuzzle extends KuzzleEventEmitter {
...
+      await this.internalIndex.refreshCollection('users');
+      this.log.info('[✔] User mappings import successful');
+    }
+  }
+
+  private async importMappings (
+    config: {
+      toImport: ImportConfig,
+      toSupport: SupportConfig
+    }
```
Aschen, line 413: Here you are not using the `ImportStatus` object to avoid re-importing mappings, is it intentional ?

lib/kuzzle/kuzzle.ts
```diff
@@ -307,118 +539,47 @@ class Kuzzle extends KuzzleEventEmitter {
...
-        this.log.info('[✔] User mappings import successful');
-      }
+      for (const [type, importMethod] of Object.entries(this.importTypes)) {
+        const mutex = new Mutex(`backend:import:${type}`, { timeout: 0 });
+        const isInitialized = await this.ask('core:cache:internal:get', `${BACKEND_IMPORT_KEY}:${type}`);
+        const isLocked = await mutex.lock();
+        
+        await importMethod(
+          { toImport, toSupport },
+          {
```
Aschen, line 556: You are passing `isFirstCall`, `isInitialized` and `isLocked` to every import function but you are only using `isFirstCall` to determine whether you should import something or not.

lib/kuzzle/kuzzle.ts
```diff
@@ -307,118 +539,47 @@ class Kuzzle extends KuzzleEventEmitter {
...
     try {
-      if (! _.isEmpty(toImport.userMappings) && await mutexes.userMappings.lock()) {
-        lockedMutexes.push(mutexes.userMappings);
-        await this.internalIndex.updateMapping('users', toImport.userMappings);
-        await this.internalIndex.refreshCollection('users');
-        this.log.info('[✔] User mappings import successful');
-      }
+      for (const [type, importMethod] of Object.entries(this.importTypes)) {
+        const mutex = new Mutex(`backend:import:${type}`, { timeout: 0 });
+        const isInitialized = await this.ask('core:cache:internal:get', `${BACKEND_IMPORT_KEY}:${type}`);
```
Aschen, line 551: Nitpicking: Usually I'm not fond of the `isXXXX` notation for boolean variable. Since they are boolean they already express a state without having to write `is`. This is particularly valid in Typescript since we have strong types.

Aschen (changes requested):

lib/kuzzle/kuzzle.ts
```diff
@@ -526,4 +685,4 @@ class Kuzzle extends KuzzleEventEmitter {
   }
 }
 
-module.exports = Kuzzle;
+module.exports = Kuzzle;
```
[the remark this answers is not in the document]
Shiranuit: This is needed by one of the lib we use to make some tests a.k.a `rewire` and the TS transpilation doesn't export the Kuzzle class the way the lib needs it.

lib/kuzzle/kuzzle.ts
Shiranuit: Oops id did rollback some of my commit by accident, this it not intended

lib/kuzzle/kuzzle.ts
Shiranuit: Oops id did rollback some of my commit by accident, this it not intended

Aschen (changes requested):

Aschen (approved):

PR #2146, opened 2021-09-23
Freeze @elastic/elasticsearch at 7.13.0

alexandrebouthinon:
## What does this PR do ?
Because of a Breaking Change introduced in the version `7.14.0` of  the package `@elastic/elasticsearch`, see the Issue [OSS Distrubtion rejected](https://github.com/elastic/elasticsearch-js/issues/1519), we have to freeze the version of the `@elastic/elasticsearch` package at version `7.13.0`

With that fix you should be able to run Kuzzle on AWS using the managed Elasticsearch service

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle",
-  "version": "2.14.4",
+  "version": "2.14.5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


Shiranuit (approved):

Aschen (changes requested):
The version should be 2.14.5

Aschen (approved):

PR #2147, opened 2021-09-27
Fix API Key being expired when refreshed

Shiranuit:
⚠️ Depends on #2145 
## What does this PR do ?

This PR fixes a bad behaviour when re-login to the same user or refreshing a token that appeared to be an API Key or an infinite duration token and was causing the token / API Key to expires even though they should not.

## To Do
- [x] Disable refreshing of API Keys / Infinite duration token
- [x] Tests


lib/kerror/codes/7-security.json
```diff
@@ -45,6 +45,12 @@
           "code": 7,
           "message": "Error verifying token: %s",
           "class": "InternalError"
+        },
+        "refresh_forbidden": {
+          "description": "The provided token cannot be refreshed.",
+          "code": 8,
+          "message": "Token cannot be refreshed.",
```
Aschen, line 52: We should provide the reason to the user so he can know if he is using an API key or a token with infinite TTL

Aschen (approved):

PR #2148, opened 2021-09-29
Convert full state to typescript

Aschen:
## What does this PR do ?

Convert the `cluster/state.js` file to Typescript

### Other changes

  - dump the full state when a node is shutting down

### DIFF
--- a/lib/cluster/state.ts
+++ b/lib/cluster/state.ts
@@ -45,14 +45,13 @@ class RoomSubscriptions {
    * @constructor
    * 
    * @param messageId ID of the last message that updated this room
-   * @param [subscribers]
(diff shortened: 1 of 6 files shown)


lib/cluster/state.ts
```diff
@@ -218,41 +285,49 @@ class State {
...
+    if (! room) {
       return 0;
     }
 
     return room.countSubscriptions();
   }
 
   listRealtimeRooms () {
-    const list = {};
+    const list: {
```
xbill82, line 324: Why not writing an interface for this? So we could annotate the return type with it?

xbill82 (approved):

PR #2185, opened 2021-11-02
Freeze Elasticsearch client version

Aschen:
## What does this PR do ?

Use the direct link to the latest version of the elasticsearch client supporting Open Search as well

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "kuzzle",
   "author": "The Kuzzle Team <support@kuzzle.io>",
-  "version": "2.14.12",
+  "version": "2.14.13",
   "description": "Kuzzle is an open-source solution that handles all the data management through a secured API, with a large choice of protocols.",
   "bin": {
     "kuzzle": "bin/start-kuzzle-server"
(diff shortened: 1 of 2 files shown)


xbill82 (approved):
Did you mean "freeze"?

package.json
```diff
@@ -39,7 +39,7 @@
     "lib": "lib"
   },
   "dependencies": {
-    "@elastic/elasticsearch": "7.15.0",
+    "@elastic/elasticsearch": "https://github.com/elastic/elasticsearch-js/archive/refs/tags/v7.13.0.tar.gz",
```
alexandrebouthinon, line 42:
```suggestion
    "@elastic/elasticsearch": "7.13.0",
```

alexandrebouthinon (changes requested):

package.json
Aschen: The whole point of this PR is to stop using NPM version because it's upgraded by npm-check-update

alexandrebouthinon (approved):

Aschen:
@xbill82
I mean always use this version because further versions are not compatible with OpenSearch. Until the SDK is forked or another solution we need to stick with this version

Issue #2163, opened 2021-10-22
Add Typescript type to config object


matissepage:
## Feature Description

Write the Typescript type of the Kuzzle config. Must have a subtype per key at the top level (server, plugins, limits etc)

## Example Use Case

[example for subtype](https://github.com/kuzzleio/kuzzle/blob/master/lib/core/plugin/pluginsManager.js#L89)

PR #2188, opened 2021-11-04
Add type for services config

matissepage:
warning Depends on #2163 

**This PR is still in progress**

## What does this PR do ?

Add type for services config 

```npm run test:unit```
or
```npm run test```

### Other changes

Add a config folder for more clarity

### CODE REVIEW

**Review by Aschen — APPROVED** [review]

**Review by alexandrebouthinon — APPROVED** [review]

File: lib/types/index.ts [hunk 1]
```diff
@@ -32,4 +32,5 @@ export * from './Global';
 export * from './config/PasswordPolicy';
 export * from './config/KuzzleConfiguration';
 export * from './config/RoleDefinition';
-export * from './config/ProfileDefinition';
\ No newline at end of file
+export * from './config/ProfileDefinition';
+export * from './config/ServicesConfiguration';
```
alexandrebouthinon, line 36:
```suggestion
export * from './config/ServicesConfiguration';

```

lib/types/config/ServicesConfiguration.ts
```diff
@@ -0,0 +1,298 @@
+import { JSONObject } from '../../../index';
+
+export type ServicesConfiguration = {
```
alexandrebouthinon, line 3: Since there are only 3 external service configurations described under the `services` key and this will be modified to support different tools (Postgres as a possible replacement for Elasticsearch as storageEngine, ...) maybe it is time to prepare the extensibility of this part of the configuration. 

For example by declaring an interface for each of these three keys (storageEngine, internalCache, memoryStorage) and use them to implement future services compatibility. 

It would be enough to make the next specific configurations inherit from these interfaces and thus avoid the problems of key collisions, silent omission of some too specific configuration fields etc... I don't know how interesting the ROI is but I think we should discuss it 🙂

alexandrebouthinon: For example:
```js
export interface IStorageEngineConfiguration { /* ... */ }
export interface IInternalCacheConfiguration { /* ... */ }
export interface IMemoryStorageConfiguration { /* ... */ }
```

Aschen: The types for services should be like this:

```
type StorageService = StorageServiceElasticsearch // in the futur we will have: StorageServiceElasticsearch | StorageServicePostgre

type StorageServiceElasticsearch = {
  backend: 'elasticsearch',
  aliases: string[],
  // etc
}
```

lib/types/config/ServicesConfiguration.ts
```diff
@@ -0,0 +1,298 @@
...
+  
+       /**
+        * The default storage layer is Elasticsearch and it is 
+        * currently the only storage layer we support.
+        */
+        storageEngine: {
+          /**
+           * @default ['storageEngine']
+           */
+          aliases: Array<string>
```
Aschen, line 181: Prefer the more concise notation `string[]`

lib/types/config/ServicesConfiguration.ts
```diff
@@ -0,0 +1,298 @@
...
+                  /**
+                   * @default 
+                   * 
+                   * [
+                   *   {
+                   *     type: 'keyword',
+                   *   }
+                   * ]
+                   */
+                  author: JSONObject
```
Aschen, line 230: This is a string

lib/types/config/ServicesConfiguration.ts
```diff
@@ -0,0 +1,298 @@
...
+           *   - "false": Stores document and does not update the collection
+           *     mapping (field are not indexed)
+           *   - "strict": Rejects document
+           * @see https://www.elastic.co/guide/en/elasticsearch/reference/7.4/dynamic-mapping.html
+           */
+           commonMapping: {
+             /**
+              * @default "false"
+              */
+             dynamic: string
```
Aschen, line 216: This is `"true" | "false" | "strict"`

lib/types/config/ServicesConfiguration.ts
```diff
@@ -0,0 +1,298 @@
...
+                  /**
+                   * @default 
+                   * 
+                   * [
+                   *   {
+                   *     type: 'date',
+                   *   }
+                   * ]
+                   */
+                   createdAt: JSONObject
```
Aschen, line 241: This is a number

lib/types/config/ServicesConfiguration.ts
```diff
@@ -0,0 +1,298 @@
...
+                   /**
+                   * @default 
+                   * 
+                   * [
+                   *   {
+                   *     type: 'keyword',
+                   *   }
+                   * ]
+                   */
+                    updater: JSONObject
```
Aschen, line 252: This is a string

lib/types/config/ServicesConfiguration.ts
```diff
@@ -0,0 +1,298 @@
...
+                    /**
+                   * @default 
+                   * 
+                   * [
+                   *   {
+                   *     type: 'date',
+                   *   }
+                   * ]
+                   */
+                     updatedAt: JSONObject
```
Aschen, line 263: This is a number

lib/types/config/ServicesConfiguration.ts
```diff
@@ -0,0 +1,298 @@
...
+                   /**
+                      * @default 
+                      * 
+                      * [
+                      *   {
+                      *     type: 'keyword',
+                      *   }
+                      * ]
+                      */
+                    profileIds: JSONObject
```
Aschen, line 292: This is a `string[]`

lib/types/config/ServicesConfiguration.ts
```diff
@@ -0,0 +1,298 @@
...
+              * @default "kuzzle"
+              */
+             name: string
+  
+             collections: {
+               users: {
+                 /**
+                  * @default 'false'
+                  */
+                 dynamic: string
```
Aschen, line 280: This is `"true" | "false" | "strict"`

Aschen (changes requested):
As @alexandrebouthinon remarked, the types for services should be like this to allows to have other services later (PostgreSQL, RAM cache, ...) :

```
type StorageService = StorageServiceElasticsearch // in the futur we will have: StorageServiceElasticsearch | StorageServicePostgre

type StorageServiceElasticsearch = {
  backend: 'elasticsearch',
  aliases: string[],
  // etc
}
```

Aschen (changes requested):

Aschen (approved):

PR #2192, opened 2021-11-08
Improve local dev stack using Docker Compose and Traefik

alexandrebouthinon:
## What does this PR do ?
Improve Dev stack using natives Docker Compose features, Traefik and a dedicated container for node_modules installation/compilation.
This should make this stack easier to run on MacOS/Windows and improve cluster management using the `--scale` Docker Compose service feature

Example to run a 3 nodes Kuzzle cluster:

```
docker compose up -d --scale kuzzle=3
```

Try to use it and play with Kuzzle nodes count using the `--scale` flag:

```
docker compose up -d --scale kuzzle=3
docker compose up -d --scale kuzzle=1 
docker compose up -d --scale kuzzle=6  
```

### Other changes

Clean `run-dev.sh` putting installation/compilation logic in `kuzzle-installer.sh`


docker-compose.yml
```diff
@@ -1,91 +1,118 @@
...
+      - "/var/run/docker.sock:/var/run/docker.sock:ro"
 
-  kuzzle_node_2:
-    <<: *kuzzle-config
-    container_name: kuzzle_node_2
-    ports:
-      - "17511:7512" # Kuzzle API port
-      - "11883:1883" # Kuzzle MQTT port
-      - "9230:9229" # Debug port
+  kuzzle_installer:
```
Aschen, line 24: I'm not found of this since it will run npm install every time we run Docker-Compose.

This file is meant to be used by core developers anyway so IMHO it's not a problem if it's not ready-to-use but that you have to read the documentation before

Aschen (changes requested):
Depending on the task we want to achieve it could be handy to be able to target specific nodes of the cluster.

Actually there is another configuration file for Docker Compose with Swarm in the `docker/` directory https://github.com/kuzzleio/kuzzle/blob/2-dev/docker/docker-compose-swarm.yml

I would rather keep the actual Docker-Compose at repository root since it runs 3 nodes by default with different ports. Otherwise I agree that using Traeffik would be better than Nginx :+1:

alexandrebouthinon:
I didn't see you commited the file, I've provided you last time. I think the new one is more accomplished and I think to remove https://github.com/kuzzleio/kuzzle/blob/2-dev/docker/docker-compose-swarm.yml . What do you think?

alexandrebouthinon:
About the specific container targeting, this is possible by getting IP address of a specific one using `docker ps` then get the IP address using `docker inspect <container_id>` but we can think about a script/alias to help core devs to get it easily

docker-compose.yml
alexandrebouthinon: Yes it runs `npm install` at each `docker-compose up` command BUT since this is not `npm ci` it take 5 seconds to get that no installation is needed. I think this is not a big deal and permit core team non Linux users to run this stack easily

alexandrebouthinon: In my opinion, we need to separate Core development environments in two types:
- Using this Docker Compose file to install and run all the components
- Using `kourou app:start-services` and local Node.js environment (using Ergol)
As it everyone can contribute easily to the projet

alexandrebouthinon: But we need to stop hybride workflows that led to packaging issue and false positive during tests

alexandrebouthinon: I can amend the CONTRIBUTING.md file to explain how to run Kuzzle with each method

xbill82: As a mac user, I welcome this stack that would make it extremely easy to run a kuzzle cluster from scratch with zero-fuss on non-linux systems.
I agree, as Kuzzle developers we are aware of how it works and are of course able to manually perform all the steps manually, but improving the DX is always a good thing, even for core-dev.

docker-compose.yml
```diff
@@ -1,91 +1,118 @@
...
+      - "traefik.http.routers.kuzzle.rule=Host(`localhost`)"
+      - "traefik.http.routers.kuzzle.service=kuzzle_http"
+      # Kuzzle MQTT
+      - "traefik.tcp.services.kuzzle_mqtt.loadbalancer.server.port=1883"
+      - "traefik.tcp.routers.kuzzle.entrypoints=kuzzle_mqtt"
+      - "traefik.tcp.routers.kuzzle.rule=HostSNI(`*`)"
+      - "traefik.tcp.routers.kuzzle.service=kuzzle_mqtt"
+    environment:
+      - kuzzle_services__storageEngine__client__node=http://elasticsearch:9200
+      - kuzzle_services__storageEngine__commonMapping__dynamic=true
```
xbill82, line 75: In development environments, it is healthier to use `dynamic: 'strict'`, in order to see all the possible errors. On the other side, I understand this is bad for rapid prototyping, so I don't really know...

docker-compose.yml
xbill82: But, at the same time, I'm not sure it's a good thing that the install performs `npm install`. If dependencies have minor/patch updates, they will be installed and the stack will not be iso-release. This goes without mentioning the risk of potential breaking-in-minor changes...
In the end, I like the idea of having an installation step, but I think we should improve it further.

docker/scripts/kuzzle-installer.sh
```diff
@@ -0,0 +1,19 @@
...
+set -e
+
+if [ -z "$NODE_VERSION" ];
+then
+  echo "Missing NODE_VERSION, use default NODE_14_VERSION"
+  n "$NODE_14_VERSION"
+fi
+
+echo "Installing dependencies"
+npm install
```
xbill82, line 12: ☝️ See my comment above.

xbill82 (changes requested):

docker-compose.yml
Aschen: This is not a development environment, this stack is intended for core development only

xbill82: ok then

docker-compose.yml
alexandrebouthinon: Yes I agree with you, the installation and build process needs some fine tuning

rolljee (approved):

Aschen (changes requested):
As we discussed, can you:
  - keep the docker-compose.yml at the root as-is to always have 3 nodes starting with different ports
  - add/update the docker/docker-compose-cluster.yml with your config

alexandrebouthinon:
I'll close this PR to get rid of unwanted commits to simplify reviews

Shiranuit (approved):

xbill82:
Any updates on this? @clementbolin

clementbolin:
> Any updates on this? @clementbolin

I closed this issue, because we have implemented this feature in this [PR](https://github.com/kuzzleio/kuzzle/pull/2253), without the Traefik service.

PR #2193, opened 2021-11-08
Add optional history of cluster state sync messages

Aschen:
## What does this PR do ?

This PR add an optional history mechanism intended to debug the cluster state sync messages.

The optional history can be activated with a new API action `cluster:debug`

When used with the `stateSyncHistory` argument, this action will enable an history of state sync messages on each node. Each node will keep the last X history messages and the full state before the message was processed.

If a node die because of a desync issue, this history will saved into Redis for further analysis.

  - Step 1 :
  - Step 2 :
  - Step 3 :

### Other changes

  - `KuzzleDocument` now use a templated type for `_source`

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -74,3 +74,4 @@ lib/core/realtime/connectionRooms.js
 lib/core/realtime/room.js
 lib/core/realtime/subscription.js
 lib/types/realtime/RoomList.js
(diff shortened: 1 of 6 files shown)


Aschen:
We don't have desync issues, this work should be paused until we have new issues or more time

Issue #2194, opened 2021-11-10
Nodes sharing same ip when reloading Kuzzle too fast


Aschen:
## Description

When the hotreloader reload Kuzzle, this error message can appear

```
 [CLUSTER] Start retrieving full state..
kuzzle_1         | [CLUSTER] 1 remote nodes discovered
kuzzle_1         | [CLUSTER] Another node share the same IP address as this one (172.19.0.5): knode-scintillating-gershwin-22963. Shutting down.
kuzzle_1         | {"authStrategies":[],"rooms":[]}
kuzzle_1         | Initiating shutdown...
kuzzle_1         | Halted.
```

### Possible solution

The previous Node ID Card is still present and thus have the same ip

PR #2205, opened 2021-11-14
Ensure ID Card is refreshed while worker thread starts

Aschen:
## What does this PR do ?

The node ID Card is refreshed by a [Worker Thread](https://nodejs.org/docs/latest-v12.x/api/worker_threads.html) in a different process (and a different event loop) to ensure that the Node is always able to make the necessary request to Redis.

Worker Thread can take a long time to start, so a timer is started from the main process to refresh the ID Card until the Worker Thread starts. 

This allows to reduce the ID Card expiration time and thus fix the error users can get when a node is restarting to fast (see #2194)

### Boyscout

 - convert IdCardHandler to Typescript

### DIFF
--- a/lib/cluster/idCardHandler.ts
+++ b/lib/cluster/idCardHandler.ts
@@ -97,7 +97,7 @@ export class ClusterIdCardHandler {
   /**
    * Local node ID Card
    */
-  private idCard: IdCard;
+  private idCard: IdCard = null;
 
   /**
    * Local node IP address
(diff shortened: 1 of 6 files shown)


lib/cluster/idCardHandler.ts
```diff
@@ -133,15 +195,44 @@ class ClusterIdCardHandler {
...
+   *
+   * Once the worker starts, this timer will be stopped.
+   */
+  private startTemporaryRefresh () {
+    this.refreshTimer = setInterval(async () => {
+      try {
+        await this.save();
+      }
+      catch (error) {
+        global.kuzzle.log.error(`Cannot refresh the ID Card before the worker starts: ${error}`);
```
xbill82, line 221:
```suggestion
        global.kuzzle.log.error(`An error occurred while refreshing the ID card during WorkerThread startup: ${error}`);
```

xbill82 (approved):

Shiranuit (approved):

PR #2210, opened 2021-11-17
Replace segfault handler dependency

Shiranuit:
## What does this PR do ?
- Replace the package `segfault-handler` by the package `node-segfault-handler`
- Update docker images to install `libunwind-dev` instead of `libexecinfo`
- Update workflows to build and run every images instead of only Kuzzle

## Why ?
The node package `segfault-handler` is using the library `libexecinfo` to work, the catch is that the library `libexecinfo` is not working on Linux Alpine meaning that since this package was installed, our Alpine images where broken, so we are migrating to the node package `node-segfault-handler` which uses the library `libunwind` instead which, unlike `libexecinfo`, do work on Linux Alpine.

The package `node-segfault-handler` should provide us about the same native stacktraces as the `segfault-handler` package, and it will also provide us Javascript stacktraces directly from V8 which was not possible with the `segfault-handler` package


Aschen (approved):
This library should also be added in our `kuzzle-runner` image from https://github.com/kuzzleio/kuzzle-containers

PR #2213, opened 2021-11-22
Add new API action `server:metrics`

### LINKED ISSUE #2206: Add new route server:metrics
## Description

The idea is to generate most of the metrics from the Kuzzle core and expose them via a new HTTP server:metrics route. This route will expose for the moment:
- The number of subscriptions on current node.
- the current info from server:getStats (https://docs.kuzzle.io/core/2/api/controllers/server/get-stats/)
- The number of realtime rooms

alexandrebouthinon:
## What does this PR do ?

Add a new route `server:metrics` to fetch relevant Kuzzle core metrics directly from related components.

Fix #2206 

You can test the route using:

- via HTTP
```http
GET "http://localhost:7512/_metrics"
```

- via WebSocket
```json
{
  "action": "metrics",
  "controller": "server"
}
```

You should fetch this kind of response:
```js
{
  "action":"metrics",
  "controller":"server",
  "result": {
    "funnel":{
      "concurrentRequests":1,
      "pendingRequests":0
    },
    "hotelClerk": {
     "rooms":1,
     "subscriptions":1
    },
    "router": {
      "connections": {
        "internal":1,
        "websocket":1,
        "HTTP/1.1":1
      }
    }
  },
  "status":200
}
```

### Other changes

- To provide such metrics, I added a new method `metrics()` to monitored components like `hotelClerk` or the `funnel`. This function can be triggered using internal event system (e.g.: `core:realtime:hotelClerk:metrics` for `hotelClerk`).
- Old routes `server:get*Stats` are now deprecated and will be removed in Kuzzle v3

### DIFF
--- a/doc/2/api/controllers/server/metrics/index.md
+++ b/doc/2/api/controllers/server/metrics/index.md
@@ -6,7 +6,7 @@ title: metrics
 
 # metrics
 
-
+<SinceBadge version="auto-version"/>
 
 Returns inner metrics directly from the current Kuzzle node core components.
 
@@ -36,14 +36,14 @@ Method: GET
 
 Returns the found statistic snapshots in the following format:
 
-- `funnel`: current node API funnel metrics
+- `api`: current node API funnel metrics
   - `concurrentRequests`: number of concurrent requests currently executed.
   - `pendingRequests`: number of requests waiting for execution
-- `hotelClerk`: current node realtime hotelClerk metrics
+- `network`: current node router metrics
+  - `connections`: number of active connections, per network protocol
+- `realtime`: current node realtime hotelClerk metrics
   - `rooms`: number of active realtime rooms
   - `subscriptions`: number of active subscriptions 
-- `router`: current node router metrics
-  - `connections`: number of active connections, per network protocol
 
 
 
@@ -55,20 +55,20 @@ Returns the found statistic snapshots in the following format:
   "controller": "server",
   "requestId": "<unique request identifier>",
   "result": {
-    "funnel":{
+    "api":{
       "concurrentRequests": 1,
       "pendingRequests": 0
     },
-    "hotelClerk": {
-     "rooms": 1,
-     "subscriptions": 1
-    },
-    "router": {
+    "network": {
       "connections": {
         "internal": 1,
         "websocket": 1,
         "HTTP/1.1": 1
       }
+    },
+    "realtime": {
+     "rooms": 1,
+     "subscriptions": 1
     }
   }
 }
--- a/doc/2/api/controllers/server/metrics/index.md
+++ b/doc/2/api/controllers/server/metrics/index.md
@@ -1,0 +1,75 @@
+---
+code: true
+type: page
+title: metrics
+---
+
+# metrics
+
+
+
+Returns inner metrics directly from the current Kuzzle node core components.
+
+---
+
+## Query Syntax
+
+### HTTP
+
+```http
+URL: http://kuzzle:7512/_metrics
+Method: GET
+```
+
+### Other protocols
+
+```js
+{
+  "controller": "server",
+  "action": "metrics"
+}
+```
+
+---
+
+## Response
+
+Returns the found statistic snapshots in the following format:
+
+- `funnel`: current node API funnel metrics
+  - `concurrentRequests`: number of concurrent requests currently executed.
+  - `pendingRequests`: number of requests waiting for execution
+- `hotelClerk`: current node realtime hotelClerk metrics
+  - `rooms`: number of active realtime rooms
+  - `subscriptions`: number of active subscriptions 
+- `router`: current node router metrics
+  - `connections`: number of active connections, per network protocol
+
+
+
+```js
+{
+  "status": 200,
+  "error": null,
+  "action": "metrics",
+  "controller": "server",
+  "requestId": "<unique request identifier>",
+  "result": {
+    "funnel":{
+      "concurrentRequests": 1,
+      "pendingRequests": 0
+    },
+    "hotelClerk": {
+     "rooms": 1,
+     "subscriptions": 1
+    },
+    "router": {
+      "connections": {
+        "internal": 1,
+        "websocket": 1,
+        "HTTP/1.1": 1
+      }
+    }
+  }
+}
+```
--- a/doc/2/api/controllers/server/metrics/index.md
+++ b/doc/2/api/controllers/server/metrics/index.md
@@ -61,9 +61,8 @@ Returns the found statistic snapshots in the following format:
     },
     "network": {
       "connections": {
-        "internal": 1,
         "websocket": 1,
-        "HTTP/1.1": 1
+        "http/1.1": 1
       }
     },
     "realtime": {
--- a/lib/api/controllers/serverController.js
+++ b/lib/api/controllers/serverController.js
@@ -42,10 +42,10 @@ class ServerController extends NativeController {
       'getStats',
       'healthCheck',
       'info',
+      'metrics',
       'now',
       'publicApi',
       'openapi',
-      'metrics'
     ]);
   }
 
@@ -266,9 +266,9 @@ class ServerController extends NativeController {
    */
   async metrics () {
     return {
-      funnel: await global.kuzzle.ask('kuzzle:api:funnel:metrics'),
-      hotelClerk: await global.kuzzle.ask('core:realtime:hotelClerk:metrics'),
-      router: await global.kuzzle.ask('core:network:router:metrics'),
+      api: await global.kuzzle.ask('kuzzle:api:funnel:metrics'),
+      network: await global.kuzzle.ask('core:network:router:metrics'),
+      realtime: await global.kuzzle.ask('core:realtime:hotelClerk:metrics'),
     };
   }
 
--- a/lib/api/controllers/serverController.js
+++ b/lib/api/controllers/serverController.js
@@ -53,6 +53,8 @@ class ServerController extends NativeController {
    *
    * @param {Request} request
    * @returns {Promise<Object>}
+   * 
+   * @deprecated
    */
   getStats (request) {
     return global.kuzzle.statistics.getStats(request);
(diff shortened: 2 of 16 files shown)


doc/2/api/controllers/server/metrics/index.md
```diff
@@ -0,0 +1,75 @@
...
+}
+```
+
+---
+
+## Response
+
+Returns the found statistic snapshots in the following format:
+
+- `funnel`: current node API funnel metrics
```
Aschen, line 39: `funnel` is an internal name. We could think about sometime more explicit like `api`

doc/2/api/controllers/server/metrics/index.md
```diff
@@ -0,0 +1,75 @@
...
+  - `concurrentRequests`: number of concurrent requests currently executed.
+  - `pendingRequests`: number of requests waiting for execution
+- `hotelClerk`: current node realtime hotelClerk metrics
```
Aschen, line 42: Same comment about this naming, `realtime` would be better

doc/2/api/controllers/server/metrics/index.md
```diff
@@ -0,0 +1,75 @@
...
+  - `rooms`: number of active realtime rooms
+  - `subscriptions`: number of active subscriptions 
+- `router`: current node router metrics
```
Aschen, line 45: Same comment, could be `network` for example

doc/2/api/controllers/server/metrics/index.md
```diff
@@ -0,0 +1,75 @@
...
+      "concurrentRequests": 1,
+      "pendingRequests": 0
+    },
+    "hotelClerk": {
+     "rooms": 1,
+     "subscriptions": 1
+    },
+    "router": {
+      "connections": {
+        "internal": 1,
```
Aschen, line 68: This `internal` protocol should not be listed. Also this protocol always have only 1 connection

doc/2/api/controllers/server/metrics/index.md
```diff
@@ -0,0 +1,75 @@
...
+        "websocket": 1,
+        "HTTP/1.1": 1
```
Aschen, line 70: Why not just the name `http` as the protocol name?

lib/api/controllers/serverController.js
```diff
@@ -44,7 +44,8 @@ class ServerController extends NativeController {
       'info',
       'now',
       'publicApi',
-      'openapi'
+      'openapi',
+      'metrics'
```
Aschen, line 48: nitpicking: should be in alphabetical order

lib/config/httpRoutes.js
```diff
@@ -88,16 +88,17 @@ const routes = [
   { verb: 'get', path: '/profiles/_scroll/:scrollId', controller: 'security', action: 'scrollProfiles' },
 
   { verb: 'get', path: '/_adminExists', controller: 'server', action: 'adminExists' },
-  { verb: 'get', path: '/_getAllStats', controller: 'server', action: 'getAllStats' },
+  { verb: 'get', path: '/_getAllStats', controller: 'server', action: 'getAllStats', deprecated: { since: '2.15.0', message: 'Use this route instead: http://kuzzle:7512/_metrics' } }, // @deprecated
```
Aschen, line 91: Wrong version number. You should use the `auto-version` tag instead which is replaced during the release process. Also the `// @deprecated` comment is redundant IMHO

lib/core/network/router.js
```diff
@@ -138,6 +138,33 @@ class Router {
...
+      () => this.metrics());
+  }
+
+  /**
+   * Returns the metrics of the router
+   * @returns {Object}
+   */
+  metrics () {
+    const connectionsByProtocol = new Map();
+    for (const [, connection] of this.connections) {
```
Aschen, line 157: You could iterate over the values only with [Map.values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/values)

lib/core/network/router.js
```diff
@@ -138,6 +138,33 @@ class Router {
...
+      'core:network:router:metrics',
+      () => this.metrics());
+  }
+
+  /**
+   * Returns the metrics of the router
+   * @returns {Object}
+   */
+  metrics () {
+    const connectionsByProtocol = new Map();
```
Aschen, line 156: Why using a map here? Usually we use Map for RAM cache or anything that need fast set/get. Here you use it as an intermediary storage object. Also you have to convert it to a POJO at the end of this method

lib/core/realtime/hotelClerk.ts
```diff
@@ -537,6 +545,17 @@ export class HotelClerk {
     });
   }
 
+  /**
+   * Returns inner metrics from the HotelClerk
+   * @returns {JSONObject}
+   */
+  metrics (): JSONObject {
```
Aschen, line 552: Nit: You could use a more precise type than JSONObject here

lib/core/realtime/hotelClerk.ts
```diff
@@ -537,6 +545,17 @@ export class HotelClerk {
     });
   }
 
+  /**
+   * Returns inner metrics from the HotelClerk
+   * @returns {JSONObject}
```
Aschen, line 550: The `returns` js doc with the type is useless when using typescript

lib/core/realtime/hotelClerk.ts
```diff
@@ -172,6 +172,14 @@ export class HotelClerk {
         return this.unsubscribe(connectionId, roomId, notify);
       });
 
+    /**
+     * Returns inner metrics from the HotelClerk
+     * @return {Object}
```
Aschen, line 177: Nit: You could use the more precise type here.

test/api/funnel/metrics.test.js
```diff
@@ -0,0 +1,21 @@
...
+    kuzzle.ask.withArgs('core:security:user:anonymous:get').resolves({_id: '-1'});
+
+    const funnel = new Funnel();
+    const metrics = funnel.metrics();
+
+    should(metrics).be.an.Object();
+    should(metrics.concurrentRequests).be.a.Number();
+    should(metrics.concurrentRequests).be.equal(0);
+    should(metrics.pendingRequests).be.a.Number();
+    should(metrics.pendingRequests).be.equal(0);
```
Aschen, line 19: Nitpicking: a more concise way to write this test could be with `should(...).match({ ... })`

test/core/network/router/router.test.js
```diff
@@ -134,4 +134,18 @@ describe('Test: router', () => {
...
+
+  describe('#metrics', () => {
+    it('should return the metrics object', () => {
+      // Fake connections
+      router.newConnection(new RequestContext({connection: {id: 'foo', protocol: 'bar'}}));
+      router.newConnection(new RequestContext({connection: {id: 'foo2', protocol: 'bar'}}));
+
+      const metrics = router.metrics();
+
+      should(metrics).be.an.Object();
```
Aschen, line 146: Same comment about the `.match()` matcher

test/core/realtime/hotelClerk/hotelClerk.test.js
```diff
@@ -47,4 +47,13 @@ describe('HotelClerk', () => {
         .be.calledWith('b');
     });
   });
+
+  describe('#metrics', () => {
+    it('should return the metrics object', () => {
+      const metrics = hotelClerk.metrics();
+      should(metrics).be.an.Object();
```
Aschen, line 54: Same comment about the `match()` matcher

Aschen (changes requested):
I think a new Advanced Guide on monitoring Kuzzle could be useful. It doesn't have to be a long one but just having this menu entry on the doc would be great :+1:

lib/config/httpRoutes.js
Aschen: Also, the API documentation should be updated with the `DeprecatedBadge` as well

doc/2/api/controllers/server/metrics/index.md
alexandrebouthinon: This is auto generated from the connections Map

alexandrebouthinon: And I think it's worth keeping this format to track HTTP/2 or HTTP/1.0 (obsolete) connections

doc/2/api/controllers/server/metrics/index.md
alexandrebouthinon: I ignored him

doc/2/api/controllers/server/metrics/index.md
Aschen: I agreed :+1: Just some nitpicking about the naming then, can it be normalized? Either everything in downcase or the real case (like `WebSocket`)

alexandrebouthinon: Yep no problem 👍

lib/core/network/router.js
```diff
@@ -138,6 +138,36 @@ class Router {
...
+    const connectionsByProtocol = {};
+
+    for (const connection of this.connections.values()) {
+      const protocol = connection.connection.protocol.toLowerCase();
+      if (protocol !== 'internal') {
+        if (connectionsByProtocol[protocol] === undefined) {
+          connectionsByProtocol[protocol] = 0;
+        }
+        connectionsByProtocol[protocol]++;
+      }
```
Aschen, line 165: Nitpicking: you could use a guard instead to avoid indenting the whole code block
```js
      if (protocol === 'internal') {
        continue;
      }
      // rest of the code
```

Aschen (approved):

lib/core/network/router.js
alexandrebouthinon: Yes good idea

alexandrebouthinon:
@Aschen I created a separated issue for the Advanced Guide. I think we can merge this PR when reviewed, what do you think?

Aschen:
@alexandrebouthinon Agreed :+1: @Shiranuit Can you review it again please?

rolljee (approved):

PR #2216, opened 2021-11-25
Remove unicode property names from KuzzleRequest

Aschen:
## What does this PR do ?

In the `KuzzleRequest` object, most of the property were using a special Unicode character to prevents user from modifying them (private attribute emulation).

This was preventing Koncorde to match content of the `KuzzleRequest` object directly.

Those properties now have ASCII names. If an user want to edit one of them he can but Kuzzle may not works after that.

### DIFF
--- a/lib/api/request/requestResponse.ts
+++ b/lib/api/request/requestResponse.ts
@@ -20,6 +20,8 @@
  */
 
 import { JSONObject } from 'kuzzle-sdk';
+
+import '../../../lib/types/Global';
 import * as assert from '../../util/assertType';
 import { Deprecation } from '../../types';
 import { KuzzleError } from '../../kerror/errors/kuzzleError';
@@ -43,7 +45,10 @@ export class Headers {
       set: (target, name, value) => this.setHeader(name as string, value),
     });
 
-    this.setHeader('X-Kuzzle-Node', (global as any).kuzzle.id);
+    // eslint-disable-next-line dot-notation
+    if (global['_kuzzle']) {
+      this.setHeader('X-Kuzzle-Node', global.kuzzle.id);
+    }
   }
 
   /**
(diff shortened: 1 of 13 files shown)


test/api/request/requestInput.test.js
```diff
@@ -64,46 +64,4 @@ describe('#RequestInput', () => {
...
-
-  it('should not overwrite the action value if it has already been set', () => {
-    let input = new RequestInput({});
-
-    input.action = 'foo';
-
-    should(input.action).eql('foo');
-
-    input.action = 'bar';
-
```
alexandrebouthinon, line 106: Can this be considered a breaking change since we are changing the KuzzleRequest properties mutation behavior?

Shiranuit (approved):

Njuelle (approved):

lib/api/request/requestResponse.ts
```diff
@@ -154,7 +159,11 @@ export class RequestResponse {
 
   constructor (request) {
     this.raw = false;
-    this[_request] = request;
+
+    Reflect.defineProperty(this, _request, {
```
xbill82, line 163: Here, `_request` is still `request\u200b`, maybe you wanted to do 
```suggestion
    Reflect.defineProperty(this, 'request', {
```

xbill82 (approved):
I don't know why you deleted those tests

Aschen:
@xbill82 I deleted the test because I also deleted the code asserting correct values for properties.

I wanted to use private property (using `Reflect.defineProperty`) so they don't get printed in the console but I noticed a performance drop:
```
Reflect x 3,194,648 ops/sec ±4.05% (81 runs sampled)
Normal x 60,356,260 ops/sec ±4.02% (83 runs sampled)
Seal x 12,397,200 ops/sec ±3.95% (82 runs sampled)
Fastest is Normal
```

Since this object is crucial and instantiated often I decided to use simple properties without getter/setters returning a private one.  
Of course someone can now mess up with the internal `Request` but since they own the code it's their responsibility

PR #2232, opened 2021-12-13
Fix OpenAPI specification check at startup

Aschen:
## What does this PR do ?

The library used to check the validity of the custom OpenAPI schema definition was returning a promise which was not awaited not catched.

### DIFF
--- a/test/core/backend/BackendController.test.js
+++ b/test/core/backend/BackendController.test.js
@@ -72,29 +72,6 @@ describe('Backend', () => {
         .be.eql(definition.actions.sayHello);
     });
 
-    it('should rejects if the controller definition is invalid', () => {
-      definition.actions.sayHello.handler = {};
-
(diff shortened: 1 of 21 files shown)


Shiranuit (approved):

test/core/backend/BackendController.test.js
```diff
@@ -72,29 +72,6 @@ describe('Backend', () => {
         .be.eql(definition.actions.sayHello);
     });
 
-    it('should rejects if the controller definition is invalid', () => {
```
alexandrebouthinon, line 75: Shouldn't this test be updated rather than deleted?

Aschen: We are not checking the definition validity here anymore because the check require a promise and we will loose a lot of DX if developers have to declare controllers with `await app.controller.use`

PR #2237, opened 2021-12-20
Support lodash syntax in request argument extractors

clementbolin:
## What does this PR do ?

We should support lodash syntax in body argument extractors

Example:

```ts
const name = request.getBodyString('person.name');
```

```
npm run test
```


lib/api/request/kuzzleRequest.ts
```diff
@@ -59,7 +60,7 @@ export class KuzzleRequest {
    */
   public id: string;
 
-  constructor (data: any, options: any) {
+  constructor(data: any, options: any) {
```
Aschen, line 63: Your formatter should not delete this space. It allows to quickly find a function definition in a codebase by searching `functionName (`

lib/api/request/kuzzleRequest.ts
```diff
@@ -102,11 +103,15 @@ export class KuzzleRequest {
       if (options.error) {
         if (options.error instanceof Error) {
           this.setError(options.error);
-        }
-        else {
-          const error = new KuzzleError(options.error.message, options.error.status || 500);
-
-          for (const prop of Object.keys(options.error).filter(key => key !== 'message' && key !== 'status')) {
+        } else {
```
Aschen, line 106: This else should be on the next line

lib/api/request/kuzzleRequest.ts
```diff
@@ -125,7 +130,7 @@ export class KuzzleRequest {
   /**
    * Request internal ID
    */
-  get internalId (): string {
+  get internalId(): string {
```
Aschen, line 133: Same comment about the space, apply to all the other methods

lib/api/request/kuzzleRequest.ts
```diff
@@ -196,19 +201,22 @@ export class KuzzleRequest {
...
    * Adds an error to the request, and sets the request's status to the error one.
    */
-  setError (error: Error) {
-    if (! error || !(error instanceof Error)) {
-      throw new InternalError('Cannot set non-error object as a request\'s error');
+  setError(error: Error) {
+    if (!error || !(error instanceof Error)) {
+      throw new InternalError(
+        'Cannot set non-error object as a request\'s error'
+      );
```
Aschen, line 208: Why reformating this? Even if it was slightly more than 80 column, it's better to keep it on a single line than to waste 2 lines

Aschen (changes requested):
Can you cancel all the modification that are not related to this PR but to your IDE coding style please?
Usually we don't use auto-formatter because they are difficult to configure exactly as we want

Aschen (changes requested):
The documentation should be updated to inform users that we support lodash path in those methods

lib/api/request/kuzzleRequest.ts
```diff
@@ -30,7 +30,8 @@ import { KuzzleError, InternalError } from '../../kerror/errors';
 import kerror from '../../kerror';
 import { Deprecation, User } from '../../types';
 import * as assert from '../../util/assertType';
-import { get, isPlainObject } from '../../util/safeObject';
+import { isPlainObject } from '../../util/safeObject';
+import * as _ from 'lodash';
```
OlivierCavadenti, line 34: this import the whole library, you could just import the single method with:

`import get from 'lodash/get';`

test/api/request/request.test.js
```diff
@@ -421,11 +421,26 @@ describe('#Request', () => {
...
+              damage: 100
+            }
+          }
         };
       });
 
       describe('#getBodyArray', () => {
+        it('extracts the required parameter (lodash request)', () => {
+          should(request.getBodyArray('relations.lebron'))
+            .exactly(request.input.body.relations.lebron);
```
OlivierCavadenti, line 441: `_.get` path accept indexes for arrays like "relations.lebron[1]", could be nice to add a test case.

lib/api/request/kuzzleRequest.ts
Aschen: I agree on this for a frontend application but I'm not sure it changes something for a Node.js application.  
I think all our dependencies should use this kind of partial import maybe. Either way, actually the whole lodash code is always loaded into RAM :facepalm: 
https://user-images.githubusercontent.com/4447392/139255742-08e819d6-5dff-4439-b341-f48004ddb45b.png

OlivierCavadenti:
By the way, few points:  

- You have a system to return a default value here : 
```js
const value = get(obj, name);

    if (value === undefined || value === null) {
      if (def !== null) {
        return def;
      }
    ...
    }
```
The lodash _.get function take a third parameter to do that for you (and simplify code): `const value = get(obj, name, def);`

- In _getString (and getBodyString, ... functions), I think the parameter "name" (and doc) is not accurate now because you can also specify path (parameter name + doc (with path example) can be rework to avoid confusion).

- To finish, lodash _.get function can also take an array to extract data in object, can be useful to add this :
`getBodyString (name: string | string[], def: string | null = null): string`

clementbolin:
> By the way, few points:
> 
> * You have a system to return a default value here :
> 
> ```js
> const value = get(obj, name);
> 
>     if (value === undefined || value === null) {
>       if (def !== null) {
>         return def;
>       }
>     ...
>     }
> ```
> 
> The lodash _.get function take a third parameter to do that for you (and simplify code): `const value = get(obj, name, def);`
> 
> * In _getString (and getBodyString, ... functions), I think the parameter "name" (and doc) is not accurate now because you can also specify path (parameter name + doc (with path example) can be rework to avoid confusion).
> * To finish, lodash _.get function can also take an array to extract data in object, can be useful to add this :
>   `getBodyString (name: string | string[], def: string | null = null): string`



> By the way, few points:
> 
> * You have a system to return a default value here :
> 
> ```js
> const value = get(obj, name);
> 
>     if (value === undefined || value === null) {
>       if (def !== null) {
>         return def;
>       }
>     ...
>     }
> ```
> 
> The lodash _.get function take a third parameter to do that for you (and simplify code): `const value = get(obj, name, def);`
> 
> * In _getString (and getBodyString, ... functions), I think the parameter "name" (and doc) is not accurate now because you can also specify path (parameter name + doc (with path example) can be rework to avoid confusion).
> * To finish, lodash _.get function can also take an array to extract data in object, can be useful to add this :
>   `getBodyString (name: string | string[], def: string | null = null): string`

Yes I had seen that it was possible to give a default value to the get function of lodash, but in the following case for example by removing the condition that he suggested me to remove, we would have an error throw that would not be desired.

If my obj = `{ from: 1, size: 11, scroll: '10m', searchBody: null }` name = `searchBody`and my def = `{}`
the lodash function will return `null` which will trigger an error, when we would have liked it to return an empty object.
This is exactly what happens on the test: `should provide empty body when the route is invoked with GET with a null search body is provided line` `1022` in the request tests.
So even if we can work around this problem for specific cases, I'm for keeping this condition.

Aschen:
@clementbolin I don't understand your point about the way we fetch the default value?

I don't see why we couldn't use lodash feature to return a default argument. Before we had this custom code because we were not using lodash

doc/2/framework/classes/kuzzle-request/get-body-array/index.md
```diff
@@ -10,6 +10,7 @@ description: KuzzleRequest class getBodyArray() method
 <SinceBadge version="2.11.0" />
 
 Gets a parameter from the request body and checks that it is an array.
+We also support lodash syntax. [(```relations.lebron[0]```)](https://lodash.com/docs/4.17.15#get)
```
Aschen, line 13: Single line code formatting is done with two backtick 
```suggestion
We also support lodash syntax. [(`relations.lebron[0]`)](https://lodash.com/docs/4.17.15#get)
```
Can you please also correct the other occurrences please?

Aschen (changes requested):

Aschen (changes requested):
Missing the `SinceBadge`. You can use the `auto-version` placeholder, it will be replaced in the release process

Shiranuit (approved):
Apply changes request by Aschen and this will be good to me

lib/api/request/kuzzleRequest.ts
```diff
@@ -849,13 +849,9 @@ export class KuzzleRequest {
     errorName: string,
     def: number | null = null
   ): number {
-    let value = get(obj, name);
+    let value = get(obj, name, def) || def;
```
Aschen, line 852: Why do we need to add the `|| def` part? Lodash does not returns the default param directly?

doc/2/framework/classes/kuzzle-request/get-body-array/index.md
clementbolin: done

Shiranuit (approved):

lib/api/request/kuzzleRequest.ts
```diff
@@ -848,13 +849,9 @@ export class KuzzleRequest {
     errorName: string,
     def: number | null = null
   ): number {
-    let value = get(obj, name);
+    let value = get(obj, name, def);
 
     if (value === undefined || value === null) {
```
Aschen, line 854: a `null` value should not be considered as missing (apply to other part of the code)

Aschen (changes requested):

Aschen (approved):

PR #2238, opened 2021-12-20
Support Elasticsearch 7.14+

alexandrebouthinon:
# Why this need some changes by our side?

There is a new feature introduced in 7.14 version of Elasticsearch and enabled by default. This one create an index called `.geoip_databases` incompatible with the custom Kuzzle data model.
It can be disabled in the `elasticsearch.yml`:
```yaml
ingest:
  geoip:
    downloader:
      enabled: false
```
But since this is enable by default, we need to ignore the automatically created index to avoid issues when using the native index:stats Elasticsearch API action

### DIFF
--- a/.ci/test-cluster.yml
+++ b/.ci/test-cluster.yml
@@ -78,7 +78,11 @@ services:
     container_name: kuzzle_redis
 
   elasticsearch:
-    image: kuzzleio/elasticsearch:7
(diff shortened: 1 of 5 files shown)


Aschen (changes requested):
Missing unit tests

Aschen (approved):

PR #2239, opened 2021-12-21
Revert `collection:truncate` behavior

alexandrebouthinon:
# Why?

The new `collection:truncate` internal behavior using Elasticsearch `deleteByQuery` API action should be revert since this ES API action can only delete 1000 documents by default and is limited by Kuzzle limits configuration.

### DIFF
--- a/doc/2/api/controllers/collection/truncate/index.md
+++ b/doc/2/api/controllers/collection/truncate/index.md
@@ -8,6 +8,11 @@ title: truncate
 
 Empties a collection by removing all its documents, while keeping any associated mapping.
 
+::: info
+This action delete then recreate the related Elasticsearch index.
+Please note that deleting/creating an index cannot be done concurrently within an Elasticsearch cluster
+:::
+
 ::: warning
 Documents removed that way do not trigger real-time notifications.
 :::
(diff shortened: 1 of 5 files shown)


Aschen (approved):

Aschen:
Can you add a small explanation in the doc about the fact that this operation will delete and recreate ES indice and that this operation is not // inside an ES cluster.  

For example in tests we often truncate many collections and in that case the refresh + deleteByQuery method should be used

Shiranuit (approved):

doc/2/api/controllers/collection/truncate/index.md
```diff
@@ -8,6 +8,11 @@ title: truncate
 
 Empties a collection by removing all its documents, while keeping any associated mapping.
 
+::: info
+This action delete then recreate the related Elasticsearch index.
+Please note that deleting/creating an index cannot be done concurrently within an Elasticsearch cluster
```
Aschen, line 13:
```suggestion
Please note that deleting/creating an index cannot be done concurrently within an Elasticsearch cluster, if you need to truncate a lot of collections (for your functional tests for example), then you should use `collection:refresh` and `document:deleteByQuery`
```

Aschen (approved):

PR #2240, opened 2021-12-21
Anonymous token if we have no token

clementbolin:
## What does this PR do ?

no longer throw an error when a browser connects to Kuzzle for the first time during the request ```auth:checkToken```

```
npm run test:unit
```

### Other changes

Update unit tests

### DIFF
--- a/lib/api/controllers/authController.js
+++ b/lib/api/controllers/authController.js
@@ -430,11 +430,11 @@ class AuthController extends NativeController {
       token = request.input.jwt;
     }
     else {
-      token = request.getBodyString('token');
+      token = request.getBodyString('token', '') || null;
     }
 
     try {
-      const { expiresAt, userId } = await this.ask('core:security:token:verify', token);
+      const { expiresAt = -1, userId } = await this.ask('core:security:token:verify', token);
 
       return { expiresAt, kuid: userId, valid: true };
     }
--- a/test/api/controllers/authController.test.js
+++ b/test/api/controllers/authController.test.js
@@ -767,12 +767,18 @@ describe('Test the auth controller', () => {
       testToken = new Token({ expiresAt: 42, userId: 'durres' });
     });
 
-    it('should reject an error if no token is provided', () => {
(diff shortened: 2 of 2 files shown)


Shiranuit (approved):

lib/api/controllers/authController.js
```diff
@@ -430,11 +430,11 @@ class AuthController extends NativeController {
       token = request.input.jwt;
     }
     else {
-      token = request.getBodyString('token');
+      token = request.getBodyString('token', '') || null;
```
Aschen, line 433:
```suggestion
      token = request.getBodyString('token', '', null);
```

Aschen (changes requested):

lib/api/controllers/authController.js
Shiranuit: @Aschen The second parameter of `getBodyString` is the default value, the goal here is to set the `token` to null if there is no body or the field `token` does not exists.

The catch is that `getBodyString` will throw if the default value is set to null and the field is missing, that's why it's defaulted to an empty string that will then be evaluated with `|| null` to retrieve a null value if the field is missing.

Aschen: Ok I see the problem. Then the modification suggested here should fix that https://github.com/kuzzleio/kuzzle/pull/2237#issuecomment-998654473

Shiranuit: I read the suggested modifications  and this will not fix the probleme, because if we give a null value to lodash, if not found lodash will return the null value causing the method getBodyString to throw because the value could not be undefined or null.

Aschen: Because we shouldn't throw exception for null value. The `missing_argument` error should only be thrown when the argument is missing

Shiranuit: Yes, so we should change the method to throw `missing_argument` not based on the returned value of `_.get` but `_.has`
to verify that the argument is not missing

alexandrebouthinon (approved):

Aschen (approved):

PR #2242, opened 2021-12-23
Add server:capabilities api action

clementbolin:
## What does this PR do ?

Exposing specific properties to the configuration because `server:getConfig` / `server:info` give too much info. We want a less verbose API action.

With the following information:

- exposing Kuzzle limits
- exposing backend for the storage and cache services (e.g. elasticsearch and redis)
- API http routes (as in the server:publicApi action)

Start Kuzzle with `docker-compose up`, then with kourou use the capabilities action: `kourou server:capabilities` or send GET request to the route `http:localhost:7512/_capabilities` route.


doc/2/api/controllers/server/capabilities/index.md
```diff
@@ -0,0 +1,73 @@
+---
+code: true
+type: page
+title: capabilities
+---
+
+# Capabilities
```
Aschen, line 7: Should match the action name
```suggestion
# capabilities
```

doc/2/api/controllers/server/capabilities/index.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+
+Returns the current Kuzzle capabilities.
+
+This route should only be accessible to administrators, as it might return sensitive information about the backend.
```
Aschen, line 13: I don't agree, this action (not route) should be accessible to most of the users. As mentioned in the original issue, it can allows SDK to behave accordingly automatically.

This action purpose is to inform the client about what kind of limits or behavior it should expect from Kuzzle.

You can look at the original issues for more info about the usage and the expected response

Aschen (changes requested):

doc/2/api/controllers/server/capabilities/index.md
Aschen, line 7: Missing the `SinceBadge`. You can use the `auto-version` placeholder, it will be replaced in the release process

Aschen (changes requested):

doc/2/api/controllers/server/capabilities/index.md
```diff
@@ -4,20 +4,21 @@ type: page
...
-This route should only be accessible to administrators, as it might return sensitive information about the backend.
+This action is accessible by all users, the purpose of this action is to inform the client what kind of limits or behavior they should expect from Kuzzle.
 
 ---
 
 ## Query Syntax
 
 ### HTTP
 
+<SinceBadge version="2.16.6"/>
```
Aschen, line 21: You need to use the `auto-version` placeholder since you cannot predict the version number of the release your feature will be on. This placeholder is replaced by the release tool in the release process.

Aschen (changes requested):

lib/api/controllers/serverController.js
```diff
@@ -110,6 +111,21 @@ class ServerController extends NativeController {
...
+   * @returns {Promise<Object>}
+   */
+  async capabilities () {
+    const config = JSON.parse(JSON.stringify(global.kuzzle.config));
+    const publicApi = await this.publicApi();
+
+    return {
+      limits: config.limits,
+      routes: publicApi,
+      services: config.services,
```
Aschen, line 125: You shouldn't expose the entire services configuration since it can contains sensitive information such as ES credentials.  

Just exposing the services backend names should be enough info

Aschen (changes requested):

doc/2/api/controllers/server/capabilities/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+This action is accessible by all users, the purpose of this action is to inform the client what kind of limits or behavior they should expect from Kuzzle.
+
+---
+
+## Query Syntax
+
+### HTTP
+
+<SinceBadge version="auto-version"/>
```
Aschen, line 21: The badge should be at the beginning of the page when the entire page is new, just after the title

lib/api/controllers/serverController.js
```diff
@@ -46,6 +47,7 @@ class ServerController extends NativeController {
       'now',
       'publicApi',
       'openapi',
+      'capabilities',
```
Aschen, line 50: Nitpicking
Can you sort this array please?

lib/api/controllers/serverController.js
```diff
@@ -110,6 +112,28 @@ class ServerController extends NativeController {
...
+  /**
+   * Returns the Kuzzle capabilities
+   * @returns {Promise<Object>}
+   */
+  async capabilities () {
+    const config = JSON.parse(JSON.stringify(global.kuzzle.config));
+    const publicApi = await this.publicApi();
+    const services = [];
+
+    _.map(config.services, (val, key) => {
```
Aschen, line 124: In general, avoid usage of lodash methods, here you could use `Object.entries(config.services).map`

Also, the purpose of `map` is to create an array from another array. Here an useless array is created and returned by `map` because you are populating the array you created l122.

Finally, we tend to avoid the usage of `map` when it's not needed because the `for (const ... of ...)` is more readable.

lib/api/controllers/serverController.js
```diff
@@ -110,6 +112,28 @@ class ServerController extends NativeController {
...
+      if (val.backend !== undefined) {
+        services.push({ [key]: val.backend });
+      }
+    });
+
+    return {
+      limits: config.limits,
+      routes: publicApi,
+      services: services,
```
Aschen, line 133: You should not expose the entire `services` configuration object, it contains sensitive values.

Here we want to know what kind of services Kuzzle is using: Elasticsearch or Postgres? Redis or RAM ?

lib/api/controllers/serverController.js
```diff
@@ -110,6 +112,28 @@ class ServerController extends NativeController {
...
+    const publicApi = await this.publicApi();
+    const services = [];
+
+    _.map(config.services, (val, key) => {
+      if (val.backend !== undefined) {
+        services.push({ [key]: val.backend });
+      }
+    });
+
+    return {
```
Aschen, line 130: I think we could also add the list of available plugins and their versions

doc/2/api/controllers/server/capabilities/index.md
```diff
@@ -0,0 +1,74 @@
...
+---
+
+# capabilities
+
+
+
+Returns the current Kuzzle capabilities.
+
+This action is accessible by all users, the purpose of this action is to inform the client what kind of limits or behavior they should expect from Kuzzle.
+
```
Aschen, line 14: You should add the list of which properties are returned and what information they are giving

doc/2/api/controllers/server/capabilities/index.md
```diff
@@ -0,0 +1,74 @@
...
+title: capabilities
+---
+
+# capabilities
+
+
+
+Returns the current Kuzzle capabilities.
+
+This action is accessible by all users, the purpose of this action is to inform the client what kind of limits or behavior they should expect from Kuzzle.
```
Aschen, line 13: Some nitpicking

```suggestion
::: info
This action should be accessible by all users, the purpose of this action is to inform the client what kind of limits or behavior they should expect from Kuzzle. 
The SDKs are using this action to modify their behavior accordingly to Kuzzle capabilities.
:::
```

Aschen (changes requested):

lib/api/controllers/serverController.js
clementbolin: This is what I do with the `map` line 124, I only get the names of the services that I store in the services variable.

Aschen: Oh yes sorry :facepalm:

Aschen: Instead of just having `serviceName: backendName` can we have an object containing `backend: <name>` ?
It will offer more flexibility in the future if we want to add other properties in services description

clementbolin: I'm not sure I really understand what you want, you would have something like these examples: `backend: 'Redis Elasticsearch`or an object like this `backend:  { serviceName: backendName, serviceName: backendName, etc ... }`?

Aschen: More something like
```services: {
  storage: {
    backend: 'elasticsearch',
    version: '7.10.2',
  },
  cache: {
    backend: 'redis',
    version: '6.1.0',
  }
}
```

Shiranuit (approved):
Apply requested changes and this will be good to me

lib/api/controllers/serverController.js
clementbolin: I have for the moment displayed the version of each service in hard in the `default.config` until I find a way to retrieve it dynamically

lib/api/controllers/serverController.js
```diff
@@ -119,6 +119,16 @@ class ServerController extends NativeController {
   async capabilities () {
     const config = JSON.parse(JSON.stringify(global.kuzzle.config));
     const publicApi = await this.publicApi();
+    const servicesVersion = {
```
Aschen, line 122: You should avoid to retrieve the info each time the action is called.

Aschen (changes requested):

lib/api/controllers/baseController.js
```diff
@@ -72,7 +76,9 @@ class NativeController extends BaseController {
    *
    * @returns {Promise}
    */
-  init () {
+  async init () {
+    this._info.elasticsearch = await global.kuzzle.ask('core:storage:public:info:get');
+    this._info.redis = await global.kuzzle.ask('core:cache:public:info:get');
```
Aschen, line 81: Those promise should be made in //

Aschen (changes requested):
Missing functional tests

features/ServerController.feature
```diff
@@ -93,3 +93,14 @@ Feature: Server Controller
   Scenario: Http call onto deprecated method should not print a warning when NODE_ENV=production
     When I execute the action "server":"publicApi"
     Then The response should contains a "deprecations" equals to undefined
+
+  # server:capabilities =====================================================================
+  @http
```
Aschen, line 98: This tag indicate that the scenario is only for HTTP protocol

lib/api/controllers/baseController.js
```diff
@@ -72,7 +76,13 @@ class NativeController extends BaseController {
    *
    * @returns {Promise}
    */
-  init () {
+  async init () {
```
Aschen, line 79: You should not put that in the BaseController because then every controller will make those 2 calls to `kuzzle.ask`

Aschen (changes requested):

lib/api/controllers/serverController.js
```diff
@@ -54,6 +55,27 @@ class ServerController extends NativeController {
...
+   * @returns {Promise}
+  */
+  async init () {
+    const [storageInfo, cacheInfo] = await Promise.all([
+      global.kuzzle.ask('core:storage:public:info:get'),
+      global.kuzzle.ask('core:cache:public:info:get'),
+    ]);
+    this._info.elasticsearch = storageInfo;
+    this._info.redis = cacheInfo;
+    return Bluebird.resolve();
```
Aschen, line 78: You don't need to return another promise, since the method is `async` it will automatically returns a promise

Aschen (changes requested):

Aschen (approved):

Issue #2073, opened 2021-05-27
Add the _source option to mCreateOrReplace


xbill82:
## Feature Description

Add the `_source` option to the `mCreateOrReplace` action. In Kuzzle v2 this option will be `true` by default, but will enable to avoid receiving the source of the concerned documents, to shorten the response.
In Kuzzle v3, this option will be `false` by default.

PR #2244, opened 2021-12-24
Add _source argument for mCreateOrReplce API action

clementbolin:
#2073 

## What does this PR do ?

Add the `_source` option to the mCreateOrReplace action. In Kuzzle v2 this option will be `true` by default, but will enable to avoid receiving the source of the concerned documents, to shorten the response.
In Kuzzle v3, this option will be `false` by default.

```
npm run test:unit
```

## Other changes

Doc openApi about `document: mCreateOrReplce`


lib/api/controllers/documentController.js
```diff
@@ -729,6 +730,11 @@ class DocumentController extends NativeController {
 
     this.assertNotExceedMaxWrite(documents.length);
 
+    if (Object.prototype.hasOwnProperty.call(request.input.args, '_source') &&
```
alexandrebouthinon, line 733: Use this to get this parameter and set it to `true` by default https://docs.kuzzle.io/core/2/framework/classes/kuzzle-request/get-string/

Aschen: Why not directly accessing the property with `request.input.args._source` ? Using `Object.prototype.hasOwnProperty.call` is very slow and not readable.

Either way you should use the `getBoolean` method to correctly retrieve boolean flags

lib/core/storage/clientAdapter.js
```diff
@@ -514,6 +514,8 @@ class ClientAdapter {
     global.kuzzle.onAsk(
       `core:storage:${this.scope}:document:mCreateOrReplace`,
       (index, collection, documents, opts) => {
+        // eslint-disable-next-line no-console
+        console.log('here');
```
Aschen, line 518: Debug leftover

lib/service/storage/elasticsearch.js
```diff
@@ -2121,7 +2121,9 @@ class ElasticSearch extends Service {
           _index: alias
         }
       });
-      esRequest.body.push(extractedDocuments[i]._source);
+      if (_source) {
```
Aschen, line 2124: A unit test should be added/modified for this

test/api/controllers/documentController.test.js
```diff
@@ -669,6 +669,68 @@ describe('DocumentController', () => {
...
+
+    it('should return success result of mCreateOrReplace with _source for each documents', async () => {
+      request.input.args.silent = true;
+      request.input.body = content;
+      
+      content = {
+        body: content,
+      };
+
+      await documentController._mChanges(
```
Aschen, line 718: You are testing the `mChanges` method rather than the `mCreateOrReplace` one

Aschen (changes requested):
Good catch to notice that the `mCreateOrReplace` method did not have any unit test. Can you add some to test the behavior please?

lib/api/controllers/documentController.js
```diff
@@ -729,6 +730,11 @@ class DocumentController extends NativeController {
 
     this.assertNotExceedMaxWrite(documents.length);
 
+    if (Object.prototype.hasOwnProperty.call(request.input.args, '_source') && 
+        !request.getBoolean('_source')) {
+      _source = false;
+    }
```
Shiranuit, line 736:
```suggestion
    const _source = request.getBoolean('_source');
```

Unlike other methods `getBoolean` does not throw if the argument is missing and returns false instead, using this is far more simpler and readable.

Shiranuit (changes requested):

lib/api/controllers/documentController.js
xbill82: What is the issue of this discussion? Why has it been marked as resolved?

xbill82: Given the surrounding conversations, I frankly do not understand why this very simple bit of code is in this shape.

lib/api/controllers/documentController.js
clementbolin: Yes, but I did that because by default `getBoolean` returns `false` if the argument is missing, and I want it to returs `true` if the `_source` param is missing. Is there a function for know if the param exist by default ?

Aschen: You must check if the argument exists in `request.input.args.source`, if it doesn't then default value is true, if it does then the is considered as false only if it's `false` (the boolean` or `"false"` (the string)

package.json
```diff
@@ -34,7 +34,8 @@
     "test:lint:js": "eslint --max-warnings=0 ./lib ./test ./features-legacy ./bin ./features ./doc/build-error-codes.js ./plugins/available/functional-test-plugin",
     "test:lint:js:fix": "eslint --max-warnings=0 --fix ./lib ./test ./features-legacy ./bin ./features ./doc/build-error-codes.js",
     "doc-error-codes": "node -r ts-node/register doc/build-error-codes",
-    "docker:install": "docker-compose run kuzzle_node_1 npm install"
+    "docker:install": "docker-compose run kuzzle_node_1 npm install",
+    "docker:run": "docker-compose run kuzzle_node_1 npm run --"
```
Aschen, line 38: The naming is not very explicit since this script is running NPM and not the project

lib/api/controllers/documentController.js
```diff
@@ -716,7 +716,7 @@ class DocumentController extends NativeController {
    * @returns {Promise.<Object>} { successes, errors }
    */
   async _mChanges (request, methodName, action) {
-    let _source = true;
+    let _source = request.input.args._source !== undefined ? request.getBoolean('_source') : true;
```
Aschen: Why are you not just using `getBoolean` like `document.update` for example?

clementbolin: Because if the value is `undefined`, the `getBoolean()` returns `false` instead of `true` the value I want from the issue #2073

package.json
clementbolin: Yes, I agree, I had not idea about naming but maybe something like this could be more explicit: `docker:script` `docker:npm` ?

Aschen (approved):

package.json
Aschen: `docker:npm` seem fine to me :+1:

.ci/test-cluster.yml
```diff
@@ -78,7 +78,7 @@ services:
     container_name: kuzzle_redis
 
   elasticsearch:
-    image: kuzzleio/elasticsearch:7.16.2
+    image: kuzzleio/elasticsearch:7
```
clementbolin, line 81: I use this version of ES because it is compatible with Mac M1 ? Is this a problem @alexandrebouthinon ?

lib/service/storage/elasticsearch.js
clementbolin: I added functionnal tests on this part of the code

features/DocumentController.feature
```diff
@@ -346,6 +346,25 @@ Feature: Document Controller
...
+    And I "create" the following documents:
+      | _id          | body                               |
+      | "document-1" | { "name": "document1", "age": 42 } |
+    When I successfully execute the action "document":"mCreateOrReplace" with args:
+      | index      | "nyc-open-data"        |
+      | collection | "yellow-taxi"          |
+      | body       | { "documents": [ { "_id": "document-1", "body": { "name": "replaced1" } } ] } |
+      | source     | false                  |
+    Then I should receive a "successes" array of objects matching:
+      | _id          |  status | result    | created |
```
Aschen, line 361: You should test that `_source` is `undefined`. You can use the `_UNDEFINED_` placeholder for this

features/DocumentController.feature
```diff
@@ -346,6 +346,25 @@ Feature: Document Controller
...
+      | "document-1" |  200    | "updated" | false   |
+    And I should receive a empty "errors" array
+    And I refresh the collection
+    And I count 1 documents
+    And The document "document-1" content match:
+      | name | "replaced1" |
```
Aschen, line 367: This is useless for your test case because it's duplicated from the previous test case. This will slow down the testing phase and should be removed

.ci/test-cluster.yml
Aschen: The `7` version is a pointer on the latest version and should always be used to automatically keep our CI up to date

doc/2/api/controllers/document/m-create-or-replace/index.md
```diff
@@ -19,7 +19,7 @@ The number of documents that can be created or replaced by a single request is l
 ### HTTP
 
 ```http
-URL: http://kuzzle:7512/<index>/<collection>/_mCreateOrReplace[?refresh=wait_for][&silent]
+URL: http://kuzzle:7512/<index>/<collection>/_mCreateOrReplace[?refresh=wait_for][&silent][&_source]
```
Aschen, line 22:
```suggestion
URL: http://kuzzle:7512/<index>/<collection>/_mCreateOrReplace[?refresh=wait_for][&silent][&source]
```

doc/2/api/controllers/document/m-create-or-replace/index.md
```diff
@@ -89,6 +89,7 @@ kourou document:mCreateOrReplace <index> <collection> <body> -a silent=true
 - `refresh`: if set to `wait_for`, Kuzzle will not respond until the created/replaced documents are indexed
 - `silent`: if set, then Kuzzle will not generate notifications <SinceBadge version="2.9.2" />
 - `strict`: if set, an error will occur if at least one document has not been created/replaced <SinceBadge version="2.11.0" />
+- `_source`: if set, at true, the response will contain the _source field of the created/replaced documents (default true) <SinceBadge version="auto-version"/>
```
Aschen, line 92:
```suggestion
- `source`: if set, at true, the response will contain the _source field of the created/replaced documents (default true) <SinceBadge version="auto-version"/>
```

features/DocumentController.feature
```diff
@@ -207,79 +207,6 @@ Feature: Document Controller
     And I refresh the collection
     Then The document "document-1" should not exist
 
-  # document:export ============================================================
```
Aschen, line 210: Why are you removing those tests?

lib/api/controllers/documentController.js
```diff
@@ -829,6 +829,7 @@ class DocumentController extends NativeController {
    * @returns {Promise.<Object>} { successes, errors }
    */
   async _mChanges (request, methodName, action) {
+    const _source = request.input.args._source !== undefined ? request.getBoolean('_source') : true;
```
Aschen, line 832: The option name should be `source`

lib/service/storage/elasticsearch.js
Aschen: I know but on Kuzzle core we need to write unit tests as well to ensure maximal quality and maintainability

Aschen (changes requested):

features/DocumentController.feature
clementbolin: It was a mistake.

lib/service/storage/elasticsearch.js
clementbolin: I talked with Yannick about the unit tests, and we think it's unnecessary because the ES tests use mock function.

test/service/storage/elasticsearch.test.js
```diff
@@ -3536,6 +3536,36 @@ describe('Test: ElasticSearch service', () => {
       elasticsearch._mExecute = sinon.stub().resolves(mExecuteResult);
     });
 
+    it('should call _mExecute with formated documents and source flag', () => {
+      const promise = elasticsearch.mCreateOrReplace(index, collection, documents, { source: false });
+
+      return promise
```
OlivierCavadenti, line 3542: I know that all unit tests are made like this here, but why not use async / await syntax ?

.ci/test-cluster.yml
alexandrebouthinon: @Aschen is right this is the same image with two different tag

test/service/storage/elasticsearch.test.js
clementbolin: I think you are right, I followed the standard of other unit tests.

alexandrebouthinon: I agree with Olivier, new unit tests should be written using async/await syntax. The old ones will be update on fly

alexandrebouthinon (approved):

test/service/storage/elasticsearch.test.js
```diff
@@ -3536,6 +3536,36 @@ describe('Test: ElasticSearch service', () => {
       elasticsearch._mExecute = sinon.stub().resolves(mExecuteResult);
     });
 
+    it('should call _mExecute with formated documents and source flag', async () => {
```
OlivierCavadenti, line 3539: My idea is to make like in redis.test.js ([line 68](https://github.com/kuzzleio/kuzzle/blob/6cbf4ad03cb5a2923863771451d1e360043e8776/test/service/cache/redis.test.js#L68)).

The goal is to avoid nested then calls like:

```js
const result = await elasticsearch.mCreateOrReplace(index, collection, documents, { source: false });
```

PR #2245, opened 2021-12-27
Create signup action in auth controller

matissepage:
## What does this PR do ?

Add a auth:signup action who is more intuitive and deprecate the security:createRestrictedUser action


doc/2/api/controllers/auth/signup/index.md
```diff
@@ -0,0 +1,114 @@
...
+
+---
+
+## Response
+
+Returns the restricted user creation status:
+
+- `_id`: new user kuid
+- `_source`: new user content and attributed profiles
+- `created`: always true
```
Aschen, line 94: Should not be present if always true

doc/2/api/controllers/auth/signup/index.md
```diff
@@ -0,0 +1,114 @@
...
+- `version`: always 1
```
Aschen, line 95: Same comment, IMHO this property is useless

lib/api/controllers/authController.js
```diff
@@ -35,6 +35,9 @@ const { NativeController } = require('./baseController');
 const formatProcessing = require('../../core/auth/formatProcessing');
 const User = require('../../model/security/user');
 const ApiKey = require('../../model/storage/apiKey');
+const { generateRandomName } = require('../../util/name-generator');
+const { v4: uuidv4 } = require('uuid');
```
Aschen, line 39: This import should be next to the external packages imports l.27 and l.28.

The module imports should be separated in 3 parts:
```
// internal node modules like 'fs'

// external packages like 'lodash'

// local packages like './services/user'
```

lib/api/controllers/authController.js
```diff
@@ -57,6 +60,7 @@ class AuthController extends NativeController {
       'getMyCredentials',
       'getMyRights',
       'getStrategies',
+      'signup',
```
Aschen, line 63: This should be sorted alphabetically

lib/api/controllers/authController.js
```diff
@@ -272,6 +279,28 @@ class AuthController extends NativeController {
     };
   }
 
+  /**
+   * Creates a new User object in Kuzzle's database layer and applies restricted profileIds
+   *
+   * @param {Request} request
+   * @returns {Promise<Object>}
+   */
+  async signup(request) {
```
Aschen, line 288:
```suggestion
  async signup (request) {
```

lib/api/controllers/authController.js
```diff
@@ -649,6 +678,151 @@ class AuthController extends NativeController {
       throw kerror.get('security', 'rights', 'unauthorized');
     }
   }
+
+  /**
+   * @param {Request} request
+   * @returns {Promise}
+   * @private
+   */
+  async _persistUser(request, profileIds, content, { humanReadableId=true } = {}) {
```
Aschen, line 687: You should avoid code duplication here since this method is also present in the `SecurityController`.  

Also, the `security.createRestrictedUser` method also contains duplicated code.

Aschen (changes requested):

lib/api/controllers/securityController.js
```diff
@@ -791,18 +792,12 @@ class SecurityController extends NativeController {
...
-    if (has(content, 'profileIds')) {
-      throw kerror.get('api', 'assert', 'forbidden_argument', 'body.content.profileIds');
-    }
-
-    return this._persistUser(
-      request,
-      global.kuzzle.config.security.restrictedProfileIds,
-      content,
-      { humanReadableId });
+    const under_request = new KuzzleRequest({
```
Aschen, line 795: You need to keep the whole original request context, you can look on how it's done for the JSON Query endpoint here: https://github.com/kuzzleio/kuzzle/blob/master/lib/core/network/router.js#L113

Aschen (changes requested):

PR #2249, opened 2022-01-12
Add a keepalive for Redis

Shiranuit:
## What does this PR do ?

Add a Keepalive mechanism that ping redis every 60s to keep the connection alive on Microsoft Azure

### DIFF
--- a/lib/service/cache/redis.js
+++ b/lib/service/cache/redis.js
@@ -102,7 +102,6 @@ class Redis extends Service {
    */
   _setupKeepAlive() {
     this.client.on('ready', async () => {
-      console.log('Redis client ready');
       await this._ping();
       this.pingIntervalID = setInterval(this._ping, 60000);
     });
--- a/lib/service/cache/redis.js
+++ b/lib/service/cache/redis.js
@@ -82,7 +82,9 @@ class Redis extends Service {
 
     this.setCommands();
     // TODO: Remove this when IORedis does support application level pings to keep connection to Azure Redis alive
-    this._setupKeepAlive();
+    if (config.pingKeepAlive && config.pingKeepAlive > 0) {
+      this._setupKeepAlive(config.pingKeepAlive);
+    }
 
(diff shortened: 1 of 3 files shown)


Aschen (approved):

lib/service/cache/redis.js
```diff
@@ -93,6 +96,34 @@ class Redis extends Service {
...
   }
 
+  /**
+   * Setup a ping interval to keep the connection alive
+   * Every 60 seconds a ping is sent to Redis
+   */
+  _setupKeepAlive() {
+    this.client.on('ready', async () => {
+      await this._ping();
+      this.pingIntervalID = setInterval(this._ping.bind(this), 60000);
```
alexandrebouthinon, line 106: I personnaly don't like hardcoded value in code. You could create a new configuration field to:
- enable/disable the ping mechanism
- configure the ping delay

alexandrebouthinon (changes requested):

lib/service/cache/redis.js
alexandrebouthinon: This new configuration should be use later when IoRedis ping implementation will be ready

alexandrebouthinon: And I think this should be disabled by default to avoid edge effects on other installation that do not require it

Shiranuit: I did add a config that enables the Ping Keep Alive and sets its delay

alexandrebouthinon (approved):

alexandrebouthinon (changes requested):
Missing default value in https://github.com/kuzzleio/kuzzle/blob/master/lib/config/default.config.js

Shiranuit:
> Missing default value in https://github.com/kuzzleio/kuzzle/blob/master/lib/config/default.config.js

I could add a default value to 0, but since the config field is optional I dont know if it makes sense

alexandrebouthinon (approved):
You're right, It's ok for me

PR #63, opened 2015-11-04
Kuz 257 authorization

benoitvidis:
First authorization layer implementation.

Missing: 
* Some functional tests
* Some documentation
* The permissions definition only allows some boolean value for the time being. Using a closure still to be implemented.

NB: This PR introduces a new package dependency. The container image will need to be updated once it is merged.

PR #2253, opened 2022-01-13
Support M1 chip

clementbolin:
Depend [Support M1 chip #63](https://github.com/kuzzleio/kuzzle-containers/pull/63)

## What does this PR do ?

Support Kuzzle core-dev for M1 chip.
I added a `kuzzle_installer` service, in order to install the dependencies directly in the container without getting any error

docker-compose up


docker/scripts/run-dev.sh
```diff
@@ -13,7 +13,7 @@ then
 fi
 
 if [ -n "$TRAVIS" ] || [ -n "$REBUILD" ]; then
-    npm ci --unsafe-perm
+    # npm ci --unsafe-perm
```
Aschen, line 16: Leftover?

Aschen (changes requested):
With this PR, dependencies have to be rebuilt each time we run the Docker Compose stack.

Aschen (changes requested):
Can you add a section in the README about m1 chip and installing dependencies from inside the container?

xbill82 (approved):

alexandrebouthinon (approved):

PR #2260, opened 2022-01-26
Document checkToken behaviour when no token is given

Shiranuit:
## What does this PR do ?

Document checkToken behaviour when no token is given

### DIFF
--- a/doc/2/api/controllers/auth/check-token/index.md
+++ b/doc/2/api/controllers/auth/check-token/index.md
@@ -41,7 +41,7 @@ Body:
 ---
 
 <SinceBadge version="2.16.8">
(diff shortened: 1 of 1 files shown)


doc/2/api/controllers/auth/check-token/index.md
```diff
@@ -40,6 +40,9 @@ Body:
 
 ---
 
+<SinceBadge version="2.16.8">
+When no token is provided the method returns information about the anonymous token, instead of throwing an error.
```
Aschen, line 44:
```suggestion
When no token is provided the method returns information about the anonymous token (`kuid` is `-1`), instead of throwing an error.
```

Aschen (approved):

PR #2263, opened 2022-02-02
Support HTTP streams

Shiranuit:
⚠️ #2266 Should be merge in this PR before merging this PR into 2-dev

## What does this PR do ?

This PR add the capability to send stream of data through the HTTP protocol,
this way we can later implement download and data processing on the fly without storing everything in RAM.

From now on returning an instance of HttpStream as a result of a controller action will make Kuzzle send data
through HTTP chunk by chunk.

Kuzzle supports streaming with a fixed and dynamic size.
Fixed size can be achieved by setting the `totalBytes` property of the `HttpStream` instance

### Examples

```js
const fs = require('fs');
async myAction (request) {
 const readStream = fs.createReadStream('./Photos.zip');
 return new HttpStream(readStream);
}
```

```js
const { PassThrough } = require('stream');
async myAction (request) {
 const stream = new PassThrough();
 setInterval(() => {
   // Process data
   stream.write('data');
  if (done) {
    stream.end();
  }
 }, 50):
 return new HttpStream(stream);
}
```

`npm run test` 

For now it's pretty hard to test this feature functionally without having any controller action using it.

### To Do
- [x] Unit Tests
- [x] Documentation


doc/2/guides/advanced/streams/index.md
```diff
@@ -0,0 +1,57 @@
+---
+code: false
+type: page
+title: Stream large volumes
+description: Create streams to send large amount of data over the network
+order: 300
+---
+
+# HTTP Streams <SinceBadge version="auto-version" />
```
Aschen, line 9: This section should rather be in the **Develop On Kuzzle** > **API Controller** section (We have a [Response Format](https://docs.kuzzle.io/core/2/guides/develop-on-kuzzle/api-controllers/#response-format) section)

doc/2/guides/advanced/streams/index.md
```diff
@@ -0,0 +1,57 @@
...
+   // Process data
+   stream.write(/* Processed data */);
+  if (done) {
+    stream.end(); // Close the stream
+  }
+ }, 50):
+
+ return new HttpStream(stream);
+}
+```
```
Aschen, line 46: IMHO this example is confusing, maybe you could replace it with a stream of CSV lines from SearchResult pagination request?

lib/api/request/kuzzleRequest.ts
```diff
@@ -246,6 +246,10 @@ export class KuzzleRequest {
       throw new InternalError('cannot set an error as a request\'s response');
     }
 
+    if (this.context.connection.protocol !== 'http' && result instanceof HttpStream) {
+      throw new InternalError('cannot set an HttpStream as a request\'s response in a non-http protocol');
```
Aschen, line 250: You should define a new error in the error manager rather than throwing a raw error

lib/core/network/protocols/httpwsProtocol.js
```diff
@@ -636,6 +623,160 @@ class HttpWsProtocol extends Protocol {
...
+   * Send stream data thorugh the HTTP response
+   * @param {HttpRequest} request
+   * @param {HttpStream} httpStream - The data stream to send
+   * @param {uWS.HttpResponse} response
+   * @param {HttpMessage} message
+   */
+  httpSendStream (request, response, httpStream, message) {
+    const streamSizeFixed = typeof httpStream.totalBytes === 'number' && httpStream.totalBytes > 0;
+
+    if (httpStream.stream._readableState.errored
```
Aschen, line 669: Nit
```suggestion
    if ( httpStream.stream._readableState.errored
```

lib/core/network/protocols/httpwsProtocol.js
```diff
@@ -636,6 +623,160 @@ class HttpWsProtocol extends Protocol {
...
+    });
+
+    
+    httpStream.stream.on('end', () => {
+      response.end();
+      httpStream.stream.destroy();
+    });
+    
+
+    httpStream.stream.on('error', err => {
```
Aschen, line 768: What happen to the response if the user stream is destroyed? Shouldn't we `end` it with an error?

Aschen (changes requested):
:tada:

lib/core/network/protocols/httpwsProtocol.js
Shiranuit: Yes you are right I forgot to end the stream in case of stream failure

doc/2/guides/advanced/streams/index.md
Shiranuit: I agree, I will replace it with a better example once I get the time

doc/2/framework/classes/http-stream/properties/index.md
```diff
@@ -0,0 +1,25 @@
...
+
+# Properties
+
+## `stream`
+
+This property is an instance of the [stream.Readable](https://nodejs.org/docs/latest-v14.x/api/stream.html#stream_class_stream_readable) class from NodeJS.
+
+| Type                                                                         | Description                |
+|------------------------------------------------------------------------------|----------------------------|
+| <pre>[BackendCluster](/core/2/framework/classes/backend-cluster)</pre> | BackendCluster instance |
```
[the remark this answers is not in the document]
Shiranuit: Oops 😅 

And yes stream and totalBytes are readonly properties that can only be set at construction

lib/core/network/protocols/httpwsProtocol.js
```diff
@@ -657,13 +657,18 @@ class HttpWsProtocol extends Protocol {
...
    * @param {HttpStream} httpStream - The data stream to send
    * @param {uWS.HttpResponse} response
+   * @param {HttpMessage} message
    */
-  httpSendStream(request, response, httpStream, message) {
+  httpSendStream (request, response, httpStream, message) {
     const streamSizeFixed = typeof httpStream.totalBytes === 'number' && httpStream.totalBytes > 0;
 
     if (httpStream.stream._readableState.errored || httpStream.stream._readableState.closed) {
+      response.writeStatus('200 OK');
```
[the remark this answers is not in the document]
Shiranuit: That's right I totally forgot to change it  😅

doc/2/guides/advanced/streams/index.md
Shiranuit: I think this section might be useful here too if someone stumble uppon this page, what do you think ?
should we duplicate this section or juste move it ?

Aschen: IMHO it's not a very common usage to manipulate HTTP streams.  Streams are more a subset of what you can do when defining API action than an entire feature on it's own.  
About duplication, it's generally not a good idea I think

doc/2/guides/advanced/streams/index.md
```diff
@@ -0,0 +1,57 @@
...
+
+Read a file from the disk and send it.
+```js
+const fs = require('fs');
+
+async myAction (request) {
+ const readStream = fs.createReadStream('./Document.zip');
```
Aschen, line 53: Too much Windows for you @Shiranuit

Aschen (approved):
Read my comment about the documentation, apart from that it's really fine work :+1: 

(Also I think an example with sending documents as CSV from search result could be a better example)

Shiranuit:
@Aschen 
> (Also I think an example with sending documents as CSV from search result could be a better example)

After implementing it in document:export I'm not sure that this is a good example, it's not that easy to do it properly and I do not want to provide a bad example that could lead to bad usage of this.

Shiranuit:
As request I moved the documentation and I added the following warning
```
::: warning
Chunks are send through the HTTP Protocol each time a chunk is emitted through the `data` event of the given stream.
It's up to you to implement a buffer mechanism to avoid sending too many small consecutive chunks through the network.

Sending too many small chunks instead of bigger chunks will increase the number of syscall made to the TCP Socket and might decrease performance and throughput.
:::
```

doc/2/guides/develop-on-kuzzle/api-controllers/index.md
```diff
@@ -585,6 +585,45 @@ app.controller.register('redirect', {
...
+and not have everything stored in ram.
+
+::: warning
+Chunks are send through the HTTP Protocol each time a chunk is emitted through the `data` event of the given stream.
+It's up to you to implement a buffer mechanism to avoid sending too many small consecutive chunks through the network.
+
+Sending too many small chunks instead of bigger chunks will increase the number of syscall made to the TCP Socket and might decrease performance and throughput.
+:::
+
+## Usage
```
Aschen, line 609: Wrong heading section.

Can you use the same format as other examples? 
image

Aschen (changes requested):

doc/2/guides/develop-on-kuzzle/api-controllers/index.md
Shiranuit: @Aschen Yes I fixed the documentation in #2266

Shiranuit: I copied back the changes made to the documentation from #2266

Aschen (approved):

PR #2264, opened 2022-02-03
Change default value for `loginPerSecond` config

Aschen:
## What does this PR do ?

If the limit is 1 then it's not possible to use Kourou to retrieve a token with `kourou auth:login`

### DIFF
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -87,7 +87,7 @@
     "concurrentRequests": 50,
     "documentsFetchCount": 10000,
     "documentsWriteCount": 200,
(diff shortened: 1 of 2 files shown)


Shiranuit:
Why can't you login with kourou when the limit is 1, does Kourou attempt to authenticate twice ?

Aschen:
> Why can't you login with kourou when the limit is 1, does Kourou attempt to authenticate twice ?

If you run `kourou auth:login ...` then Kourou log itself a first time with provided username and password and then another time because that is the executed command.

But I just realized that I can simply not use a loggued Kourou to execute `auth:login` :facepalm:

PR #2265, opened 2022-02-03
Fix mExists action when using the HTTP protocol

matissepage:
## What does this PR do ?

This PR add functional test for the mExists API action

`npm run build`
`docker-compose up`
`npm run test:functional`

### Boyscout

change the definition of the route verb mExist (`get` to `post`).


features/step_definitions/documents-steps.js
```diff
@@ -122,6 +122,21 @@ Then(/The document "(.*?)" should( not)? exist/, async function (id, not) {
   }
 });
 
+Then(/The document "(.*?)" should( not)? multiple exist:$/, async function (id, not) {
```
Aschen, line 125: If you are adding a step using `mExists` then the step should be able to check multiple documents existence.  

Otherwise you can also use the [generic step to execute and API action](https://github.com/kuzzleio/kuzzle/blob/master/features/step_definitions/controllers-steps.js#L30) and then check the result with [another generic step checking array result](https://github.com/kuzzleio/kuzzle/blob/master/features/step_definitions/controllers-steps.js#L64

features/step_definitions/documents-steps.js
```diff
@@ -122,6 +122,21 @@ Then(/The document "(.*?)" should( not)? exist/, async function (id, not) {
...
+  const mExists = await this.sdk.document.mExists(
```
Aschen, line 126: I don't think the `document.mExists` method is implemented in the SDK, no?

features/DocumentController.feature
```diff
@@ -194,6 +194,21 @@ Feature: Document Controller
 
   # document:exists ============================================================
 
+  @mappings
+  Scenario: Check document existence
```
Aschen, line 198: A better test scenario would be one where at least 4 documents are checked in the same time with 2 existing and 2 not existing so you are testing both cases with more than 1 documents (which is the purpose of mExists, test documents by batch)

Aschen (changes requested):

Aschen (changes requested):
You tests case are perfect but you are using the `mGet` action, I think I understood your struggle here.

You can not use the predefined steps because it's using the SDK and the `document.mExists` method does not exists yet in the SDK.

You should modify the corresponding step and make it use the `sdk.query` method instead so it will supports `mExists` as well: https://github.com/kuzzleio/kuzzle/blob/master/features/step_definitions/documents-steps.js#L126

Aschen:
@matissepage Any update on this?

features/DocumentController.feature
```diff
@@ -262,15 +251,14 @@ Feature: Document Controller
       | "document-1" | { "name": "document1" } |
       | "document-2" | { "name": "document2" } |
       | "document-3" | { "name": "document3" } |
-    When I "get" the following document ids:
+    When I execute the "Exists" action on the following document ids:
```
Aschen, line 254: This should raise an error since the `document:Exists` action does not exists :thinking: 

Also, you don't need a new step, you can use this one 
```
    When I successfully execute the action "document":"mExits" with args:
      | body.ids[0]     | "id1"                      |
      | body.ids[2]     | "id1"                      |
      | body.ids[3]     | "id1"                      |
      | body.ids[4]     | "id1"                      |
```

Aschen (changes requested):

features/DocumentController.feature
clementbolin: If you look in [document-steps.js](https://github.com/kuzzleio/kuzzle/pull/2265/files#diff-9f541c6eda56ab57543db808194ae15c2b9f448967279667815b2eed019e604bR67), I used the `sdk.query()` method.

Aschen: Ah I saw, the first line transform it into `mExists` so it was misleading to me.

Actually you don't need an extra step, you can just use a standard one. We should be careful to not add too many different steps because it increase maintainability

Aschen (approved):

alexandrebouthinon (approved):

alexandrebouthinon:
I updated the title and label for this PR since we found a bug on this API action when we called it using the HTTP protocol

Issue #1659, opened 2020-06-18
As an end user, I'd like to have my request timeout when internal request to service take too long


berthieresteban:
Internal core modules makes requests to external services (ES and Redis), if the request take too long, the original request must be responded with an ExternalServiceError

Timeout for each services must be configurable

Aschen:
Rajouter à kuzzle.ask

Shiranuit:
Today it's very rare when requests to our external services do timeout, the Javascript SDK do support adding a timeout to a request to avoid waiting too long.

PR #2266, opened 2022-02-04
Add API action to export documents in CSV or JSONL format

Shiranuit:
:warning: Depends on #2263 
## What does this PR do ?

This PR adds a new document action `document:export`, this action will help the export of search queries in multiple format (`jsonl`, `csv`, ...).
This action use the new stream mechanism to allow browser to download exported documents directly from the backend without exploding the Ram and CPU usage.

`npm run test`

### To Do

- [x] Documentation
- [x] Functionnal Tests
- [x] ~Wait for cluster status to be green on cluster timeout during scroll~ (See #1659)
- [x] Reduce syscall by buffering lines before sending it as a chunk through the network


lib/api/controllers/documentController.js
```diff
@@ -173,6 +176,40 @@ class DocumentController extends NativeController {
     };
   }
 
+  async export (request) {
+    const { index, collection } = request.getIndexAndCollection();
+    const batchSize = request.getInteger('batchSize', 2000);
+    const query = request.getBody({});
+    const format = request.getString('format', 'jsonl');
+    const fields = request.getArray('fields', []);
```
Aschen, line 184: This should be a body array to work with http (it's the same subject as for targets list in HTTP)
For this action it's more important to have something working with HTTP since it's the targeted protocol

lib/types/HttpStream.ts
```diff
@@ -42,5 +43,45 @@ export class HttpStream {
...
     this.stream = readableStream;
     this.totalBytes = totalBytes;
+    this.destroyed = readableStream.destroyed;
+  }
+
+  /**
+   * Returns if the stream is errored
+   */
+  isErrored (): boolean {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
Aschen, line 53: I think we should disable this eslint rule since it's prevent use to disable typescript when needed and we end up with 2 lines of comments

lib/types/HttpStream.ts
```diff
@@ -42,5 +43,45 @@ export class HttpStream {
   ) {
     this.stream = readableStream;
     this.totalBytes = totalBytes;
+    this.destroyed = readableStream.destroyed;
+  }
+
+  /**
+   * Returns if the stream is errored
+   */
+  isErrored (): boolean {
```
Aschen, line 52: Nitpicking / coding style : you could use a getter named `errored` here for example

lib/types/HttpStream.ts
```diff
@@ -42,5 +43,45 @@ export class HttpStream {
...
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    return this.stream._readableState.errored !== null;
+  }
+
+  /**
+   * Get the error
+   */
+  getError (): Error {
```
Aschen, line 61: Same comment about getter usage

lib/types/HttpStream.ts
```diff
@@ -42,5 +43,45 @@ export class HttpStream {
...
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    return this.stream._readableState.errored;
+  }
+
+  /**
+   * Returns if the stream has been destroyed
+   */
+  isDestroyed (): boolean {
```
Aschen, line 70: Same comment about getter usage

lib/util/dump-collection.ts
```diff
@@ -0,0 +1,337 @@
...
+/**
+ * Formats the value for correct CSV output, avoiding to return
+ * values that would badly serialize in CSV.
+ *
+ * @param value The value to format
+ * @returns The value or a string telling the value is not scalar
+ */
+function formatValueForCSV (value: any) {
+  if (isObject(value)) {
+    return '[NOT_SCALAR]';
```
Aschen, line 75: Maybe we could replace this by `[OBJECT] ` because this term is more common than scalar. Especially because users that will manipulate the CSV will not always be developers

lib/util/dump-collection.ts
```diff
@@ -0,0 +1,337 @@
...
+
+  constructor (
+    protected readonly index: string,
+    protected readonly collection: string,
+    protected readonly batchSize: number,
+    protected readonly query: any = {},
+    protected readonly writeStream: stream.Writable,
+  ) {
+    this.options = {
+      scroll: '2s',
```
Aschen, line 95: Will it only took 2sec to write the n documents from the batch in an http stream? This value was here because writing the filesystem is quite predictable but here I think we should have a bigger value, what do you think?

lib/util/dump-collection.ts
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    try {
+      return await global.kuzzle.ask(
+        'core:storage:public:document:scroll',
+        scrollId,
+        { scrollTTL: this.options.scroll }
+      );
+    }
+    catch {
```
Aschen, line 149: What if the error is a timeout from the ES cluster for example?

lib/util/dump-collection.ts
```diff
@@ -0,0 +1,337 @@
...
+  }
+}
+
+class JSONLDumper extends AbstractDumper {
+  protected ndjsonStream = ndjson.stringify();
+
+  async setup () {
+    this.ndjsonStream.on('data', (line: string) => {
+      if (! this.writeStream) {
+        throw new Error('Cannot write data: WriteStream is not initialized.');
```
Aschen, line 204: Error should be handled through the error manager to have associated error code

lib/util/dump-collection.ts
```diff
@@ -0,0 +1,337 @@
...
+      }
+      this.writeStream.write(line);
+    });
+    return;
```
Aschen, line 208: Useless return

lib/util/dump-collection.ts
```diff
@@ -0,0 +1,337 @@
...
+      collection: this.collection,
+      index: this.index,
+      type: 'collection',
+    });
+  }
+
+  writeLine (content: any): Promise<void> {
+    return new Promise(resolve => {
+      if (this.ndjsonStream.write(content)) {
+        resolve(undefined);
```
Aschen, line 222:
```suggestion
        resolve();
```

lib/util/dump-collection.ts
```diff
@@ -0,0 +1,337 @@
...
+      body: document._source,
+    });
+  }
+
+  protected get fileExtension () {
+    return 'jsonl';
+  }
+}
+
+class KuzzleDumper extends JSONLDumper {
```
Aschen, line 242: This dumper cannot be used with streams and should be removed

lib/util/dump-collection.ts
```diff
@@ -0,0 +1,337 @@
...
+      // If no field has been selected, then all fields are selected.
+      const mappings = await global.kuzzle.ask(
+        'core:storage:public:mappings:get',
+        this.index,
+        this.collection
+      );
+      if (! mappings.properties) {
+        return;
+      }
+      this.fields = Object.keys(flattenObject(mappings.properties));
```
Aschen, line 295: Nested objects will have the `properties` from es mappings in their path

Aschen (changes requested):
I think we should use a CSV library (which is mandatory when parsing but maybe we can handle writing ourselves) or at least handle the case when either a property (rare case but can happen) or a value contains the separator.

Also I think we should let the user choose the separator like many other CSV exports do (with a default to ",")

lib/api/controllers/documentController.js
```diff
@@ -194,14 +194,14 @@ class DocumentController extends NativeController {
...
-    let mimeType = 'application/json';
+    let mimeType = 'html/text';
 
     switch (format) {
       case 'csv':
         mimeType = 'text/csv';
         break;
       default:
-        mimeType = 'application/json';
+        mimeType = 'html/text';
```
Aschen: Why this type?

Shiranuit: Because JSONL is not really a JSON file and could not be parsed like so, if it's defined as `application/json` browser and other HTTP libraries might want to parse it as JSON and will fail

doc/2/api/controllers/document/export/index.md
```diff
@@ -0,0 +1,128 @@
+---
+code: true
+type: page
+title: export
+---
+
+# export <SinceBadge version="auto-version"/>
```
Aschen, line 7: Usually we put the badge under the title. Also I think this will mess up with the anchor if you put it inside title

```suggestion
# export

<SinceBadge version="auto-version"/>
```

doc/2/api/controllers/document/export/index.md
```diff
@@ -0,0 +1,128 @@
...
+## Arguments
+
+- `collection`: collection name
+- `index`: index name
+
+### Optionnal:
+
+- `separator`: This option is only supported for the `CSV` format, it defines which character sequence will be used to format the CSV documents
+- `scroll`: creates a forward-only result cursor. This option must be set with a [time duration](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/common-options.html#time-units), at the end of which the cursor is destroyed. If set, a cursor identifier named `scrollId` is returned in the results. This cursor can then be moved forward using the [scroll](/core/2/api/controllers/document/scroll) API action
+- `size`: set the maximum number of documents returned per result page. By default it's `10000`.
```
Aschen, line 103: Is it still used since we don't have pagination here?

doc/2/api/errors/error-codes/api/index.md
```diff
@@ -31,6 +31,7 @@ description: Error codes definitions
 | api.assert.koncorde_restricted_keyword<br/><pre>0x0201000d</pre>  | [BadRequestError](/core/2/api/errors/error-codes#badrequesterror) <pre>(400)</pre> | The %s "%s" of Koncorde DSL is not supported for search queries. | A restricted keyword has been provided in filters |
 | api.assert.koncorde_dsl_error<br/><pre>0x0201000e</pre>  | [BadRequestError](/core/2/api/errors/error-codes#badrequesterror) <pre>(400)</pre> | %s | Provided filters contains an error |
 | api.assert.forbidden_stream<br/><pre>0x0201000f</pre>  | [InternalError](/core/2/api/errors/error-codes#internalerror) <pre>(500)</pre> | Cannot set an HttpStream as a request's response in a non-http protocol | A forbidden HttpStream has been provided in a response |
+| api.assert.unsupported_protocol<br/><pre>0x02010010</pre>  | [InternalError](/core/2/api/errors/error-codes#internalerror) <pre>(500)</pre> | The protocol "%s" is not supported. | A request has been made with an unsupported protocol |
```
Aschen, line 34: Can you add the name of the API for which the protocol is not supported in the error message? It will help developer to debug their application

Aschen (changes requested):

doc/2/api/controllers/document/export/index.md
Shiranuit: Yes it use internally, I will add a small paragraph that talks about thoses parameters and explains that are they used for.

lib/util/dump-collection.ts
Shiranuit: As discussed, this should be implemented globally rather than in this PR, see https://github.com/kuzzleio/kuzzle/issues/1659

doc/2/api/controllers/document/export/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+This method behaves like a `document:search` except that it scrolls and formats the searched results in one of the supported format (`csv`, `jsonl`) then returns everything as an HTTP Stream,
+that can either be downloaded directly from a browser or any scripts.
+
+This method also supports the [Koncorde Filters DSL](/core/2/api/koncorde-filters-syntax) to match documents by passing the `lang` argument with the value `koncorde`.
+Koncorde filters will be translated into an Elasticsearch query.
+
+::: info
+If you are struggling to retrieve all your documents, try tweaking the [size](#optional) and [scroll](#optional) parameters.
+Thoses parameters affects how many documents the internal scroll retrieves and processes per page and how long the scroll will exists.
```
Aschen, line 21: Suggestion: explain how they are used
```suggestion
The `scroll` parameter represents the maximum time needed for the client to download a page of `size` results.  
You should try with smaller pages of results if you experienced download problems.
```

lib/api/controllers/documentController.js
```diff
@@ -173,6 +176,49 @@ class DocumentController extends NativeController {
...
+      throw kerror.get('api', 'assert', 'unsupported_protocol', request.context.connection.protocol, 'document:export');
+    }
+
+    if (lang === 'koncorde') {
+      searchBody.query = await this.translateKoncorde(searchBody.query || {});
+    }
+
+    const stream = new BufferedPassThrough({ highWaterMark: 16384 });
+
+    dumpCollectionData(
```
Aschen, line 197: I think it would make more sense if the `dumpCollectionData` would returns the stream. Also it could directly returns an HttpStream

lib/util/bufferedPassThrough.ts
```diff
@@ -0,0 +1,186 @@
...
+      const remainingBufferSize = this.bufferSize - this.offset;
+
+      // If the remaining bytes in the chunk exceed the remaining bytes in the internal buffer
+      if (remainingChunkSize >= remainingBufferSize) {
+        // Write as much as the remaining bytes as possible to the buffer  from t
+        chunkOffset += this.writeToBuffer(chunk, chunkOffset, chunkOffset + remainingBufferSize, encoding);
+
+        // Sends the whole buffer to the stream since it is full
+        if (! this.push(this.buffer)) {
+          await new Promise((res) => {
```
Aschen, line 102: We should add the [arrow-parens](https://eslint.org/docs/rules/arrow-parens) rule to eslint for this one

```suggestion
          await new Promise(res => {
```

lib/util/dump-collection.ts
```diff
@@ -0,0 +1,327 @@
...
+    });
+  }
+
+  onResult (document: { _id: string; _source: any }): Promise<void> {
+    const values = [document._id, ...pickValues(document._source, this.fields)];
+    return this.writeLine(values.join(this.options.separator));
+  }
+}
+
+export async function dumpCollectionData (writableStream: stream.Writable, index: string, collection: string, query: any = {}, format = 'jsonl', fields: string[] = [], options: JSONObject = {}) {
```
Aschen, line 315: We should try to avoid abstract terms as "data" 
```suggestion
export async function dumpCollectionDocument (writableStream: stream.Writable, index: string, collection: string, query: any = {}, format = 'jsonl', fields: string[] = [], options: JSONObject = {}) {
```

Aschen (changes requested):

lib/api/controllers/documentController.js
```diff
@@ -173,6 +174,45 @@ class DocumentController extends NativeController {
...
+
+    return dumpCollectionDocument(
+      index,
+      collection,
+      searchBody,
+      format,
+      fields,
+      {
+        lang,
+        scroll: scrollTTL || '1m',
```
Aschen, line 210: Since the default size is `10` documents per page, I think we can safely have a smaller value like `5s` for the scroll

lib/util/dump-collection.ts
```diff
@@ -0,0 +1,334 @@
...
+    });
+  }
+
+  onResult (document: { _id: string; _source: any }): Promise<void> {
+    const values = [document._id, ...pickValues(document._source, this.fields)];
+    return this.writeLine(values.join(this.options.separator));
+  }
+}
+
+export function dumpCollectionDocument (index: string, collection: string, query: any = {}, format = 'jsonl', fields: string[] = [], options: JSONObject = {}): HttpStream {
```
Aschen, line 317: Nitpicking: this should be pluralized
```suggestion
export function dumpCollectionDocuments (index: string, collection: string, query: any = {}, format = 'jsonl', fields: string[] = [], options: JSONObject = {}): HttpStream {
```

Aschen (approved):
Lats nitpickings

Shiranuit:
I added new test cases for the CSV export
As request by @etrousset I added the ability to map a field path to a name for the CSV export

lib/api/controllers/documentController.js
```diff
@@ -181,6 +181,12 @@ class DocumentController extends NativeController {
     const fields = request.getBodyArray('fields', []);
     const lang = request.getLangParam();
     const separator = request.getString('separator', ',');
+    const fieldsName = request.getBodyObject('fieldsName', {});
+
+    // Remove "fields" and "fieldsName" from searchBody to avoid ES throwing an error
+    // since those properties are not allowed in the searchBody
+    delete searchBody.fields;
+    delete searchBody.fieldsName;
```
Aschen, line 189:
```suggestion
    searchBody.fields = undefined;
    searchBody.fieldsName = undefined;
```

doc/2/api/controllers/document/export/index.md
```diff
@@ -92,7 +92,10 @@ Method: GET
   "size": <page size>,
   "scroll": "<scroll duration>",
   "lang": "<query language>",
-  "format": "<export format>"
+  "format": "<export format>",
+  "fieldsName": {
```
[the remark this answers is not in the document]
Shiranuit: Otherwise if we wanted this to work within the query string we should take an array and do some manual splitting, wdyt  ?

PR #2281, opened 2022-03-01
Support Fermium, Gallium and 18 Node.js

alexandrebouthinon:
## What does this PR do ?

Update our processes to match the [newly adopted Node.js LTS version support decisions](https://docs.google.com/document/d/1KNd0Hmll8_7nGXVGWl4xubHbzD_E9YuzSwlZ_DOgDnw/edit#).

- [x] Update Github Actions Workflows
- [x] Update Docker base images
- [x] Build and deploy Docker image for both amd64 and arm64
- [x] Rework CI pipeline to make it one file using condition

DISCLAIMER: This is normal all the deployment jobs (Docker, Doc) can be seen in the Pull request workflow. This a weird way Github display them even if the running condition is `false`

Check CI/CI pipelines


Aschen (changes requested):
You also have to modify the scripts for functional tests, they are using the `NODE_VERSION` and `NODE_12_VERSION` env variables.

Example https://github.com/kuzzleio/kuzzle/blob/master/.ci%2Fscripts%2Frun-test-cluster.sh

alexandrebouthinon:
Yep @Aschen, I'm publishing a version of the `kuzzleio/core-dev` (even before the 19th April) to start using NVM instead of `n` in tests scripts

alexandrebouthinon:
This PR is completely bugged caused by the March 17th Github outage. I'm going to create a new one

PR #2290, opened 2022-03-18
Update Node.js LTS support policy

alexandrebouthinon:
## What does this PR do ?

Update our processes to match the newly adopted Node.js LTS version support decisions.

Kuzzle will follow the same LTS policy as Node.js and thus always support 3 version at the same time.

The 12 April 2022, Kuzzle will support only the following versions:
 - Node.js 14
 - Node.js 16
 - Node.js 18

- [x] Update Github Actions Workflows
- [x] Update Docker base images
- [x] Build and deploy Docker image for both amd64 and arm64
- [x] Rework CI pipeline to make it one file using condition

DISCLAIMER: This is normal all the deployment jobs (Docker, Doc) can be seen in the Pull request workflow. This a weird way Github display them even if the running condition is `false`

Check CI/CI pipelines


.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,252 @@
...
+        node-version: ${{ fromJson(needs.prepare-matrix.outputs.matrix).node-version }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/install-packages
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+      - uses: ./.github/actions/cache-node-modules
+        with:
+          NODE_VERSION: ${{ matrix.node-version }}
```
Aschen, line 77: Is it normal to have uppercase and lowercase?

.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,252 @@
...
+        with:
+          KUZZLE_IMAGE: ${{ matrix.kuzzle-image }}
+
+
+  functional-tests:
+    name: Functional Tests - ${{ matrix.test_set }} - Node.js v${{ matrix.node-version }}
+    needs: [unit-tests, prepare-matrix]
+    strategy:
+      matrix:
+        test_set: [http, websocket, "legacy:mqtt", "legacy:http", "legacy:websocket"]
```
Aschen, line 125: Mix of strings with `"` and without

alexandrebouthinon: This is normal, Github Actions try to interpret the `:` as a YAML property, I don't know really why 😓

.github/workflows/workflow.yaml
alexandrebouthinon: We could use `node-version` in our custom cache action but this is not the main goal of this PR. I can update it if it is a big concern for you 🙂

alexandrebouthinon: Updated 😉

Aschen (approved):

Shiranuit (approved):

docker/images/kuzzle/Dockerfile
```diff
@@ -1,20 +1,18 @@
 ################################################################################
 # Production build image
 ################################################################################
-FROM node:12.18.1-stretch-slim as builder
+FROM node:12-bullseye-slim as builder
```
Aschen, line 4: Shouldn't we build those images with newer version of node?

PR #2291, opened 2022-03-21
Improve cache to speed up CI

alexandrebouthinon:
# Why?
To improve and simplify CI jobs duration, I setup Yarn linker to benefit its native cache capabilities


Aschen (commented):
I don't see any reference to yarn?

Aschen (approved):

Issue #2293, opened 2022-03-22
Add upsertUser action


Aschen:
## Feature Description

There is no `createOrReplaceUser` action because replacing an user will invalidate current sessions or API keys.

Instead we could have an `upsertUser` action that perform a create or an update of an user.

This is similar to [document:upsert](https://docs.kuzzle.io/core/2/api/controllers/document/upsert/)

PR #2299, opened 2022-03-28
Add security:upsertUser action

ChillPC:
#2293 

## What does this PR do ?

Add the action `upsertUser` to the `securityController` controller which update a user or create if it does not exist.

  `npm test:unit`

### Other changes

`npm clean` before running the linter with `npm run test` to avoid js code generated by ts.


clementbolin (changes requested):

clementbolin:
Do you think it is possible to write OpenAPI documentation for your new api action ? @Aschen

Aschen (approved):

Aschen:
> Do you think it is possible to write OpenAPI documentation for your new api action ? @Aschen

I agree, it's a good idea to write associated OpenAPI doc for every new action :+1:

doc/2/api/controllers/security/upsert-user/index.md
```diff
@@ -0,0 +1,126 @@
+---
+code: true
+type: page
+title: upsertUser
+---
+
+# upsertUser
```
clementbolin, line 7: You need to add the `<SinceBadge version="auto-version"/>` element to inform the user that this action is available since the last release. the `auto-version` field will be automatically changed during release.

clementbolin (changes requested):

lib/api/controllers/securityController.js
```diff
@@ -845,21 +845,19 @@ class SecurityController extends NativeController {
...
-          ...defaultValues,
-          ...content, // Order important, content erase default duplicates
-        };
-
-        return this._persistUser(request, profileIds, creatingContent);
-      }
-
-      throw error;
-    }
+    return this._changeUser(request, id, content, userId, profileIds)
```
Aschen, line 848: Why changing to promise chain? `async/await` should be preferred

Aschen (changes requested):

lib/api/controllers/securityController.js
ChillPC: Since the function would send a Promise, I thought that `then` would be preferred as it does not block waiting for the result (or error) of the promise

clementbolin (approved):

lib/api/controllers/securityController.js
Aschen: It will not change anything because the result of the `upsertUser` will not be available until the result of the `_changeUser` method is available (=resolved from the promise).

So you can await the promise resolution in this method with `await` or return a promise who will be awaited later by another method. It does not change the way any subsequent caller will be blocked when having to wait for the promise to be resolved.

clementbolin:
Does this feature miss the functional test ? @Aschen

Aschen:
> Does this feature miss the functional test ? @Aschen

Damn! You're right Bolin :facepalm:

Aschen (changes requested):
Missing functional test (thx @clementbolin )

Aschen (changes requested):

doc/2/api/controllers/security/update-user/index.md
```diff
@@ -68,6 +68,10 @@ Returns the update user kuid and version number.
   "requestId": "<unique request identifier>",
   "result": {
     "_id": "<kuid>",
+    "_source": {
+      "profileIds": ["<profileId>"],
```
Aschen, line 72: Nitpicking: since you are giving an example fullname below you can give an example profile here instead of a placeholder

features/SecurityController.feature
```diff
@@ -413,6 +413,27 @@ Feature: Security Controller
       | "test-user"  |
       | "test-user2" |
 
+  @security
+  Scenario: Upsert user
+    When I successfully execute the action "security":"upsertUser" with args:
+      | _id        | "user-test"           |
+      | body       | { "content": { "profileIds": ["default"]}, "default": { "name": "default-name", "credentials": { "local": { "username": "user-test", "password": "user-test" } } } } |
```
Aschen, line 420: Instead of a huge one lined body you could build your object property by property to have more readable step case
```
| body.content.profileIds | ["default"] |
```
etc

features/SecurityController.feature
```diff
@@ -413,6 +413,27 @@ Feature: Security Controller
...
+    Then I should receive a result matching:
+      | _id        | "user-test"           |
+      | _source    | { "profileIds": ["default"], "name": "default-name"} |
```
Aschen, line 423: Same comment about more readable tests cases (and for the steps below)

Aschen (changes requested):

doc/2/api/controllers/security/update-user/index.md
Aschen: Let's keep consistency then :+1:

lib/api/controllers/securityController.js
```diff
@@ -819,20 +820,46 @@ class SecurityController extends NativeController {
...
+   * If the user doesn't already exist, a new user is created.
+   *
+   * @param {Request} request
+   * @returns {Promise}
+   */
+  async upsertUser (request) {
+    const id = request.getId();
+    const content = request.getBodyObject('content');
+    const userId = request.getKuid();
+    const profileIds = get(content, 'profileIds');
```
Aschen, line 837: I just realized that this and l.840 to 846 are leftovers from a code before we had argument extractors in the request (`request.getBodyXXX`)

You should retrieve the `profileIds` array with `request.getBodyArray` method instead

Aschen (changes requested):

features/SecurityController.feature
ChillPC: Not supported by cucumber, sadly :(

ChillPC: Finally needs a rewrite of some test

features/support/world.js
```diff
@@ -40,37 +41,48 @@ class KuzzleWorld {
...
-      const rawObject = dataTable.rawTable[i];
+      const
+        object = {},
+        rawObject = dataTable.rawTable[i];
 
       for (let j = 0; j < keys.length; j++) {
         if (rawObject[j] !== '-') {
-          // eslint-disable-next-line no-eval
-          object[keys[j]] = eval(`const o = ${rawObject[j]}; o`);
+          _.set(object, keys[j], eval(`var o = ${rawObject[j]}; o`));
```
Aschen: Same comment, why keeping both lines? The result shouldn't be different no?

Aschen (approved):

PR #2303, opened 2022-04-01
Added strict and test

### LINKED ISSUE #2298: Support `strict` option for document:updateByQuery
## Feature Description

The `document:updateByQuery` action should support the `strict` option like the other `document:m*` actions

See https://docs.kuzzle.io/core/2/api/controllers/document/m-create-or-replace/#arguments

ChillPC:
## What does this PR do ?

Add strict argument for action document:updateByQuery

fix #2298

### DIFF
--- a/doc/2/api/controllers/document/update-by-query/index.md
+++ b/doc/2/api/controllers/document/update-by-query/index.md
@@ -93,6 +93,7 @@ kourou document:updateByQuery <index> <collection> <body> -a silent=true
 - `refresh`: if set to `wait_for`, Kuzzle will not respond until the update is indexed
 - `source`: if set to `true` Kuzzle will return the updated documents body in the response.
 - `lang`: specify the query language to use. By default, it's `elasticsearch` but `koncorde` can also be used. <SinceBadge version="2.8.0"/>
+- `strict`: if set, an error will occur if at least one document has not been created/replaced
 - `silent`: if set, then Kuzzle will not generate notifications <SinceBadge version="2.9.2" />
 
 ## Body properties
@@ -106,7 +107,7 @@ kourou document:updateByQuery <index> <collection> <body> -a silent=true
 
 Returns an object containing 2 arrays: `successes` and `errors`
 
-Each updated document is an object of the successes array with the following properties:
+Each updated document is an object of the `successes` array with the following properties:
 
 - `_id`: document unique identifier
 - `_source`: document content
@@ -118,6 +119,7 @@ Each errored document is an object of the `errors` array with the following prop
 - `status`: HTTP error status code
 - `reason`: human readable reason
 
+If `strict` mode is enabled, will rather return an error if at least one document has not been updated.
 ```js
 {
   "status": 200,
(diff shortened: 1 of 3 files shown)


test/api/controllers/documentController.test.js
```diff
@@ -1285,6 +1285,56 @@ describe('DocumentController', () => {
...
+        changes: {
+          bar: 'foo'
+        }
+      };
+      request.input.args.refresh = 'wait_for';
+      request.input.args.source = true;
+      request.input.args.strict = true;
+      request.context.user = { _id: 'aschen' };
+
+      const methodName = 'core:storage:public:document:updateByQuery';
```
clementbolin, line 1314: Same thing here, if you use the `methodName` variable why not also use it on the `1290` line?

clementbolin (changes requested):

doc/2/api/controllers/document/update-by-query/index.md
```diff
@@ -118,6 +119,7 @@ Each errored document is an object of the `errors` array with the following prop
 - `status`: HTTP error status code
 - `reason`: human readable reason
 
+If `strict` mode is enabled, will rather return an error if at least one document has not been updated.
```
clementbolin, line 122: You must add the `<SinceBadge version="auto-version"/>` to inform the user.

clementbolin (changes requested):

test/api/controllers/documentController.test.js
ChillPC: Was for debugging purpose I think, useless now

lib/api/controllers/documentController.js
```diff
@@ -767,25 +768,31 @@ class DocumentController extends NativeController {
...
     let query = request.getBodyObject('query');
     const changes = request.getBodyObject('changes');
+    const strict = request.getBoolean('strict');
     const silent = request.getBoolean('silent');
     const userId = request.getKuid();
     const refresh = request.getString('refresh', 'false');
     const source = request.getBoolean('source');
     const { index, collection } = request.getIndexAndCollection();
     const lang = request.getLangParam();
+    const methodName = 'core:storage:public:document:updateByQuery';
```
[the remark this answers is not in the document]
ChillPC: To avoid duplication of string in the ask (line `785`) and throw of error (line `793`). Everything is inspired from the function `_mChange` (line `838`) which take an argument `methodName`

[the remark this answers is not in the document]
clementbolin: Yes indeed, I did not have its use in the error my bad

clementbolin (approved):

lib/api/controllers/documentController.js
Aschen, line 778: Captain nitpicker here but it's considered a good practice (for readability) to split the code into logical blocks separated by spaces.  

For example, here the first part of the function is only about extracting arguments from the request, then we are modifying some of those arguments if needed and finally we are calling external services.

IMHO, the `methodName` declaration should be made in another "logical block", just before being use for the first name for example.

Also, it should rather `eventName` so you can even more simplify it with simply `event`

Aschen (approved):

lib/api/controllers/documentController.js
ChillPC: Separated by spaces or lines ?
I wanted to keep consistency with the whole code. Should I change `methodName` to `event` in all other methods ?

PR #2305, opened 2022-04-01
Fix typo in openapi for the document:get API action

ChillPC:


Aschen (changes requested):
This PR should be assigned to you and have the appropriate label before being merged.

Also, nitpicking on the title, it should avoid parenthesis and clearly state whats it's modifying (on this you are almost good :+1: )

Aschen:
Last nitpicking:
 - we use the "API action" term instead of "API Call", "method call", "api route", etc
 - we usually represents API action with `<controller>:<action>`

`Fix typo for openapi for the document:get API action`

Shiranuit (approved):

Aschen (approved):

Issue #2300, opened 2022-03-29
HTTP POST on /_query with empty json body returns HTTP 102 code causing HTTP Parsers to fail to parse response


Shiranuit:
## Current Behaviour

Making an HTTP request to /_query with an empty JSON body returns an HTTP reponse code `102` which causes HTTP Parsers to fail to parse the Kuzzle Response.

## Expected Behavior

Making HTTP requests to /_query with an empty JSON body should not return the HTTP response code 102 and should returns `200 OK` instead since there is no error.

PR #2308, opened 2022-04-05
HTTP POST on /_query with empty json body returns HTTP 102 code causing HTTP Parsers to fail to parse response

clementbolin:
#2300

## What does this PR do ?

return an error when you send an empty JSON to `/_query`

HTTP request to `POST` `localhost:7512/_query`

body: `{}`

### DIFF
--- a/.ci/test-cluster.yml
+++ b/.ci/test-cluster.yml
@@ -78,7 +78,7 @@ services:
     container_name: kuzzle_redis
 
   elasticsearch:
-    image: kuzzleio/elasticsearch:7.16.2
+    image: kuzzleio/elasticsearch:7
     container_name: kuzzle_elasticsearch
     ulimits:
       nofile: 65536
--- a/features/step_definitions/controllers-steps.js
+++ b/features/step_definitions/controllers-steps.js
@@ -197,7 +197,7 @@ Then('The raw response should match:', function (dataTable) {
   should(this.props.rawResponse).matchObject(expectedResult);
 });
 
-Then('I send a HTTP {string} request with:', async function (method, dataTable) {
+Then(/I send a (bad )?HTTP "(.*?)" request with:$/, async function (expectError, method, dataTable) {
   const body = this.parseObject(dataTable);
 
   const options = {
@@ -207,7 +207,7 @@ Then('I send a HTTP {string} request with:', async function (method, dataTable) 
     method,
     body: {
       jwt: this.sdk.jwt,
-      ...body
+      ...body,
     },
     headers: body.headers,
   };
@@ -222,7 +222,12 @@ Then('I send a HTTP {string} request with:', async function (method, dataTable) 
     this.props.rawResponse = response;
   }
   catch (err) {
-    this.props.error = err.error.error;
+    if (expectError) {
+      this.props.error = err.error.error;
+    }
+    else {
+      throw err;
+    }
   }
 });
 
--- a/features/step_definitions/controllers-steps.js
+++ b/features/step_definitions/controllers-steps.js
@@ -207,18 +207,23 @@ Then('I send a HTTP {string} request with:', async function (method, dataTable) 
     method,
     body: {
       jwt: this.sdk.jwt,
-      ...body,
+      ...body
     },
     headers: body.headers,
   };
 
-  delete body.headers;
+  body.headers = undefined;
 
-  const response = await requestPromise(options);
+  try {
+    const response = await requestPromise(options);
 
-  this.props.result = response.body.result;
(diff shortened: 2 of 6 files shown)


lib/core/network/router.js
```diff
@@ -189,7 +190,18 @@ class Router {
         cb(request);
       }
       else {
-        global.kuzzle.funnel.execute(mutatedRequest, (err, result) => cb(result));
+        global.kuzzle.funnel.execute(mutatedRequest, (err, result) => {
+          const _res = result || request;
+
+          if (err && ! _res.error) {
+            console.log('[Router] ERRRO R:', err);
```
OlivierCavadenti, line 197: Use global.kuzzle.log.error function instead ?

clementbolin: Yes, I was debugging and forgot to delete this console.log

.ci/test-cluster.yml
```diff
@@ -78,7 +78,7 @@ services:
     container_name: kuzzle_redis
 
   elasticsearch:
-    image: kuzzleio/elasticsearch:7.16.2
+    image: kuzzleio/elasticsearch:7
```
OlivierCavadenti, line 81: Why change the version of es here ?

package.json
```diff
@@ -35,7 +35,8 @@
     "test:lint:js:fix": "eslint --max-warnings=0 --fix ./lib ./test ./features-legacy ./bin ./features ./doc/build-error-codes.js",
     "doc-error-codes": "node -r ts-node/register doc/build-error-codes",
     "docker:install": "docker-compose run kuzzle_node_1 npm install",
-    "docker:test:unit": "docker-compose run kuzzle_node_1 npm run test:unit"
+    "docker:test:unit": "docker-compose run kuzzle_node_1 npm run test:unit",
+    "docker:npm": "docker-compose run kuzzle_node_1 npm run --"
```
OlivierCavadenti, line 39: why add this script in this PR ?

clementbolin: On Mac M1, you have to run the command inside the docker, so to be easier we add this script

.ci/test-cluster.yml
clementbolin: Now, we use this version, because it is built for amd64/arm64

features/step_definitions/controllers-steps.js
```diff
@@ -207,18 +207,23 @@ Then('I send a HTTP {string} request with:', async function (method, dataTable)
     method,
     body: {
       jwt: this.sdk.jwt,
-      ...body,
+      ...body
```
Aschen, line 210: It's nitpicking but the comma should actually stays here.  
It makes refactor more easier because you can cut/copy + paste any line anywhere

features/step_definitions/controllers-steps.js
```diff
@@ -207,18 +207,23 @@ Then('I send a HTTP {string} request with:', async function (method, dataTable)
...
 
-  this.props.result = response.body.result;
-  this.props.response = response.body;
-  this.props.rawResponse = response;
+    this.props.result = response.body.result;
+    this.props.response = response.body;
+    this.props.rawResponse = response;
+  }
+  catch (err) {
+    this.props.error = err.error.error;
```
Aschen, line 225: If you are silently catch the error then you have to modify the step so it can accept a "successfully" modifier.  
In this way, you can ensure the request is successful with ` I successfully send a HTTP "post" request with:`.  
Previous usage of the step were assuming success so they should be updated accordingly.

Example here https://github.com/kuzzleio/kuzzle/blob/2-dev/features/step_definitions/controllers-steps.js#L12

Aschen (changes requested):

features/step_definitions/controllers-steps.js
clementbolin: I fixed it, but instead of using `I send a (successfully ) ?HTTP {string} request with:` I replace the `successfully` by `bad`, that way I only need to change only my new test

Aschen (approved):

test/service/storage/elasticsearch.test.js
```diff
@@ -3536,7 +3536,7 @@ describe('Test: ElasticSearch service', () => {
       elasticsearch._mExecute = sinon.stub().resolves(mExecuteResult);
     });
 
-    it('should call _mExecute with formated documents', () => {
+    it('should call _mExecute with formated documents', async () => {
```
alexandrebouthinon, line 3539: Useless async keyword
```suggestion
    it('should call _mExecute with formated documents', () => {
```

alexandrebouthinon (approved):

Issue #2226, opened 2021-12-07
Cannot use auth.getCurrentUser with EmbeddedSDK


Aschen:
## Description

```js
  const kuzzle = this.context.accessors.sdk

  await kuzzle.auth.login('local', { username, password })
  const user = await kuzzle.auth.getCurrentUser()
```

Error `Caught an unexpected plugin error: Cannot read property 'userId' of null\nThis is probably not a Kuzzle error, but a problem with a plugin implementation.`

ChillPC:
Need to remove `auth` method throwing error and deprecate others

Aschen:
> Need to remove `auth` methods throwing error and deprecate others

See my comment on your PR, there is a way to forbid and deprecate those actions in Kuzzle code https://github.com/kuzzleio/kuzzle/pull/2312#pullrequestreview-936794621

PR #2312, opened 2022-04-08
Deprecate auth controller in the embeded sdk

ChillPC:
#2226 

## What does this PR do ?

Fix documentation for the embedded sdk with the auth controller.

Removing accessibility of functions from the embedded sdk would need rewrite in the [sdk-javascript](https://github.com/kuzzleio/sdk-javascript) because :
- The embedded sdk import `Kuzzle` from the sdk-javascript [here](https://github.com/kuzzleio/kuzzle/blob/48a8f38b0e73caa0840bf11a14cfe96ffb9e3be9/lib/core/shared/sdk/embeddedSdk.ts#L28)
- The sdk-javascript import all controllers in the [/src/Kuzzle.ts](https://github.com/kuzzleio/sdk-javascript/blob/1d262832395753a7b4f43d7de3b5766ce79bd2fa/src/Kuzzle.ts#L4)


Aschen (changes requested):
You can check which action is going to be executed in the [EmbeddedSDK.query](https://github.com/kuzzleio/kuzzle/blob/master/lib/core/shared/sdk/embeddedSdk.ts#L136) method and forbid selected actions from auth controller.

For the action that need to be deprecated to avoid breaking change, you can define a new type `EmbeddedAuth` inheriting from the SDK `AuthController` to add the `@deprecated` tags in the JS doc.
Look at how the [RealtimeController.subscribe](https://github.com/kuzzleio/kuzzle/blob/master/lib/core/shared/sdk/embeddedSdk.ts#L39) method documentation has been overloaded using this technique.

Last point, don't forget the `DeprecatedBadge` in the documentation

ChillPC:
> You can check which action is going to be executed in the [EmbeddedSDK.query](https://github.com/kuzzleio/kuzzle/blob/master/lib/core/shared/sdk/embeddedSdk.ts#L136) method and forbid selected actions from auth controller.

Finally, everything throw an error except login which send the token used by the embeddedSdk.

> For the action that need to be deprecated to avoid breaking change, you can define a new type `EmbeddedAuth` inheriting from the SDK `AuthController` to add the `@deprecated` tags in the JS doc. Look at how the [RealtimeController.subscribe](https://github.com/kuzzleio/kuzzle/blob/master/lib/core/shared/sdk/embeddedSdk.ts#L39) method documentation has been overloaded using this technique.

I don't see the need to create a new class just to overload the method if they will at the end be intercepted by the query method. For the documentation, everything new/different from the sdk is listed on the [embeddedSdk page](https://docs.kuzzle.io/core/2/guides/develop-on-kuzzle/embedded-sdk/). Hence the warning at the top of the section.

> Last point, don't forget the `DeprecatedBadge` in the documentation

Since there is not a whole section for the `authController` in the `embeddedSdk` page, I thought a simple warning should be enough

clementbolin (approved):

doc/2/guides/develop-on-kuzzle/embedded-sdk/index.md
```diff
@@ -39,6 +39,10 @@ The behavior of the [realtime:subscribe](/sdk/js/7/controllers/realtime) method
 Learn more about [Backend Realtime Subscriptions](/core/2/guides/develop-on-kuzzle/embedded-sdk#backend-realtime-subscriptions)
 :::
 
+::: warning
+The [auth controller](/sdk/js/7/controllers/auth/) is deprecated in the embedded sdk. [User impersonation](/core/2/guides/develop-on-kuzzle/embedded-sdk#user-impersonation) is preferred.
```
Aschen, line 43:
```suggestion
The [auth controller](/sdk/js/7/controllers/auth/) is deprecated in the embedded sdk. [User impersonation](/core/2/guides/develop-on-kuzzle/embedded-sdk#user-impersonation) or usage of the security controller is preferred.
```

lib/core/shared/sdk/embeddedSdk.ts
```diff
@@ -36,13 +37,39 @@ import ImpersonatedSDK from './impersonatedSdk';
 
 const contextError = kerror.wrap('plugin', 'context');
 
+const deprecatedForbiddenEmbeddedActions = {
+  'auth': new Set([
```
Aschen, line 41: Why using a Set here? A simple array will also works and it's more "portable"

lib/core/shared/sdk/embeddedSdk.ts
```diff
@@ -36,13 +37,39 @@ import ImpersonatedSDK from './impersonatedSdk';
 
 const contextError = kerror.wrap('plugin', 'context');
 
+const deprecatedForbiddenEmbeddedActions = {
```
Aschen, line 40: Naming: Same comment about deprecated and forbidden

doc/2/api/errors/error-codes/api/index.md
```diff
@@ -53,5 +53,6 @@ description: Error codes definitions
...
 | api.process.unauthorized_origin<br/><pre>0x0202000c</pre>  | [UnauthorizedError](/core/2/api/errors/error-codes#unauthorizederror) <pre>(401)</pre> | The origin "%s" is not authorized. | The domain reaching out to Kuzzle is not authorized |
 | api.process.too_many_logins_requests<br/><pre>0x0202000d</pre>  | [TooManyRequestsError](/core/2/api/errors/error-codes#toomanyrequestserror) <pre>(429)</pre> | Rejected: Too many login attempts per second | The request was denied because the maximum ("limits.loginsPerSecond") number of login attempts per second has been exceeded. |
+| api.process.deprecated_embedded_sdk_action<br/><pre>0x0202000e</pre>  | [BadRequestError](/core/2/api/errors/error-codes#badrequesterror) <pre>(400)</pre> | The action %s of the controller %s has been called while it is deprecated %s. | A deprecated sdk action has been called |
```
Aschen, line 56: Technically something deprecated is something you can use even if you shouldn't because it will be removed in a future version so it's weird to have something deprecated and forbidden

lib/core/shared/sdk/embeddedSdk.ts
```diff
@@ -134,6 +161,24 @@ export class EmbeddedSDK extends Kuzzle {
         : options.propagate;
     }
 
+    if (deprecatedForbiddenEmbeddedActions[request.controller] !== undefined
+      && deprecatedForbiddenEmbeddedActions[request.controller].has(request.action)
+    ) {
+      throw kerror.get(
```
Aschen, line 167: This errors should tell the user that those methods cannot be used with the EmbeddedSDK

Aschen (changes requested):
Missing unit tests

lib/core/shared/sdk/embeddedSdk.ts
ChillPC: Because it is an unordered collection of unique elements, thus a set. For portability, Set is a native javascript object

lib/core/shared/sdk/embeddedSdk.ts
```diff
@@ -154,7 +154,7 @@ export class EmbeddedSDK extends Kuzzle {
     options: { propagate?: boolean } = {}
   ): Promise<ResponsePayload> {
     // By default, do not propagate realtime notification accross cluster nodes
-    if ( isPlainObject(request)
+    if (isPlainObject(request)
```
Aschen, line 157: This was the good indentation so every condition start on the same column
```suggestion
    if ( isPlainObject(request)
```

lib/core/shared/sdk/embeddedSdk.ts
Aschen: Good point :+1: Since the initialization is not that different from native arrays it's ok to me

lib/core/shared/sdk/embeddedSdk.ts
```diff
@@ -136,6 +163,24 @@ export class EmbeddedSDK extends Kuzzle {
         : options.propagate;
     }
 
+    if (forbiddenEmbeddedActions[request.controller] !== undefined
```
Aschen, line 166: Nitpicking
```suggestion
    if ( forbiddenEmbeddedActions[request.controller] !== undefined
```

lib/core/shared/sdk/embeddedSdk.ts
```diff
@@ -136,6 +163,24 @@ export class EmbeddedSDK extends Kuzzle {
...
+      && forbiddenEmbeddedActions[request.controller].has(request.action)
+    ) {
+      throw kerror.get(
+        'api',
+        'process',
+        'forbidden_embedded_sdk_action',
+        request.action,
+        request.controller,
+        ', use user impersonation or security controller instead',
```
Aschen, line 175: Why having a custom sentence as a placeholder and not a static one since you are using it here only?

lib/kerror/codes/2-api.json
```diff
@@ -182,6 +182,12 @@
           "code": 13,
           "message": "Rejected: Too many login attempts per second",
           "class": "TooManyRequestsError"
+        },
+        "forbidden_embedded_sdk_action": {
+          "description": "A forbidden embdedded sdk action has been called",
```
Aschen, line 187: Nitpicking
```suggestion
          "description": "A forbidden EmbeddedSDK action has been called",
```

lib/kerror/codes/2-api.json
```diff
@@ -182,6 +182,12 @@
...
+          "code": 14,
+          "message": "The action %s of the controller %s has been called while it is forbidden in the embedded sdk%s.",
```
Aschen, line 189: Nitpicking: usually we represents controller action with `<controller>:<action>`
```suggestion
          "message": "The action "%s:%s" has been called while it is forbidden in the EmbeddedSDK%s.",
```

lib/kerror/codes/2-api.json
```diff
@@ -182,6 +182,12 @@
...
+          "class": "BadRequestError"
```
Aschen, line 190: It's a PluginImplementationError since the EmbeddedSDK is used by developers

test/core/shared/sdk/embeddedSdk.test.js
```diff
@@ -65,5 +66,20 @@ describe('EmbeddedSDK', () => {
       should(embeddedSdk.protocol.query)
         .be.calledWithMatch({ ...request, propagate: true });
     });
+
+    it('should throw an error if the action is forbidden', async () => {
+      const forbiddenRequest = { controller: 'auth', action: 'createApiKey' };
+
+      should(() => embeddedSdk.query(forbiddenRequest)).throw(Error);
```
Aschen, line 73: Usually we try to check the error code (at least) and error class since changing them is a breaking change

test/core/shared/sdk/embeddedSdk.test.js
```diff
@@ -65,5 +66,20 @@ describe('EmbeddedSDK', () => {
       should(embeddedSdk.protocol.query)
         .be.calledWithMatch({ ...request, propagate: true });
     });
+
+    it('should throw an error if the action is forbidden', async () => {
```
Aschen, line 70:
```suggestion
    it('should throw an error if the action is forbidden', () => {
```

test/core/shared/sdk/embeddedSdk.test.js
```diff
@@ -65,5 +66,20 @@ describe('EmbeddedSDK', () => {
...
+      const forbiddenRequest = { controller: 'auth', action: 'createApiKey' };
+
+      should(() => embeddedSdk.query(forbiddenRequest)).throw(Error);
+    });
+
+    it('should warn if the action is not supported', async () => {
```
Aschen, line 76:
```suggestion
    it('should warn if the action is not supported', () => {
```

Aschen (changes requested):

lib/core/shared/sdk/embeddedSdk.ts
```diff
@@ -37,13 +38,39 @@ import ImpersonatedSDK from './impersonatedSdk';
...
+    'searchApiKeys',
+    'updateMyCredentials',
+    'updateSelf',
+    'validateMyCredentials',
+  ]),
+};
+
+const warnEmbeddedActions = {
+  'auth': {
+    'login': 'EmbeddedSdk\'s auth:login is deprecated, use user impersonation instead',
```
Aschen, line 63:
```suggestion
    'login': 'EmbeddedSDK.login is deprecated, use user impersonation instead',
```

lib/core/shared/sdk/embeddedSdk.ts
Aschen: Ok I see why now

Aschen (approved):

PR #2321, opened 2022-04-11
Date extractor in Kuzzle request

OlivierCavadenti:
## What does this PR do ?

At this time, no parameter extractor exists for date.
This commit adds a first extractor for dates : 
- getDate to extract and valid dates in ISO8601 or custom format.
- getTimestamp to extract a valid timestamp


lib/api/request/kuzzleRequest.ts
```diff
@@ -604,6 +604,34 @@ export class KuzzleRequest {
...
+   * @param options Additional options
+   *    - `format`: date format timestamp, date or ISO8061 (default: 'date')
+   * @throws {api.assert.missing_argument} If parameter not found and no default
+   *                                       value provided
+   */
+  getDate (name: string,
+    options: {
+            format?: 'timestamp' | 'date' | 'ISO8061';
+          } = { format: 'date' }
+  ): number | Date | string {
```
Aschen, line 620: Nitpicking: coding standard: when the line is too long, each argument should be on it's own line
```suggestion
function getDate (
  name: string,
  options: { format?: 'timestamp' | 'date' | 'ISO8061' } = { format: 'date' },
): number | Date | string { }
```

lib/api/request/kuzzleRequest.ts
```diff
@@ -604,6 +604,34 @@ export class KuzzleRequest {
...
+    const args = this.input.args;
```
Aschen, line 621: Why the intermediary reference?

Aschen (changes requested):
Regarding the format option I was more thinking about the input format because we often manipulate timestamp or ISO8061 dates.  

Then I think the extractor should just returns the number (timestamp) or the string (iso8601) even if I don't really like this option because it breaks the monomorphism of the function.  

Or we could have a separate `getTimestamp`. What do you think?

lib/api/request/kuzzleRequest.ts
```diff
@@ -1007,6 +1001,21 @@ export class KuzzleRequest {
...
+   *
+   * @param name parameter name
+   * @private
+   */
+  private _getDate (name: string): Date {
+    const args = this.input.args;
+    if (args[name] === undefined) {
+      throw assertionError.get('missing_argument', name);
+    }
+    return new Date(args[name]);
```
Aschen: What if the format is incorrect?

lib/api/request/kuzzleRequest.ts
```diff
@@ -1007,6 +1001,21 @@ export class KuzzleRequest {
...
     return value;
   }
+
+  /**
+   * Generic getter value: date value.
+   *
+   * @param name parameter name
+   * @private
+   */
+  private _getDate (name: string): Date {
```
Aschen: Usually we don't add `_` when the `private` modifier is already used

Aschen (changes requested):

lib/api/request/kuzzleRequest.ts
OlivierCavadenti: this function is dropped since I don't cast to Date now

lib/api/request/kuzzleRequest.ts
```diff
@@ -604,6 +606,50 @@ export class KuzzleRequest {
     return this._getObject(this.input.args, name, name, def);
   }
 
+  /**
+   * Gets a parameter from a request arguments and check if the date is an ISO8601 format date or is valid regarding
+   * a given custom format (example : YYYY-MM-DD).
```
Aschen, line 611: The description should precise that moment format is used

lib/api/request/kuzzleRequest.ts
```diff
@@ -604,6 +606,50 @@ export class KuzzleRequest {
...
+   */
+  getDate (name: string, format?: string): string {
+    const args = this.input.args;
+    if (args[name] === undefined) {
+      throw assertionError.get('missing_argument', name);
+    }
+    if (format && moment(args[name], format, true).isValid() === false) {
+      throw assertionError.get('invalid_type', name, 'date');
+    }
+    if ( moment(args[name], moment.ISO_8601).isValid() === false) {
```
Aschen, line 628: Nitpicking: 
 - usually we don't add blank space after parenthesis
 - usually we use the `!` modifier instead of `=== false`
 
```suggestion
    if (! moment(args[name], moment.ISO_8601).isValid()) {
```

Shiranuit (approved):

PR #2322, opened 2022-04-12
Fix multi-search

Aschen:
## What does this PR do ?

Fix a bug in multi-search where `request` object was not passed as argument

```
import { Kuzzle, WebSocket } from 'kuzzle-sdk';

const kuzzle = new Kuzzle(new WebSocket('localhost'));

async function run () {
  await kuzzle.connect();

  const res = await kuzzle.query({
    controller: 'document',
    action: 'search',
    targets: [
      {
        index: 'index1',
        collections: ['collection1']
      },
      {
        index: 'index2',
        collections: ['collection1']
      },
    ],
    body: {
     query: {
      term: { name: 'Adrien' }
     }
    },
    from: 2,
    size: 2
  })

  console.log(res.result.hits)
}

run();
```

### DIFF
--- a/doc/2/api/controllers/document/search/index.md
+++ b/doc/2/api/controllers/document/search/index.md
@@ -33,12 +33,17 @@ Koncorde filters will be translated into an Elasticsearch query.
 Koncorde `bool` operator and `regexp` clause are not supported for search queries.
 :::
 
-## Multi Search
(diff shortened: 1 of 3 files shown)


Aschen:
Fixed by https://github.com/kuzzleio/kuzzle/pull/2331

Issue #2219, opened 2021-11-29
Generic events are not triggered on some actions


Aschen:
## Expected Behavior

Latest added actions does not trigger generic events: `mUpsert` (maybe more)

## Current Behavior

Should trigger generic events

## Possible Solution

Add in the `lib/config/documentEventAliases.js` file + find a way to ensure actions of the document controller MUST be listed in this file

## Steps to Reproduce

1.
2.
3.

## Context (Environment)

Kuzzle version:
Node.js version:
SDK version:

PR #2323, opened 2022-04-12
Generic events are not triggered on some actions

clementbolin:
#2219 

## What does this PR do ?

Automatically add document events to EventAliases.

new events:
- [x] mUpsert
- [x] deleteFields
- [x] upsert
- [x] exists
- [x] export
- [x] mExists
- [x] search

unit tests: `npm run test:unit`
functional tests: `npm run test:functional`

### Other changes

- [x] Convert `lib/config/documentEventAliases.js` -> `.ts`


lib/api/controllers/documentController.js
```diff
@@ -926,4 +933,4 @@ function hasMultiTargets (str) {
   return [',', '*', '+'].some(chr => str.includes(chr)) || str === '_all';
 }
 
-module.exports = DocumentController;
+module.exports = { DocumentController, actions };
```
Aschen, line 936: Instead of modifying the export and declaring the actions separately, you could declare a static property on the DocumentController containing your action <-> event mappings

lib/api/controllers/documentController.js
```diff
@@ -31,37 +31,44 @@ const {
 const extractFields = require('../../util/extractFields');
 const { dumpCollectionDocuments } = require('../../util/dump-collection');
 
+/**
+ * @description actions available on the document Controller
```
Aschen, line 35: Nit: you could specify that this is used by generic events

Aschen (changes requested):
Few comments, also can you update the title please?

clementbolin:
I don't know if it's useful, but I added `EventAliases` type. @Aschen

Aschen (changes requested):

lib/api/controllers/documentController.js
```diff
@@ -31,37 +31,52 @@ const {
...
-      'validate'
-    ]);
+    super([ ...actions.keys() ]);
+  }
+
+  /**
+   * @description return the document actions Map
+   * @returns {Map<string, string>}
+   */
+  static getActions () {
```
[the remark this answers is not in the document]
clementbolin: Small question, is there a difference in performance or memory usage between what I did and your example? Or is it only for DX?

lib/api/controllers/documentController.js
```diff
@@ -31,37 +31,52 @@ const {
 const extractFields = require('../../util/extractFields');
 const { dumpCollectionDocuments } = require('../../util/dump-collection');
 
+/**
+ * @description actions available on the document Controller (used by generic events)
+ * @key actions
+ * @value event type (empty for actions that are not linked to events)
+ */
+const actions = new Map([
```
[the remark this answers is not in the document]
Aschen: I understand this point but IMHO the declaration of the Map is less obvious that an object

lib/api/controllers/documentController.js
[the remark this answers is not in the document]
Aschen: It's only DX

lib/api/controllers/documentController.js
```diff
@@ -36,46 +36,45 @@ const { dumpCollectionDocuments } = require('../../util/dump-collection');
...
+  'validate': '',
+};
 
 /**
  * @class DocumentController
  */
 class DocumentController extends NativeController {
   constructor () {
-    super([ ...actions.keys() ]);
+    super([ ...Object.keys(actions) ]);
```
Aschen, line 71: Now it can just be
```suggestion
    super(Object.keys(actions));
```

lib/api/documentExtractor.js
```diff
@@ -39,7 +39,7 @@ const extractors = ([
         return request;
       }
     },
-    targets: ['delete', 'get']
+    targets: ['delete', 'get', 'exists']
```
Aschen, line 42: Well `exists` did not technically returns a document but it still get info about this document so we can add it, good job :+1:

lib/config/documentEventAliases.ts
```diff
@@ -22,12 +22,12 @@
...
 import DocumentController from '../api/controllers/documentController';
 
-function filter (map: Map<string, string>, expectValue: string) {
+function filter (obj: Record<string, unknown>, expectValue: string): Array<string> {
   const result = [];
 
-  for (const [key, value] of map) {
-    if (value === expectValue) {
-      result.push(key);
+  for (const action in obj) {
```
Aschen, line 28: You should avoid `for in` loops https://bitsofco.de/for-in-vs-for-of/

lib/api/controllers/documentController.js
```diff
@@ -36,46 +36,45 @@ const { dumpCollectionDocuments } = require('../../util/dump-collection');
...
-  ['replace', 'write'],
-  ['scroll', ''],
-  ['search', 'get'],
-  ['update', 'update'],
-  ['upsert', 'update'],
-  ['updateByQuery', 'update'],
-  ['validate', ''],
-]);
+const actions = {
+  'count': '',
```
Aschen, line 40: Keys should not be strings

Aschen (changes requested):

lib/config/documentEventAliases.ts
clementbolin: That's what I did, I didn't push yet.

lib/api/controllers/documentController.js
```diff
@@ -37,30 +37,30 @@ const { dumpCollectionDocuments } = require('../../util/dump-collection');
...
+  export: 'get',
+  get: 'get',
+  mCreate: 'write',
+  mCreateOrReplace: 'write',
+  mDelete: 'delete',
+  mExists: 'get',
+  mGet: 'get',
+  mReplace: 'write',
+  mUpdate: 'update',
+  mUpsert: '',
```
Aschen: This should also trigger the update generic event

Aschen (changes requested):
I know it's not an easy part to deal with events but can you please add functional tests to ensure the generic events are working with `upsert` and `deleteFields` since those action are slightly different from the other ones?

Tests are written here https://github.com/kuzzleio/kuzzle/blob/2-dev/features/PluginEvents.feature

lib/api/controllers/documentController.js
clementbolin: Are you sure you are reviewing the right commit? because mUpsert triggers a generic update event.
`mUpsert: 'update',`

lib/config/documentEventAliases.ts
Aschen: You can put the PR in draft mode then so other reviewers knows it still wip

lib/api/documentExtractor.js
```diff
@@ -26,6 +26,27 @@ const kerror = require('../kerror');
 const assertionError = kerror.wrap('api', 'assert');
 
 const extractors = ([
+  {
+    methods: {
+      extractFromRequest: request => [
+        {
+          _id: request.input.args._id,
+          fields: request.input.body.fields,
```
Aschen, line 34: Deleted fields should be in `_source`, for example you could recreate an object from lodash path with properties set to `undefined`

lib/api/documentExtractor.js
```diff
@@ -139,9 +168,18 @@ const extractors = ([
         for (let it = 0; it < documents.length; it++) {
           const document = documents[it];
 
-          request.input.body.documents.push({
-            _id: document._id,
-            body: document._source });
+          if (request.input.action === 'mUpsert') {
+            request.input.body.documents.push({
+              _id: document._id,
+              changes: document._source.changes,
```
Aschen, line 174: Same comment here you should only have a `_source` property containing the changes

lib/api/documentExtractor.js
```diff
@@ -139,9 +168,18 @@ const extractors = ([
...
+              default: document._source.default
+            });
+          }
+          else {
+            request.input.body.documents.push({
+              _id: document._id,
+              body: document._source });
```
Aschen, line 181: wrong wrapping
```suggestion
              body: document._source 
            });
```

lib/api/documentExtractor.js
```diff
@@ -127,7 +148,15 @@ const extractors = ([
         for (let it = 0; it < request.input.body.documents.length; it++) {
           const document = request.input.body.documents[it];
 
-          documents.push({ _id: document._id, _source: document.body });
+          if (request.input.action === 'mUpsert') {
+            documents.push({
+              _id: document._id,
+              _source: { changes: document.changes, default: document.default }
```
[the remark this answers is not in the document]
Aschen: Also, the `upsert` action should be modified accordingly

lib/config/documentEventAliases.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  return result;
+}
+
+export const documentEventAliases: EventAliases = {
+  list: {
+    'delete': filter(DocumentController.actions, 'delete'),
+    'get': filter(DocumentController.actions, 'get'),
+    'update': filter(DocumentController.actions, 'update'),
+    'write': filter(DocumentController.actions, 'write'),
```
Aschen, line 42: Object keys should not have quotes

lib/types/config/EventAliases.ts
```diff
@@ -19,15 +19,8 @@
...
-  list: {
-    'delete': ['delete', 'deleteByQuery', 'mDelete'],
-    'get': ['get', 'mGet', 'search'],
-    'update': ['update', 'mUpdate', 'updateByQuery', 'upsert'],
-    'write': ['create', 'createOrReplace', 'mCreate', 'mCreateOrReplace', 'mReplace', 'replace']
-  },
-  namespace: 'generic:document',
-  notBefore: ['search', 'deleteByQuery', 'updateByQuery'],
-};
+export interface EventAliases {
```
Aschen, line 22: This interface should be declared in the file it is used and not exported

Aschen (changes requested):

lib/api/documentExtractor.js
clementbolin: I don't really understand what you want here? I did this, is this okay ? :
```js
      insertInRequest: ([documents], request) => {
        request.input.args._id = documents._id;
        request.input.body.fields = documents._source;
        return request;
      },
```

Aschen (approved):

PR #2328, opened 2022-04-13
Fix race condition when creating index hidden indice

Aschen:
## What does this PR do ?

When creating a new index, an special ES indice is created (`&<index>._kuzzle_keep`) to reserve the index name.

If `index:create` is called in the same time, then we have a race condition when trying to create this indice since it could have been created between the check for existence and the creation.

This PR use a mutex to prevent such race conditions

### DIFF
--- a/lib/service/storage/elasticsearch.js
+++ b/lib/service/storage/elasticsearch.js
@@ -1819,21 +1819,13 @@ class ElasticSearch extends Service {
       index: await this._getIndice(index, collection)
     };
 
-    const mutex = new Mutex(`hiddenCollection/${index}`);
     try {
-      await mutex.lock();
-
       await this._client.indices.delete(esRequest);
 
-      if (! await this._hasHiddenCollection(index)) {
-        await this._createHiddenCollection(index);
-      }
+      await this._createHiddenCollection(index);
     }
     catch (e) {
       throw this._esWrapper.formatESError(e);
-    }
-    finally {
-      await mutex.unlock();
     }
 
     return null;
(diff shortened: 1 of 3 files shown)


lib/service/storage/elasticsearch.js
```diff
@@ -1819,7 +1819,7 @@ class ElasticSearch extends Service {
       index: await this._getIndice(index, collection)
     };
 
-    const mutex = new Mutex(`hiddenCollection/delete/${index}`);
+    const mutex = new Mutex(`hiddenCollection/${index}`);
     try {
       await this._client.indices.delete(esRequest);
```
Shiranuit, line 1825: shouldn't the mutex be acquired before deleting the indice in ES ?

Shiranuit, line 1825: I see the mutex is acquired after deleting the index in ES, shouldn't the mutex be acquired before ?

Shiranuit (changes requested):

lib/service/storage/elasticsearch.js
```diff
@@ -1819,12 +1819,12 @@ class ElasticSearch extends Service {
...
+    const mutex = new Mutex(`hiddenCollection/${index}`);
     try {
-      await this._client.indices.delete(esRequest);
-
       await mutex.lock();
 
+      await this._client.indices.delete(esRequest);
+
       if (! await this._hasHiddenCollection(index)) {
         await this._createHiddenCollection(index);
```
Shiranuit, line 1829: A bit tricky but `deleteCollection` uses the mutex `hiddenCollection/${index}` then later calls `_createHiddenCollection` which also use the mutex `hiddenCollection/${index}` which is already locked at the moment it is called, which causes your test to be stuck in an infinite loop

Shiranuit (approved):

clementbolin (approved):

Issue #2275, opened 2022-02-21
Be able to force number of replicas and primary shards using the Kuzzle configuration


alexandrebouthinon:
## Feature Description
In order to match every specific Elasticsearch cluster specs, we need to be able to choose how many primary and replicas we want for Elasticsearch indexes. For now we can only give the `number_of_replicas` and `number_of_shards` index settings when creating Kuzzle collection. This is fine but not enough since we cannot modify it for Kuzzle indexes. IT teams should be able to enforce those parameters to match their Elasticsearch cluster topology and settings using the Kuzzle configuration

## Possible solution
- Add dedicated param in the Kuzzle configuration
- Rework the Kuzzle index creation to use these settings
- Rework the Kuzzle collection create to use these settings

alexandrebouthinon:
Added in https://github.com/kuzzleio/kuzzle/pull/2391

Issue #2208, opened 2021-11-16
Make the number of shards configurable for internal collections


Aschen:
## Feature Description

Allows to modify the default number of shards for internal collections of the `kuzzle` index such as `users`.

:warning: The number of shards can only be set when the collection is created for the first time

PR #2329, opened 2022-04-14
Add kuzzle config settings for internal index's number of shards and replicas

ChillPC:
#2208 #2275

## What does this PR do ?

Add the possibility to give ES index settings for internal index's collection in kuzzle configuration. Also update dynamic ES settings of internal index's collection when restarting kuzzle.

The changes also affect plugins as there ES indexes are treated by kuzzle as internal index's collection.

  For creation :

- Step 1 : Set your number of replicas and shards for internal collection
- Step 2 : Make your elasticsearch service port 9200 accessible
- Step 3 : Start kuzzle
- Step 4 : `curl -X GET "127.0.0.1:9200/_cat/shards/%25kuzzle*"`

Update number of replicas :

- Step 1 : Set a different number of replicas
- Step 2 : Make your elasticsearch service port 9200 accessible
- Step 3 : Restart kuzzle
- Step 4 : `curl -X GET "127.0.0.1:9200/_cat/shards/%25kuzzle*"`

### DIFF
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -774,54 +774,90 @@
       "internalIndex": {
         "name": "kuzzle",
         "collections": {
-          "dynamic": "false",
           "users": {
-            "properties": {
-              "profileIds": { "type": "keyword" }
+            "settings": {
+              "number_of_shards": 1,
+              "number_of_replicas": 1
+            },
+            "mappings": {
+              "dynamic": "false",
+              "properties": {
+                "profileIds": { "type": "keyword" }
+              }
             }
           },
           "profiles": {
-            "dynamic": "false",
-            "properties": {
-              "policies": {
-                "properties":  {
-                  "roleId": { "type": "keyword" }
+            "settings": {
+              "number_of_shards": 1,
+              "number_of_replicas": 1
+            },
+            "mappings": {
+              "dynamic": "false",
+              "properties": {
+                "policies": {
+                  "properties":  {
+                    "roleId": { "type": "keyword" }
+                  }
                 }
               }
             }
           },
           "roles": {
-            "dynamic": "false",
-            "properties": {
-              "controllers": {
-                "dynamic": "false",
-                "properties": {}
+            "settings": {
+              "number_of_shards": 1,
+              "number_of_replicas": 1
+            },
+            "mappings": {
+              "dynamic": "false",
+              "properties": {
+                "controllers": {
+                  "dynamic": "false",
+                  "properties": {}
+                }
               }
             }
           },
           "validations": {
-            "properties": {
-              "index": { "type": "keyword" },
-              "collection": { "type": "keyword" },
-              "validations": {
-                "dynamic": "false",
-                "properties": {}
+            "settings": {
+              "number_of_shards": 1,
+              "number_of_replicas": 1
+            },
+            "mappings": {
+              "properties": {
+                "index": { "type": "keyword" },
+                "collection": { "type": "keyword" },
+                "validations": {
+                  "dynamic": "false",
+                  "properties": {}
+                }
               }
             }
           },
           "config": {
-            "dynamic": "false",
-            "properties": {}
+            "settings": {
+              "number_of_shards": 1,
+              "number_of_replicas": 1
+            },
+            "mappings": {
+              "dynamic": "false",
+              "properties": {}
+            }
           },
           "api-keys": {
-            "dynamic": "false",
-            "properties": {
-              "userId": { "type": "keyword" },
-              "hash": { "type": "keyword" },
-              "expiresAt": { "type": "long" },
-              "ttl": { "type": "keyword" },
-              "description": { "type": "text" },
-              "token": { "type": "keyword" }
+            "settings": {
+              "number_of_shards": 1,
+              "number_of_replicas": 1
+            },
+            "mappings": {
+              "dynamic": "false",
+              "properties": {
+                "userId": { "type": "keyword" },
+                "hash": { "type": "keyword" },
+                "expiresAt": { "type": "long" },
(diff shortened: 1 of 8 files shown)


ChillPC:
Manually tested with creation and update of internal index's creation with :
- New deprecated friendly config with `undefined` (this branch)
- New clean config (`test/new-kuzzle-config`)
- Old config (`test/deprecated-kuzzle-config`)

Aschen (changes requested):
`lib/core/shared/store.js` is a good candidate for a typescript conversion ;-)

.kuzzlerc.sample
```diff
@@ -771,57 +771,98 @@
...
+      // See https://www.elastic.co/guide/en/elasticsearch/reference/7.5/index-modules.html#index-modules-settings
       "internalIndex": {
         "name": "kuzzle",
         "collections": {
-          "dynamic": "false",
           "users": {
-            "properties": {
-              "profileIds": { "type": "keyword" }
+            "settings": {
+              "number_of_shards": 1,
```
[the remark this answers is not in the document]
alexandrebouthinon: 1 shard is ok for maximum 1 million documents. If we have to go beyond that, I think we start looking at fine-tuning and it just needs to be documented

lib/core/shared/store.js
```diff
@@ -114,17 +135,101 @@ class Store {
...
+              // Create call will update collection and cache it internally
+              return global.kuzzle.ask(
+                `core:storage:${this.scope}:collection:create`,
+                this.index,
+                collection,
+                // @deprecated
+                (isConfigDeprecated)
+                  ? { mappings: config.mappings }
+                  : { mappings: config.mappings, settings: dynamicSettings },
+              );
```
alexandrebouthinon, line 211: I don't get the purpose of all of this. IMO if a collection already exists we only need to reject the collection creation process didn't we?

lib/core/shared/store.js
```diff
@@ -92,18 +94,37 @@ class Store {
...
+    else {
+      // Ensure that cached collection are used after being properly
+      // created by a kuzzle cluster
+      const cachingMutex = new Mutex(`Store.init(${this.index})`, {
+        timeout: -1,
+        ttl: 30000,
+      });
+
+      await cachingMutex.lock();
+      cachingMutex.unlock();
```
Aschen, line 123: Why locking the mutex to unlock it in the next instruction?

lib/core/shared/store.js
```diff
@@ -114,17 +135,101 @@ class Store {
...
-      { concurrency: 10 });
+  createCollections (
+    collections,
+    { indexCacheOnly = false } = {}
+  ) {
+    return promiseAllN(
+      Object.entries(collections).map(
+        ([collection, config]) => async () => {
+          // @deprecated
+          if (config.mappings !== undefined && config.settings !== undefined) {
```
Aschen, line 146: You should use a guard instead to avoid indenting the whole block

lib/core/shared/store.js
```diff
@@ -114,17 +135,101 @@ class Store {
...
+            }
+
+            const exist = await global.kuzzle.ask(
+              `core:storage:${this.scope}:collection:exist`,
+              this.index,
+              collection);
+
+            if (exist) {
+              // @deprecated
+              const dynamicSettings = (isConfigDeprecated)
```
Aschen, line 168: Useless parenthesis
```suggestion
              const dynamicSettings = isConfigDeprecated
```

lib/core/shared/store.js
```diff
@@ -114,17 +135,101 @@ class Store {
...
+                }
+                global.kuzzle.log.warn([
+                  'Attempt to recreate',
+                  `an existing collection ${collection}`,
+                  `of index ${this.index} of scope ${this.scope}`,
+                  'with non matching',
+                  'static setting : number_of_shards',
+                  `at ${config.settings.number_of_shards} while existing one`,
+                  `is at ${existingSettings.number_of_shards}`,
+                ].join('\n'));
```
Aschen, line 199: Log line should be on a single line

lib/core/shared/store.js
```diff
@@ -114,17 +135,101 @@ class Store {
...
+              }
+
+              return global.kuzzle.ask(
+                `core:storage:${this.scope}:collection:create`,
+                this.index,
+                collection,
+                // @deprecated
+                (isConfigDeprecated)
```
Aschen, line 207: Useless parenthesis
```suggestion
                isConfigDeprecated
```

lib/util/async.ts
```diff
@@ -0,0 +1,57 @@
+// Taken from :
+// https://gist.github.com/jcouyang/632709f30e12a7879a73e9e132c0d56b
+export function promiseAllN<T> (collection: Array<() => Promise<T>>, n = 100): Promise<T[]> {
```
Aschen, line 3: Can you add a more explicit comment and also give an usage example please?

package-lock.json
```diff
@@ -429,6 +429,14 @@
         "js-yaml": "^4.1.0",
         "minimatch": "^3.0.4",
         "strip-json-comments": "^3.1.1"
+      },
+      "dependencies": {
+        "strip-json-comments": {
```
Aschen, line 434: Why did the `package-lock.json` file changed but not the `package.json` ?

Aschen (changes requested):

lib/core/shared/store.js
Aschen, line 123: Why locking the mutex to unlock it in the next instruction?

Aschen, line 123: This should be awaited

Aschen (changes requested):

Shiranuit (approved):

PR #2331, opened 2022-04-14
Support for multiple alias pointing to one index

Shiranuit:
## What does this PR do ?

This PR allows to have multiple alias pointing to a single ES index.

Create some alias on an ES Index, you should be able to push document in a collections using one of the many aliases of the collection
`POST localhost:9200/%26index1.collection1/_alias/%40%26myindex.mycollection`

### Other changes

Fix an issue in multisearch where the request was not given to a function that requires it for permission checking on indexes.
Add functional tests to ensure multisearch is working
Add caching to index resolution from alias
Change how index resultion from alias is done to support the case where multiple alias are pointing to a same index


lib/model/security/user.ts
```diff
@@ -87,16 +87,16 @@ export class User {
...
 
   /**
    * Verifies that every targets are allowed by at least one profile,
    * while skipping the ones that includes a wildcard since they will be expanded
    * later on, based on index and collections authorized for the given user.
    */
-  private async areTargetsAllowed (profiles: Profile[], targets: Target[]) {
-    const profilesPolicies = await Bluebird.map(profiles, profile => profile.getAllowedPolicies());
+  private async areTargetsAllowed (request: KuzzleRequest, profiles: Profile[], targets: Target[]) {
+    const profilesPolicies = await Bluebird.map(profiles, profile => profile.getAllowedPolicies(request));
```
Aschen, line 99: `Bluebird.map` should be replaced by `Promise.all`, it would have detected the error

lib/service/storage/elasticsearch.js
```diff
@@ -401,35 +402,106 @@ class ElasticSearch extends Service {
...
   }
 
-  async _formatSearchResult (body) {
+  /**
+   * Generate a map that associate an alias to a pair of index and collection
+   * 
+   * @param {*} targets 
+   * @returns 
+   */
+  mapTargetsToAlias (targets) {
```
Aschen, line 433: This method should starts with `_`

lib/service/storage/elasticsearch.js
```diff
@@ -401,35 +402,106 @@ class ElasticSearch extends Service {
...
+      for (const targetCollection of target.collections) {
+        const alias = this._getAlias(target.index, targetCollection);
+
+        if (! aliasToTargets[alias]) {
+          aliasToTargets[alias] = {
+            collection: targetCollection,
+            index: target.index,
+          };
+        }
+
```
Aschen, line 446:
```suggestion
```

lib/service/storage/elasticsearch.js
```diff
@@ -401,35 +402,106 @@ class ElasticSearch extends Service {
...
+      }
+    }
+
+    return aliasToTargets;
+  }
+
+  async _formatSearchResult (body, searchInfo = {}) {
+
```
Aschen, line 454:
```suggestion
```

lib/service/storage/elasticsearch.js
```diff
@@ -401,35 +402,106 @@ class ElasticSearch extends Service {
...
-      if (hit._index) {
-        alias = await this._getAliasFromIndice(hit._index);
+        /**
+         * Since multiple alias can point to the same index in ES, we need to
+         * find the first alias that exists in the map of aliases associated
+         * to the targets.
+         */
+        const alias = aliases.find(_alias => {
+          return aliasToTargets[_alias];
+        });
```
Aschen, line 492:
```suggestion
        const alias = aliases.find(_alias => aliasToTargets[_alias]);
```

Aschen (changes requested):
I started a PR to fix the multi search as well, can you use the documentation fix I made as well ? 
https://github.com/kuzzleio/kuzzle/pull/2322/files

Also, I noticed that `User.areTargetsAllowed` did not had unit tests, it could be good to add them for future maintainers

Aschen (approved):

clementbolin (approved):

PR #2332, opened 2022-04-15
Improve security message for SDK usage

Aschen:
## What does this PR do ?

Improve security log message when the action come from the EmbeddedSDK

```
// before
[SECURITY] User "null" applied action "createUser" on user "kuid-precious-lovecraft-84193."

// after
[SECURITY] EmbeddedSDK applied action "createUser" on user "kuid-precious-lovecraft-84193."
```


lib/api/controllers/securityController.js
```diff
@@ -38,6 +38,11 @@ const { generateRandomName } = require('../../util/name-generator');
  * @class SecurityController
  */
 class SecurityController extends NativeController {
+
```
clementbolin, line 41:
```suggestion

```

clementbolin (approved):

Shiranuit (approved):

PR #2333, opened 2022-04-16
Mappings update trigger a reindex

clementbolin:
## What does this PR do ?

reindex documents with the new mappings (`collection:update`)

Add a few documents in it.
Then add mappings to describe the fields in the added documents.

Expected: running a document:search in the collection should return the searched documents.

or run the functional tests

### DIFF
--- a/features/CollectionController.feature
+++ b/features/CollectionController.feature
@@ -38,6 +38,29 @@ Feature: Collection Controller
     Then I should receive a "hits" array of objects matching:
       | _id          | _source      |
       | "document-1" | { "age": 2 } |
+  
+  Scenario: Update a collection and search document (dyanmic false)
+    Given an index "nyc-open-data"
+    And I "create" the collection "nyc-open-data":"green-taxi" with:
+      | mappings | { "dynamic": "false", "properties": { "name": { "type": "keyword" }, "metadata": {"properties": {}, "dynamic": "false"} } } |
+    And I "create" the following documents:
+      | _id          | body        |
+      | "document-1" | {"age": 2} |
+    And I refresh the collection
+    When I "update" the collection "nyc-open-data":"green-taxi" with:
+      | mappings | { "properties": { "age": { "type": "long" } } } |
+    When I search documents with the following query:
+      """
+      {
+        "match": {
+          "age": 2
+        }
+      }
+      """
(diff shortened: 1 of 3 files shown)


features/CollectionController.feature
```diff
@@ -38,6 +38,29 @@ Feature: Collection Controller
     Then I should receive a "hits" array of objects matching:
       | _id          | _source      |
       | "document-1" | { "age": 2 } |
+  
+  Scenario: Update a collection and search document (dyanmic false)
```
Shiranuit, line 42:
```suggestion
  Scenario: Update a collection and search document (dynamic false)
```

lib/service/storage/elasticsearch.js
```diff
@@ -1387,6 +1387,20 @@ class ElasticSearch extends Service {
 
         await this.updateMapping(index, collection, mappings);
 
+        // update documents with new mappings
+        if (previousMappings.dynamic === 'false' && (await this._client.search(esRequest)).body.hits.hits.length > 0) {
```
Shiranuit, line 1391: If think it's better to call `count` than `search` if you just want to know if there are documents to update

Side note:
Avoid putting long calls with chained property access like this in condition for clarity `await this._client.search(esRequest)).body.hits.hits.length` prefer storing the result in a const variable.
Conditions that are too long and uses `&&` or `||` should be splitted on multiple lines like so
```js
const documentsCount = await this._client.search(esRequest)).body.hits.hits.length;
if ( previousMappings.dynamic === 'false'
  && documentsCount > 0
) {
```

features/CollectionController.feature
```diff
@@ -38,6 +38,29 @@ Feature: Collection Controller
...
+    When I "update" the collection "nyc-open-data":"green-taxi" with:
+      | mappings | { "properties": { "age": { "type": "long" } } } |
+    When I search documents with the following query:
+      """
+      {
+        "match": {
+          "age": 2
+        }
+      }
+      """
```
Shiranuit, line 59: It's good but you also need to verify that the search does not match the field `age` before updating the mapping and triggering the reindex

Shiranuit (changes requested):

Aschen:
Stale PR

PR #2338, opened 2022-04-21
Expose nodeId to applications

Aschen:
## What does this PR do ?

Expose the cluster node ID to applications

### DIFF
--- a/lib/kuzzle/kuzzle.ts
+++ b/lib/kuzzle/kuzzle.ts
@@ -164,8 +164,9 @@ class Kuzzle extends KuzzleEventEmitter {
   };
 
   private koncorde: Koncorde;
-  private id: string;
   private secret: string;
+
+  public id: string;
 
   constructor (config: KuzzleConfiguration) {
     super(
--- a/test/core/backend/Backend.test.js
+++ b/test/core/backend/Backend.test.js
@@ -53,6 +53,21 @@ describe('Backend', () => {
     });
   });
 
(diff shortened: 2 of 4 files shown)


lib/kuzzle/kuzzle.ts
```diff
@@ -164,9 +164,10 @@ class Kuzzle extends KuzzleEventEmitter {
   };
 
   private koncorde: Koncorde;
-  private id: string;
   private secret: string;
 
+  public id: string;
```
clementbolin, line 169: I think it's a good idea to add a comment to say that `id` defines the node ID.

test/core/backend/Backend.test.js
```diff
@@ -53,6 +53,21 @@ describe('Backend', () => {
     });
   });
 
+  describe.only('#nodeId', () => {
```
clementbolin, line 56: forgot to remove the `only()`

clementbolin (changes requested):

test/core/backend/Backend.test.js
OlivierCavadenti: Can we add not only rule ?
https://www.npmjs.com/package/eslint-plugin-no-only-tests

Aschen: Good idea!

OlivierCavadenti (approved):

clementbolin (approved):

PR #2340, opened 2022-04-25
Add headers in response content

Shiranuit:
## What does this PR do ?

Adds the headers directly into the content of the response this way protocols like websocket can also read the headers.
Some headers like `set-cookie` are removed from the response to let the browser deciding if the cookie should be readable or not by a script, preventing leakage of cookie if the request is not from the same domain.

`npm run test`

### Other changes

Instead of selecting headers individually during the websocket upgrade to add them each time into the headers of the request, the list of headers is now retrieved entirely and added to the request.
So we are not limited to `origin` and `cookie` anymore.


Aschen (commented):
So if I understand well, standard HTTP headers such as `content-type` or `content-length` will also be added into the WebSocket / MQTT / ... response payload.

I'm not sure we want to add useless overhead from the HTTP protocol into other protocols.

Is it difficult to only copy the custom headers added by the user when using `request.response.configure` for example?

Shiranuit:
> So if I understand well, standard HTTP headers such as `content-type` or `content-length` will also be added into the WebSocket / MQTT / ... response payload.
> 
> I'm not sure we want to add useless overhead from the HTTP protocol into other protocols.
> 
> Is it difficult to only copy the custom headers added by the user when using `request.response.configure` for example?

Here is a list of all the standards header we do return actually, this list can be added to the list of restrictedHeaders to not return if you think it will cause a useless overhead.
Or like you said we could only return a list of headers added using `request.response.configure`
```js
{
  'X-Kuzzle-Node': 'knode-depressed-ernst-92256',
  'Accept-Encoding': 'gzip,deflate,identity',
  'Access-Control-Allow-Headers': 'Content-Type, Access-Control-Allow-Headers, Authorization, X-Requested-With, Content-Encoding, Content-Length, X-Kuzzle-Volatile',
  'Access-Control-Allow-Methods': 'GET,POST,PUT,PATCH,DELETE,OPTIONS,HEAD',
  'content-type': 'application/json',
  'Access-Control-Allow-Credentials': 'true',
 }
```

Aschen:
I think it's better to return only those added by the user than restrict our default ones because the user may want to return a different `Access-Control-Allow-Methods` for example

Aschen (changes requested):
Can we have at least a functional test that ensure:
 - set-cookie is not leaked
 - default headers are not added to the response

Shiranuit:
@Aschen 
As requested functional tests have been added to ensure:
- set-cookie is not leaked
- default headers are not added to the response

Aschen (approved):

Issue #2297, opened 2022-03-27
admin:resetDatabase action should also reset the mappings loading lock


Aschen:
## Expected Behavior

The `admin:resetDatabase` action should also reset the lock that is preventing to load default mappings on each node when Kuzzle is starting.

See https://github.com/kuzzleio/kuzzle/blob/master/lib/kuzzle/kuzzle.ts#L441

ChillPC:
Can you give a more detailed description?

Aschen:
Actually we talked about this issue before (ping @clementbolin )

Instead of removing the lock, the lock has to be an hash of the mappings that has been charged by Kuzzle so if a new version of the mappings is available at startup, the lock became obsolete

PR #2341, opened 2022-04-25
Remove import mappings lock on `admin:resetDatabase` action

ChillPC:
#2297

# How do you test it

1. Create new app
2. In `app.ts`:
  ```js
  import { Backend } from './kuzzle';

  const app = new Backend('issue-2341');

  app.import.mappings({
    foo: {
      bar: {
        mappings: {
          properties: {
            first_mapping: { type: 'keyword'}
          }
        }
      }
    }
  });
  ```
3. Execute: `kourou admin:resetDatabase`
4. Replace in `app.ts`: `first_mapping` by `second_mapping`
5. Check mappings, `second_mapping` will be there and not `first_mapping`


lib/kuzzle/kuzzle.ts
```diff
@@ -570,15 +572,14 @@ class Kuzzle extends KuzzleEventEmitter {
   private async _waitForImportToFinish () {
     const importTypes = Object.keys(this.importTypes);
 
-    while (importTypes.length) {
+    importTypes.forEach(async importType => {
```
Aschen, line 575: We try to avoid `forEach` loop because they are not respecting the synchronicity of the `await` keyword

lib/kuzzle/kuzzle.ts
```diff
@@ -606,8 +607,32 @@ class Kuzzle extends KuzzleEventEmitter {
 
     try {
       for (const [type, importMethod] of Object.entries(this.importTypes)) {
+        const importPayload = {};
+
+        switch (type) {
+          case 'fixtures':
+            _.set(importPayload, ['toSupport', 'fixtures'], toSupport.fixtures);
```
Aschen, line 614: You should use the dot notation to access object elements
```suggestion
            _.set(importPayload, 'toSupport.fixtures', toSupport.fixtures);
```

lib/kuzzle/kuzzle.ts
```diff
@@ -606,8 +607,32 @@ class Kuzzle extends KuzzleEventEmitter {
...
+            break;
+          case 'permissions':
+            _.set(importPayload, ['toSupport', 'securities'], toSupport.fixtures);
+            _.set(importPayload, ['toImport', 'profiles'], toImport.profiles);
+            _.set(importPayload, ['toImport', 'roles'], toImport.roles);
+            _.set(importPayload, ['toImport', 'users'], toImport.users);
+            break;
+        }
+
+        const importPayloadHash = sha256(JSON.stringify(importPayload));
```
Aschen, line 628: You should use `stable-json-stringify` here to ensure the key order will be the same (even if with Node.js there is very little chances to have a key order inconsistency)

lib/kuzzle/kuzzle.ts
```diff
@@ -606,8 +607,32 @@ class Kuzzle extends KuzzleEventEmitter {
...
         const mutex = new Mutex(`backend:import:${type}`, { timeout: 0 });
-        const initialized = await this.ask('core:cache:internal:get', `${BACKEND_IMPORT_KEY}:${type}`) === '1';
+
+        const existingHash = await this.ask(
+          'core:cache:internal:get',
+          `${BACKEND_IMPORT_KEY}:${type}`);
+
+        const initialized = existingHash && existingHash === importPayloadHash;
```
Aschen, line 635: Can be simply
```suggestion
        const initialized = existingHash === importPayloadHash;
```

Aschen (changes requested):

Aschen (changes requested):
Missing unit tests for this PR

test/kuzzle/kuzzle.test.js
```diff
@@ -386,6 +396,21 @@ describe('/lib/kuzzle/kuzzle.js', () => {
...
+
+      should(kuzzle.ask).be.calledWith(
+        'core:cache:internal:get',
+        'backend:init:import:permissions');
+
+      should(kuzzle.ask).be.calledWith(
+        'core:cache:internal:get',
+        'backend:init:import:fixtures');
+
+      should(kuzzle.ask).be.calledWithMatch(
```
Aschen, line 411: Same comment, the ask function should have been called with 3 args

Aschen (approved):

OlivierCavadenti (approved):

Shiranuit (approved):

PR #2342, opened 2022-04-26
Standardize arrays and objects in querystring

Shiranuit:
## What does this PR do ?

`request.getArray` and `request.getObject` are now able to retrieve array and object from the query parameters.
Example of process for `request.getArray`:
- The retrieved value is an array -> returns it directly
- The retrieved value is a string:
  - If the request is an HTTP Request -> JSON Parse the string:
    - If the parsed result is an array returns it directly
    - If not throw an invalid_type error
  - If not throw an invalid_type error
-  The retrieved value is something else -> throw an invalid_type error

The same process applies for `getObject`

Documentation on this behaviour has been added

There was some controller or pipe process that where splitting strings on `,` to obtain arrays, a new method has been added to the request object named `getArrayLegacy`, this method exist to ensure retrocompatibility, the method is deprecated and should not be used.

`getArrayLegacy` behaves pretty much like `getArray` but instead of throwing if the string cannot be parsed as a JSON Array
the string is simply splitted each time a `,` is encountered, to keep retrocompatibility.

`getArrayLegacy` process:
- The retrieved value is an array -> returns it directly
- The retrieved value is a string:
  - If the request is an HTTP Request -> JSON Parse the string:
    - If the parsed result is an array returns it directly
    - If not returns the string splitted on `,`
  - If not returns the string splitted on `,`
-  The retrieved value is something else -> throw an invalid_type error

Tests have been added to cover this new cases

`npm run test`

### Other changes

There was an undetected bug where calling `getBodyBoolean` in an HTTP Request was not returning the value of the field in the body but was returning `true` or `false` if the field was present or not.
This is now fixed.


alexandrebouthinon (commented):
I'm quite a fan of this addition but for the sake of backwards compatibility, I would tend to do the opposite:
- Keep the old behavior for `getObject` and `getArray`.
- Create two new functions with the new behavior

lib/api/request/kuzzleRequest.ts
```diff
@@ -589,12 +592,63 @@ export class KuzzleRequest {
...
+      throw assertionError.get('missing_argument', name);
+    }
+
+    if (Array.isArray(value)) {
+      return value;
+    }
+
+    // If we are using the HTTP protocol and we have a string instead of an Array
+    // we try to parse it as JSON
+    if (typeof value === 'string') {
```
Aschen, line 627: You could throw here and avoid indenting this whole code

lib/api/request/kuzzleRequest.ts
```diff
@@ -993,6 +1054,23 @@ export class KuzzleRequest {
     }
 
     if (! Array.isArray(value)) {
+      // If we are using the HTTP protocol and we have a string instead of an Array
+      // we try to parse it as JSON
+      if (this.context.connection.protocol === 'http'
```
Aschen, line 1059:
```suggestion
      if ( this.context.connection.protocol === 'http'
```

lib/api/request/kuzzleRequest.ts
```diff
@@ -1020,6 +1100,23 @@ export class KuzzleRequest {
     }
 
     if (! isPlainObject(value)) {
+      // If we are using the HTTP protocol and we have a string instead of an Array
+      // we try to parse it as JSON
+      if (this.context.connection.protocol === 'http'
```
Aschen, line 1105:
```suggestion
      if ( this.context.connection.protocol === 'http'
```

Aschen (approved):

Shiranuit:
> I'm quite a fan of this addition but for the sake of backwards compatibility, I would tend to do the opposite:
> 
>     * Keep the old behavior for `getObject` and `getArray`.
> 
>     * Create two new functions with the new behavior

I'm not sure to understand why would you do that ?
I mean `getArray` and `getObject` are the same has before they just have a another way of retrieving an array or object from a string.
deprecating those  functions to add two new functions that do basicaly the same things with something more would mean replacing every getArray or getObject in the code of Kuzzle and everywhere it's used in the client code just to support something new that doesn't cause a breaking change

Issue #2348, opened 2022-04-28
Check type for `index`, `collection` and `_id` params


Aschen:
## Description

We should ensure those param are string, otherwise we might have other errors like this one

```
kourou sdk:execute 'return sdk.document.search(42, 42, {})'

Caught an unexpected plugin error: str.includes is not a function
This is probably not a Kuzzle error, but a problem with a plugin implementation.
      at /var/app/node_modules/kuzzle/lib/api/controllers/documentController.js:918:42
      at Array.some (<anonymous>)
      at hasMultiTargets (/var/app/node_modules/kuzzle/lib/api/controllers/documentController.js:918:26)
      at DocumentController.search (/var/app/node_modules/kuzzle/lib/api/controllers/documentController.js:86:11)
      at doAction (/var/app/node_modules/kuzzle/lib/api/funnel.js:940:47)
      at Funnel.processRequest (/var/app/node_modules/kuzzle/lib/api/funnel.js:560:34)
```

PR #2349, opened 2022-05-02
Ensure usage of string for index and collection for `document:search` action 

ChillPC:
#2348

```
kourou sdk:execute 'return sdk.document.search(42, 42, {})'
```

Will not throw an error.


lib/api/request/kuzzleRequest.ts
```diff
@@ -710,27 +710,33 @@ export class KuzzleRequest {
   /**
    * Returns the index specified in the request
    */
-  getIndex (): string {
+  getIndex (throwOnMissing = true): string {
```
Aschen, line 713: Naming proposal, also it's better to use object named properties for readability but also to be future proof 
```suggestion
  getIndex ({ required = true } = {}): string {
```

Aschen (changes requested):

lib/api/request/kuzzleRequest.ts
```diff
@@ -710,27 +710,33 @@ export class KuzzleRequest {
   /**
    * Returns the index specified in the request
    */
-  getIndex (): string {
+  getIndex ({ required = true } = {}): string {
```
OlivierCavadenti, line 713: required not used

lib/api/request/kuzzleRequest.ts
```diff
@@ -805,7 +811,7 @@ export class KuzzleRequest {
       throw assertionError.get('invalid_type', '_id', 'string');
     }
 
-    return id;
+    return String(id);
```
OlivierCavadenti, line 814: getIndex and getCollection are nearly identical, cant mutualize ?

lib/api/controllers/documentController.js
```diff
@@ -69,8 +69,8 @@ class DocumentController extends NativeController {
    */
   async search (request) {
     const { from, size, scrollTTL, searchBody } = request.getSearchParams();
-    const index = request.input.args.index;
-    const collection = request.input.args.collection;
+    const index = request.getIndex(false);
+    const collection = request.getCollection(false);
```
OlivierCavadenti, line 73: not quite high with boolean in function that change the behavior... When this is the case, it's show that you need another function in reality.
Maybe do another function (getRequiredIndex) ?

ChillPC: With a default struct with the field `required`, it would achieve the same goal in a clean way I think

OlivierCavadenti (approved):

Aschen (approved):
After this comment, LGTM

Shiranuit (approved):

PR #2354, opened 2022-05-09
Split jobs doc

clementbolin:
## What does this PR do ?

Splitting of CI jobs into a new file.

**Why ?**
During a PR of a 2-dev branch we have seen this because the doc deployment jobs are skipped automatically because of this lines:
`if: github.event_name != 'pull_request'`
It also allows to have a more readable CI by not having all our jobs in the same file.


.github/workflows/workflow-doc.yaml
```diff
@@ -0,0 +1,99 @@
+name: Workflow Doc
```
alexandrebouthinon, line 1: This workflow also deploy packages and Docker image maybe you could find a better name

clementbolin: Maybe something like: `Workflow deploy` or `Workflow deploy npm / docker / doc` I prefer the first option

alexandrebouthinon: I personally like "workflow deploy" or "Deployments"

Aschen (approved):
👍

PR #2366, opened 2022-07-07
Adding parameters to OpenAPI requests

CustomEntity:
## What does this PR do ?

Add "All" scope for OpenApi requests
Add unit tests for this scope

  - Step 1 : Start Kuzzle instance
  - Step 2 : Send a GET request on https://{baseUrl}:{port}/_openapi?scope=all

### Other changes

See Commit: [#8a7f0c6](https://github.com/kuzzleio/kuzzle/commit/8a7f0c691e48719c17cecd53b594782bb402090d)

### TODO

@MaximCosta is working on "exclude" parameter to exclude tags

### DIFF
--- a/lib/api/controllers/serverController.js
+++ b/lib/api/controllers/serverController.js
@@ -299,8 +299,11 @@ class ServerController extends NativeController {
   async openapi (request) {
     const format = request.getString('format', 'json');
     const scope = request.getString('scope', 'all');
+    const exclude = request.getString('exclude', '').split(',').map(e => e.trim()).filter(e => e);
+    const only = request.getString('only', '').split(',').map(e => e.trim()).filter(e => e);
 
     let definition;
+
 
     if (scope === 'kuzzle') {
       definition = await global.kuzzle.ask('core:api:openapi:kuzzle');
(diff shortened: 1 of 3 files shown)


lib/api/controllers/serverController.js
```diff
@@ -299,9 +299,12 @@ class ServerController extends NativeController {
   async openapi (request) {
     const format = request.getString('format', 'json');
     const scope = request.getString('scope', 'all');
+    const exclude = request.getString('exclude', '').split(',').map(e => e.trim()).filter(e => e);
```
Aschen, line 302: You should use [request.getArray](https://docs.kuzzle.io/core/2/framework/classes/kuzzle-request/get-array/) instead

lib/api/controllers/serverController.js
```diff
@@ -299,9 +299,12 @@ class ServerController extends NativeController {
...
+    const only = request.getString('only', '').split(',').map(e => e.trim()).filter(e => e);
```
Aschen, line 303: You should use [request.getArray](https://docs.kuzzle.io/core/2/framework/classes/kuzzle-request/get-array/) instead

lib/api/controllers/serverController.js
```diff
@@ -315,6 +318,34 @@ class ServerController extends NativeController {
       throw kerror.get('api', 'assert', 'unexpected_argument', scope, ['kuzzle', 'app', 'all']);
     }
 
+    if (exclude.length) {
+      definition.tags = definition.tags.filter(tag => ! exclude.includes(tag.name));
+      Object.entries(definition.paths).forEach(([path, pathDefinition]) => {
```
Aschen, line 323: You should use `for ... of ` loop instead of `Array.forEach`.
`Array.forEach` does not comply with `await` keyword and can leads to unwanted behavior so we stick to the ` for ... of ` loop only

lib/api/controllers/serverController.js
```diff
@@ -315,6 +318,34 @@ class ServerController extends NativeController {
...
+        Object.entries(pathDefinition).forEach(([method, methodDefinition]) => {
+          if (methodDefinition.tags.some(tag => exclude.includes(tag))) {
+            delete pathDefinition[method];
```
Aschen, line 326: Avoid the `delete` keyword because it causes change in the object hidden class tree 

(you can read about [hidden classes(https://www.freecodecamp.org/news/javascript-essentials-why-you-should-know-how-the-engine-works-c2cc0d321553/) and [inline cache](https://mrale.ph/blog/2015/01/11/whats-up-with-monomorphism.html))

Aschen (changes requested):
I don't think it's really useful to filter the output of the OpenAPI response since it will be used in some kind of external tool like Swagger UI (which allows to collapse certain tags) or code generator.

The code added to the codebase does not bring enough value to me to justify it's existence and maintenance.

PR #2367, opened 2022-07-08
Fixed getArray function not working with only one element

CustomEntity:
## What does this PR do ?

Fixes an error that is thrown when a single parameter is input and an array is requested.
This will allow to avoid having to fill at least 2 elements.

  - Step 1 : Start Kuzzle instance
  - Step 2 : Send a request with one array parameter. 
 
Example: 

    https://{baseUrl}:{port}/_openapi?{arrayparameter}=one //NOT WORKING
    https://{baseUrl}:{port}/_openapi?{arrayparameter}=one&{arrayparameter}=two //WORKING

### DIFF
--- a/lib/api/request/kuzzleRequest.ts
+++ b/lib/api/request/kuzzleRequest.ts
@@ -1063,7 +1063,8 @@ export class KuzzleRequest {
           }
         }
         catch (e) {
-          // Do nothing, let the error be thrown below
+          // Handle when only one parameter is provided
+          return [value];
         }
       }
       throw assertionError.get('invalid_type', errorName, 'array');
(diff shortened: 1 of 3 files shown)


lib/api/request/kuzzleRequest.ts
```diff
@@ -1063,7 +1063,8 @@ export class KuzzleRequest {
           }
         }
         catch (e) {
-          // Do nothing, let the error be thrown below
+          // Handle when only one parameter is provided
```
Aschen, line 1066:
```suggestion
          // Handle when only one element is provided
```

Aschen (changes requested):

Aschen:
Closed due to inactivity

PR #2370, opened 2022-07-12
Add eslint-plugin-kuzzle

Aschen:
## What does this PR do ?

Add and use our common eslint plugin


Aschen:
https://github.com/kuzzleio/kuzzle/pull/2371

PR #2375, opened 2022-08-02
Add VirtualIndex to store multiple Kuzzle collection in the same ES index

TisserantG:
## What does this PR do ?
Add virtual tenant 

See functionnal test 

### Other changes

js -> ts for many files


kuzzle:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Found some issues !</td>
    </tr>
  </tbody>
</table>



## Executed Checks

### :x: Changelog Tag Check

Pull Requests must have a `changelog:XXX` label.
If you want to allow this PR to not have a changelog tag, add the following line to the PR body:


```
ci-no-changelog-tag
```

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 4a62a79e2ed3f375fc3c453623d9e19da8007111
</p>

lib/kerror/codes/1-services.json
```diff
@@ -293,6 +293,30 @@
           "message": "Attempt to recreate an existing collection %s of index %s of scope %s with non matching static setting : %s at %s while existing one is at %s",
           "code": 51,
           "class": "BadRequestError"
+        },
+        "delete_virtual_collection": {
+          "description": "Remove virtual collection is forbidden",
+          "message": "You can't remove a virtual collection. ",
```
OlivierCavadenti, line 299: remove whitespace at the end

lib/kuzzle/event/kuzzleEventEmitter.ts
```diff
@@ -45,7 +49,15 @@ class PluginPipeDefinition {
...
-class KuzzleEventEmitter extends EventEmitter {
+export class KuzzleEventEmitter extends EventEmitter {
+  public superEmit: any;
+  public pipeRunner: any;
+  public pluginPipes: any;
+  public pluginPipeDefinitions: any;
+  //eslint-disable-next-line @typescript-eslint/ban-types
+  public corePipes: Map<string, Function[]>;
+  public coreAnswerers: any;
+  public coreSyncedAnswerers: any;
```
OlivierCavadenti, line 60: I suggest to modify types like : 

```
  public superEmit: any;
  public pipeRunner: PipeRunner;
  public pluginPipes: Map<any, any>;
  public pluginPipeDefinitions: Map<any, PluginPipeDefinition>;
  //eslint-disable-next-line @typescript-eslint/ban-types
  public corePipes: Map<string, Function[]>;
  public coreAnswerers: Map<any, any>;
  public coreSyncedAnswerers: Map<any, any>;
```

maybe you can replace more any too

lib/kuzzle/event/kuzzleEventEmitter.ts
```diff
@@ -19,24 +19,28 @@
...
-const Promback = require("../../util/promback");
-const memoize = require("../../util/memoize");
-const PipeRunner = require("./pipeRunner");
-const kerror = require("../../kerror");
+import PipeRunner from "./pipeRunner";
 
 class PluginPipeDefinition {
+  public event: any;
+  public handler: any;
+  public pipeId: any;
```
OlivierCavadenti, line 42: string ?

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,157 @@
...
+    "core:storage:virtualindex:create:after";
+
+  public static deleteVirtualIndexEvent =
+    "core:storage:virtualindex:delete:after";
+
+  constructor() {
+    super("VirtualIndex", global.kuzzle.config.services.storageEngine);
+  }
+
+  public softTenant: Map<string, string> = new Map(); //Key : virtual index, value : real index //TODO rename!
```
OlivierCavadenti, line 19: resolve this TODO before merge ?

test/core/storage/clientAdapter.test.js
```diff
@@ -570,6 +587,8 @@ describe("#core/storage/ClientAdapter", () => {
       });
 
       it('should register a "mappings:import" event', async () => {
+        //sinon.stub(kuzzle, "ask").resolves();
```
OlivierCavadenti, line 590: remove comment

test/core/storage/storageEngine.test.js
```diff
@@ -7,17 +7,35 @@ const { PreconditionError } = require("../../../index");
...
+  });
+
   before(() => {
+    /*
     mockRequire("../../../lib/core/storage/clientAdapter", ClientAdapterMock);
-    StorageEngine = mockRequire.reRequire(
+    ({ StorageEngine } = mockRequire.reRequire(
       "../../../lib/core/storage/storageEngine"
-    );
+    ));
```
OlivierCavadenti, line 29: remove comments

test/kuzzle/kuzzle.test.js
```diff
@@ -12,15 +11,31 @@ const MutexMock = require("../mocks/mutex.mock.js");
...
+class ProcessMock {
+  on() {}
+  exit() {}
+  //removeAllListeners() {}
+  async emit() {}
+  constructor() {
+    sinon.spy(this, "on");
+    sinon.spy(this, "exit");
+    this.removeAllListeners = sinon.spy();
+    //sinon.spy(this, "removeAllListeners");
```
OlivierCavadenti, line 29: remove comments

test/kuzzle/kuzzle.test.js
```diff
@@ -65,17 +86,49 @@ describe("/lib/kuzzle/kuzzle.js", () => {
     };
 
     mockrequire("../../lib/core/cache/cacheEngine", coreModuleStub);
-    mockrequire("../../lib/core/storage/storageEngine", coreModuleStub);
     mockrequire("../../lib/core/security", coreModuleStub);
     mockrequire("../../lib/core/realtime", coreModuleStub);
     mockrequire("../../lib/cluster", clusterModuleStub);
-    mockrequire("../../lib/util/mutex", { Mutex: MutexMock });
+    //mockrequire("../../lib/util/mutex", { Mutex: MutexMock });
```
OlivierCavadenti, line 92: remove comment

test/kuzzle/kuzzle.test.js
```diff
@@ -94,144 +147,121 @@ describe("/lib/kuzzle/kuzzle.js", () => {
 
   describe("#start", () => {
     it("should init the components in proper order", async () => {
-      const Koncorde = sinon.stub();
+      /*
       const stubbedKuzzle = Kuzzle.__with__({
         koncorde_1: { Koncorde },
         vault_1: { default: { load: () => {} } },
       });
+      */
```
OlivierCavadenti, line 155: remove comment

test/kuzzle/kuzzle.test.js
```diff
@@ -94,144 +147,121 @@ describe("/lib/kuzzle/kuzzle.js", () => {
...
-      const Koncorde = sinon.stub();
-      const stubbedKuzzle = Kuzzle.__with__({
-        koncorde_1: { Koncorde },
-        vault_1: { default: { load: () => {} } },
-      });
+      const baseKuzzle = _mockKuzzle();
 
-      await stubbedKuzzle(async () => {
-        const baseKuzzle = _mockKuzzle(Kuzzle);
+      //const baseKuzzle = new Kuzzle();
```
OlivierCavadenti, line 201: remove comment

test/mocks/service/elasticsearchClient.mock.js
```diff
@@ -31,6 +31,7 @@ class ElasticsearchClientMock {
     this.updateByQuery = sinon.stub().resolves();
 
     this.cat = {
+      //aliases: sinon.stub().resolves({ body: [] }),
```
OlivierCavadenti, line 34: remove comment

test/service/storage/elasticsearch.test.js
```diff
@@ -27,13 +29,16 @@ describe("Test: ElasticSearch service", () => {
...
-  let ES;
+  let ES = Elasticsearch;
 
   before(() => {
+    /*
     mockRequire("../../../lib/util/name-generator", {
       randomNumber: randomNumberMock,
     });
     ES = mockRequire.reRequire("../../../lib/service/storage/elasticsearch");
+     */
```
OlivierCavadenti, line 40: remove comment

test/service/storage/elasticsearch.test.js
```diff
@@ -3762,6 +3783,7 @@ describe("Test: ElasticSearch service", () => {
         _kuzzle_info: {
           author: null,
           createdAt: timestamp,
+          index: index,
```
OlivierCavadenti, line 3786: index: index => index

test/service/storage/elasticsearch.test.js
```diff
@@ -3931,6 +3953,7 @@ describe("Test: ElasticSearch service", () => {
     beforeEach(() => {
       kuzzleMeta = {
         _kuzzle_info: {
+          index: index,
```
OlivierCavadenti, line 3956: index: index => index

lib/api/controllers/indexController.js
```diff
@@ -92,8 +112,14 @@ class IndexController extends NativeController {
    */
   async list(request) {
     const countCollection = request.getBoolean("countCollection");
+    const onlyVirtual = request.getBoolean("onlyVirtual");
+    const onlyPhysical = request.getBoolean("onlyPhysical");
```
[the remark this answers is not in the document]
TisserantG: You can't have the two boolean to true, but you can have the two boolean to false.

test/service/storage/elasticsearch.test.js
TisserantG: done

test/service/storage/elasticsearch.test.js
TisserantG: done

test/service/storage/elasticsearch.test.js
TisserantG: done

test/mocks/service/elasticsearchClient.mock.js
TisserantG: done

test/kuzzle/kuzzle.test.js
TisserantG: done

test/kuzzle/kuzzle.test.js
TisserantG: done

test/kuzzle/kuzzle.test.js
TisserantG: done

test/kuzzle/kuzzle.test.js
TisserantG: done

test/core/storage/clientAdapter.test.js
TisserantG: done

lib/kuzzle/event/kuzzleEventEmitter.ts
TisserantG: Done (best effort to find real Class)

lib/kerror/codes/1-services.json
TisserantG: done

lib/core/storage/clientAdapter.ts
```diff
@@ -330,8 +399,19 @@ class ClientAdapter {
...
-      this.cache.listIndexes()
+    global.kuzzle.onAsk(
+      `core:storage:${this.scope}:index:list`,
+      (onlyVirtual?, onlyPhysical?) => {
+        const indexes = this.cache.listIndexes();
+        if (!onlyPhysical && !onlyVirtual) {
+          return indexes;
+        } else if (onlyPhysical && !onlyVirtual) {
+          return indexes.filter((index) => !this.virtualIndex.isVirtual(index));
+        } else if (!onlyPhysical && onlyVirtual) {
```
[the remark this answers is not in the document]
TisserantG: Good question. I think we need case if people send two boolean with value true

lib/service/storage/elasticsearch.ts
```diff
@@ -1787,52 +1882,79 @@ class ElasticSearch extends Service {
...
 
-      if (item.status >= 400) {
-        const error = {
-          _id: item._id,
-          status: item.status,
-        };
+  /**
+   * return the last action
+   */
+  prepareItem(item, actionNames, lastAction, alias, kuzzleMeta) {
```
Aschen, line 1895: Should be private

lib/service/storage/elasticsearch.ts
```diff
@@ -1897,7 +2024,10 @@ class ElasticSearch extends Service {
 
     const aliases = body.map(({ alias }) => alias);
 
-    const schema = this._extractSchema(aliases, { includeHidden: true });
+    const schema = this._extractSchema(aliases, {
+      includeHidden: true,
+      includeVirtual: includeVirtual,
+    }); //TODO => true
```
Aschen, line 2030: What is this todo about?

lib/api/controllers/indexController.js
```diff
@@ -68,8 +76,20 @@ class IndexController extends NativeController {
    */
   async create(request) {
     const index = request.getIndex();
+    const options =
+      request.input.body && request.input.body.options
```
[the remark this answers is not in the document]
TisserantG: I don't know why but Eslint do not accept this optional chaining operation :s

lib/api/controllers/indexController.js
```diff
@@ -29,7 +29,15 @@ const { NativeController } = require("./baseController");
  */
 class IndexController extends NativeController {
   constructor() {
-    super(["create", "delete", "exists", "list", "mDelete", "stats"]);
+    super([
+      "create",
+      "createvirtual",
```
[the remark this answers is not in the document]
TisserantG: Done

lib/cluster/publisher.js
```diff
@@ -210,6 +210,27 @@ class ClusterPublisher {
...
   }
 
+  /**
+   * Publishes an event about a new virtual index being added
+   *
+   * @param  {string} index
+   * @param  {string} virtual
+   * @returns {Long} ID of the message sent
+   */
+  sendAddVirtualIndex(real, virtual) {
```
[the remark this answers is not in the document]
TisserantG: done

lib/service/storage/elasticsearch.ts
```diff
@@ -1966,12 +2099,19 @@ class ElasticSearch extends Service {
...
+  async deleteIndexes(indexes = []): Promise<string[]> {
     if (indexes.length === 0) {
       return Bluebird.resolve([]);
     }
-    const deleted = new Set();
-
+    const deleted = new Set<string>();
+    for (const index of indexes) {
+      if (this.virtualIndex.isVirtual(index)) {
+        //TODO : remove second await?
```
Aschen, line 2109: You can leave it, actually this is really not a critical part of the code since this action is not used very often

lib/service/storage/elasticsearch.ts
```diff
@@ -1966,12 +2099,19 @@ class ElasticSearch extends Service {
...
    * @returns {Promise.<String[]>}
    */
-  async deleteIndexes(indexes = []) {
+  async deleteIndexes(indexes = []): Promise<string[]> {
     if (indexes.length === 0) {
       return Bluebird.resolve([]);
     }
-    const deleted = new Set();
-
+    const deleted = new Set<string>();
```
Aschen, line 2106: I think it could be a simple array here and it will avoid the `Array.from` at the end

lib/service/storage/elasticsearch.ts
```diff
@@ -2088,7 +2231,7 @@ class ElasticSearch extends Service {
     const esRequest = {
       _source: false,
       body: {
-        docs: ids.map((_id) => ({ _id })),
+        docs: ids.map((_id) => ({ _id: this.virtualIndex.getId(index, _id) })), //TODO : test
```
Aschen, line 2234: TODO leftover

lib/service/storage/elasticsearch.ts
```diff
@@ -2856,10 +3000,57 @@ class ElasticSearch extends Service {
...
 
+  /**
+   * Return true if the document is rejected, and add reject reason in rejected array
+   * else, return false
+   * @param document
+   * @param rejected
+   * @param prepareMUpsert
+   * @param requireId
+   */
+  _isRejectedDocument(document, rejected, prepareMUpsert, requireId): boolean {
```
Aschen, line 3014: Should be `private isRejectedDocument`

lib/service/storage/elasticsearch.ts
```diff
@@ -2762,6 +2900,39 @@ class ElasticSearch extends Service {
     };
   }
 
+  processExecutedItem(item, partialErrors, successes, document, source) {
```
Aschen, line 2903: Should be private

lib/service/storage/elasticsearch.ts
```diff
@@ -2918,7 +3116,8 @@ class ElasticSearch extends Service {
    * @throws If there is not exactly one indice associated
    */
   async _getIndice(index, collection) {
```
Aschen, line 3118: All those private method should lost their underscore and gain the `private` modifier

lib/service/storage/elasticsearch.ts
```diff
@@ -3083,7 +3284,7 @@ class ElasticSearch extends Service {
   _extractCollection(alias) {
     const separatorPos = alias.indexOf(NAME_SEPARATOR);
 
-    return alias.substr(separatorPos + 1, alias.length);
+    return alias.substring(separatorPos + 1, alias.length + 1);
```
Aschen, line 3287: Why those changes?

lib/service/storage/elasticsearch.ts
```diff
@@ -3094,38 +3295,55 @@ class ElasticSearch extends Service {
...
+  }
+
   /**
    * Creates the hidden collection on the provided index if it does not already
    * exists
    *
    * @param {String} index Index name
    */
   async _createHiddenCollection(index) {
+    //TODO
```
Aschen, line 3345: Todo leftover

lib/service/storage/elasticsearch.ts
```diff
@@ -3218,9 +3436,48 @@ class ElasticSearch extends Service {
...
+  _sanitizeSearchBodyForVirtualIndex(searchBody, index) {
+    const filteredQuery = {
+      bool: {
+        filter: [],
+        must: {},
+        should: null,
+      },
+    };
+    if (searchBody.query.bool) {
+      filteredQuery.bool = searchBody.query.bool;
```
Aschen, line 3455: You should make a deep copy with `JSON.parse(JSON.stringify())` here to avoid mutate the original query

lib/service/storage/elasticsearch.ts
```diff
@@ -3218,9 +3436,48 @@ class ElasticSearch extends Service {
...
     }
 
+    if (index && this.virtualIndex.isVirtual(index)) {
+      this._sanitizeSearchBodyForVirtualIndex(searchBody, index);
+    }
+
     return searchBody;
   }
 
+  _sanitizeSearchBodyForVirtualIndex(searchBody, index) {
```
Aschen, line 3446: I think this method should be tested unitary with as most as possible usecases. Also, it should be in the `VirtualIndex` service

lib/service/storage/elasticsearch.ts
```diff
@@ -3218,9 +3436,48 @@ class ElasticSearch extends Service {
...
+    };
+    if (searchBody.query.bool) {
+      filteredQuery.bool = searchBody.query.bool;
+      if (!filteredQuery.bool.filter) {
+        filteredQuery.bool.filter = []; //add query bool in filter
+      } else if (!Array.isArray(filteredQuery.bool.filter)) {
+        filteredQuery.bool.filter = [filteredQuery.bool.filter];
+      }
+    } else {
+      filteredQuery.bool.must = searchBody.query;
```
Aschen, line 3462: I think this create a circular dependency no?

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,157 @@
+import Crypto from "crypto";
+
+import { Service } from "../service";
+
+export class VirtualIndex extends Service {
+  public static createEvent = "core:storage:virtualindex:cache:create";
```
Aschen, line 6: We use kebab case in events name (apart from public events because we don't want breaking)
```suggestion
  public static createEvent = "core:storage:virtual-index:cache:create";
```

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  constructor() {
+    super("VirtualIndex", global.kuzzle.config.services.storageEngine);
+  }
+
+  public softTenant: Map<string, string> = new Map(); //Key : virtual index, value : real index //TODO rename!
+
+  async init(): Promise<void> {
+    await this.buildCollection();
+    await this.initVirtualTenantList();
```
Aschen, line 23: The methods in this class should not use the term `tenant`

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    global.kuzzle.emit(VirtualIndex.createVirtualIndexEvent, {
+      real: index,
+      virtual: virtualIndex,
+    });
+
+    this.softTenant.set(virtualIndex, index);
+    await global.kuzzle.ask(
+      "core:storage:private:document:create",
+      "virtualindexes",
```
Aschen, line 82: words should be separated by `-`, `_` or `.`

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,157 @@
...
+    await global.kuzzle.ask(
+      "core:storage:private:document:delete",
+      "virtualindexes",
+      "list",
+      id
+    );
+  }
+
+  async initVirtualTenantList() {
+    //TODO : from database
```
Aschen, line 105: todo leftover

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,157 @@
...
+    if (this.softTenant.size === 0) {
+      this.softTenant = new Map<string, string>();
```
Aschen: Use a guard and return early

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  async buildCollection() {
+    try {
+      await global.kuzzle.ask(
+        "core:storage:private:index:create",
+        "virtualindexes",
+        {}
+      );
+    } catch (e) {
+      //already created
```
Aschen, line 135: Explicitely catch the corresponding error code and throw if it's not the expected one

lib/types/config/storageEngine/StorageEngineElasticsearchConfiguration.ts
```diff
@@ -297,6 +310,13 @@ export type StorageEngineElasticsearch = {
           };
         };
       };
+      "virtual-indexes": {
+        dynamic: "strict";
+        properties: {
+          real: { type: "text" };
+          virtual: { type: "text" };
```
Aschen, line 317:
```suggestion
        dynamic: "strict"|"false"|"true";
        properties: {
          physical: { type: string };
          virtual: { type: string };
```

Aschen (changes requested):
From what I see there is a lack of support of the `m*` method (`mCreate` etc)

Can you add at least 2 functional tests to ensure they are working? One with `mCreate` and one with any other action (because mCreate have optional IDs while all the other works the same with IDs)

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  getId(index: string, id: string): string {
+    if (this.isVirtual(index)) {
+      return index + id;
+    }
+    return id;
+  }
+
+  randomString(size = 20) {
+    return Crypto.randomBytes(size).toString("base64").slice(0, size);
```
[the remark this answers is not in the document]
Shiranuit: > It would be better to generate sequentials IDs so they will be sorted by default by ES directly (and thus in the Admin Console for example)

I'm not  really sure about that since having sequential ID's when not having any form of transaction in ES could result in the same ID generated twice, which could lead to two nodes each creating a document in the same virtual collection and both having the same id which will cause the creation of the document to fail in one of the two nodes.

That said I'm not sure the usage of Crypto is really necessary, there is no sensitive information and crypto is a bit slow but that's not what will bottleneck Kuzzle, so I'm not against it.
Just for reference
```bash
Crypto x 632,951 ops/sec ±0.83% (90 runs sampled)
Without Crypto x 3,115,597 ops/sec ±0.33% (93 runs sampled)
```

**Code without crypto**
```js
  const buf = Buffer.alloc(20);
  for (let i = 0 ; i < 20; i++) {
    buf.writeUInt8((Math.random() * 255)&255, i);
  }
  const id = buf.toString('base64').slice(0, 20);
```

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,157 @@
...
+      "virtualindexes",
+      "list",
+      id
+    );
+  }
+
+  async initVirtualTenantList() {
+    //TODO : from database
+    if (this.virtualIndexMap.size === 0) {
+      this.virtualIndexMap = new Map<string, string>();
```
Shiranuit, line 107: Why are you recreating a new map if the one actually used is empty ?

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,157 @@
...
+      let from = 0;
+      let total;
+
+      do {
+        const list = await global.kuzzle.ask(
+          "core:storage:private:document:search",
+          "virtualindexes",
+          "list",
+          { from: from, size: 100 }
```
Shiranuit, line 116: Doing a search using `from` + `size` is limited to the first 10000 elements (by default), if there is more than 10000 virtual index this could be a problem.
That said 10000 is already a big number of virtual indexes, I dont know if there will be cases of that happening but it's good to keep in mind.

Shiranuit (changes requested):

lib/kuzzle/kuzzle.ts
```diff
@@ -147,7 +147,7 @@ class Kuzzle extends KuzzleEventEmitter {
   /**
    * Vault component (will be initialized after bootstrap)
    */
-  private vault: vault;
+  private _vault: vault;
```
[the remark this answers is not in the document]
TisserantG: Because it's instancied in 'start()' function and not in constructor.

lib/service/storage/elasticsearch.ts
TisserantG: substr is deprecated

lib/service/storage/elasticsearch.ts
TisserantG: I don't understand why.

lib/core/storage/clientAdapter.ts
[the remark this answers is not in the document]
TisserantG: Done (replace booleans by "type")

TisserantG:
> From what I see there is a lack of support of the `m*` method (`mCreate` etc)
> 
> Can you add at least 2 functional tests to ensure they are working? One with `mCreate` and one with any other action (because mCreate have optional IDs while all the other works the same with IDs)

Done

lib/api/controllers/indexController.js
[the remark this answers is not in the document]
TisserantG: Done

lib/api/controllers/indexController.js
```diff
@@ -68,8 +76,20 @@ class IndexController extends NativeController {
    */
   async create(request) {
     const index = request.getIndex();
+    const options =
```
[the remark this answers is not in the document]
TisserantG: done

lib/api/controllers/indexController.js
```diff
@@ -68,8 +76,20 @@ class IndexController extends NativeController {
...
+      request.input.body && request.input.body.options
+        ? request.input.body.options
+        : {};
+    await this.ask("core:storage:public:index:create", index, options);
+  }
 
-    await this.ask("core:storage:public:index:create", index);
+  async createvirtual(request) {
+    const index = request.getIndex();
+    const options =
```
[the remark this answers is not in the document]
TisserantG: done

lib/api/controllers/indexController.js
```diff
@@ -92,8 +112,14 @@ class IndexController extends NativeController {
    */
   async list(request) {
     const countCollection = request.getBoolean("countCollection");
+    const onlyVirtual = request.getBoolean("onlyVirtual");
```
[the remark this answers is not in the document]
TisserantG: done.

lib/cluster/node.js
```diff
@@ -871,6 +872,15 @@ class ClusterNode {
       this.onIndexAdded(scope, index)
     );
 
+    global.kuzzle.on(
+      VirtualIndex.createVirtualIndexEvent,
+      ({ real, virtual }) => this.onVirtualIndexAdded(real, virtual)
```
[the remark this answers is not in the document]
TisserantG: done

lib/cluster/protobuf/sync.proto
```diff
@@ -159,6 +159,17 @@ message AddCollection {
   string collection = 4;
 }
 
+message AddVirtualIndex {
+      uint64 messageId = 1;
+      string virtual = 2;
+      string real = 3;
```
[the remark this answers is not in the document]
TisserantG: done

lib/cluster/protobuf/sync.proto
```diff
@@ -159,6 +159,17 @@ message AddCollection {
...
+}
+
+message RemoveVirtualIndex {
+    uint64 messageId = 1;
+    string virtual = 2;
```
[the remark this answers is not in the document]
TisserantG: done

lib/cluster/publisher.js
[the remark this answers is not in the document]
TisserantG: done

lib/cluster/subscriber.js
```diff
@@ -643,6 +646,30 @@ class ClusterSubscriber {
...
   }
 
+  /**
+   * Adds a new virtual index in the map of virtual index
+   *
+   * @param {Object} message
+   * @return {void}
+   */
+  async handleVirtualIndexAddition(message) {
+    const { real, virtual } = message;
```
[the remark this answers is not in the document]
TisserantG: done

lib/config/default.config.ts
```diff
@@ -418,6 +419,13 @@ const defaultConfig: KuzzleConfiguration = {
               },
             },
           },
+          "virtual-indexes": {
+            dynamic: "strict",
+            properties: {
+              real: { type: "text" },
+              virtual: { type: "text" },
```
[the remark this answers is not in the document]
TisserantG: done

lib/core/storage/clientAdapter.ts
```diff
@@ -84,16 +100,62 @@ class ClientAdapter {
...
+      physicalIndex: null,
+      propagate,
+    });
+  }
+
+  async createVirtualIndex(
+    index: string,
+    { indexCacheOnly = false, propagate = true, physicalIndex = null } = {}
+  ) {
+    if (!physicalIndex) {
```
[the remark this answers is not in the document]
TisserantG: done

lib/core/storage/clientAdapter.ts
```diff
@@ -84,16 +100,62 @@ class ClientAdapter {
...
+    index: string,
+    { indexCacheOnly = false, propagate = true, physicalIndex = null } = {}
+  ) {
     if (this.cache.hasIndex(index)) {
       throw servicesError.get("index_already_exists", this.scope, index);
     }
 
-    if (!indexCacheOnly) {
-      await this.client.createIndex(index);
+    if (physicalIndex) {
```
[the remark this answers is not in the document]
TisserantG: done

lib/core/storage/clientAdapter.ts
```diff
@@ -84,16 +100,62 @@ class ClientAdapter {
...
+      if (!this.cache.hasIndex(physicalIndex)) {
+        throw new KuzzleError("The given physicalIndex does not exist.", 404);
```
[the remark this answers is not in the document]
TisserantG: done

lib/core/storage/clientAdapter.ts
```diff
@@ -84,16 +100,62 @@ class ClientAdapter {
...
+      }
     }
-
+    this.client._assertValidIndexAndCollection(index);
     this.cache.addIndex(index);
+    if (physicalIndex) {
+      this.client.createVirtualIndex(index, physicalIndex);
```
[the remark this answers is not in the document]
TisserantG: done

lib/core/storage/clientAdapter.ts
```diff
@@ -330,8 +399,19 @@ class ClientAdapter {
      * Return a list of all indexes within this adapter's scope
      * @returns {string[]}
      */
-    global.kuzzle.onAsk(`core:storage:${this.scope}:index:list`, () =>
-      this.cache.listIndexes()
+    global.kuzzle.onAsk(
+      `core:storage:${this.scope}:index:list`,
+      (onlyVirtual?, onlyPhysical?) => {
```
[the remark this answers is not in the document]
TisserantG: done

lib/core/storage/storageEngine.ts
```diff
@@ -0,0 +1,79 @@
...
+  get private() {
+    return this.privateClient;
+  }
+
+  // to be mocked in Unit Test
+  static initClientAdapters(scopeEnum: string, virtualIndex: VirtualIndex) {
+    return new ClientAdapter(scopeEnum, virtualIndex);
+  }
+
+  constructor(virtualIndex: VirtualIndex) {
```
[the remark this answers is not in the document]
TisserantG: Done

lib/kerror/codes/1-services.json
```diff
@@ -293,6 +293,30 @@
           "message": "Attempt to recreate an existing collection %s of index %s of scope %s with non matching static setting : %s at %s while existing one is at %s",
           "code": 51,
           "class": "BadRequestError"
+        },
+        "delete_virtual_collection": {
```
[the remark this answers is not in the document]
TisserantG: IMHO have different error code is always better to test and to use the API.

lib/kuzzle/event/kuzzleEventEmitter.ts
```diff
@@ -139,13 +151,14 @@ class KuzzleEventEmitter extends EventEmitter {
...
    * @param  {*} data
    */
-  emit(event, data) {
+  emit(event, data): boolean {
     const events = getWildcardEvents(event);
     debug('Triggering event "%s" with data: %o', event, data);
 
-    for (let i = 0; i < events.length; i++) {
-      super.emit(events[i], data);
+    for (const subevent of events) {
```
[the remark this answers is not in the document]
TisserantG: Done

lib/kuzzle/event/kuzzleEventEmitter.ts
```diff
@@ -183,8 +196,8 @@ class KuzzleEventEmitter extends EventEmitter {
     const events = getWildcardEvents(event);
     const funcs = [];
 
-    for (let i = 0; i < events.length; i++) {
-      const targets = this.pluginPipes.get(events[i]);
+    for (const subevent of events) {
```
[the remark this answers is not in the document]
TisserantG: done

lib/kuzzle/kuzzle.ts
```diff
@@ -100,29 +100,29 @@ type ImportStatus = {
   firstCall?: boolean;
 };
 
-class Kuzzle extends KuzzleEventEmitter {
-  private config: KuzzleConfiguration;
+export class Kuzzle extends KuzzleEventEmitter {
+  readonly config: KuzzleConfiguration;
```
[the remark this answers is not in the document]
TisserantG: done

lib/kuzzle/kuzzle.ts
```diff
@@ -230,32 +268,38 @@ class Kuzzle extends KuzzleEventEmitter {
...
+
       await this.internalIndex.init();
 
-      await new SecurityModule().init();
+      await Kuzzle.initSecurityModule();
 
       // This will init the cluster module if enabled
       this.id = await this.initKuzzleNode();
 
+      await virtualIndex.init();
```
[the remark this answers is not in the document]
TisserantG: Done

lib/service/storage/elasticsearch.ts
```diff
@@ -19,27 +19,32 @@
...
-const extractFields = require("../../util/extractFields");
-const { Mutex } = require("../../util/mutex");
-const { randomNumber } = require("../../util/name-generator");
+import { VirtualIndex } from "./virtualIndex";
+import { ScopeEnum } from "../../core/storage/storeScopeEnum";
+import { Client } from "@elastic/elasticsearch";
+
+import assert from "assert";
+import _ from "lodash";
+//import { Client as StorageClient } from "@elastic/elasticsearch";
```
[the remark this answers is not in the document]
TisserantG: done

lib/service/storage/elasticsearch.ts
```diff
@@ -62,6 +67,8 @@ const NAME_SEPARATOR = ".";
 const FORBIDDEN_CHARS = `\\/*?"<>| \t\r\n,+#:${NAME_SEPARATOR}${PUBLIC_PREFIX}${PRIVATE_PREFIX}`;
 const DYNAMIC_PROPERTY_VALUES = ["true", "false", "strict"];
 
+const debug = new DebugBuilder("kuzzle:services:elasticsearch");
```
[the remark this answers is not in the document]
TisserantG: done

lib/service/storage/elasticsearch.ts
```diff
@@ -97,15 +99,35 @@ class ElasticSearch extends Service {
       return { promise, reject, resolve };
     };
 
-    return new StorageClient({ defer, ...config });
+    return new Client({ defer, ...config });
+    //return new StorageClient({ defer, ...config });
```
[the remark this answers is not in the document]
TisserantG: done

lib/service/storage/elasticsearch.ts
```diff
@@ -97,15 +99,35 @@ class ElasticSearch extends Service {
...
   }
 
-  constructor(config, scope = scopeEnum.PUBLIC) {
-    super("elasticsearch", config);
+  _client: Client;
```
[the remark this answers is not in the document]
TisserantG: Done

lib/service/storage/elasticsearch.ts
```diff
@@ -367,12 +405,15 @@ class ElasticSearch extends Service {
    * @param {String} index - Index name
    * @param {String} collection - Collection name
    * @param {Object} searchBody - Search request body (query, sort, etc.)
+   * @param target contain index and collections for multisearch. If target is not null, index and collection sould be null.
```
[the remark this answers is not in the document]
TisserantG: done

lib/service/storage/virtualIndex.ts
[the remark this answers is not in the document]
TisserantG: done

lib/service/storage/elasticsearch.ts
```diff
@@ -921,6 +977,7 @@ class ElasticSearch extends Service {
     esRequest.body._kuzzle_info = {
       author: getKuid(userId),
       createdAt: Date.now(),
+      index: index,
```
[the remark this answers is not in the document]
TisserantG: done

lib/service/storage/elasticsearch.ts
```diff
@@ -1345,8 +1413,19 @@ class ElasticSearch extends Service {
   async createCollection(
     index,
     collection,
-    { mappings = {}, settings = {} } = {}
+    {
+      mappings = {
```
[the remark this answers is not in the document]
TisserantG: done

lib/service/storage/elasticsearch.ts
```diff
@@ -1484,8 +1561,14 @@ class ElasticSearch extends Service {
   async updateCollection(
     index,
     collection,
-    { mappings = {}, settings = {} } = {}
+    {
+      mappings = { _meta: null, dynamic: null, properties: null },
```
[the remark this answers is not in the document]
TisserantG: done

lib/service/storage/elasticsearch.ts
TisserantG: done

lib/service/storage/elasticsearch.ts
TisserantG: ok

lib/service/storage/elasticsearch.ts
TisserantG: done

lib/service/storage/elasticsearch.ts
TisserantG: removed

lib/service/storage/elasticsearch.ts
TisserantG: done

lib/service/storage/elasticsearch.ts
TisserantG: done

lib/service/storage/elasticsearch.ts
TisserantG: Done for most of them, BUT some methods exist both with and without underscore, so I do not rename them.

lib/service/storage/elasticsearch.ts
```diff
@@ -2918,7 +3116,8 @@ class ElasticSearch extends Service {
    * @throws If there is not exactly one indice associated
    */
   async _getIndice(index, collection) {
-    const alias = `${ALIAS_PREFIX}${this._indexPrefix}${index}${NAME_SEPARATOR}${collection}`;
+    const realIndex = this.virtualIndex.getRealIndex(index);
```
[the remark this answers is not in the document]
TisserantG: done

lib/service/storage/elasticsearch.ts
TisserantG: done

lib/service/storage/elasticsearch.ts
TisserantG: Done

lib/service/storage/elasticsearch.ts
TisserantG: Right, I will correct that!

lib/service/storage/virtualIndex.ts
TisserantG: done

lib/service/storage/virtualIndex.ts
TisserantG: done

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  getVirtualId(index: string, id: string): string {
+    if (this.isVirtual(index) && id.startsWith(index)) {
+      return id.substring(index.length, id.length);
+    }
+    return id;
+  }
+
+  public async createVirtualIndex(virtualIndex: string, index: string) {
+    //TODO : cluster //TODO : throw exception if "index" is virtual
```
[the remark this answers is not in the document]
TisserantG: done

lib/service/storage/virtualIndex.ts
TisserantG: done

lib/service/storage/virtualIndex.ts
TisserantG: done

lib/service/storage/virtualIndex.ts
TisserantG: done

lib/service/storage/virtualIndex.ts
TisserantG: done

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,157 @@
...
+            dynamic: "strict",
+            properties: {
+              real: { type: "text" },
+              virtual: { type: "text" },
+            },
+          },
+        }
+      );
+    } catch (e) {
+      //already created
```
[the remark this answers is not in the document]
TisserantG: done

lib/types/config/storageEngine/StorageEngineElasticsearchConfiguration.ts
TisserantG: done

lib/service/storage/virtualIndex.ts
TisserantG: Correction done

lib/service/storage/elasticsearch.ts
TisserantG: If I remember will, we said that we do not continue to create unit test. 
Is it ok if I test all possible usecase with functionnal test?

lib/service/storage/elasticsearch.ts
TisserantG: Undo : That can create a fail in an unit test because it return many time the same name in the deleted array.

lib/api/controllers/indexController.js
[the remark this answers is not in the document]
Aschen: This only works in Typescript unfortunately

lib/api/controllers/indexController.js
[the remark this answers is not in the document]
Aschen: What about having a `filter` parameter:
 - if it's `virtual` then list only virtual
 - if it's `physical` then list only physical
 - if it's not set, list everything
 
 wdyt?

lib/service/storage/elasticsearch.ts
Aschen: In some case unit tests may be more easy to write and fast to execute. This is a good example because you would have to try many differents search requests as input and doing so with functionnal tests would take a lot of extra time:
 - having to send network request
 - having to play it 6 times in total: Node.js 14 and 16 with HTTP, WebSocket and MQTT

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,200 @@
...
+    global.kuzzle.onAsk(VirtualIndex.createEvent, (physical, virtual) =>
+      this.addInSoftIndexMap(physical, virtual)
+    );
+
+    global.kuzzle.onAsk(VirtualIndex.deleteEvent, (info) =>
+      this.removeInSoftIndexMap(info)
+    );
+  }
+
+  addInSoftIndexMap(physical, virtual) {
```
Aschen, line 32: Nitpicking: here the naming should be `addInVirtualIndexMap` instead

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,200 @@
...
+    this.virtualIndexMap.set(virtual, physical);
+  }
+
+  removeInSoftIndexMap(notification) {
```
Aschen, line 36: Nitpicking: here the naming should be `removeInVirtualIndexMap` instead

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,200 @@
...
+    this.virtualIndexMap.delete(notification);
+  }
+
+  getRealIndex(name: string): string {
```
Aschen, line 40: Nitpicking: here the naming should be `getPhysicalIndex`

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,200 @@
...
+  getId(index: string, id: string): string {
+    if (this.isVirtual(index)) {
+      return index + id;
+    }
+    return id;
+  }
+
+  randomString(size = 20) {
+    const buf = Buffer.alloc(size);
+    for (let i = 0; i < 20; i++) {
```
Aschen, line 60:
```suggestion
    for (let i = 0; i < size; i++) {
```

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,200 @@
...
+      "core:storage:private:document:create",
+      "virtual-indexes",
+      "list",
+      { physical: index, virtual: virtualIndex },
+      { id: index + virtualIndex }
+    );
+  }
+
+  async removeVirtualIndex(index: string) {
+    const realIndex = this.virtualIndexMap.get(index);
```
Aschen, line 90: Nit: should be `physicalIndex`

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,200 @@
...
+    this.virtualIndexMap = new Map<string, string>();
+    let from = 0;
+    let total;
+
+    do {
+      const list = await global.kuzzle.ask(
+        "core:storage:private:document:search",
+        "virtual-indexes",
+        "list",
+        { from: from, size: 100 }
```
Aschen, line 117:
```suggestion
        { from, size: 100 }
```

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,200 @@
...
+    await global.kuzzle.ask(
+      "core:storage:private:document:delete",
+      "virtual-indexes",
+      "list",
+      id
+    );
+  }
+
+  async initVirtualIndexList() {
+    if (this.virtualIndexMap.size === 0) {
```
Aschen, line 105: I believe that if you are returning when it's empty then when Kuzzle starts you are not going to fill the `virtualIndexMap`

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,200 @@
...
+      );
+      total = list.total;
+      for (const hit of list.hits) {
+        this.virtualIndexMap.set(hit._source.virtual, hit._source.physical);
+      }
+      from += 100;
+    } while (from < total);
+  }
+
+  async buildCollection() {
```
Aschen, line 127: Rather than creating a new internal index, you should create your collection in the internal `kuzzle` index

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,200 @@
...
+        "core:storage:private:collection:create",
+        "virtual-indexes",
+        "list",
+        {
+          mappings: {
+            _meta: undefined,
+            dynamic: "strict",
+            properties: {
+              physical: { type: "text" },
+              virtual: { type: "text" },
```
Aschen, line 151:
```suggestion
              physical: { type: "keyword" },
              virtual: { type: "keyword },
```

package.json
```diff
@@ -41,6 +43,8 @@
   },
   "dependencies": {
     "@elastic/elasticsearch": "https://github.com/elastic/elasticsearch-js/archive/refs/tags/v7.13.0.tar.gz",
+    "@types/bluebird": "^3.5.36",
+    "@types/js-yaml": "^4.0.5",
```
Aschen, line 47: Those should be devDependencies

Aschen (changes requested):

lib/service/storage/virtualIndex.ts
TisserantG: Done

lib/service/storage/virtualIndex.ts
TisserantG: Done

lib/service/storage/virtualIndex.ts
TisserantG: Done

lib/service/storage/virtualIndex.ts
TisserantG: Done

lib/service/storage/virtualIndex.ts
TisserantG: Done

TisserantG: Done

lib/service/storage/virtualIndex.ts
TisserantG: Done

lib/service/storage/virtualIndex.ts
TisserantG: Done

lib/service/storage/virtualIndex.ts
TisserantG: Done

lib/service/storage/virtualIndex.ts
TisserantG: Done

package.json
TisserantG: Done

lib/core/storage/clientAdapter.ts
```diff
@@ -19,36 +19,51 @@
...
 /**
  * Storage client adapter to perform validation on index/collection existence
  * and to maintain the index/collection cache.
  */
-class ClientAdapter {
+export class ClientAdapter {
+  client: Elasticsearch;
+  cache: IndexCache;
+  scope: string;
+  virtualIndex: VirtualIndex;
```
Aschen, line 39: Those should be private

lib/core/storage/clientAdapter.ts
```diff
@@ -84,17 +99,45 @@ class ClientAdapter {
...
+        index,
+        scope: this.scope,
+      });
+    }
+  }
 
-    this.cache.addIndex(index);
+  async createVirtualIndex(
+    index: string,
+    physicalIndex = null,
```
Aschen, line 120: You shouldn't have a default value to null if it's not allowed (see the error l124)

lib/core/storage/clientAdapter.ts
```diff
@@ -178,8 +230,8 @@ class ClientAdapter {
    *
    * @returns {Promise}
    */
-  async populateCache() {
-    const schema = await this.client.getSchema();
+  async populateCache(includeVirtual) {
```
Aschen, line 233: It would be better to have a named property here to give context when calling the function:
- `populateCache({ includeVirtual: true })` instead of `populateCache(true)`

lib/core/storage/clientAdapter.ts
```diff
@@ -330,9 +389,15 @@ class ClientAdapter {
      * Return a list of all indexes within this adapter's scope
      * @returns {string[]}
      */
-    global.kuzzle.onAsk(`core:storage:${this.scope}:index:list`, () =>
-      this.cache.listIndexes()
-    );
+    global.kuzzle.onAsk(`core:storage:${this.scope}:index:list`, (type?) => {
```
Aschen, line 392:
```suggestion
    global.kuzzle.onAsk(`core:storage:${this.scope}:index:list`, (type: string) => {
```

lib/core/storage/storageEngine.ts
```diff
@@ -0,0 +1,92 @@
...
+    const privateIndexes = this.privateClient.cache.listIndexes();
+
+    await this.virtualIndex.init();
+    await Promise.all([
+      this.publicClient.populateCache(true),
+      this.privateClient.populateCache(true),
+    ]);
+
+    //await global.kuzzle.ask("core:storage:private:cache:refresh");
+    //await global.kuzzle.ask("core:storage:public:cache:refresh");
```
Aschen, line 83: Those comments should be removed

lib/kuzzle/kuzzle.ts
```diff
@@ -230,32 +272,34 @@ class Kuzzle extends KuzzleEventEmitter {
...
         maxConditions: this.config.limits.subscriptionConditionsCount,
         regExpEngine: this.config.realtime.pcreSupport ? "js" : "re2",
         seed: this.config.internal.hash.seed,
       });
 
-      await new CacheEngine().init();
-      await new StorageEngine().init();
+      await Kuzzle.initCacheEngine();
+
+      const storageEngine = Kuzzle.createStorageEngine();
```
Aschen, line 287: Why not having only one function call like `initCacheEngine` ?

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,190 @@
...
+    "core:storage:virtual-index:create:after";
+
+  public static deleteVirtualIndexEvent =
+    "core:storage:virtual-index:delete:after";
+
+  constructor() {
+    super("VirtualIndex", global.kuzzle.config.services.storageEngine);
+  }
+
+  public virtualIndexMap: Map<string, string> = new Map<string, string>(); //Key : virtual index, value : real index
```
Aschen, line 17: Comment should be js-doc so they appear in the IDE

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  getId(index: string, id: string): string {
+    if (this.isVirtual(index)) {
+      return index + id;
+    }
+    return id;
+  }
+
+  randomString(size = 20) {
```
Aschen, line 58: This should rather be in the `NameGenerator` util

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  getVirtualId(index: string, id: string): string {
+    if (this.isVirtual(index) && id.startsWith(index)) {
+      return id.substring(index.length, id.length);
+    }
+    return id;
+  }
+
+  public async createVirtualIndex(virtualIndex: string, index: string) {
+    global.kuzzle.emit(VirtualIndex.createVirtualIndexEvent, {
```
Aschen, line 74: I think the event should be dispatched after the successful creation in the `virtual-indexes` collection top avoid unconsistent states

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,190 @@
...
+      "kuzzle",
+      "virtual-indexes",
+      { physical: index, virtual: virtualIndex },
+      { id: index + virtualIndex }
+    );
+  }
+
+  async removeVirtualIndex(index: string) {
+    const physicalIndex = this.virtualIndexMap.get(index);
+    global.kuzzle.emit(VirtualIndex.deleteVirtualIndexEvent, {
```
Aschen, line 91: Same comment about the event after deletion

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,190 @@
...
+    await global.kuzzle.ask(
+      "core:storage:private:document:delete",
+      "kuzzle",
+      "virtual-indexes",
+      id
+    );
+  }
+
+  async initVirtualIndexList() {
+    //this.virtualIndexMap = new Map<string, string>();
```
Aschen, line 105: Comment should be removed

lib/service/storage/virtualIndex.ts
```diff
@@ -0,0 +1,190 @@
...
+            dynamic: "strict",
+            properties: {
+              physical: { type: "keyword" },
+              virtual: { type: "keyword" },
+            },
+          },
+        }
+      );
+    } catch (e) {
+      if (e.status !== 412) {
```
Aschen, line 147: You should use the correct error code instead of the error status. They are more specific and also more explicit

Aschen (changes requested):

lib/service/storage/virtualIndex.ts
Shiranuit: Use a scroll instead, but only if there is more than 10000 documents

lib/core/storage/clientAdapter.ts
```diff
@@ -19,36 +19,51 @@
...
  * and to maintain the index/collection cache.
  */
-class ClientAdapter {
+export class ClientAdapter {
+  client: Elasticsearch;
+  cache: IndexCache;
+  scope: string;
+  virtualIndex: VirtualIndex;
+
+  static getMutex() {
```
Shiranuit, line 41: I will re state what I've said before
> I dont think this a good design to create a static function with a generic  name to make a mutex,
mutexes have specifics names that are used to lock specific actions / ressources but having something like `ClientAdapter.getMutex()` might lead someone to think that if a mutex is needed in one of the ClientAdapter function they should use this method to acquire the mutex, which is wrong because the mutex will lock a ressource with the same name (`loadMappings`) which might be totally unrelated to mappings loading and impact an other node that might be starting and checking if one of the node is finishing or has already finished the initialization sequence.

You can create a generic `getMutex` which take mutex name and mutex options this way you keep it clear when acquiring mutexes and you can mock it without difficulty

**example**
```ts
  static getMutex(mutexName: string, options: MutexOptions) {
```

Shiranuit (changes requested):

Shiranuit:
There is also a lack of documentation on this feature

lib/kuzzle/event/kuzzleEventEmitter.ts
```diff
@@ -139,13 +152,14 @@ class KuzzleEventEmitter extends EventEmitter {
    * @param  {string} event
    * @param  {*} data
    */
-  emit(event, data) {
+  emit(event, data): boolean {
```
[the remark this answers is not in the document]
TisserantG: To be compliant with the  superclass. If I remove the return, I have : 

Type '(event: any, data: any) => void' is not assignable to type '<T extends string | symbol>(event: T, ...args: any[]) => boolean'.

Aschen:
Closed due to inactivity and de-priorized feature

Issue #2382, opened 2022-08-23
Add strong typing to pipe and hook registering


Aschen:
## Feature Description

In order to ensure a correct usage of pipe and hooks plugged to the event system, we should be able to ensure the interface contract is respected.

Events should provide and expose a type for their handler and strong typing should ensure the good handler is registered.

## Example Use Case

Given this event, we should declare the associated handler:
```js
export type EventWorkflowErrorHandler = (error: Error, message: string) => Promise<void>

this.app.trigger('workflows:execution:error', new Error('BAD'), 'Something bad happen bru');
```

Then, we should register a hook on the event like this:

```js
this.app.hook.register<EventWorkflowErrorHandler>('workflows:execution:error', (error: Error, message: string) => {
  console.log(error, message)
});
```

PR #2383, opened 2022-08-23
Strong typing for pipe and hooks

Aschen:
## What does this PR do ?

Implements https://github.com/kuzzleio/kuzzle/issues/2382

This PR allows to specify type when registering pipes or hooks.

This can be used to have strong interface contracts between modules emitting events and modules listening to those events.

Modules emitting events should define the type of the arguments received by a listening handler function:

```js
  type EventFoobar = {
    name: 'event:foobar';

    args: [number, string];
  }

  await app.trigger<EventFoobar>('event:foobar', 30, 'Tuan');
```

On the other hand, the listening module can use this type to ensure that the handler will have the correct arguments:
```js
app.pipe.register<EventFoobar>('event:foobar', async (age: number, name: string) => {
  return age;
});
```

### Other changes

 - The type for pipes handler function now ensure that the first argument is returned in a promise as expected
 - also did this for cluster events
 - add types for generic events
```js
app.pipe.register<EventGenericDocumentBeforeUpdate<PersonContent>>(
  'generic:document:beforeUpdate',
  async (documents: KDocument<PersonContent>[], request: KuzzleRequest) => {
    return documents
  });
```

### DIFF
--- a/doc/2/framework/types/event-handler/index.md
+++ b/doc/2/framework/types/event-handler/index.md
@@ -7,7 +7,7 @@ description: EventHandler type definition
 
 # EventHandler
 
-The `EventHandler` type represents a handler function used with a pipe or a hook to listen to events.
+The `XXXEventHandler` typse represents a handler function used with a pipe or a hook to listen to events.
 
 See [Event System](/core/2/guides/develop-on-kuzzle/event-system)
 
@@ -16,7 +16,7 @@ See [Event System](/core/2/guides/develop-on-kuzzle/event-system)
 **Example:**
 
 ```js
-import { EventHandler, KuzzleRequest } from 'kuzzle';
+import { PipeEventHandler, KuzzleRequest } from 'kuzzle';
 
-const pipeHandler: EventHandler = async (request: KuzzleRequest) => request
+const pipeHandler: PipeEventHandler = async (request: KuzzleRequest) => request
 ```
--- a/docker/scripts/start-kuzzle-dev.ts
+++ b/docker/scripts/start-kuzzle-dev.ts
@@ -6,7 +6,7 @@
 import should from 'should/as-function';
 import { omit } from 'lodash';
 
-import { Backend, KuzzleRequest, Mutex } from '../../index';
+import { Backend, EventDefinition, KuzzleRequest, Mutex } from '../../index';
 import { FunctionalTestsController } from './functional-tests-controller';
 import functionalFixtures from '../../features/fixtures/imports.json';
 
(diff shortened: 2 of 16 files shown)


kuzzle:
## Executed Checks

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0209845fd846bc8f59ba13acd9d14d663cc8a37b
</p>

TisserantG (approved):

Aschen:
> I suggest having something like this instead
> 
> ```ts
> // TYPES
> 
> type PipeEvent = {
>   name: string; // Event Name
>   args: any[]; // Arguments types
> }
> 
> type EventHandler<T extends PipeEvent> = (...payload: T["args"]) => T["args"][0];
> 
> function register<T extends PipeEvent>(name: T["name"], handler: EventHandler<T>) {
> 
> }
> 
> // EXAMPLE
> type MyEvent = {
>   name: "my:kuzzle:event";
>   args: [string, number];
> }
> 
> register<MyEvent>("my:kuzzle:event", (message: string, age: number) => {
>   return 'ok';
> });
> ```
> 
> ### Why
> 
> Event names can sometimes be complicated and difficult to memorize, having them embedded in the type will force the user to use the proper event name otherwise typescript will complain, this could help reduce mistakes.
> 
> WDYT ?

That's amazing, I love it

doc/2/framework/types/event-handler/index.md
```diff
@@ -7,7 +7,7 @@ description: EventHandler type definition
 
 # EventHandler
 
-The `EventHandler` type represents a handler function used with a pipe or a hook to listen to events.
+The `XXXEventHandler` typse represents a handler function used with a pipe or a hook to listen to events.
```
Shiranuit, line 10: Typo
```suggestion
The `XXXEventHandler` type represents a handler function used with a pipe or a hook to listen to events.
```

docker/scripts/start-kuzzle-dev.ts
```diff
@@ -86,6 +86,41 @@ app.controller.register('pipes', {
 
 /* Actual code for tests start here */
 
+/**
+ * This function is never call bu simply ensure the correctness of types definition
```
Shiranuit, line 90: Typo
```suggestion
 * This function is never call but simply ensure the correctness of types definition
```

lib/types/events/EventGenericDocument.ts
```diff
@@ -0,0 +1,43 @@
+import { KuzzleRequest, KDocument, JSONObject } from "../../../";
+
+/**
+ * Events with documents having only have the `_id`
```
Shiranuit, line 4:
```suggestion
 * Events with documents only having the `_id`
```

Shiranuit (approved):

PR #2387, opened 2022-09-12
Debug prevent node removal

Shiranuit:
## What does this PR do ?

- Convert IDCardRenewer from a WorkerThread to a ChildProcess to prevent it from being frozen when doing a coredump of kuzzle
- Add a packet to synchronize the Kuzzle cluster nodes and notify them to not evict a specific node
- Refactor the debug controller to extract the logic into a dedicated service
- Add a Cluster debug module that will help debug specific things about the cluster


kuzzle:
## Executed Checks

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 37e79db356529aea8798d3a49b45fc4e7b5845c1
</p>

rolljee (approved):

lib/api/controllers/debugController.ts
```diff
@@ -117,110 +49,25 @@ export class DebugController extends NativeController {
...
-              })
-            );
-          }
-
-          // No need to catch, notify is already try-catched
-          await Promise.all(promises);
-        });
-      }
-    }
+    global.kuzzle.ask("core:debugger:enable");
```
Aschen, line 52: This should be awaited I think

lib/api/controllers/debugController.ts
```diff
@@ -117,110 +49,25 @@ export class DebugController extends NativeController {
...
-        module.removeAllListeners(eventName);
-      }
-      await module.cleanup();
-    }
-
-    this.inspector.disconnect();
-    this.debuggerStatus = false;
-    this.events.clear();
-    this.kuzzlePostMethods.clear();
+    global.kuzzle.ask("core:debugger:disable");
```
Aschen, line 59: This should be awaited I think

lib/api/controllers/debugController.ts
```diff
@@ -238,19 +85,13 @@ export class DebugController extends NativeController {
...
     const event = request.getBodyString("event");
 
-    let listeners = this.events.get(event);
-    if (!listeners) {
-      listeners = new Set();
-      this.events.set(event, listeners);
-    }
-
-    listeners.add(request.context.connection.id);
+    global.kuzzle.ask(
```
Aschen, line 90: This should be awaited I think

lib/api/controllers/debugController.ts
```diff
@@ -267,85 +108,12 @@ export class DebugController extends NativeController {
...
-        this.notifyConnection(connectionId, DEBUGGER_EVENT, {
-          event,
-          result: payload,
-        })
-      );
-    }
-
-    // No need to catch, notify is already try-catched
-    await Promise.all(promises);
+    global.kuzzle.ask(
```
Aschen, line 113: This should be awaited I think

lib/core/debug/kuzzleDebugger.ts
```diff
@@ -0,0 +1,232 @@
...
+   * Connect the debugger
+   */
+  async enable() {
+    if (this.debuggerStatus) {
+      return;
+    }
+
+    this.inspector.connect();
+    this.debuggerStatus = true;
+    global.kuzzle.ask("cluster:node:preventEviction", true);
```
Aschen, line 86: Same comment about `ask` (here and the rest of the file)

lib/api/controllers/debugController.ts
Shiranuit: Every "Ask" call you described are sync, but for the sake of every other ask being awaited I will add the await keyword, this way there will not be any issue if somehow the ask functions are changed to async

alexandrebouthinon (approved):

PR #2391, opened 2022-10-18
Allow to force number of replicas and shards per index from Kuzzle configuration

alexandrebouthinon:
## Why ?
Because a system administrator would ensure index and collection created by users match the Elasticsearch architecture (number of nodes). This in order to provide the best Elasticsearch settings by default for the current cluster topology

## How ?

Add a field `globalSettings` in the `storageEngine` configuration of the KuzzleRC. This field could contains for now:
- `number_of_replicas`: number of replicas to use by default when creating hidden and public index/collection
- `number_of_shards`: number of shards to use by default when creating hidden and public index/collection

## How to test it?
You can start a fresh Kuzzle cluster (with a fresh Elasticsearch cluster) using these command:
```bash
kourou app:start-services
kuzzle_services__storageEngine__globalSettings__number_of_replicas=3 kuzzle_services__storageEngine__globalSettings__number_of_shards=1 npm run dev 
```

Check the number of shards and replicas for newly created hidden collections (eg: `%kuzzle.config` etc...) and try to create new ones using the Admin Console.
By default, using Admin Console regular UI, all the created collection/index will get the `globalSettings` number of shards/replicas. If you want to test to override these values, you can use the `API Action` feature and provide a complete custom request as this one:

```http
{
  "controller": "collection",
  "action": "create",
  "index": "titi",
  "collection": "tutu",
  "body": {
    "settings": {
      "number_of_replicas": 2,
      "number_of_shards" : 2
    }
  }
}
```

### DIFF
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -779,7 +779,7 @@
         }
       },
       // Global settings to apply by default (can be overriden by user requests)
-      "globalSettings": {
+      "defaultSettings": {
         "number_of_shards": 1,
         "number_of_replicas": 1
       },
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -778,6 +778,11 @@
           }
         }
       },
+      // Global settings to apply by default (can be overriden by user requests)
+      "globalSettings": {
+        "number_of_shards": 1,
+        "number_of_replicas": 1
+      },
       // Internal index default name and collections
       // When starting, at initialisation, kuzzle will attempt
       // to update all ES dynamic settings of existing internal
--- a/lib/types/config/storageEngine/StorageEngineElasticsearchConfiguration.ts
+++ b/lib/types/config/storageEngine/StorageEngineElasticsearchConfiguration.ts
@@ -104,6 +104,26 @@ export type StorageEngineElasticsearch = {
     };
   };
 
+  /**
+   * Global settings to apply by default (if not overridden by user request ones)
+   * @default
+   * {
+   *   number_of_shards: 1,
+   *   number_of_replicas: 0
+   * }
+   */
+  globalSettings: {
+    /**
+     * @default 1
+     */
+    number_of_shards: number;
+
+    /**
+     * @default 1
+     */
+    number_of_replicas: number;
+  };
+
   internalIndex: {
     /**
      * @default "kuzzle"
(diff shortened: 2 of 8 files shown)


kuzzle:
## Executed Checks

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 17b171f8943365499c1e68fef0f04b0e407faae4
</p>

.kuzzlerc.sample
```diff
@@ -778,6 +778,11 @@
           }
         }
       },
+      // Global settings to apply by default (can be overriden by user requests)
+      "globalSettings": {
+        "number_of_shards": 1,
+        "number_of_replicas": 1
```
OlivierCavadenti, line 784: properties with underscore but all other properties in this file are kebab case

lib/types/config/storageEngine/StorageEngineElasticsearchConfiguration.ts
```diff
@@ -104,6 +104,26 @@ export type StorageEngineElasticsearch = {
     };
   };
 
+  /**
+   * Global settings to apply by default (if not overridden by user request ones)
+   * @default
+   * {
+   *   number_of_shards: 1,
+   *   number_of_replicas: 0
```
OlivierCavadenti, line 112: 0 or 1 ? default is different from below doc

.kuzzlerc.sample
alexandrebouthinon: This was THE potential debate of this PR 😄 
@kuzzleio/core what do you prefer? Vote on this comment with 🐍 for snake case (the current format) or 🐫 for camel case

Aschen: Since it's a native option for ES I prefer to keep the same syntax

rolljee: :snake: because even if it's not the style we use in kuzzle (coding standard)

It is still the syntax awaited by ES and i believe it should be enough to accept this format

.kuzzlerc.sample
```diff
@@ -778,6 +778,11 @@
           }
         }
       },
+      // Global settings to apply by default (can be overriden by user requests)
+      "globalSettings": {
```
Shiranuit, line 782: Maybe `defaultSettings` or `defaultShardSettings` would be better

alexandrebouthinon: I like defaultSettings

Aschen (approved):

rolljee (approved):

PR #2399, opened 2022-11-22
Fix set-cookie header not respecting the RFC when having multiple of them

Shiranuit:
## What does this PR do ?

Fix the `Set-Cookie` header not respecting the RFC when there is multiple cookie to set.
When that's the case the RFC says that there should be one header per cookie to set instead of having one `Set-Cookie` with each cookie separated by a comma


kuzzle:
## Executed Checks

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a7526163bb216a8f971a40db489951ca285573e0
</p>

rolljee (approved):

OlivierCavadenti (approved):

PR #2401, opened 2022-11-28
Add wait_for_active_shards settings to all ES indices creation

### LINKED ISSUE #2400: Creating multiple collections and documents using a multinode Elasticsearch cluster lead to an error
## Expected Behavior
Creating an index, multiple collections in this index then multiples documents in these collections should work.

## Current Behavior
Randomly failed during collections/documents creation on this line resulting in a 404 Not Found error on the previously created collections (indices in Elasticsearch)

## Possible Solution

This is probably due to the fact that the Elasticsearch cluster is composed of several nodes and the existence of indexes does not have time to propagate to all nodes. As a result, the creation of the documents fails because the node that receives the request can potentially not find the index on its instance. Maybe we should wait for complete shard propagation before running the document creation.

## Steps to Reproduce

1. Launch a Kuzzle stack using a multinode Elasticsearch cluster
2. Create a bunch of collection (at least 25)
3. Then try to create documents in these collection right after the `collection:create` actions
4. You should see the previously described error on some of them

## Context (Environment)
Elasticsearch: 7.17.3 (with at least 3 nodes)
Kuzzle version: 2.19.9
Node.js version:

alexandrebouthinon:
# Why?

When using an Elasticsearch cluster we need to ensure a newly created index is completely propagated to all the ES cluster nodes. For now when using an ES cluster and we run an index create followed by other action on this index this can lead to a 404 error.

Fix #2400


kuzzle:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Found some issues !</td>
    </tr>
  </tbody>
</table>



## Executed Checks

### :x: Branch Merge Check

Merging "wait-for-active-shards" into "master" is not allowed.
If you want to allow this PR to be merged into "master", add the following line to your PR body:


```
ci-allow-merge-into: master
```

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 17bdba1c6e4d2498a94c66bd87147c8b46cd36f0
</p>

Aschen (changes requested):
Can you change the target to `2-dev` so we will make only one release with the Cookie Header fix?

lib/service/storage/elasticsearch.js
```diff
@@ -3154,14 +3156,36 @@ class ElasticSearch extends Service {
           },
         },
         index: await this._getAvailableIndice(index, HIDDEN_COLLECTION),
-      });
+      };
+
+      esRequest.wait_for_active_shards = await this._getWaitForActiveShards();
```
Aschen, line 3161: This should be added directly at `esRequest` initialisation to avoid unecessary hidden class

lib/service/storage/elasticsearch.js
```diff
@@ -3393,6 +3417,12 @@ class ElasticSearch extends Service {
       }
     }
   }
+
+  async _getNumberOfNodes() {
```
Aschen, line 3421: IMHO you could retrieve the number of nodes directly in `_getWaitForActiveShards` and avoid an extra function

Aschen (changes requested):

lib/service/storage/elasticsearch.js
```diff
@@ -3154,14 +3155,37 @@ class ElasticSearch extends Service {
           },
         },
         index: await this._getAvailableIndice(index, HIDDEN_COLLECTION),
-      });
+        wait_for_active_shards: await this._getWaitForActiveShards(),
+      };
+
+      await this._client.indices.create(esRequest);
```
Aschen, line 3161: This also need to be done when updating a collection `updateCollection`

Aschen (changes requested):

lib/service/storage/elasticsearch.js
alexandrebouthinon: I think you were thinking about `collection:truncate` because `collection:updateCollection` do not create a new Elasticsearch index. I'm going to update `collection:truncate` according to your suggestion 👍

Aschen: Ok I didn't realize it was only for creation and not for mappings update. But good catch for `truncate` ;-)

Aschen (approved):

PR #2409, opened 2022-12-06
Ensure alias is being deleted when removing a collection

alexandrebouthinon:
# Why?

According to Elasticsearch documentation, an index associated alias is deleted when the associated index got removed. But there are a couple edge case where it is not properly done.

# How?

After the index deletion we now check if the previously linked alias not exists anymore. If it exists we remove it using the `indices.deleteAlias` action


kuzzle:
## Executed Checks

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against cfa00f5f2e598b9326a8fe52c97bc753d606873c
</p>

PR #2426, opened 2023-01-19
Allows to pass arguments for GET `document:export`

Aschen:
## What does this PR do ?

The `document:export` action is meant to be used as a `a href` link to benefits from the HTTP Stream.

Thus, it is not possible to pass body arguments with the `GET` method. This PR fix that by allowing passing those arguments in the query string in JSON format.

Example: `http://localhost:7512/iot/measures/_export?format=csv&query={}&fields=[%22date%22,%22temperature%22]`

### DIFF
--- a/doc/2/api/controllers/document/export/index.md
+++ b/doc/2/api/controllers/document/export/index.md
@@ -57,17 +57,25 @@ Body:
     // ...
   ],
   "fields": [
-    // ...
-  ]
+    // ["name", "age"]
+  ],
+  "fieldsName": {
+    // "name": "Customer Name"
+  }
 }
 ```
 
 You can also access this route with the `GET` verb:
 
 ```http
-URL: http://kuzzle:7512/<index>/<collection>/_export[?format=<export format>][&size=<int>][&scroll=<time to live>][&lang=<query language>]
+URL: http://kuzzle:7512/<index>/<collection>/_export[?format=<export format>][&size=<int>][&scroll=<time to live>][&lang=<query language>][&searchBody=<query, sort>][&fields=<fields to export>][&fieldsName=<header of each exported field>]
 Method: GET
 ```
+
+::: info
+It's possible to pass arguments that are usually in the body into the query string in JSON format.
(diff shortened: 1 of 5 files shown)


kuzzle:
## Executed Checks

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a3e89273a0a711b73309ba98c4f5c6659b5a0da3
</p>

sebtiz13:
Why not use POST request with body ?
The json in GET arguments could cause problem with url length limitation (2048 characters).

Aschen:
> Why not use POST request with body ? The json in GET arguments could cause problem with url length limitation (2048 characters) or we have a solution to prevent eventual problem with this ?

Because you cannot create a `a href` link with POST so we need to allows arguments to be passed in the query string.

If the URL is bigger than 2048 then we don't have a solution yet. The best solution would be 2 action:
- POST action to send filters and options, return an ID to execute the search
- GET action with this ID to execute the search and export documents as a stream

This could be done but honestly it require a lot of additional work (2 actions, caching in Redis) and we need to release a functional export.

sebtiz13:
> > Why not use POST request with body ? The json in GET arguments could cause problem with url length limitation (2048 characters) or we have a solution to prevent eventual problem with this ?
> 
> Because you cannot create a `a href` link with POST so we need to allows arguments to be passed in the query string.
> 
> If the URL is bigger than 2048 then we don't have a solution yet. The best solution would be 2 action:
> 
>     * POST action to send filters and options, return an ID to execute the search
> 
>     * GET action with this ID to execute the search and export documents as a stream
> 
> 
> This could be done but honestly it require a lot of additional work (2 actions, caching in Redis) and we need to release a functional export.

Ok it's good for the moment, but I think we need to remind this solution for future or possibly plan this.

Aschen:
> 
> But can't you use a form component that forge the request using POST and then start the export on form success action ?

I tried this as well but it doesn't works with Stream apparently

PR #2437, opened 2023-02-28
Overall improvement of the Debug Controller

Shiranuit:
## What does this PR do ?

This PR makes a lot of improvement to the Kuzzle Cluster and Debugger stability while debugging:

### Node eviction prevention

The node eviction prevention mode has been improved, before when a node was used to debug the cluster was notified to not evict this node even if its heartbeat was not received in time by the other nodes that were subscribed.
This is necessary when doing a snapshot of the Node process which causes the node process to not be active during the snapshot.
However when the Node that was prevented from being evicted while debugging had recovered after taking a heap snapshot, there was a case when this Node would evict other nodes because it had not received any heartbeat during the time it was taking the heap snapshot since the node threads were frozen.

A change has been made to now revoke the right of a node to evict other nodes when this node is currently prevented from being evicted from the cluster.

Another changed has been made to the ID Card Renewer, before this PR the ID Card Renewer was a worker thread that had for unique goal to maintain the ID Card data alive inside Redis.
But when taking heap snapshot the debugger is taking a snapshot of each threads owned by the process which was freezing the ID Card Renewer and preventing it from renewing the ID Card in time, which was causing the node to evict himself, and not being seen by new nodes.
Now the ID Card Renewer is created inside another process via a fork and maintains a communication channel with the main process.

### Segfault when taking a Heap Snapshot

We discovered an issue when taking a Heap Snapshot via the `Inspector` that was causing NodeJS to segfault when the `reportProgress` flag was set to `true` because the `inspector` was trying to make a call into the JS Heap while the heap is being inspected for the snapshot.
See this issue: https://github.com/nodejs/node/issues/44634

Now, when `reportProgress` is set to `true` we override it and change it to `false`, but this causes the `Chrome Inspector` to not parse the snapshot even though it has received every chunk of it.
The `Chrome Inspector` relies on the event emitted by `reportProgress` to show a progress bar that represent the state of the ongoing snapshot, and to know when the snapshot is finished this way it can wait for all the snapshot chunks to be received to start the parsing.
In order to force the Chrome Inspector to parse the snapshot we fake the `HeapProfiler.reportHeapSnapshotProgress` and send the following params directly after receiving `reportProgress:true`:
```js
{
  done: 0,
  finished: true,
  total: 0,
}
```

The `done` and `total` parameters are only used to show a progress bar and nothing else, so they're not important, only the `finished` property is used by the Chrome Inspector to spawn a worker that will wait for all the chunks to be received and start parsing the snapshot.
Sending the event `HeapProfiler.reportHeapSnapshotProgress` before any of the `HeapProfiler.addHeapSnapshotChunk` events are sent has no impact, infact this what happens all the time when taking a snapshot, first multiple `HeapProfiler.reportHeapSnapshotProgress` are emitted until one is emitted with `finished:true`, then all the `HeapProfiler.addHeapSnapshotChunk` are emitted with the snapshot chunks.

### Allow socket connections used for debugging to bypass some of Kuzzle's limitations

When taking a Heap Snapshot the debugger emits a lot of `HeapProfiler.addHeapSnapshotChunk` event which contains the snapshot chunks, those chunks are often bigger than 1mb and they're a lot of them, so much than it can cause them to accumulate in the BackPressure buffer dedicated to the socket connection and even exceed the BackPressure buffer limit from the KuzzleRC causing the socket to be closed by Kuzzle.
This is a huge problem since it disconnects the debugger, and prevents us from receiving all the chunks of a snapshot.

To avoid this, when a websocket connection is using the debugger and listen to any event from the debugger by calling `debug:addListener` the websocket connection is marked as used for debugging and is allowed to exceed such limitations.

Another issue that could occur is when Kuzzle request are added to the pending buffer if there are too many request to process, what could happens is request issued to the debug controller meant to debug Kuzzle might be put at the end of the buffer and never be processed or be processed so much later that it renders their result useless.
To counter that, requests from websocket connections previously marked as used for debugging are put at the top of the queue allowing them to be processed first by Kuzzle when Kuzzle is under pressure.

### DIFF
--- a/lib/api/funnel.js
+++ b/lib/api/funnel.js
@@ -179,6 +179,13 @@ class Funnel {
       throw processError.get("not_enough_nodes");
     }
 
+    const isRequestFromDebugSession =
+      request.context &&
+      request.context.connection &&
+      request.context.connection.misc &&
+      request.context.connection.misc.internal &&
+      request.context.connection.misc.internal.debugSession;
+
     if (this.overloaded) {
       const now = Date.now();
 
@@ -226,7 +233,8 @@ class Funnel {
      */
     if (
       this.pendingRequestsQueue.length >=
-      global.kuzzle.config.limits.requestsBufferSize
+        global.kuzzle.config.limits.requestsBufferSize &&
+      !isRequestFromDebugSession
     ) {
       const error = processError.get("overloaded");
       global.kuzzle.emit("log:error", error);
@@ -240,13 +248,7 @@ class Funnel {
         new PendingRequest(request, fn, context)
       );
 
-      if (
-        request.context &&
-        request.context.connection &&
-        request.context.connection.misc &&
-        request.context.connection.misc.internal &&
-        request.context.connection.misc.internal
-      ) {
+      if (isRequestFromDebugSession) {
         // Push at the front to prioritize debug requests
         this.pendingRequestsQueue.unshift(request.internalId);
       } else {
--- a/lib/api/funnel.js
+++ b/lib/api/funnel.js
@@ -239,7 +239,19 @@ class Funnel {
         request.internalId,
         new PendingRequest(request, fn, context)
       );
-      this.pendingRequestsQueue.push(request.internalId);
+
+      if (
+        request.context &&
+        request.context.connection &&
+        request.context.connection.misc &&
+        request.context.connection.misc.internal &&
+        request.context.connection.misc.internal
+      ) {
+        // Push at the front to prioritize debug requests
+        this.pendingRequestsQueue.unshift(request.internalId);
+      } else {
+        this.pendingRequestsQueue.push(request.internalId);
+      }
 
       if (!this.overloaded) {
         this.overloaded = true;
--- a/lib/api/funnel.js
+++ b/lib/api/funnel.js
@@ -51,6 +51,7 @@ const debug = require("../util/debug")("kuzzle:funnel");
 const processError = kerror.wrap("api", "process");
 const { has } = require("../util/safeObject");
 const { HttpStream } = require("../types");
+const get = require("lodash/get");
 
 // Actions of the auth controller that does not necessite to verify the token
 // when cookie auth is active
@@ -179,12 +180,7 @@ class Funnel {
       throw processError.get("not_enough_nodes");
     }
 
-    const isRequestFromDebugSession =
-      request.context &&
-      request.context.connection &&
-      request.context.connection.misc &&
-      request.context.connection.misc.internal &&
-      request.context.connection.misc.internal.debugSession;
+    const isRequestFromDebugSession = get(request, "context.connection.misc.internal.debugSession", false);
 
     if (this.overloaded) {
       const now = Date.now();
--- a/lib/api/funnel.js
+++ b/lib/api/funnel.js
@@ -180,7 +180,11 @@ class Funnel {
       throw processError.get("not_enough_nodes");
     }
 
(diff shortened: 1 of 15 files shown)


kuzzle:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Found some issues !</td>
    </tr>
  </tbody>
</table>



## Executed Checks

### :x: Changelog Tag Check

Pull Requests must have a `changelog:XXX` label.
If you want to allow this PR to not have a changelog tag, add the following line to the PR body:


```
ci-no-changelog-tag
```

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3b62a74bd9ced2a46e944a088cda3ee03b78f288
</p>

lib/cluster/subscriber.js
```diff
@@ -680,7 +680,15 @@ class ClusterSubscriber {
...
-    if (this.state === stateEnum.EVICTED || this.remoteNodeEvictionPrevented) {
+    if (this.remoteNodeEvictionPrevented) {
+      // Fake the heartbeat while the node eviction prevention is enabled
+      // otherwise when the node eviction prevention is disabled
+      // the node will be evicted if it did not send a heartbeat before disabling the protection.
+      this.lastHeartbeat = Date.now();
+      return;
+    }
+
+    if (this.state === stateEnum.EVICTED) {
```
alexandrebouthinon, line 691: Instead of faking the heartbeat in order to prevent the node to kill itself, could we add a new special state in stateEnum like `DEBUGGING` to ignore a missing heartbeat in the condition branches below?

lib/cluster/workers/IDCardRenewer.js
```diff
@@ -29,6 +29,10 @@ class IDCardRenewer {
       const redisConf = config.redis || {};
       await this.initRedis(redisConf.config, redisConf.name);
     } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(
+        `Failed to connect to redis, could not refresh ID card: ${error.message}`
+      );
```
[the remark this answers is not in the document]
alexandrebouthinon: Needed because it in a separeted worker_thread as @Shiranuit explained to me

lib/cluster/subscriber.js
Shiranuit: It's already the case, but when exiting this special state all other nodes that had not received the heartbeat round from the node that was debugging will evict it if the node that was previously used to debug did not have the opportunity to send a heartbeat before their check, that's why we are faking it meanwhile, this way the node will have the time to send a proper heartbeat when exiting his state.

lib/cluster/workers/IDCardRenewer.js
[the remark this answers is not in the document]
Shiranuit: It's intended since there is no logger in the fork process this log is here to let us know of an issue happening inside the fork, even though it should be able to send the information back to the main process, in case something prevents it from notifying the main process there will also be a log from the fork.

lib/api/funnel.js
```diff
@@ -179,6 +179,13 @@ class Funnel {
       throw processError.get("not_enough_nodes");
     }
 
+    const isRequestFromDebugSession =
+      request.context &&
+      request.context.connection &&
+      request.context.connection.misc &&
+      request.context.connection.misc.internal &&
+      request.context.connection.misc.internal.debugSession;
```
Aschen, line 187: You could replace this with `_.get`

Aschen (approved):

PR #2440, opened 2023-03-20
Allows to register realtime subscriptions from the backend

Aschen:
## What does this PR do ?

Realtime subscription can be added on a connection directly from the backend. This allows to restrict the usage of the [realtime:subscribe](/core/2/api/controllers/realtime/subscribe) and to add realtime subscription from a custom API action.

The [Backend.subscription.add](/core/2/framework/classes/backend-subscription/add) method takes the same arguments as the [realtime:subscribe](/core/2/api/controllers/realtime/subscribe) API action but allows to register the subscription on any connection.

**Example: _Subscribe to a room from a custom API action_**

```js
async customSubscribe (request: Kuzzle Request) {
  const { roomId, channel } = await app.subscription.add(
    request.context.connection,
    "lamaral",
    "windsurf",
    {
      range: { wind: { gte: 20 } },
    },
    {
      users: "all",
      scope: "in",
      volatile: { name: "Aschen" },
    }
  );

  return { roomId, channel };
}

```

Then, the specified connection will start to receive realtime notifications about the subscription. In the SDKs, those messages need to be listened.

For example, with the Javascript SDK:

```js
const { result } = await sdk.query({
  controller: "custom",
  action: "subscribe",
});

sdk.protocol.on(result.channel, (notification) => {
  console.log(notification);
});
```

### DIFF
--- a/lib/core/backend/backendSubscription.ts
+++ b/lib/core/backend/backendSubscription.ts
@@ -1,0 +1,94 @@
+/*
+ * Kuzzle, a backend software, self-hostable and ready to use
+ * to power modern apps
+ *
+ * Copyright 2015-2022 Kuzzle
+ * mailto: support AT kuzzle.io
+ * website: http://kuzzle.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { JSONObject } from "kuzzle-sdk";
+import { Connection, KuzzleRequest } from "../../api/request";
+import * as kerror from "../../kerror";
+import { ApplicationManager } from "./index";
+
+const runtimeError = kerror.wrap("plugin", "runtime");
+
+export class BackendSubscription extends ApplicationManager {
+  /**
+   * Registers a new realtime subscription on the specified connection
+   *
+   * @param connection Connection to register the subscription on
+   * @param index Index name
+   * @param collection Collection name
+   * @param filters Subscription filters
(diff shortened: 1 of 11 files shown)


kuzzle:
## Executed Checks

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8221c59de859ad2edc8d6d5dd67f4cffeff2544e
</p>

lib/core/backend/backendSubscription.ts
```diff
@@ -0,0 +1,94 @@
...
+   *
+   * @param connection Connection to register the subscription on
+   * @param index Index name
+   * @param collection Collection name
+   * @param filters Subscription filters
+   * @param options.volatile Volatile data
+   * @param options.scope Subscription scope
+   * @param options.users Option for users notifications
+   */
+  async add(
```
alexandrebouthinon, line 41: I can't remember if this could be a problem but what about the `sdk.as` function of the backend SDK instance? Should this function support it?

lib/core/backend/backendSubscription.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    const { channel, roomId } = await global.kuzzle.ask(
+      "core:realtime:subscribe",
+      subscriptionRequest
+    );
+
+    return { channel, roomId };
+  }
+
+  async remove(connection: Connection, roomId: string): Promise<void> {
```
alexandrebouthinon, line 84: Ditto

thomas-mauran (approved):

lib/core/backend/backendSubscription.ts
Aschen: I don't see any usage of impersonation with the realtime engine since we don't change document metadata. Are you thinking about something else?

lib/core/backend/backendSubscription.ts
Aschen: ditto ;p

lib/core/backend/backendSubscription.ts
alexandrebouthinon: No I'm not. Just wanted to make you think about it just in case 🙂

alexandrebouthinon (approved):

lib/core/backend/backendSubscription.ts
Aschen: Ok, I don't see any then :+1:

PR #2459, opened 2023-05-15
Delete ID Card directly from Redis when Kuzzle is shutting down

Shiranuit:
## What does this PR do ?

This PR forces the ID Card Renewer process to delete the node ID Card from Redis when Kuzzle is shutting down for whatever reason, this will allow nodes to be restarted much faster and avoid the error `[CLUSTER] Another node share the same IP address as this one (10.84.65.9): knode-regular-greer-69286. Shutting down.` because of the ID Card not being expired before the node had restarted.

### DIFF
--- a/lib/cluster/idCardHandler.ts
+++ b/lib/cluster/idCardHandler.ts
@@ -216,7 +216,11 @@ export class ClusterIdCardHandler {
 
     const exitHandler = () => {
       if (!childProcess.killed || childProcess.connected) {
-        childProcess.kill();
+        try {
+          childProcess.disconnect();
+        } catch (e) {
+          // It could happens that the worker has been killed before the dispose causing disconnect to fail
+        }
       }
       process.exit();
     };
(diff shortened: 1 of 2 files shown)


kuzzle:
## Executed Checks

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e58bfda115bf54a34efab55cfcbfd45cb04a4f7f
</p>

lib/cluster/idCardHandler.ts
```diff
@@ -216,7 +216,11 @@ export class ClusterIdCardHandler {
 
     const exitHandler = () => {
       if (!childProcess.killed || childProcess.connected) {
-        childProcess.kill();
+        try {
+          childProcess.disconnect();
+        } catch (e) {
+          // It could happens that the worker has been killed before the dispose causing disconnect to fail
```
OlivierCavadenti, line 222: no necessary to add a debug log at least here instead of silent ?

Shiranuit: No not really, even though the process fails to close the connection for some reason, the call to process.exit will stop the process and then the OS is going to close the IPC between them, leading the child process to stop himself and clear the ID Card inside Redis

PR #2461, opened 2023-05-15
Add http protocol parameter additionalContentTypes

fmauNeko:
## What does this PR do ?
Implements a new `kuzzlerc` settings for the http protocol, that allows specifying additional accepted `Content-Type`s.
Note: This requires the matching implementation of a `protocol:http:beforeParsingPayload` pipe that is able to convert to JSON the payloads received with those new Content-Types.

### DIFF
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -405,14 +405,14 @@
     "protocols": {
       "http": {
         // * additionalContentTypes:
-        //    Allows additional Content-Types to be accepted by Kuzzle.
+        //    Enables Kuzzle to accept additional Content-Types.
+        //    Note: This relies on the implementation of a
+        //    "protocol:http:beforeParsingPayload" pipe that implements
+        //    the formatting of the additional content types to JSON.
         //    The default content types are:
         //      * application/json
         //      * application/x-www-form-urlencoded
         //      * multipart/form-data
-        //    Note: This relies on the implementation of a
-        //    "protocol:http:beforeParsingPayload" pipe that implements
-        //    the formatting of the additional content types to JSON.
         // * allowCompression:
         //    Enable support for compressed requests, using the
         //    Content-Encoding header
--- a/.kuzzlerc.sample
+++ b/.kuzzlerc.sample
@@ -404,6 +404,15 @@
     // protocols can be extended and configured in this section.
     "protocols": {
       "http": {
+        // * additionalContentTypes:
+        //    Allows additional Content-Types to be accepted by Kuzzle.
+        //    The default content types are:
+        //      * application/json
+        //      * application/x-www-form-urlencoded
+        //      * multipart/form-data
+        //    Note: This relies on the implementation of a
+        //    "protocol:http:beforeParsingPayload" pipe that implements
+        //    the formatting of the additional content types to JSON.
         // * allowCompression:
         //    Enable support for compressed requests, using the
         //    Content-Encoding header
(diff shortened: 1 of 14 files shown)


kuzzle:
## Executed Checks

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7e95c5fdfb32f5432ce4f0af5c231d9911a276db
</p>

.kuzzlerc.sample
```diff
@@ -404,6 +404,15 @@
     // protocols can be extended and configured in this section.
     "protocols": {
       "http": {
+        // * additionalContentTypes:
+        //    Allows additional Content-Types to be accepted by Kuzzle.
```
rolljee, line 408:
```suggestion
        //    Enables Kuzzle to accept additional Content-Types.
```

.kuzzlerc.sample
```diff
@@ -404,6 +404,15 @@
...
+        //    The default content types are:
+        //      * application/json
+        //      * application/x-www-form-urlencoded
+        //      * multipart/form-data
+        //    Note: This relies on the implementation of a
```
rolljee, line 413: This part is the most important, it should be placed L409

lib/core/network/protocols/httpwsProtocol.js
```diff
@@ -502,7 +502,7 @@ class HttpWsProtocol extends Protocol {
 
     if (
       contentType &&
-      !HTTP_ALLOWED_CONTENT_TYPES.some((allowed) =>
+      !this.httpConfig.opts.allowedContentTypes.some((allowed) =>
```
rolljee, line 505: Why this variable has changed ?

lib/types/config/ServerConfiguration.ts
```diff
@@ -74,6 +74,19 @@ export type ServerConfiguration = {
    */
   protocols: {
     http: {
+      /**
+       * Allows additional Content-Types to be accepted by Kuzzle.
```
rolljee, line 78:
```suggestion
       * Enables Kuzzle to accept additional Content-Types.
```

lib/types/config/ServerConfiguration.ts
```diff
@@ -74,6 +74,19 @@ export type ServerConfiguration = {
...
+       * The default content types are:
+       *   * application/json
+       *   * application/x-www-form-urlencoded
+       *   * multipart/form-data
+       * Note: This relies on the implementation of a
```
rolljee, line 83: This is the most important part it should be place L79

rolljee (changes requested):
You should add a test that confirm that the right values is passed to this parameters

`lib/config/index.js` the assert you added here is not tested.

Also, can you find somewhere to add an example of a value you can pass in this parameter and how to implement the underlying pipe to process it ? 

It would simplify the implementation for anyone 👍🏼

lib/core/network/protocols/httpwsProtocol.js
fmauNeko: The default allowed content types and the additional ones from the config are merged into this variable L1197

test/core/network/protocols/websocket.test.js
```diff
@@ -40,6 +40,7 @@ describe("core/network/protocols/websocket", () => {
       port: 7512,
       protocols: {
         http: {
+          additionalContentTypes: [],
```
OlivierCavadenti, line 43: why additionalContentTypes here and allowedContentTypes before?

Shiranuit (changes requested):
Could you add some tests using Jest to ensure this is working as intended ?

rolljee (approved):

Shiranuit (approved):

PR #2462, opened 2023-05-16
Mandatory query param in deleteByQuery

OlivierCavadenti:
Actually if you call deleteByQuery without query param, you can delete all documents because default query value is set to "{}" which is equals to "match_all" predicate.

Remove default value for query to avoid that.

### DIFF
--- a/lib/api/controllers/documentController.js
+++ b/lib/api/controllers/documentController.js
@@ -765,7 +765,7 @@ class DocumentController extends NativeController {
    * @returns {Promise<Object>}
    */
   async deleteByQuery(request) {
-    let query = request.getBodyObject("query", {});
+    let query = request.getBodyObject("query");
     const silent = request.getBoolean("silent");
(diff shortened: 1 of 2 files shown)


rolljee:
It's a tricky one, if we had this, since it is a breaking change, we must update to kuzzle v3 😕

Shiranuit (changes requested):
Like it has been said by @rolljee we cannot do that, even though I do want to because it adds safety.
Doing it would mean upgrading Kuzzle to V3 which is far from what is planned right now.

I think we should think of a way to add this behind a flag that enable some more strict  behaviour for safety
we might need to do a workshop to address those kind of issue in the future

OlivierCavadenti:
> I think we should think of a way to add this behind a flag that enable some more strict behaviour for safety
> we might need to do a workshop to address those kind of issue in the future

By the way, I see that we already have a mention of that in the documentation:

image

PR #2476, opened 2023-09-12
Update/uwebsocket

rolljee:
## What does this PR do ?

* This PR update uwebsocket version
* Remove nodeJS 14 from the matrix of tests
* <s>Activate dead-links job that was removed</s>

  - Step 1 :
  - Step 2 :
  - Step 3 :

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -15,7 +15,7 @@ env:
 jobs:
   prepare-matrix:
     name: Forge Node LTS Matrix
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - id: set-matrix
         run: |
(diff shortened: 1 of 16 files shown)


tdislay (approved):

.github/workflows/workflow.yaml
```diff
@@ -8,9 +8,8 @@ on:
   pull_request: # This syntax stand for all PRs events
 
 env:
-  NODE_LTS_MAINTENANCE_VERSION: "14"
```
fmauNeko, line 11: According to the Node.js version nomenclature, we should keep MAINTENANCE and have MAINTENANCE = 16, ACTIVE = 18 and CURRENT = 20, and once Node.js 21 is released (which means 20 becomes Active LTS and 18 Maintenance LTS), we should have MAINTENANCE = 18, ACTIVE = 20, CURRENT = 21

fmauNeko (approved):

.github/workflows/workflow.yaml
rolljee: I'll modify that 👍🏼

alexandrebouthinon (approved):

PR #2490, opened 2023-10-23
Fix incorrect HTTP Content-Type headers being accepted

fmauNeko:
## What does this PR do ?

This fixes successful matches of `Content-Type` headers with extraneous content.
Previously, `Content-Type: canardapplication/jsonziufhzihuchevalzef izufh` would be considered valid because it contains `application/json`.

  - Step 1 : Send any request with a `Content-Type` header containing a valid Content-Type surrounded by random characters.
  - Step 2 : Verify that it is now refused.

### DIFF
--- a/test/core/network/protocols/http.test.js
+++ b/test/core/network/protocols/http.test.js
@@ -359,7 +359,7 @@ describe("core/network/protocols/http", () => {
 
     it("should reject requests with supported content types with extraneous characters", () => {
       httpWs.server._httpOnMessage("get", "/", "", {
-        "content-type": "canard application/jsoncheval",
+        "content-type": "serge application/jsoncheval",
       });
 
       should(entryPoint.newConnection).not.called();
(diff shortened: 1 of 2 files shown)


rolljee (approved):
LGTM, but PR miss some label and I believe the title is not appropriate

test/core/network/protocols/http.test.js
```diff
@@ -357,6 +357,20 @@ describe("core/network/protocols/http", () => {
         });
     });
 
+    it("should reject requests with supported content types with extraneous characters", () => {
+      httpWs.server._httpOnMessage("get", "/", "", {
+        "content-type": "canard application/jsoncheval",
```
tdislay, line 362:
```suggestion
        "content-type": "serge application/jsoncheval",
```

tdislay (approved):

Shiranuit (approved):

PR #2496, opened 2023-11-14
added the alias that is preblematic in the error message

etrousset:
## What does this PR do ?

Add more detail in the error triggered by es service in _getIndice method

### DIFF
--- a/lib/service/storage/elasticsearch.js
+++ b/lib/service/storage/elasticsearch.js
@@ -40,7 +40,6 @@ const scopeEnum = require("../../core/storage/storeScopeEnum");
 const extractFields = require("../../util/extractFields");
 const { Mutex } = require("../../util/mutex");
 const { randomNumber } = require("../../util/name-generator");
-const { error } = require("console");
 
 const SCROLL_CACHE_PREFIX = "_docscroll_";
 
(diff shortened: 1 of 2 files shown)


lib/service/storage/elasticsearch.js
```diff
@@ -2965,7 +2965,7 @@ class ElasticSearch extends Service {
     });
 
     if (body.length < 1) {
-      throw kerror.get("unknown_index_collection");
+      throw kerror.get(`unknown_index_collection : ${alias}`);
```
alexandrebouthinon, line 2968: This method does not work like a console log, you must edit the associated error code description message to add your variable (https://github.com/kuzzleio/kuzzle/blob/75cb185aed40c3d4cf6dad33eac905404c723549/lib/kerror/codes/1-services.json#L88). You could find such a message syntax here https://github.com/kuzzleio/kuzzle/blob/75cb185aed40c3d4cf6dad33eac905404c723549/lib/kerror/codes/1-services.json#L100

alexandrebouthinon (changes requested):

alexandrebouthinon:
Close this since it disclose information about inner database architecture directly in the error message

PR #2502, opened 2023-11-28
Update workflows, merge repository kuzzle-containers

rolljee:
## What does this PR do ?

* This PR update workflows files to be more readable and up to date. This will prepare the work for NodeJS 20
* Merging Kuzzle-containers repository here, because it make more sense now that kuzzle build and push all image that he needs.


.github/workflows/kuzzle-runner.workflow.yml
```diff
@@ -0,0 +1,101 @@
...
+on:
+  push:
+    branches:
+      - master
+
+env:
+  DOCKER_PLATFORMS: "linux/amd64,linux/arm64,linux/arm/v7"
+  NODE_LTS_MAINTENANCE_VERSION: "16"
+  NODE_LTS_ACTIVE_VERSION: "18"
+  NODE_LTS_CURRENT_VERSION: "20"
```
Kuruyia, line 12: Why not use a [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) for doing the same thing?

.github/workflows/workflow.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
     steps:
       - id: set-matrix
         run: |
-          echo "::set-output name=matrix::{\"node-version\": [\"$NODE_LTS_MAINTENANCE_VERSION\", \"$NODE_LTS_ACTIVE_VERSION\"]}"
+          echo "matrix={\"node-version\": [\"$NODE_LTS_MAINTENANCE_VERSION\", \"$NODE_LTS_ACTIVE_VERSION\"]}" >> $GITHUB_OUTPUT
```
Kuruyia, line 23: Is it possible to set up the matrix of Nodejs versions normally (with an array of numbers) instead of constructing it like this? This seems rather "hacky".

docker/images/kuzzle-runner/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+      g++ \
+      git \
+      libzmq3-dev \
+      libunwind-dev \
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
+
+RUN npm i -g kourou
+
+WORKDIR /var/app
```
Kuruyia, line 27: The working directory was already set above.

```suggestion
```

Kuruyia (changes requested):

rolljee:
We are going to squash this one 🥶

.github/workflows/workflow.yaml
rolljee: Well, I tried a bunch of thing, seems that this is the way to go if we don't want to repeat the matrix over and over for each jobs

Kuruyia (approved):

fmauNeko (approved):

PR #2508, opened 2023-12-05
Update workflow to have kuzzle-core and plugin-dev

rolljee:
## What does this PR do ?

* This PR moves the creation of the docker image kuzzleio/kuzzle to a dedicated workflow run like other created image.
* This PR re-add the creation of kuzzleio/plugin-dev witch apparently stopped being created 2 year ago 😊 

I also deleted some files (scripts mostly) that wasn't used to remove little by little all noise in the repository


docker/images/kuzzle/Dockerfile
```diff
@@ -14,36 +15,29 @@ RUN  set -x \
...
 
-ENV NODE_ENV=production
+ADD ./bin bin
+ADD ./config config
+ADD ./lib lib
+ADD ./package.json package.json
+ADD ./package-lock.json package-lock.json
+ADD ./index.ts index.ts
+ADD ./tsconfig.json tsconfig.json
+ADD ./.kuzzlerc.sample.jsonc .kuzzlerc.sample.jsonc
```
alexandrebouthinon, line 27: I know this was already like that but I think we can get rid of it
```suggestion
```

alexandrebouthinon (approved):

.github/workflows/kuzzle-core.yml
```diff
@@ -0,0 +1,42 @@
...
+          echo "major-version=$(jq -r .version package.json | cut -d. -f 1)" >> $GITHUB_OUTPUT
+
+      - name: Build and push
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: ./docker/images/kuzzle/Dockerfile
+          push: true
+          platforms: "linux/amd64,linux/arm64"
+          tags: kuzzleio/kuzzle:${{ steps.get-version.outputs.major-version }},kuzzleio/kuzzle:latest,kuzzleio/kuzzle:${{ steps.get-version.outputs.version }}
```
fmauNeko, line 42: https://github.com/docker/metadata-action could be interesting to use, because it autogenerate tags and opencontainers labels.

fmauNeko (approved):

.github/workflows/kuzzle-core.yml
rolljee: It would, I'll keep this in mind for the futur implementation of semantic release

PR #2521, opened 2024-01-31
fix(semantic-release): fix an issue where semver would not satisfies …

rolljee:
…pre-release versions

## What does this PR do ?

This PR allow kuzzle to work with pre release version of plugin and kuzzle itself


Kuruyia (approved):

Kuruyia:
Maybe we can add a test for this case?

rolljee:
> Maybe we can add a test for this case?

That's a good idea, I'll do it 👍🏼

fmauNeko (approved):

PR #2530, opened 2024-05-08
Fix documentation example for using `searchUsersByCredentials` action. 

azagaya:
## What does this PR do ?

Fixes a small mistake in the documentation.

Step 1: Try to do a `searchUsersBycredentials` query as the documentation states, see it fails.
Step 2: Try with my proposed changes, and it works.


rolljee (approved):
Thank you for your contribution on the documentation 🎉

fmauNeko (approved):

fmauNeko (approved):

PR #2537, opened 2024-06-05
build: no strict version for kuzzle-sdk

sebtiz13:
## What does this PR do ?

Change `kuzzle-sdk` version restriction to allow npm to dedup the package when it's required in other package.

## Explanation
Currently, Kuzzle require strictly `kuzzle-sdk@7.11.1`, if a package like plugin require `kuzzle` and `kuzzle-sdk@7.11.3`

With strict restriction, NPM produce this install

- kuzzle-sdk@7.11.3
- kuzzle
  - kuzzle-sdk@7.11.1
  
With package restriction like `kuzzle-sdk@^7.11.3` (so accept greater than `7.11.3` but less than `8`), NPM produce this install

- kuzzle-sdk@7.11.3
- kuzzle

Here, kuzzle-sdk is not installed in kuzzle directory because `kuzzle-sdk@7.11.3` satisfy the restriction.
Also, normally in future if I want to install `kuzzle-sdk@8.0.0` in my package, npm will throw an error for incompatibility (need to verify, because npm haven't consistent behavior through the versions)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,7 @@
         "koncorde": "4.3.0",
         "kuzzle-plugin-auth-passport-local": "6.4.0",
         "kuzzle-plugin-logger": "3.0.3",
-        "kuzzle-sdk": "7.11.1",
+        "kuzzle-sdk": "^7.11.3",
         "kuzzle-vault": "2.0.4",
         "lodash": "4.17.21",
         "long": "5.2.3",
@@ -10325,9 +10325,9 @@
       }
     },
     "node_modules/kuzzle-sdk": {
-      "version": "7.11.1",
-      "resolved": "https://registry.npmjs.org/kuzzle-sdk/-/kuzzle-sdk-7.11.1.tgz",
-      "integrity": "sha512-1nPZmekFE2Q1gNhPisn9kg1/LsDRzDnl4GIYGdjIHISK5BhiAICMEsmMMyTiYqja42/mdMqm+M5bFx4p3wMJdQ==",
+      "version": "7.11.3",
+      "resolved": "https://registry.npmjs.org/kuzzle-sdk/-/kuzzle-sdk-7.11.3.tgz",
+      "integrity": "sha512-YHPSE99GRg1GT9UD5hOuJnt5vc/Gqmp705id4sCjyRJNRaq9uf3j3OzM6bvD+4YPZB7LTim53z66CFQP7IV7Qg==",
       "dependencies": {
         "min-req-promise": "^1.0.1",
         "ws": "^8.13.0"
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "koncorde": "4.3.0",
     "kuzzle-plugin-auth-passport-local": "6.4.0",
     "kuzzle-plugin-logger": "3.0.3",
-    "kuzzle-sdk": "7.11.1",
+    "kuzzle-sdk": "^7.11.3",
     "kuzzle-vault": "2.0.4",
     "lodash": "4.17.21",
     "long": "5.2.3",


fmauNeko (approved):

fmauNeko (approved):

sebtiz13:
> > ## What does this PR do ?
> > Change `kuzzle-sdk` version restriction to allow npm to dedup the package when it's required in other package.
> > ## Explanation
> > Currently, Kuzzle require strictly `kuzzle-sdk@7.11.1`, if a package like plugin require `kuzzle` and `kuzzle-sdk@7.11.3`
> > With strict restriction, NPM produce this install
> > 
> > * kuzzle-sdk@7.11.3
> > * kuzzle
> >   
> >   * kuzzle-sdk@7.11.1
> > 
> > With package restriction like `kuzzle-sdk@^7.11.3` (so accept greater than `7.11.3` but less than `8`), NPM produce this install
> > 
> > * kuzzle-sdk@7.11.3
> > * kuzzle
> > 
> > Here, kuzzle-sdk is not installed in kuzzle directory because `kuzzle-sdk@7.11.3` satisfy the restriction. Also, normally in future if I want to install `kuzzle-sdk@8.0.0` in my package, npm will throw an error for incompatibility (need to verify, because npm haven't consistent behavior through the versions)
> 
> Wouldn't it be a best practice to update the sdk to his latest version but still left fixed to avoid issue ?
> 
> Like we could launch an action that launch `ncu` and update if all tests aren't failing ?

The problem it's if we use the version `7.11.1` in the core and then we have a `7.11.3` update. We need to release the core only to update the package version. Instead, with not strict restriction, It's not mandatory to release the core only for that.
It's for this reason we restrict the version range of package with caret `^` to accept only from `7.11.1` to `8` (not include) to ensure the compatibility (in theory no breaking change), but we can be stricter and use tilde to restrict the range only on patch so from `7.11.1` to `7.12` (not include) to be safer.

rolljee:
> > > ## What does this PR do ?
> > > Change `kuzzle-sdk` version restriction to allow npm to dedup the package when it's required in other package.
> > > ## Explanation
> > > Currently, Kuzzle require strictly `kuzzle-sdk@7.11.1`, if a package like plugin require `kuzzle` and `kuzzle-sdk@7.11.3`
> > > With strict restriction, NPM produce this install
> > > 
> > > * kuzzle-sdk@7.11.3
> > > * kuzzle
> > >   
> > >   * kuzzle-sdk@7.11.1
> > > 
> > > With package restriction like `kuzzle-sdk@^7.11.3` (so accept greater than `7.11.3` but less than `8`), NPM produce this install
> > > 
> > > * kuzzle-sdk@7.11.3
> > > * kuzzle
> > > 
> > > Here, kuzzle-sdk is not installed in kuzzle directory because `kuzzle-sdk@7.11.3` satisfy the restriction. Also, normally in future if I want to install `kuzzle-sdk@8.0.0` in my package, npm will throw an error for incompatibility (need to verify, because npm haven't consistent behavior through the versions)
> > 
> > 
> > Wouldn't it be a best practice to update the sdk to his latest version but still left fixed to avoid issue ?
> > Like we could launch an action that launch `ncu` and update if all tests aren't failing ?
> 
> The problem it's if we use the version `7.11.1` in the core and then we have a `7.11.3` update. We need to release the core only to update the package version. Instead, with not strict restriction, It's not mandatory to release the core only for that. It's for this reason we restrict the version range of package with caret `^` to accept only from `7.11.1` to `8` (not include) to ensure the compatibility (in theory no breaking change), but we can be stricter and use tilde to restrict the range only on patch so from `7.11.1` to `7.12` (not include) to be safer.

If I understand in this case,

Does it mean in the case we have kuzzle requesting the 7.11.1 and another repo is requesting 7.11.3
It will install the 7.11.3 right ?

So it also mean that kuzzle will too run in 7.11.3 even tho it as only be tested with the 7.11.1 ?


I think I am okay with it, but I just want to be fully aware of all that :)

rolljee (approved):

alexandrebouthinon (approved):

PR #2538, opened 2024-06-24
docs(log): add a log to know which ES configuration is set up

cyrng:
## What does this PR do ?

Add log to display which ES version is configured


alexandrebouthinon (commented):
you need to apply the linter requested change

PR #2542, opened 2024-07-15
Feat(funnel): allow pluginRequest to trigger events

QuentinRousselet:
## What does this PR do ?
This PR introduces an optional parameter "allowTriggerEvents" in KuzzleRequest to allow or not API actions to trigger events and pipes.
In the funnel if a pluginRequest holds this parameter, the request is processed as a normal request otherwise it is treated as usally without triggering events.

In a project using Kuzzle:
-Set up a custom controller creating a document for example
-Set up a pipe that triggers on the creation of document ('document:beforeCreate/document:afterCreate)
-Send a request to your controller without "allowTriggerEvents": the pipe shouldn't trigger
-Send a request to your controller with  allowTriggerEvents:true  the pipe should trigger.

### Other changes
With this process the request also triggers statistic counts which it wouldn't have in the normal pluginRequest execution

### DIFF
--- a/doc/2/guides/develop-on-kuzzle/event-system/index.md
+++ b/doc/2/guides/develop-on-kuzzle/event-system/index.md
@@ -188,7 +188,7 @@ await app.trigger('app-name/file-available', fileUrl);
 If an internal event is triggered, the payload must be the same as the original event.
 :::
 
-**Events can be triggered** with SDK and controllers actions.
+**Events are not triggered** with SDK and controllers actions.
 
 ::: warning
 By default, actions executed through the embedded SDK or controllers won't trigger any events and thus no pipe or hooks will be called.
@@ -196,7 +196,7 @@ By default, actions executed through the embedded SDK or controllers won't trigg
 
 This behaviour is set in order to prevent an infinite loop in which a pipe calls a controller generating an event calling this same pipe again and again. 
 
-It is nonetheless possible to pass the flag `allowTrigggerEvents` to the options parameters of the controller to allow events firing :
+It is nonetheless possible to pass the flag `triggerEvents` to the options parameters of the controller to fire events :
 
 ```ts
 await this.sdk.document.create(
@@ -206,6 +206,6 @@ await this.sdk.document.create(
                 contentOfDocument: 'CREATED VIA CONTROLLER',
               },
               _idOfDocument,
-              { allowTriggerEvents: true },
+              { triggerEvents: true },
             );
 ```
--- a/doc/2/guides/develop-on-kuzzle/event-system/index.md
+++ b/doc/2/guides/develop-on-kuzzle/event-system/index.md
@@ -187,3 +187,25 @@ await app.trigger('app-name/file-available', fileUrl);
 ::: warning
 If an internal event is triggered, the payload must be the same as the original event.
 :::
+
+**Events can be triggered** with SDK and controllers actions.
+
+::: warning
+By default, actions executed through the embedded SDK or controllers won't trigger any events and thus no pipe or hooks will be called.
+:::
+
+This behaviour is set in order to prevent an infinite loop in which a pipe calls a controller generating an event calling this same pipe again and again. 
+
+It is nonetheless possible to pass the flag `allowTrigggerEvents` to the options parameters of the controller to allow events firing :
+
+```ts
+await this.sdk.document.create(
+              "index",
+              'collection',
+              {
+                contentOfDocument: 'CREATED VIA CONTROLLER',
+              },
+              _idOfDocument,
+              { allowTriggerEvents: true },
+            );
+```
--- a/doc/2/guides/develop-on-kuzzle/event-system/index.md
+++ b/doc/2/guides/develop-on-kuzzle/event-system/index.md
@@ -188,10 +188,10 @@ await app.trigger('app-name/file-available', fileUrl);
 If an internal event is triggered, the payload must be the same as the original event.
 :::
 
-**Events are not triggered** with SDK and controllers actions.
+**Events are not triggered** with the embedded SDK and controllers actions.
 
 ::: warning
-By default, actions executed through the embedded SDK or controllers won't trigger any events and thus no pipe or hooks will be called.
+By default, actions executed through the embedded SDK or controllers won't trigger any events and thus no pipe or hooks will be called. On the contrary, controllers accessed by the external SDK through HTTP or Websocket requests will always fire events.
 :::
 
 This behaviour is set in order to prevent an infinite loop in which a pipe calls a controller generating an event calling this same pipe again and again. 
--- a/lib/api/funnel.js
+++ b/lib/api/funnel.js
@@ -753,6 +753,9 @@ class Funnel {
    */
   async executePluginRequest(request) {
     try {
+      if (request.input.allowTriggerEvents) {
+        return await this.processRequest(request);
+      }
       return await doAction(this.getController(request), request);
     } catch (e) {
       this.handleErrorDump(e);
--- a/lib/api/funnel.js
+++ b/lib/api/funnel.js
@@ -753,7 +753,7 @@ class Funnel {
    */
   async executePluginRequest(request) {
     try {
-      if (request.input.allowTriggerEvents) {
+      if (request.input.triggerEvents) {
         return await this.processRequest(request);
       }
       return await doAction(this.getController(request), request);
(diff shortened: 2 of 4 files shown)


etrousset (changes requested):
We should get a way to had this flag through calling internal sdk API such as this.sdk.document.create( "my_index", "my_collection", {...} , { 'allowTriggerEvent' : true}

QuentinRousselet:
> We should get a way to had this flag through calling internal sdk API such as this.sdk.document.create( "my_index", "my_collection", {...} , { 'allowTriggerEvent' : true}

It is already the case, here is the code in the Iot platform with the custom controller that I used to test the feature
```ts
customCreateDocument: {
          handler: async (request: KuzzleRequest) => {
            const res = await this.sdk.document.create(
              request.getIndex(),
              'measures',
              {
                values: { zone: 'CREATED VIA CONTROLLER' },
              },
              undefined,
              { allowTriggerEvents: request.input.allowTriggerEvents },
            );
            return res;
          },
        },
        ```
        the methods on sdk.document takes an option object (here the 5th argument) in which we can set the flag

**Comment by etrousset** [comment]
> > We should get a way to had this flag through calling internal sdk API such as this.sdk.document.create( "my_index", "my_collection", {...} , { 'allowTriggerEvent' : true}
> 
> It is already the case, here is the code in the Iot platform with the custom controller that I used to test the feature
> 
> ```ts
> customCreateDocument: {
>           handler: async (request: KuzzleRequest) => {
>             const res = await this.sdk.document.create(
>               request.getIndex(),
>               'measures',
>               {
>                 values: { zone: 'CREATED VIA CONTROLLER' },
>               },
>               undefined,
>               { allowTriggerEvents: true }, // ici on devrait plutôt avoir ça à l'usage
>             );
>             return res;
>           },
>         },
>         ```
>         the methods on sdk.document takes an option object (here the 5th argument) in which we can set the flag 
> ```

File: doc/2/guides/develop-on-kuzzle/event-system/index.md [hunk 1]
```diff
@@ -187,3 +187,25 @@ await app.trigger('app-name/file-available', fileUrl);
 ::: warning
 If an internal event is triggered, the payload must be the same as the original event.
 :::
+
+**Events can be triggered** with SDK and controllers actions.
```
fmauNeko, line 191:
```suggestion
**Events are always triggered** from SDK and controller actions.
```

doc/2/guides/develop-on-kuzzle/event-system/index.md
```diff
@@ -187,3 +187,25 @@ await app.trigger('app-name/file-available', fileUrl);
...
+
+::: warning
+By default, actions executed through the embedded SDK or controllers won't trigger any events and thus no pipe or hooks will be called.
+:::
+
+This behaviour is set in order to prevent an infinite loop in which a pipe calls a controller generating an event calling this same pipe again and again. 
+
+It is nonetheless possible to pass the flag `allowTrigggerEvents` to the options parameters of the controller to allow events firing :
```
fmauNeko, line 199: There are three `g`s, only two is enough.
Also the parameter should in my opinion be simply called `triggerEvents`, as it instructs the funnel to actually trigger events, and not just allow it to trigger events or not.

fmauNeko, line 199:
```suggestion
It is nonetheless possible to pass the flag `triggerEvents` to the options parameters of the controller to allow events firing :
```

doc/2/guides/develop-on-kuzzle/event-system/index.md
```diff
@@ -187,3 +187,25 @@ await app.trigger('app-name/file-available', fileUrl);
...
+
+```ts
+await this.sdk.document.create(
+              "index",
+              'collection',
+              {
+                contentOfDocument: 'CREATED VIA CONTROLLER',
+              },
+              _idOfDocument,
+              { allowTriggerEvents: true },
```
fmauNeko, line 209:
```suggestion
              { triggerEvents: true },
```

lib/api/funnel.js
```diff
@@ -753,6 +753,9 @@ class Funnel {
    */
   async executePluginRequest(request) {
     try {
+      if (request.input.allowTriggerEvents) {
```
fmauNeko, line 756:
```suggestion
      if (request.input.triggerEvents) {
```

lib/api/request/requestInput.ts
```diff
@@ -32,6 +32,7 @@ const _body = "body\u200b";
 const _headers = "headers\u200b";
 const _controller = "controller\u200b";
 const _action = "action\u200b";
+const _allowTriggerEvents = "allowTriggerEvents\u200b";
```
fmauNeko, line 35:
```suggestion
const _triggerEvents = "triggerEvents\u200b";
```

lib/api/request/requestInput.ts
```diff
@@ -155,6 +156,7 @@ export class RequestInput {
     this[_body] = null;
     this[_controller] = null;
     this[_action] = null;
+    this[_allowTriggerEvents] = null;
```
fmauNeko, line 159:
```suggestion
    this[_triggerEvents] = null;
```

lib/api/request/requestInput.ts
```diff
@@ -181,6 +183,7 @@ export class RequestInput {
     this.body = data.body;
     this.controller = data.controller;
     this.action = data.action;
+    this.allowTriggerEvents = data.allowTriggerEvents;
```
fmauNeko, line 186:
```suggestion
    this.triggerEvents = data.triggerEvents;
```

lib/api/request/requestInput.ts
```diff
@@ -263,7 +266,12 @@ export class RequestInput {
       this[_action] = assert.assertString("action", str);
     }
   }
-
+  get allowTriggerEvents(): boolean | undefined {
+    return this[_allowTriggerEvents];
+  }
+  set allowTriggerEvents(bool: boolean) {
+    this[_allowTriggerEvents] = bool === true ? true : undefined;
+  }
```
fmauNeko, line 274:
```suggestion
  get triggerEvents(): boolean | undefined {
    return this[_triggerEvents];
  }
  set triggerEvents(bool: boolean) {
    this[_triggerEvents] = bool === true ? true : undefined;
  }
```

test/api/funnel/executePluginRequest.test.js
```diff
@@ -102,4 +101,31 @@ describe("funnel.executePluginRequest", () => {
...
+      controller: "testme",
+      action: "succeed",
+    });
+
+    return funnel.executePluginRequest(request).then((response) => {
+      should(response).be.exactly(request);
+      should(kuzzle.pipe).not.calledWith("testme:beforeSucceed");
+      should(kuzzle.pipe).not.calledWith("testme:afterSucceed");
+    });
+  });
```
fmauNeko, line 130:
```suggestion
  it("should trigger pipes if triggerEvent is enabled", async () => {
    const request = new Request({
      controller: "testme",
      action: "succeed",
      triggerEvents: true,
    });

    return funnel.executePluginRequest(request).then((response) => {
      should(response).be.exactly(request);
      should(kuzzle.pipe).calledWith("testme:beforeSucceed");
      should(kuzzle.pipe).calledWith("testme:afterSucceed");
    });
  });

  it("should not trigger pipes if triggerEvent is disabled", async () => {
    const request = new Request({
      controller: "testme",
      action: "succeed",
    });

    return funnel.executePluginRequest(request).then((response) => {
      should(response).be.exactly(request);
      should(kuzzle.pipe).not.calledWith("testme:beforeSucceed");
      should(kuzzle.pipe).not.calledWith("testme:afterSucceed");
    });
  });
```

fmauNeko (changes requested):

doc/2/guides/develop-on-kuzzle/event-system/index.md
QuentinRousselet: But since the default behavior without the flag is to not trigger events wouldn't it be confusing ?

fmauNeko: Well they are triggered from external SDK and controller actions, so I think that's a bit misleading

QuentinRousselet: True...
What about :
**Events are not triggered** with the embedded SDK and controllers actions.

::: warning
By default, actions executed through the embedded SDK or controllers won't trigger any events and thus no pipe or hooks will be called. On the contrary, controllers accessed by the external SDK through HTTP or Websocket requests will always fire events.
:::

rolljee (approved):

fmauNeko (approved):

etrousset (approved):

PR #2547, opened 2024-08-20
Fix global types

fmauNeko:
`global.kuzzle` typing is broken, this PR fixes it.

It also includes a conversion of the `KuzzleEventEmitter` class to TypeScript in order to provide better typings.

### DIFF
--- a/lib/kuzzle/event/KuzzleEventEmitter.ts
+++ b/lib/kuzzle/event/KuzzleEventEmitter.ts
@@ -19,25 +19,43 @@
  * limitations under the License.
  */
 
-"use strict";
-
 // Most of the functions exposed in this file should be viewed as
 // critical section of code.
 
-const assert = require("assert");
+import assert from "assert";
 
-const EventEmitter = require("eventemitter3");
-const Bluebird = require("bluebird");
-const debug = require("debug")("kuzzle:events");
-const { v4: uuidv4 } = require("uuid");
+import Bluebird from "bluebird";
+import EventEmitter from "eventemitter3";
+import { v4 as uuidv4 } from "uuid";
+import { default as kuzzleDebug } from "../../util/debug";
 
-const Promback = require("../../util/promback");
-const memoize = require("../../util/memoize");
-const PipeRunner = require("./pipeRunner");
-const kerror = require("../../kerror");
+import * as kerror from "../../kerror";
+import {
+  AskEventDefinition,
+  AskEventHandler,
+  CallEventHandler,
+  EventDefinition,
+  HookEventHandler,
+  PipeEventHandler,
+} from "../../types/EventHandler";
+import memoize from "../../util/memoize";
+import Promback from "../../util/promback";
+import PipeRunner from "./pipeRunner";
 
-class PluginPipeDefinition {
-  constructor(event, handler, pipeId = null) {
+const debug = kuzzleDebug("kuzzle:events");
+
+class PluginPipeDefinition<
+  TEventDefinition extends EventDefinition = EventDefinition,
+> {
+  public event: TEventDefinition["name"];
+  public handler: PipeEventHandler<TEventDefinition>;
+  public pipeId: string | null;
+
+  constructor(
+    event: TEventDefinition["name"],
+    handler: PipeEventHandler<TEventDefinition>,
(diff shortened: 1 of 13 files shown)


lib/kuzzle/event/KuzzleEventEmitter.ts
```diff
@@ -46,39 +64,43 @@ class PluginPipeDefinition {
...
-     * @type Map<string, Function[]>
      */
     this.pluginPipes = new Map();
 
     /**
      * Map of plugin pipe definitions by pipeId
-     *
-     * @type Map<string, PluginPipeDefinition>
      */
     this.pluginPipeDefinitions = new Map();
```
Kuruyia, line 88: Why didn't you specify the generics explicitly here as well?
```
this.pluginPipes = new Map<string, PipeEventHandler[]>();
...
this.pluginPipeDefinitions = new Map<string, PluginPipeDefinition>();
```

lib/types/Global.ts
```diff
@@ -1,5 +1,5 @@
 import { Backend } from "../core/backend";
-import { Kuzzle } from "../kuzzle";
+import { Kuzzle } from "../kuzzle/kuzzle";
```
Kuruyia, line 2: (If that doesn't break anything), maybe convert the [index.js](https://github.com/kuzzleio/kuzzle/blob/4ef12c7654f2a9af437054ebbae0324a8d481ea0/lib/kuzzle/index.js) file to TypeScript instead?

etrousset (approved):

Kuruyia (approved):

lib/kuzzle/event/KuzzleEventEmitter.ts
```diff
@@ -183,11 +208,13 @@ class KuzzleEventEmitter extends EventEmitter {
     const events = getWildcardEvents(event);
     const funcs = [];
 
-    for (let i = 0; i < events.length; i++) {
```
alexandrebouthinon, line 186: This old for loop was choose to increase performance, I don't know if `for (let i = 0; i < ...)` is still faster than `for (const ... of ...)` in later Node.js versions

lib/kuzzle/event/KuzzleEventEmitter.ts
```diff
@@ -360,8 +403,8 @@ async function pipeCallback(error, ...updated) {
     await Bluebird.map(corePipes, (fn) => fn(...updated));
   }
 
-  for (let i = 0; i < this.events.length; i++) {
```
alexandrebouthinon, line 363: This old for loop was choose to increase performance, I don't know if `for (let i = 0; i < ...)` is still faster than `for (const ... of ...)` in later Node.js versions

lib/kuzzle/event/KuzzleEventEmitter.ts
```diff
@@ -135,17 +157,21 @@ class KuzzleEventEmitter extends EventEmitter {
...
     const events = getWildcardEvents(event);
-    debug('Triggering event "%s" with data: %o', event, data);
+    debug('Triggering event "%s" with data: %o', event, args);
+
+    if (events.length === 0) {
+      debug('No listeners for event "%s"', event);
+      return false;
+    }
 
-    for (let i = 0; i < events.length; i++) {
```
alexandrebouthinon, line 146: This old for loop was choose to increase performance, I don't know if `for (let i = 0; i < ...)` is still faster than `for (const ... of ...)` in later Node.js versions

alexandrebouthinon (approved):
Just a little concern about loop performances since for loops still have better performances than `for (const...)` on latest Node.js versions

lib/kuzzle/event/KuzzleEventEmitter.ts
fmauNeko: I did that change because it was a warning from Sonarqube, and after looking at some benchmarks, it takes almost the same time for the low iteration count we have. 
Here are the results for 10 elements which we probably won't even reach in most cases:
image

fmauNeko: The related Sonar rule: https://rules.sonarsource.com/typescript/RSPEC-4138/

PR #2548, opened 2024-08-28
Elasticsearch 8

rolljee:
## What does this PR do ?

* This PR adds the already tested Elasticsearch 8 Support into the standard workflow


lib/service/storage/commons/queryTranslator.js
Kuruyia, line 1: Because KDM uses it directly, we'll probably want to export the `QueryTranslator` class to avoid breaking it. We'll still need to update the import in KDM tho': https://github.com/kuzzleio/kuzzle-device-manager/blob/9990e57db9bc845d6d6df1e64bed05e88e00f377/lib/modules/measure/MeasureExporter.ts#L5-L6

Kuruyia (changes requested):

fmauNeko (approved):

lib/service/storage/commons/queryTranslator.js
rolljee: I'll add it

Kuruyia (approved):

Kuruyia:
Please squash this PR when merging :)

PR #2555, opened 2024-10-15
feat: do not store clear api keys in internal storage

Juiced66:
KUZSUPPORT-68

# Goals: 
- remove the seed from internal storage in case we have secret provided by config :heavy_check_mark: 
  - removed `_persistSecret` method to use a `_initSecret` that will load secret at backend start
  - set some changes in `TokenRepository` `init` method to detect seed changes and invalidate all tokens
  - refactor validation errors throw order to avoid sending `expired` if token is invalid && expired _(not fully sure it is possible but hey, costs nothing)_
  - Warn users at start when using generated seed image
  - add `authToken` config support because `jwt` is deprecated but `authToken` is not usable 
  - Updated tests to reflect new uses cases.
  - remove test  `_getSecret` wich is not necessary anymore.
  - add documentation warning to encourage use a custom seed when deploying to production using `ENV`
- do not store clear api keys in internal storage :heavy_check_mark: 
  - ensure no breaking changes
  - ensure old api keys still works 

review point of interest : 
- I decided to remove apikeys from cache at it was stored in clear as in the database. It could be kept using fingerprint as a part of the key instead of api key itself ? May need some refactor

### DIFF
--- a/doc/2/guides/getting-started/deploy-your-application/index.md
+++ b/doc/2/guides/getting-started/deploy-your-application/index.md
@@ -51,7 +51,7 @@ A production deployment must include a reverse proxy to securize the connection 
 ::: warning
 #### Production Deployment: Auth Token Secret
 
-For every production deployment of Kuzzle, it is essential to set the kuzzle_security__authToken__secret environment variable. This ensures that the JWT secrets used for authenticating requests are generated externally and not stored in Elasticsearch. By managing the secret through an environment variable, you enhance security, prevent potential data exposure, and ensure tokens remain valid only as long as the secret remains unchanged.
+For every production deployment of Kuzzle, it is essential to set the `kuzzle_security__authToken__secret` environment variable. This ensures that the JWT secrets used for authenticating requests are generated externally and not stored in Elasticsearch. By managing the secret through an environment variable, you enhance security, prevent potential data exposure, and ensure tokens remain valid only as long as the secret remains unchanged.
 
 Important: If the `kuzzle_security__authToken__secret` value is changed when Kuzzle restarts, all existing tokens will be invalidated. This ensures that only tokens signed with the current secret remain valid, adding an extra layer of security.
 
--- a/doc/2/guides/getting-started/deploy-your-application/index.md
+++ b/doc/2/guides/getting-started/deploy-your-application/index.md
@@ -49,17 +49,25 @@ A production deployment must include a reverse proxy to securize the connection 
 :::
 
 ::: warning
-#### Production Deployment: Auth Token Secret
+# Authentication Security in Production
 
-For every production deployment of Kuzzle, it is essential to set the `kuzzle_security__authToken__secret` environment variable. This ensures that the JWT secrets used for authenticating requests are generated externally and not stored in Elasticsearch. By managing the secret through an environment variable, you enhance security, prevent potential data exposure, and ensure tokens remain valid only as long as the secret remains unchanged.
+## ⚠️ Important Security Requirement
 
-Important: If the `kuzzle_security__authToken__secret` value is changed when Kuzzle restarts, all existing tokens will be invalidated. This ensures that only tokens signed with the current secret remain valid, adding an extra layer of security.
+You must set the `kuzzle_security__authToken__secret` environment variable before deploying Kuzzle to production. This secret is used to sign and verify JSON Web Tokens (JWTs) for user authentication.
 
-For default configuration values, you can refer to [the sample Kuzzle configuration file](https://github.com/kuzzleio/kuzzle/blob/master/.kuzzlerc.sample.jsonc).
+## Why This Matters
+- Prevents tokens from being stored in Elasticsearch
+- Improves overall security
+- Gives you direct control over token management
 
-Note: If the secret is not set, Kuzzle will fallback to a less secure method of generating and storing the secret, which is not recommended for production environments.
+## Security Notes
+1. **Fallback Warning**: If you don't set this variable, Kuzzle will use a less secure fallback method (not recommended for production)
+2. **Token Invalidation**: Changing the secret value will immediately invalidate all existing authentication tokens
+3. **User Impact**: Users will need to log in again if the secret changes
+
+## Additional Resources
+For other configuration options, see the [sample configuration file](https://github.com/kuzzleio/kuzzle/blob/master/.kuzzlerc.sample.jsonc).
 :::
-
 ## Prepare our Docker Compose deployment
 
 We are going to write a `docker-compose.yml` file that describes our services.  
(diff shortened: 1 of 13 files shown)


doc/2/guides/getting-started/deploy-your-application/index.md
```diff
@@ -48,6 +48,18 @@ This deployment does not use any SSL encryption (HTTPS).
 A production deployment must include a reverse proxy to securize the connection with SSL.
 :::
 
+::: warning
+#### Production Deployment: Auth Token Secret
+
+For every production deployment of Kuzzle, it is essential to set the kuzzle_security__authToken__secret environment variable. This ensures that the JWT secrets used for authenticating requests are generated externally and not stored in Elasticsearch. By managing the secret through an environment variable, you enhance security, prevent potential data exposure, and ensure tokens remain valid only as long as the secret remains unchanged.
```
Juiced66, line 54:
```suggestion

For every production deployment of Kuzzle, it is essential to set the `kuzzle_security__authToken__secret` environment variable. This ensures that the JWT secrets used for authenticating requests are generated externally and not stored in Elasticsearch. By managing the secret through an environment variable, you enhance security, prevent potential data exposure, and ensure tokens remain valid only as long as the secret remains unchanged.
```

lib/core/security/tokenRepository.ts
```diff
@@ -136,6 +136,24 @@ export class TokenRepository extends ObjectRepository<Token> {
     global.kuzzle.onAsk("core:security:token:verify", (hash) =>
       this.verifyToken(hash),
     );
+
+    // ? those checks are necessary to detect JWT seed changes and delete existing token if necessary
```
Juiced66, line 140:
```suggestion
    // ? those checks are necessary to detect JWT seed changes and delete existing tokens if necessary
```

lib/core/security/tokenRepository.ts
```diff
@@ -352,6 +387,38 @@ export class TokenRepository extends ObjectRepository<Token> {
     return userToken;
   }
 
+  async _verifyApiKey(decoded, token: string) {
+    const fingerprint = sha256(token);
+
+    const userApiKeys = await ApiKey.search({
+      query: {
+        term: {
+          userId: decoded._id,
```
fmauNeko, line 396: Why not searching by `fingerprint` first and then checking that the `userID` matches?
Then the search would only return 0 or 1 result, and it would save the `Array.find` call.

lib/kuzzle/internalIndexHandler.js
```diff
@@ -202,24 +203,24 @@ class InternalIndexHandler extends Store {
     await Bluebird.all(promises);
   }
 
-  async getSecret() {
-    const response = await this.get("config", this._JWT_SECRET_ID);
+  async _initSecret() {
+    const { authToken, jwt } = global.kuzzle.config.security;
+    const configSeed =
+      authToken && authToken.secret ? authToken.secret : jwt && jwt.secret;
```
fmauNeko, line 209: Could be simplified in `authToken?.secret ?? jwt?.secret`

fmauNeko (changes requested):

lib/core/security/tokenRepository.ts
Juiced66: Because it lead to introduce a breaking (fingerprint is not indexed atm and changes in mappings are not taken in account at start)

lib/kuzzle/internalIndexHandler.js
Juiced66: applied !

lib/core/security/tokenRepository.ts
fmauNeko: Good call! Nevermind then.

fmauNeko (approved):

etrousset (changes requested):
Kuzzle need to log a big WARNING message when it uses a seed that is stored in ES to informa that it's ok on dev environements, but NEVER on production environement

sebtiz13 (approved):

etrousset (approved):

lib/core/security/tokenRepository.ts
```diff
@@ -308,36 +338,41 @@ export class TokenRepository extends ObjectRepository<Token> {
...
+
       throw securityError.getFrom(err, "verification_error", err.message);
     }
 
-    let userToken: Token;
+    if (isApiKey) {
+      return this._verifyApiKey(decoded, token);
+    }
+
+    let userToken;
```
rolljee, line 368: After discussing with @Juiced66. This function should be reworked to give two different path.

1. Path one, will only use the _verifyApiKey(token) and decode the token the way it need to be from this function
1. Path two, add a new function _verifyAuthToken(token) that also check the verification the way it need.

We may need to have a common jwt.verify function as well.

This will allow us to have two distinct ways of testing the validity of theses two way of logging in.

rolljee (changes requested):

alexandrebouthinon (approved):
Just tested on a local environment with already created API Keys by deleting the doc containing seed and pass it have env variable. Worked like a charm

lib/core/security/tokenRepository.ts
```diff
@@ -136,6 +134,24 @@ export class TokenRepository extends ObjectRepository<Token> {
...
+    // ? those checks are necessary to detect JWT seed changes and delete existing tokens if necessary
+    const existingTokens = await global.kuzzle.ask(
+      "core:cache:internal:searchKeys",
+      "repos/kuzzle/token/*",
+    );
+
+    if (existingTokens.length > 0) {
+      try {
+        const [, token] = existingTokens[0].split("#");
+        await this.verifyToken(token);
```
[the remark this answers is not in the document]
rolljee: This is not useful since you test the first one

lib/core/security/tokenRepository.ts
Juiced66: As discussed this mornig, those changes will be done in another PR as this one is working as intended

lib/kuzzle/internalIndexHandler.js
```diff
@@ -202,24 +203,29 @@ class InternalIndexHandler extends Store {
...
 
-    await this.create(
-      "config",
-      { seed },
-      {
-        id: this._JWT_SECRET_ID,
-      },
-    );
+      global.kuzzle.log.warn(
+        "[!] Kuzzle is using generated seed for authentication. This is suitable for development but should NEVER be use in Production. See https://docs.kuzzle.io/core/2/guides/getting-started/deploy-your-application/",
```
fmauNeko, line 223:
```suggestion
        "[!] Kuzzle is using a generated seed for authentication. This is suitable for development but should NEVER be used in production. See https://docs.kuzzle.io/core/2/guides/getting-started/deploy-your-application/",
```

fmauNeko (changes requested):

fmauNeko (approved):

PR #2566, opened 2024-12-18
fix: deployment workflow run rework

rolljee:
## What does this PR do ?

This PR rework the way the deployment workflow runs.

We had an issue where some docker images were not properly tagged. For example, [this](https://github.com/kuzzleio/kuzzle/actions/runs/11724916099/job/32659955410#step:7:7) run of CD should have created the 2.34.0 tag instead of the 2.34.0-beta.1 tag for the plugin-dev.

Turns out, the deployment using semantic was launched at the same time as the deployment docker images.

When semantic release ends the process it makes a commit to update the NPM version of Kuzzle. (You start to see what's coming?)

This commit was not always pushed when the process read the current package.json

```yaml
      - name: Get current version from package.json
        shell: bash
        id: get-version
        run: |
          echo "version=$(jq -r .version package.json)" >> $GITHUB_OUTPUT
          echo "major-version=$(jq -r .version package.json | cut -d. -f 1)" >> $GITHUB_OUTPUT
```

Since the package version isn't updated, we still have the -beta.x suffix. Which result in a wrongly tagged docker image.

This PR aim to fix this issue making sure the semantic release process is ended when we start creating all docker images.

Well, you can't.

### Other changes

Nope, nothing


fmauNeko (commented):
Is there a specific reason to use `ubuntu-20.04` instead of a more recent version ?

rolljee:
> Is there a specific reason to use `ubuntu-20.04` instead of a more recent version ?

I made some test trying to fix an issue with npm install. but i found the guilty one. I'll run using ubuntu 24.04 instead

fmauNeko (approved):

alexandrebouthinon (approved):

PR #2572, opened 2025-01-06
feat: remove murmurhash native in favor of murmurhashJS

rolljee:
# Overview

This PR aim to remove murmurash native in favor of murmurhashJS to avoid building murmurhash-native witch aim to fail a lot in CI/CD because it need to be build each time we npm install.

### DIFF
--- a/lib/kuzzle/kuzzle.ts
+++ b/lib/kuzzle/kuzzle.ts
@@ -25,7 +25,7 @@ import Bluebird from "bluebird";
 import stringify from "json-stable-stringify";
 import { Koncorde } from "koncorde";
 import _ from "lodash";
-import { murmurHash128 as murmur } from "murmurhash-native";
+import murmur from "murmurhash";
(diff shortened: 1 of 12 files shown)


Juiced66 (approved):

lib/kuzzle/kuzzle.ts
```diff
@@ -823,9 +823,8 @@ class Kuzzle extends KuzzleEventEmitter {
         inString = stringify(input);
     }
 
-    return murmur(
+    return murmur.v3(
```
fmauNeko, line 826: This library generates 32-bit hashes instead of the 128-bit ones we had previously.
See https://www.npmjs.com/package/murmurhash

rolljee: Okay, you are right i found one that uses 128Bit w/o binaries. But i wonder if this is really usefull ?

fmauNeko: It is not really indeed.

PR #2587, opened 2025-01-30
chore(npm): bump dumpme to solve intermittent build issues

fmauNeko:
## What does this PR do ?

Update [`dumpme`](https://github.com/kuzzleio/dumpme) to solve intermittent build issues.


etrousset (approved):

etrousset (changes requested):
Change commit type so that it appears in the release note, some think like maintenance ou fix

PR #2593, opened 2025-03-03
fix: update es version to avoid anyController error

rolljee:
# overview

With newer versions of docker desktop an error started to occur. This update the Elasticsearch image to fix this issue


Juiced66 (approved):

fmauNeko (approved):

fmauNeko (approved):
Isn't it overriden in the `release.workflow.yaml` workflow tho ? Which is set to .26, and according to Docker Hub we already build .26

rolljee:
> Isn't it overriden in the `release.workflow.yaml` workflow tho ? Which is set to .26, and according to Docker Hub we already build .26

Yes you are right this PR isn't useful in itself but at least we are coherent in the whole process

PR #2605, opened 2025-05-07
fix(funnel): handle errors for trigger event requests

QuentinRousselet:
## What does this PR do ?
This PR fix an issue where some errors thrown from a pipe triggered by a plugin request with `triggerEvents` flag weren't caught.

The error handling is the same for plugin requests with triggerEvents as for basic requests.

### DIFF
--- a/lib/api/funnel.js
+++ b/lib/api/funnel.js
@@ -754,6 +754,8 @@ class Funnel {
    */
   async executePluginRequest(request) {
     if (request.input.triggerEvents) {
+      let error;
+      let res;
       try {
         const { result } = await this.processRequest(request);
         debug(
@@ -761,17 +763,17 @@ class Funnel {
           request.id,
           result,
         );
-        return global.kuzzle.pipe("request:afterExecution", {
-          request,
-          result,
-          success: true,
-        });
-      } catch (error) {
+        res = result;
+        return { ...result };
+      } catch (e) {
+        error = e;
         debug("Error processing request %s: %a", request.id, error);
-        return global.kuzzle.pipe("request:afterExecution", {
+      } finally {
+        global.kuzzle.pipe("request:afterExecution", {
           error: error,
           request: request,
-          success: false,
+          result: res,
+          success: error === undefined ? true : false,
         });
       }
     }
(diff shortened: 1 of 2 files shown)


rolljee (approved):

lib/api/funnel.js
```diff
@@ -753,11 +753,29 @@ class Funnel {
...
+          request,
+          result,
+          success: true,
+        });
+      } catch (error) {
+        debug("Error processing request %s: %a", request.id, error);
+        return global.kuzzle.pipe("request:afterExecution", {
+          error: error,
+          request: request,
+          success: false,
```
alexandrebouthinon, line 774: IMO this is the wrong events, maybe you should use the `request:onError` event instead. Since the request has not be executed in the right controller

lib/api/funnel.js
```diff
@@ -753,11 +753,29 @@ class Funnel {
...
+        debug(
+          "Request %s successfully executed. Result: %a",
+          request.id,
+          result,
+        );
+        return global.kuzzle.pipe("request:afterExecution", {
+          request,
+          result,
+          success: true,
+        });
```
alexandrebouthinon, line 768: Wrong location for this event, should be triggered after the `doAction` method

alexandrebouthinon (changes requested):

lib/api/funnel.js
```diff
@@ -753,11 +753,29 @@ class Funnel {
...
+      } catch (error) {
+        debug("Error processing request %s: %a", request.id, error);
+        return global.kuzzle.pipe("request:afterExecution", {
```
alexandrebouthinon, line 771: We need to rethrow the error after the error event trigger. If we let this as it, this will change the final pipe error behavior which results in a hidden edge effect/breaking in pipe development experience. You should add a note to update the behavior on next major since we should do as it on next major

QuentinRousselet: After checking the behaviour without the triggerEvent flag it appears that any error inside the pipe is caught
For this reason we shouldn't rethrow the error to have the same behaviour when the flag is enabled

PR #2607, opened 2025-05-15
fix(elasticsearch): user provided _kuzzle_info for M operations

inkedsquid:
This PR makes the multiples operations (mCreate, mUpdate, mUpsert, etc.) ignore user provided _kuzzle_info, which should not be modified by anyone but the backend.

### DIFF
--- a/lib/service/storage/7/elasticsearch.ts
+++ b/lib/service/storage/7/elasticsearch.ts
@@ -3080,6 +3080,8 @@ export class ES7 {
     extractedDocuments: JSONObject[],
     documentsToGet: JSONObject[],
   ) {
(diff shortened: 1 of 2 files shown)


rolljee (changes requested):
You should add a new test that confirm that the property is now deleted at this level and that it is a wanted behavior

rolljee:
There is something about this PR that does not feel right. 

It seems to me that this is a breaking change of a behavior that was here for a long time. With this, people that are relying on this will need to modify their codebase.

To avoid this, the default behavior should remain the same as before and you should add a new option that will allow us to choose whether we want to enable or disable mCreate, mUpdate, mUpsert kuzzle_info modifications

PR #2616, opened 2025-08-07
feat: use the Kuzzle logger on the Application

alexandrebouthinon:
## What does this PR do ?

This permits to use a scoped logger in the application. Note that as the Kuzzle logger itself, it cannot be use before the `start()` method has been called.

Create an application, call the start method, log using `this.log`. You should see your log rightly scoped to your application name:

```js
// Starts a Kuzzle Backend application tailored for development
import { Backend } from "../../index";

const app = new Backend("development-app");

app.start()
  .then(() => {
    app.log.info("Kuzzle Backend application started successfully");
  })
  .catch((error) => {
    app.log.error(`Failed to start Kuzzle Backend application: ${error}`);
    process.exit(1);
});
```
image

## Other changes

- Add the correct logger namespace for each native Kuzzle component (minus Plugin and PluginContext to guarantee compatibility)
- Do not add the namespaced logger to the IDCardRenewer and EmbeddedSDK class since they use an exotic execution context

### Kuzzle start log flow after changes

image

> All the components logs are namespaced

Linked to KZLPRD-916

### DIFF
--- a/test/api/funnel/executePluginRequest.test.js
+++ b/test/api/funnel/executePluginRequest.test.js
@@ -54,7 +54,7 @@ describe("funnel.executePluginRequest", () => {
     const callback = () => {
       setTimeout(() => {
         try {
-          should(kuzzle.log.error).be.calledOnce();
+          should(funnel.logger.error).be.calledOnce();
           should(kuzzle.dump).be.called();
           done();
         } catch (e) {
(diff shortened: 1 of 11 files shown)


lib/core/backend/backend.ts
```diff
@@ -298,7 +297,7 @@ export class Backend {
     }
 
     this._kuzzle = new Kuzzle(this.config.content);
-
+    this.log = this._kuzzle.log.child(`app:${this.name}`);
```
etrousset, line 300:
```suggestion
    this.log = this._kuzzle.log.child(`app`);
```
I would just create the "app" level and let the developper add its how sublevel as he wishes

etrousset (changes requested):

rolljee (approved):

PR #2617, opened 2025-09-10
feat(log): allow Kuzzle application devs to use the logger before the application startup

alexandrebouthinon:
## What does this PR do ?

Generate the nodeId before the Kuzzle cluster initialisation to allow the loggers to access to the value and use it on every log record

Start the dev application using:

```bash
npm i
kourou app:start-services
npm run dev
```

> NOTE: you can try to customize the dev application in `docker/scripts/start-kuzzle-dev.ts` to add logs right after the constructor call, create child loggers .... 

You should see the nodeId label on every log lines even if the application didn't start yet.
### Other changes

Create a `global.nodeId` variable to access the nodeId value for the current Kuzzle node

### DIFF
--- a/lib/core/backend/backend.ts
+++ b/lib/core/backend/backend.ts
@@ -220,7 +220,7 @@ export class Backend {
    */
   public _support: JSONObject = {};
 
-  private nodeId: string;
+  private readonly nodeId: string;
 
   /**
    * Instantiates a new Kuzzle application


lib/core/backend/backend.ts
```diff
@@ -247,6 +251,15 @@ export class Backend {
       writable: true,
     });
 
+    Reflect.defineProperty(global, "nodeId", {
+      get: () => {
+        return this.nodeId;
+      },
+      set: () => {
+        throw new Error("nodeId is read-only");
```
rolljee, line 259: Can it just be writable: false here ?

rolljee (approved):

lib/core/backend/backend.ts
alexandrebouthinon: Nop can't define setter/getter and use writable property (I tried your suggestion first while developing)

PR #2663, opened 2026-06-25
fix(securityController): token revocation

inkedsquid:
## What does this PR do ?

Fixes an issue where `revokeTokens` only deleted session tokens (JWTs) from the cache, leaving persistent API keys fully intact and visible when calling `searchAPIKeys`. 

  - Step 1: Create a new user in Kuzzle.
  - Step 2: Create a new API key for this user.
  - Step 3: Run `searchAPIKeys` for the user and verify the key is present in the list.
  - Step 4: Revoke the user's tokens using `revokeTokens`.
  - Step 5: Run `searchAPIKeys` again and verify that the API key is no longer present in the list.

### DIFF
--- a/lib/api/controllers/securityController.ts
+++ b/lib/api/controllers/securityController.ts
@@ -1332,8 +1332,11 @@ export default class SecurityController extends NativeController {
    */
   async revokeTokens(request: KuzzleRequest) {
     const id = request.getId();
+    const refresh = request.getRefresh("wait_for");
 
     await this.ask("core:security:token:deleteByKuid", id);
+
+    await ApiKey.deleteByUser({ _id: id } as any, { refresh });
 
     return null;
   }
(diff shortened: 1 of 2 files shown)


lib/api/controllers/securityController.ts
```diff
@@ -1332,9 +1332,12 @@ export default class SecurityController extends NativeController {
    */
   async revokeTokens(request: KuzzleRequest) {
     const id = request.getId();
+    const refresh = request.getRefresh("wait_for");
 
     await this.ask("core:security:token:deleteByKuid", id);
 
+    await ApiKey.deleteByUser({ _id: id } as any, { refresh });
```
etrousset, line 1339: This doesn't to be the right function to call, as this will deletes all the tokens linked to a user!!! 

We should revoke only the requested token!

etrousset (changes requested):

PR #2664, opened 2026-06-25
feat(utils): distributed reentrant mutex

inkedsquid:
## What does this PR do ?
This PR introduces a distributed reentrant mutex utilizing `redlock-universal`.

Specifically, it:
- Adds the `redlock-universal` dependency to handle distributed locking via Redis.
- Implements `withLock` in `lib/util/distributedLock.ts`, utilizing Node's `AsyncLocalStorage` to keep track of acquired locks within the current execution context, achieving lock reentrancy. Lock loss during a held callback (e.g. the key expiring or being stolen) is surfaced as a `MutexLockLostError` instead of silently letting the callback keep running unprotected.
- Deprecates the existing `Mutex` class in `lib/util/mutex.ts` in favor of `withLock`, with a note not to mix the two on the same resource key.
- Adds one internal ask handler (`core:cache:internal:client:get`) to `lib/core/cache/cacheEngine.js`, exposing the raw ioredis client so `distributedLock.ts` can build its adapter in-process only, never exposed over the network.

  - Step 1 : Write an endpoint that call a mutex with `withLock` for the first time
  - Step 2 : Call it again in its context
  - Step 3 : It should not block until the first one timeout

### Boyscout

- Converted `service` and `redis` cache service files to TypeScript (`.ts`) to improve the codebase's maintainability and developer experience, with no remaining `any` usage.

KZLPRD-1241


drichard-k (approved):

drichard-k (approved):
Awaiting a review by someone else since I contributed to this

etrousset (approved):

PR #2684, opened 2026-08-10
chore(deps): update kuzzle vault

drichard-k:
- Upgrades the kuzzle vault version
- Added an option: whether or not to use the new version. If not present use the old version to avoid a breaking change


package.json
```diff
@@ -49,7 +49,7 @@
     "kuzzle-logger": "1.4.0",
     "kuzzle-plugin-auth-passport-local": "6.4.1",
     "kuzzle-sdk": ">=7.17.1 <8.0.0",
-    "kuzzle-vault": "2.1.0",
+    "kuzzle-vault": "2.1.0-beta.2",
```
rolljee, line 52: We should not use this version, please make a stable version of this, then use it here :)

rolljee (changes requested):

rolljee (approved):

Issue #2692, opened 2026-09-09
SonarCloud measures no coverage on .ts — every TS conversion voids its own coverage gate (TD-24)


rolljee:
Tracked as **TD-24** in [`docs/adr-001/type-debt-register.md`](https://github.com/kuzzleio/kuzzle/blob/2-dev/docs/adr-001/type-debt-register.md#td-24). **This blocks the start of ADR-0001 Sprint 5 (`lib/core`).**

`sonar-project.properties` has:

```
sonar.coverage.exclusions=**/*.ts,**/*.vue
```

added by #2658 on 2026-05-21, before the TypeScript-migration ADR. It excludes **all** TypeScript from coverage measurement, so every `.js` → `.ts` rename **removes its file from the "Coverage on New Code" gate**. The gate is therefore vacuous on exactly the lines a conversion PR touches — and it has already silently swallowed the **30 files converted so far**. The conversion PRs noted this in passing ("`.ts` is coverage-excluded") as if it were neutral; it is the largest hole in the migration's safety story, and it widens with every sprint.

Two compounding problems in the pipeline:

- Only `npm run test:unit:mocha:coverage` runs in the `sonarqube` job, so **vitest coverage never reaches the scanner** — which also means the ADR's "every new unit test in vitest" produces no measurable coverage at all.
- `c8` (mocha, over the compiled `dist/`) and vitest both write to `coverage/lcov.info`, so wiring vitest in naively would have one report **overwrite** the other rather than add to it.

### Plan

1. Drop `**/*.ts` from `sonar.coverage.exclusions`.
2. Verify where the c8 lcov actually points: mocha runs `dist/**/*.test.js` with `sourceMap: true`, so c8 *should* remap onto `lib/**/*.ts` — confirm empirically, and remap explicitly otherwise.
3. Give each runner its own report path and list both in `sonar.javascript.lcov.reportPaths`.
4. Record the measured coverage of the already-converted files in ADR step 06 — that number is the missing piece of the mid-course review.

Scheduled as **PR F2** of [ADR-0001 step 06](https://github.com/kuzzleio/kuzzle/blob/2-dev/docs/adr-001/steps/06-hardening-mid-course.md). Converting the 50 critical files of `lib/core` under a gate that measures nothing is the risk this issue exists to prevent.

rolljee:
Done and shipped — closing.

[...]

PR #2702, opened 2026-09-09
chore(td-22): type memoryStorageController's Redis-command table

### LINKED ISSUE #2690: TS migration: memoryStorageController was renamed, not typed (TD-22)
Tracked as **TD-22** in [`docs/adr-001/type-debt-register.md`](https://github.com/kuzzleio/kuzzle/blob/2-dev/docs/adr-001/type-debt-register.md#td-22).

`lib/api/controllers/memoryStorageController.ts` was converted in Sprint 4 PR #2681 with **0 written `any`** — and **54 implicit-`any` diagnostics** (`TS7xxx`), the largest pocket among the converted files. They cascade into **37 `TS2339`/`TS2551`** errors, i.e. 91 diagnostics for this file alone, and it is the **worst file of the converted set under strict: 122 errors**.

Concretely:

- `let mapping;` — an untyped mutable module binding, assigned by `initMapping()` and read by the constructor.
- 38 un-annotated `map: (val, request) => …` closures in the Redis-command table.
- 5 `result` accumulators inferring `any[]`.
- `this[command] = buildCommandFn(command)` indexing the class dynamically.

Step 05 recorded that the table and the arg-extraction internals "stay inferred", which was a deliberate call — but nothing recorded the size of it, and the `any` ratchet charged nothing for it. It is now visible in the new `implicit-any` ratchet (#2689).

### Suggested approach

- Type the command table as a named interface (`RedisCommandMapping`, one entry type with the `skip` / `merge` / `path` / `map` shape).
- Make `mapping` a `const` built at module load instead of a `let` + `initMapping()`.
- Annotate the `map` closures.

Fixing `mapping`'s type is what clears the 37 cascades — they are not independent findings.

### Constraint to respect

`test/api/controllers/memoryStorageController.test.js` drives `rewire(...).__get__("mapping" | "extractArgumentsFromRequest")` and `__set__({ mapping })` on the **compiled** CJS, so `mapping` must stay a top-level binding and the module must keep its `export =` shape.

Good candidate for a first `implicit-any` reduction PR: the file is self-contained and its spec is thorough.

rolljee:
Closes #2690 — type-debt register [TD-22](https://github.com/kuzzleio/kuzzle/blob/2-dev/docs/adr-001/type-debt-register.md#td-22). First `implicit-any` reduction PR.

## What

`memoryStorageController` was converted in Sprint 4 with **0 written `any`** and **54 implicit ones** — the largest inferred-`any` pocket of the converted set — which cascaded into **37 `TS2339`** errors, because `let mapping;` gave the whole command table the type `any`.

The table now has names:

- `CommandArgumentPath` — where an argument sits in the request (`["body", "keys"]`)
- `CommandArgumentSpec` — the `path` / `merge` / `skip` / `map` shape
- `CommandArguments` — one command's arguments, `null` when it takes none (`dbsize`, `flushdb`)
- `RedisCommandMapping` — command name → its arguments
- `GeoPoint`, `FieldEntry`, `KeyEntry` — the three shapes the `geoadd` / `hmset` / `mset` closures assert at runtime

The `map` closures, the `result` accumulators, the arg-extraction `reduce` and the four helpers (`toArray`, `sanitizeArrayArgument`, `processLimit`, `assertFloat`/`assertInt`) are annotated.

## Numbers

| | before | after |
|---|---|---|
| implicit-any ratchet (repo) | 518 | **464** |
| diagnostics in this file | 91 | **0** |
| strict errors, this file | 122 | **54** |
| strict errors, repo | 1344 | **1309** |
| written `any` ratchet | 208 | **208** (unchanged) |

No hatch was used to get there: no `any`, no `as unknown as`, no `@ts-ignore`.

## Three deviations from the issue's plan, on purpose

1. **`mapping` stays a `let` assigned by `initMapping()`** rather than becoming a `const`. The Mocha spec does `__set__({ mapping })` on the *compiled* CJS, and assigning to a `const` throws at runtime. The annotation alone is what cleared the `TS7034`/`TS7005` and all 37 cascades — the `const` bought nothing.
2. **The `map` closures are annotated individually.** An entry is `CommandArgumentPath | CommandArgumentSpec` (the bare path is used 100+ times) and TypeScript does not contextually type a callback through a union.
3. **`this[command] = …` is typed by a class index signature** (`[command: string]: unknown`), not a cast: type-only, emitted JS unchanged, and it costs the `any` ratchet nothing where `as unknown as Record<string, unknown>` would have cost 1 (measured — the ratchet caught it).

`assertFloat`/`assertInt` take `unknown` and parse `String(value)`: `Number.parseFloat` agrees on every input the old untyped signature accepted (arrays are rejected before the parse, as before).

## Validation

- mocha: **3025 passing**, 0 failing — including the `rewire`-based spec that drives `mapping` and `extractArgumentsFromRequest` directly, which is the real guard on this change
- vitest: 166 passing
- ratchets: js 50 / mocha 151 / any 208 / implicit-any **464** (baseline updated in the same commit, as the ratchet requires)
- `npm run test:strict`: 101 adopted files pass; no new strict candidate (the file still has 54)
- lint 0 errors, `tsc --noEmit` clean, build green

Its `sonar.cpd.exclusions` entry stays: what CPD scores there is the command table's near-identical entries, which is not what this PR touches.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rolljee:
The SonarCloud gate failed on **new coverage: 62.4% < 80%** — worth recording why, because it is the first time the gate has caught something real since #2692 removed `**/*.ts` from `sonar.coverage.exclusions`.

Every line this PR annotates counts as new code, and `test/api/controllers/memoryStorageController.test.js` replaces the real command table with a 6-command fixture (`__set__({ mapping: testMapping })`), so **not one of the table's own `map` closures ever ran** — the file read as 76.8% covered while the parts this PR touches were not covered at all.

Fixed by a spec rather than by an exclusion: `tests/api/controllers/memoryStorageController.test.ts` drives the real table through the public actions (`geoadd`, `hmset`, `mset`, `scan`, `zrangebyscore`, `zadd`, and an argument-less command) with a stubbed `global.kuzzle.ask`, asserting the argument list handed to Redis plus the rejection paths (empty point list, malformed entry, non-string `MATCH`, non-integer `COUNT`, one-element `LIMIT`). 17 tests; vitest 166 → 183, mocha unchanged at 3025.

Locally the two lcov reports now cover 83.3% of the file's lines. What is left uncovered on the new lines is multi-line function *signatures* — c8 over the compiled `dist/` marks a declaration line as executable but never hit, and prettier splits any signature past 80 columns.

Issue #2776, opened 2026-09-17
evictSelf neither stops the node nor resyncs its counter: one lost sync message permanently poisons a node that keeps serving traffic (TD-67)


rolljee:
Found in CI on [#2775](https://github.com/kuzzleio/kuzzle/pull/2775) — `Functional tests (http, 24, 8)`, run 35230334541. This is the first occurrence of [TD-33](https://github.com/kuzzleio/kuzzle/issues/2715) where the log ties the lost message to the failing assertion end to end, and it names a **second** defect, downstream of [TD-65](https://github.com/kuzzleio/kuzzle/issues/2773).

## What the log shows

`kuzzle_node_2` is `knode-wrathful-potamoi-85816`.

```
14:01:46.507  node_1  [CLUSTER] Successfully completed the handshake with node knode-solid-peacock-99998
14:01:47.276  node_2  ERROR [CLUSTER] Node out-of-sync: 1 messages lost from node knode-solid-peacock-99998
14:01:47.280  node_1  WARN  [CLUSTER] Node "knode-wrathful-potamoi-85816" evicted.
                            Reason: Node out-of-sync: 1 messages lost from node knode-solid-peacock-99998
…           node_2  the same ERROR, 9 times, through 14:01:52.422
14:02:19      the step `Given an existing collection "nyc-open-data":"yellow-taxi"`
              fails with `Error: Index nyc-open-data does not exist`
```

**`kuzzle_node_2` never shut down.** It was evicted from every other node's view at 14:01:47 and went on answering HTTP behind nginx for the remaining 33 seconds of the run, with state that had stopped tracking the cluster. The failing step is simply the first one nginx happened to route to it after `loadFixtures` created the index elsewhere.

## Why the node survives its own eviction

`node.js`:

```js
async evictSelf(reason, error = null) {
  this.logger.error(`[CLUSTER] ${reason}`);
  if (error) { this.logger.error(error.stack); }
  this.publisher.sendNodeEvicted(this.nodeId, this.nodeId, reason);
}
```

It logs, and broadcasts `NodeEvicted` naming itself. Every **other** node receives that and takes the `message.nodeId !== this.localNode.nodeId` branch of `handleNodeEviction`, dropping it from the cluster. The broadcasting node never receives its own publication — a ZeroMQ `PUB` does not loop back — so the `global.kuzzle.shutdown()` branch, the one written for exactly this case, is unreachable for the node that needs it.

[...]

Issue #2773, opened 2026-09-17
A joining node resumes from a cluster message id its sync subscription had not started tracking yet (TD-65)


rolljee:
Found while converting `lib/cluster/publisher.js` in sprint 8 J1 ([ADR-0001 step 11](https://github.com/kuzzleio/kuzzle/blob/2-dev/docs/adr-001/steps/11-sprint-8-cluster.md)). This is a reading of [TD-33](https://github.com/kuzzleio/kuzzle/issues/2715)'s surviving half, derived from the code rather than from a reproduction.

## The two channels

`publisher.ts` binds a ZeroMQ `PUB` on `config.ports.sync` (7511). `command.js` binds its request/response sockets on `config.ports.command` (7510). They share no ordering — a message sent on one establishes nothing about the other.

## The handshake uses both, in this order

In `node.js`, `handshake()`:

1. For every discovered node: `new ClusterSubscriber(...)` then `await subscriber.init()`, which does `socket.connect(syncAddress)` + `socket.subscribe()`.
2. `await command.getFullState(nodes)` — a round trip on the **command** channel. Its answer carries each node's `lastMessageId`, read off `this.node.publisher.lastMessageId` (`command.js:129`).
3. `await command.broadcastHandshake(nodes)`.
4. `subscriber.sync(lastMessageId)` — the subscriber adopts that id and from then on expects `lastMessageId + 1`.

## Step 1 is not finished when it returns

In ZeroMQ, `connect()` is asynchronous, and a subscription is a message the subscriber sends **to the publisher**, which performs the filtering. Until it arrives, the remote `PUB` has no subscriber on that connection and **silently drops** what it publishes. This is the *slow joiner*, the documented failure mode of PUB/SUB.

Let the snapshot in step 2 be taken at T1 and the subscription register at Ts. Nothing orders Ts before T1 — different sockets, different connections. If **Ts > T1**, every message published in (T1, Ts) is dropped, while step 4 has told the subscriber to resume at T1. Its next message is not T1+1, `subscriber.js`'s gap check fires, and the node self-evicts with `Node out-of-sync: N messages lost`.

## Why this fits what TD-33 recorded

- **Exactly one message lost**, every time. At formation the only periodic traffic on the sync channel is `sendHeartbeat`, so the window straddles zero or one publication.
- **Right after the handshakes complete** — step 4 is the first moment the counter is enforced.
[...]

Issue #2715, opened 2026-09-10
Functional test matrix: one flaky variant blocks unrelated PRs (cluster readiness is not gated)


rolljee:
The functional matrix (30 variants: protocol × node × ES) fails on a single variant often enough to block unrelated PRs. Because the matrix is `fail-fast`, **one flake cancels the other 29 jobs** and the PR has to be re-run in full.

## Occurrences

Three so far, all on PRs that cannot plausibly cause them:

| When | PR | Variant | Symptom |
|---|---|---|---|
| 2026-09-09 | [#2696](https://github.com/kuzzleio/kuzzle/pull/2696) (sprint 5 G2) | one of 30 | `api.process.not_enough_nodes` in a `BeforeAll` hook |
| 2026-09-10 | [#2708](https://github.com/kuzzleio/kuzzle/pull/2708) (**docs-only**, touches no code) | `legacy:websocket, 22, 7` — [job](https://github.com/kuzzleio/kuzzle/actions/runs/34462395554/job/102824156737) | `wait-kuzzle` timeout on port 17510 |
| 2026-09-10 | [#2712](https://github.com/kuzzleio/kuzzle/pull/2712) | `websocket, 24, 7` — [job](https://github.com/kuzzleio/kuzzle/actions/runs/34466855502/job/102838510401) | `not_enough_nodes` in a `Before` hook |

Both 2026-09-10 occurrences passed on re-run with **no code change**, and #2708 is a documentation-only PR — so the cause is the harness, not the diff.

## Two distinct symptoms, one root cause

`.ci/scripts/run-test-cluster.sh` gates the suite on four `bin/wait-kuzzle` calls (ports 17510/17511/17512 and nginx 7512). `wait-kuzzle` resolves on the SDK's **`connected` event** — i.e. as soon as the WebSocket handshake succeeds. That says the transport is listening. It says nothing about the cluster having formed a quorum.

**Symptom A — `not_enough_nodes` (#2696, #2712).** The four waits pass in a couple of seconds, cucumber starts, and the very first `auth:login` in a `Before` hook is rejected:

```
ServiceUnavailableError: Rejected: this cluster is disabled
because there aren't enough nodes connected.
159 scenarios (1 failed, 157 skipped, 1 passed) — 0m01.770s
```

The sockets were open; the nodes had not finished discovering each other. This is a straightforward race the gate does not cover.

[...]

PR #2777, opened 2026-09-17
fix(cluster): close TD-33's cluster half — TD-65 and TD-67

rolljee:
## Overview

The two defects sprint 8's J1 found by typing `publisher.js`, fixed. Together they are the non-harness half of [TD-33](https://github.com/kuzzleio/kuzzle/issues/2715): *why one sync message is lost at cluster formation*, and *why one lost message costs a whole run*.

They ship together because **fixing TD-67 alone would make things worse**: it turns a spurious gap from a limp into a hard kill. TD-65 is what makes the gap non-spurious.

No protocol change, no new message type, no config. `lib/cluster/node.js` and `lib/cluster/subscriber.js` only. Mocha **3066** (+8 specs), vitest **233**, preflight green, ratchets unmoved.

## TD-65 (#2773) — a resume point taken on a channel that was not yet listening

The handshake read a peer's `lastMessageId` over the **command** socket (7510) and resumed the **sync** subscription (7511) from it. ZeroMQ propagates a subscription as a message sent *to* the publisher, which does the filtering; until it lands, the publisher silently drops what it sends. Nothing ordered that arrival before the snapshot — different sockets, different connections — so anything published in between was lost, and the subscriber was then told to expect the first of it.

**The fix:**

```js
const live = await subscriber.waitForSubscription(this.heartbeatDelay * 2);
```

`waitForSubscription()` resolves once **any** message has been received from the peer, which is the only observable proof the subscription is live on its side, and `handshake()` awaits it for every peer **before** `getFullState`.

Cheap, by construction: every node publishes a heartbeat every `cluster.heartbeat` ms (2 s by default) from **before** it is discoverable — `startHeartbeat()` runs ahead of the discovery loop — so the wait normally costs less than one round. Joins are serialised by the `clusterHandshake` mutex, so the cost is per joining node, not per peer. A peer that never answers gets a warning and is left to the existing `fullState === null` retry, which is what that path is for.

### A residual race, named rather than hidden

The resume point still comes from `fullState.nodesState` — from **whichever node answered**, not from the node being resumed. If that third party's record of a peer is older than the first message this node received from it, the gap reappears.

It is a far narrower window than the one closed here — one node's processing lag, against a full TCP connect plus subscription propagation — and closing it for good means resynchronising instead of evicting, which is a design decision (below).

## TD-67 (#2776) — a node that survives its own eviction

`evictSelf` logged the reason and broadcast `NodeEvicted` naming itself. A `PUB` socket does not deliver to its own process, so `handleNodeEviction`'s `global.kuzzle.shutdown()` branch — written for exactly this case — was unreachable by the only node that needed it. **The node left the cluster and kept answering requests behind the load balancer, from state that had stopped advancing.** That is what turned one dropped message into `Index nyc-open-data does not exist` 32 seconds later, on #2775's own CI run.

It now calls `shutdown()` itself. All four callers — two in `idCardHandler` for a dead ID-card renewer, two in `subscriber` — mean *this node can no longer be a cluster member*, so all four were leaving a node running that had asked everyone to forget it.

Three smaller defects fell out of the same reading:

| | Before | Now |
|---|---|---|
| gap branch | left the subscriber running, so one drop re-tripped on every later message — **nine log lines for one event** in CI | sets `state = EVICTED`, ending `listen()`'s loop |
| `sync()` | applied the rest of the buffer after a gap had evicted the node, then **overwrote `EVICTED` with `SANE`** on the way out | returns a verdict and stops at the eviction |
| `handshake()` | fired `sync()` without awaiting it and printed *Successfully completed the handshake* unconditionally — 2 ms **after** the eviction contradicting it | awaits it, logs success only on success |

The middle row was a latent bug of its own: nothing could reach `EVICTED` inside `sync()` before this PR, so it had never fired.

## Left open on purpose

**Preferring resynchronisation to eviction** — a node that lost a sync message could request a fresh full state instead of leaving the cluster. It is what would close TD-65's residual race, it is a behaviour and capacity decision rather than a bug fix, and it wants its own discussion. TD-67 keeps it as its third piece.

## Tests

8 new specs, all mocha in place (no new spec file, so the mocha ratchet is untouched at 149):

- `waitForSubscription` resolves true on the first received message, true immediately when already confirmed, false on timeout, and is confirmed by `listen()` itself.
- `handshake` calls it for every peer **before** `getFullState` — the ordering is the fix, so it is pinned.
- `handshake` warns and carries on when a subscription cannot be proven live.
- the gap branch evicts the subscriber, so one drop is reported once.
- `sync()` stops replaying and reports failure once the replay evicts this node.
- `evictSelf` shuts this node down.

## Risk

This changes cluster formation, which is the highest-risk code in the repo, so what is *not* changed is worth stating: no message type, no socket, no config default, no change to the discovery or quorum logic. The added wait is bounded by `heartbeatDelay * 2` and non-fatal on expiry. The shutdown in `evictSelf` is the behaviour `handleNodeEviction` already applies for a remotely-initiated eviction — this makes the self-initiated path agree with it.

Closes #2773, closes #2776 — both by hand, `Closes` does not fire on `2-dev`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rolljee:
## CI evidence, and what it is worth

**3 runs, 3 green, 0 failed jobs.** (Run 35239346272, plus two full re-runs.)

For comparison, the same 30-variant matrix on #2775 immediately before this fix: **2 failures in 3 runs**, both on `Functional tests (http, 24, 8)`, both with the `Node out-of-sync: 1 messages lost` signature.

### How much this proves

Not as much as it looks. Taking the prior failure rate at face value, three clean runs would happen about 4% of the time by chance — suggestive, not conclusive. And that 2-in-3 rate was itself measured over three runs, so the baseline is noisy too.

One thing it is explicitly **not**: I first checked the green runs for `out-of-sync` log lines and found zero. Then I checked the control — #2775's own green run, before the fix — and found zero there too. A green run has no desync by construction, since the desync is what fails it. That check was worthless and is not part of the evidence above.

The honest summary: consistent with the fix working, and the first clear departure from the measured behaviour. The signature to watch for in future runs is `Node out-of-sync` at formation; if it returns, TD-65's documented residual race (a third party's record of a peer being older than this node's first received message) is the first place to look.

